PR #362, opened 2018-08-02
CHECKOUT-3393 Multi-shipping convenience methods

capsula4:
## What?
Provides convenience methods for assigning items to an address. `CheckoutService#assignItemsToConsignment`

Also adds possibility to check error/status by address.

## Why?
Because having to use `updateConsignment` / `createConsignment` was a bit complex to use when implementing the UI we have designed for UCO.

## Testing / Proof
unit
manual

@bigcommerce/checkout


capsula4:
closed in favour of https://github.com/bigcommerce/checkout-sdk-js/pull/369

PR #374, opened 2018-08-16
CHECKOUT-3011: Initialize Braintree PayPal checkout button

davidchin:
## What?
* Ability to initialize Braintree PayPal checkout button.

## Why?
* Please see the following PRs:
  * #372
  * #370 
  * #366
  * #363

## Testing / Proof
* Unit


src/address/legacy-address.ts
```diff
@@ -0,0 +1,12 @@
+export default interface LegacyAddress {
```
capsula4: what's the difference between internal and legacy address? is this maybe `PaymentLegacyAddress` ?

src/checkout-buttons/checkout-button-error-selector.ts
```diff
@@ -0,0 +1,24 @@
...
+    private _checkoutButton: CheckoutButtonSelector;
+
+    /**
+     * @internal
+     */
+    constructor(selectors: InternalCheckoutSelectors) {
+        this._checkoutButton = selectors.checkoutButton;
+    }
+
+    getInitializeButtonError(methodId?: string): Error | undefined {
```
capsula4: should you document these methods (what happens if methodId is falsy?)

src/checkout-buttons/create-checkout-button-registry.ts
```diff
@@ -0,0 +1,43 @@
...
+            new PaypalScriptLoader(getScriptLoader()),
+            createFormPoster()
+        )
+    );
+
+    registry.register('braintreepaypalcredit', () =>
+        new BraintreePaypalButtonStrategy(
+            store,
+            checkoutActionCreator,
+            new BraintreeSDKCreator(new BraintreeScriptLoader(getScriptLoader())),
```
capsula4: Is it possible that a store uses both paypal buttons? if we instantiate `BraintreeScriptLoader`, could we be loading the library twice?

src/payment/strategies/paypal-express-payment-strategy.ts
```diff
@@ -56,13 +52,15 @@ export default class PaypalExpressPaymentStrategy extends PaymentStrategy {
 
         if (this._isInContextEnabled() && this._paypalSdk) {
             this._paypalSdk.checkout.closeFlow();
-            this._paypalSdk = null;
+            this._paypalSdk = undefined;
```
capsula4: I guess this might get tricky as it clears the paypalSDK attached to the class, but not from the window. We might be a wider issue with other strategies that use a similar approach.

src/address/legacy-address.ts
davidchin: Legacy address is the format `/checkout.php` expects.

src/checkout-buttons/create-checkout-button-registry.ts
davidchin: `getScriptLoader` returns a singleton, which internally manages the caching. But I can move it to a variable so it's clearer.

src/checkout-buttons/checkout-button-error-selector.ts
davidchin: Yeah good point. I haven't documented anything yet.

davidchin: The documentation might have to come as a separate PR.

PR #379, opened 2018-08-23
CHECKOUT-3461 Add ConsignmentSelector#getUnassignedItems method

capsula4:
## What?
Adds `ConsignmentSelector#getUnassginedItems` method

## Why?
Convenience method

## Testing / Proof
unit

@bigcommerce/checkout

### DIFF
--- a/src/shipping/consignment-selector.ts
+++ b/src/shipping/consignment-selector.ts
@@ -1,6 +1,7 @@
 import { find } from 'lodash';
 
 import { isAddressEqual, AddressRequestBody } from '../address';
+import { CartSelector, PhysicalItem } from '../cart';
 import { selector } from '../common/selector';
 
 import Consignment from './consignment';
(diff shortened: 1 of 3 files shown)


src/shipping/consignment-selector.ts
```diff
@@ -59,6 +61,25 @@ export default class ConsignmentSelector {
...
+            return [];
+        }
+
+        const assignedLineItemIds = reduce(
+            this.getConsignments(),
+            (itemIds, consignment) => itemIds.concat(consignment.lineItemIds),
+            [] as string[]
+        );
+
+        return filter(
```
davidchin: Can we just use `Array#filter`?

src/shipping/consignment-selector.ts
```diff
@@ -59,6 +61,25 @@ export default class ConsignmentSelector {
...
     }
 
+    getUnassignedItems(): PhysicalItem[] {
+        const cart = this._cart.getCart();
+
+        if (!cart) {
+            return [];
+        }
+
+        const assignedLineItemIds = reduce(
```
davidchin: Can we just use `Array#reduce`?

PR #389, opened 2018-08-29
[WIP] feat(payment): INT-774 Show Masterpass button in payments section into Stripe form

michxff:
## What? [INT-774](https://jira.bigcommerce.com/browse/INT-774)

- Add stripe strategy

## Why?

As a Shopper when Masterpass is enabled on Stripe I want to be able to use the MasterPass button to open the Master Pass page to select a credit card from my wallet

## Testing / Proof
...

@bigcommerce-labs/checkout


src/payment/strategies/masterpass/masterpass-payment-strategy.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+import { Observable } from 'rxjs';
+
+import { getCartState } from '../../../cart/carts.mock';
+import { createCheckoutStore, CheckoutRequestSender, CheckoutStore, CheckoutValidator } from '../../../checkout';
+import { getCheckoutState } from '../../../checkout/checkouts.mock';
+import { InvalidArgumentError } from '../../../common/error/errors';
+import { getConfigState } from '../../../config/configs.mock';
+import { getCustomerState } from '../../../customer/customers.mock';
+import { OrderActionCreator, OrderActionType, OrderRequestSender } from '../../../order';
+import {
```
vmparra: No need to add Index

src/payment/strategies/masterpass/masterpass-payment-strategy.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+    PaymentActionCreator, PaymentInitializeOptions, PaymentMethod,
+    PaymentMethodActionCreator, PaymentMethodActionType,
+    PaymentMethodRequestSender,
+    PaymentRequestSender
+} from '../../index';
+import { PaymentActionType } from '../../payment-actions';
```
vmparra: Merge with previous import

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,125 @@
+import { Checkout, CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
+import {
+    InvalidArgumentError,
+    MissingDataError,
+    MissingDataErrorType
+} from '../../../common/error/errors';
+import { StoreConfig } from '../../../config';
+import { OrderActionCreator, OrderRequestBody } from '../../../order';
+import { PaymentArgumentInvalidError } from '../../errors';
+import { PaymentMethod, PaymentMethodActionCreator } from '../../index';
```
vmparra: No need to  mention index (Applies to entire PR)

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,125 @@
...
+import PaymentActionCreator from '../../payment-action-creator';
```
vmparra: Merge  with next import

Sure-Round:
@davidchin @icatalina ♻️  This PR is now ready for the Sydney team to review

src/payment/strategies/masterpass/masterpass-payment-initialize-options.ts
```diff
@@ -0,0 +1,11 @@
+export interface MasterpassPaymentInitializeOptions {
+    /**
+     * The gateway where the masterpass payment will be processed.
+     */
+    gateway?: string;
```
davidchin: You shouldn't need to pass the `gateway` field again, because it should already be passed when you call `initialize`. Could you please remove?

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,127 @@
...
+        const { methodId, masterpass: masterpassOptions } = options;
+
+        if (!masterpassOptions) {
+            throw new InvalidArgumentError('Unable to initialize payment because masterpass options is missing');
+        }
+
+        const state = this._store.getState();
+        this._paymentMethod = state.paymentMethods.getPaymentMethod(methodId);
+
+        return Promise.resolve(this._hasPaymentInfo()).then(hasPaymentInfo => {
```
davidchin: I don't see any reason why you need to wrap the call inside `Promise.resolve`. Could you please remove?

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,127 @@
...
+            if (!hasPaymentInfo) {
+                return this._masterpassClientSetup().then(checkoutCallback => checkoutCallback());
+            }
+
+            this._paymentGateway = masterpassOptions.gateway;
```
davidchin: As I said before, you should have access to `gateway` field under `options.gateway`, you shouldn't need to introduce a new one.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,127 @@
...
+        if (!masterpassOptions) {
+            throw new InvalidArgumentError('Unable to initialize payment because masterpass options is missing');
+        }
+
+        const state = this._store.getState();
+        this._paymentMethod = state.paymentMethods.getPaymentMethod(methodId);
+
+        return Promise.resolve(this._hasPaymentInfo()).then(hasPaymentInfo => {
+            if (!hasPaymentInfo) {
+                return this._masterpassClientSetup().then(checkoutCallback => checkoutCallback());
```
davidchin: Sorry, I don't 100% follow here. You want to launch the Masterpass modal (or redirect to Masterpass website) when `initialize` is called? This can't happen because `initialize` is suppose to be a preparation step, i.e.: loading some scripts, attaching some event listeners. Nothing should actually happen until `execute` is called, or trigger any other event listeners themselves.

davidchin: If your intention is to prompt the shopper if they haven't selected a payment option previously, you need to trigger the modal / perform the redirect in the `execute` step. You'll know what I mean if you take a look at the implementation of Masterpass provided by Square.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,127 @@
...
+            }
+
+            this._paymentGateway = masterpassOptions.gateway;
+
+            if (masterpassOptions.onPaymentSelect) {
+                masterpassOptions.onPaymentSelect();
```
davidchin: I'm not sure why you're calling `onPaymentSelect` here because I'm pretty sure at this point the shopper hasn't selected a payment option from their wallet.

src/payment/create-payment-strategy-registry.ts
```diff
@@ -253,5 +254,15 @@ export default function createPaymentStrategyRegistry(
         )
     );
 
+    registry.register('masterpass', () =>
```
davidchin: Does this work with all Masterpass providers? i.e.: Braintree and Stripe? Do you require any provider-specific logic?

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,127 @@
...
+
+        const { paymentData } = payment;
+        const methodId = this._paymentGateway;
+
+        if (!methodId) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        return this._store.dispatch(this._orderActionCreator.submitOrder(order, options))
+            .then(() => this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId)))
```
davidchin: Is there any reason why you need to reload the payment method again?

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,127 @@
...
+            cartId: checkout.cart.id,
+            suppressShippingAddress: true,
+        };
+    }
+
+    private _hasPaymentInfo(): boolean {
+        return this._paymentMethod
+            && this._paymentMethod.initializationData
+            && this._paymentMethod.initializationData.paymentData
+            && this._paymentMethod.initializationData.paymentData.nonce;
```
davidchin: Why are you introducing a new field called `nonce` under `initializationData`? If you look at our API contract, you'll see that under `PaymentMethod` already has a `nonce` field. You shouldn't need to introduce a new one.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
michxff: i removed onPaymentSelect, at some point of the development was necessary but with the latest changes on ng-checkout isn't required anymore

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
michxff: Here the payment method which uses the masterpass wallet is going to be loaded, i.e stripe

src/payment/create-payment-strategy-registry.ts
michxff: We expect to use the same logic for all providers at least on the front-end

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
michxff: The only reason why we are doing this is that we are following the same pattern used on Braintree Visa Checkout https://github.com/bigcommerce/checkout-sdk-js/blob/fd43113e2c4c91c3f630949b3dcdb7b860cac73f/src/payment/strategies/braintree/braintree-visacheckout-payment-strategy.ts#L103 and that way is how the stripe payment provider expects to receive the payment data.

The only provider which uses the nonce at the root level is braintree but it has a custom code which is not the same approach that we are trying to follow for the masterpass implementation.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
michxff: Yes when the masterpass payment provider (wallet) is selected on the payment form the redirect is made, the following link contains a video showing that functionality https://drive.google.com/open?id=1NWpm69noGMhjsg09kB1-Vb5v3H4a_G-1

michxff:
@davidchin I did all the changes, it is ready for another look ♻️

Sure-Round:
@davidchin @icatalina ♻️ This PR is ready for another pass.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
davidchin: I see. In fact, Braintree is the correct way to expose `nonce`. All other methods (Stripe, Visa Checkout) replicate the wrong approach, which unfortunately becomes the new norm. Anyway, considering it is not something new that you've introduced, I guess we can address this later.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
davidchin: Yeah, so why do you need to do that? Unless I'm missing something, it doesn't seem you're using the data at all? We want to save a network call if it's possible.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
icatalina: Actually, using the `initializationData` is what's really custom. It was made as a way of keeping extra data for the payment method. The `nonce` is part of the normal `paymentProvider` payload and it should be contained there.

VisaCheckout is an old implementation, not attached to any standard. We shouldn't consider it a model.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,121 @@
...
+        const { methodId } = options;
+
+        this._paymentMethod = this._store.getState().paymentMethods.getPaymentMethod(methodId);
+
+        if (!this._paymentMethod) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        if (!this._hasPaymentInfo()) {
+            return this._masterpassClientSetup().then(checkoutCallback => checkoutCallback()).then(() => super.initialize(options));
```
davidchin: Thank you for the screen capture. That video shows exactly what it shouldn't happen. We shouldn't be redirecting the shopper in this method. As I said before, this method is only responsible for initialising the payment method (i.e.: loading the Masterpass client JS); not prompting the shopper to enter their payment details.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,121 @@
...
+
+        if (!this._paymentMethod) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        const methodId = this._paymentMethod.initializationData.gateway;
+        const paymentData = this._paymentMethod.initializationData.paymentData;
+
+        if (!methodId) {
+            const msg = 'Unable to proceed because "payload.payment.methodId" argument is not provided.';
```
davidchin: This temporary variable seems unnecessary. Could you please remove?

icatalina: This message is not actually true, what's missing if you throw this error is `this._paymentMethod.initializationData.gateway`.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        const order = { useStoreCredit: payload.useStoreCredit };
+
+        if (!this._paymentMethod) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        const methodId = this._paymentMethod.initializationData.gateway;
```
icatalina: `MethodId` already has a meaning. It might be a good idea to rename this to something different so it is not a source of confusion.

src/payment/strategies/masterpass/masterpass-payment-strategy.spec.ts
```diff
@@ -0,0 +1,265 @@
...
+
+describe('MasterpassPaymentStragegy', () => {
+    // Described class
+    let strategy: MasterpassPaymentStrategy;
+
+    // Strategy's constructor dependencies
+    let orderRequestSender: OrderRequestSender;
+    let requestSender: RequestSender;
+
+    // Strategy's constructor arguments
```
davidchin: What's the difference between `strategy's constructor dependencies` and `strategy's constructor arguments`? Anyway,  these comments seem redundant, could you please remove unless you really want to keep them.

src/payment/strategies/masterpass/masterpass-payment-strategy.spec.ts
```diff
@@ -0,0 +1,265 @@
...
+            it('does not load the masterpass script', async () => {
+                await strategy.initialize(initOptions);
+                expect(scriptLoader.load).not.toHaveBeenLastCalledWith(true);
+                expect(masterpassScript.checkout).not.toHaveBeenCalled();
+            });
+        });
+    });
+
+    describe('#execute', () => {
+        let payload: any;
```
davidchin: Could you please provide a proper type?

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
icatalina: can you break the `then` into different lines?

```js
return this._masterpassClientSetup()
	.then(checkoutCallback => checkoutCallback())
	.then(() => super.initialize(options));
```

Also, if all you are doing with the callback is calling it immediately? isn't that part of the `clientSetup` process?

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,121 @@
...
+        const paymentData = this._paymentMethod.initializationData.paymentData;
+
+        if (!methodId) {
+            const msg = 'Unable to proceed because "payload.payment.methodId" argument is not provided.';
+            throw new InvalidArgumentError(msg);
+        }
+
+        if (!paymentData) {
+            const msg = 'Unable to proceed because "payload.payment.paymentData" argument is not provided.';
```
icatalina: Same here, we are not reading this value from `payload.payment.paymentData`, but from `this._paymentMethod.initializationData.paymentData`.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
icatalina: oh, I see, you are swapping the `methodId` for the `gateway` and then using that in the `paymentActionCreator`, which will try to get the gateway as a payment provider. is that right?

I don't really know how are we making these `gateways` be valid payment methods but not to show in the `/payments` endpoint...

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
icatalina: Could you please move every `then` to a different line?

```js
return this._masterpassClientSetup()
	.then(checkoutCallback => checkoutCallback())
	.then(() => super.initialize(options));
```

Also, since all you are doing with the callback is calling it, it might be a good idea to make it part of `_masterpassClientSetup`... what do you think?

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,121 @@
...
+
+        if (!checkout) {
+            throw new MissingDataError(MissingDataErrorType.MissingCheckout);
+        }
+
+        if (!storeConfig) {
+            throw new MissingDataError(MissingDataErrorType.MissingCheckoutConfig);
+        }
+
+        if (!(this._paymentMethod && this._paymentMethod.initializationData)) {
```
icatalina: 🍹Just use an `||`

```js
if (!this._paymentMethod || !this._paymentMethod.initializationData) {
```

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,121 @@
...
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        const payload = this._createMasterpassPayload(this._paymentMethod, checkout, storeConfig);
+
+        return this._masterpassScriptLoader.load(this._paymentMethod.config.testMode)
+            .then(masterpassClient => () => masterpassClient.checkout(payload));
```
icatalina: Actually, reading this code, why are we returning a promise that we then call? Can't we just call the thing directly?

```js
return this._masterpassScriptLoader.load(this._paymentMethod.config.testMode)
	.then(masterpassClient => masterpassClient.checkout(payload));
```

src/payment/strategies/masterpass/masterpass-payment-strategy.spec.ts
michxff: that was a typo i will removed

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
michxff: @icatalina, @davidchin yes that's right from masterpass we only get the nonce (payment token) that will be used with the payment provider which has the wallet configured, for instance if the gateway is stripe, we load the stripe payment method in order to be able to do the submitPayment call using stripe.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
michxff: @davidchin Basically if we move the logic to the execute the flow will change

This is the expected flow

1. The user goes to the checkout page
2. It selects masterpass as payment method
3. It gets redirected to the masterpass page to pick a card
4. It returns to bcapp checkout page, the wallet is displayed.
5. The user click on place order (payment strategy execute method) and the order and the payment is created.

If we move the redirect logic to the execute method:

1. The user goes to the checkout page
2. It selects masterpass as payment option
3. The user click on place order (payment strategy execute method)
4. It gets redirected to the masterpass page to pick a card.
5. It returns to bcapp checkout page, the wallet is displayed.
6. The user click on place order again (payment strategy execute method) and the order and the payment is created.

FYI: @Sure-Round

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
michxff: Yes i will removed, but was added to avoid having lines too long

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
michxff: renamed to gateway

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
michxff: What I'm returning in the resolved promise is a function which has the masterpass client configured with the checkout method ready to be called, this is useful for delay the execution, i.e if we pass it to an event handler.

However if we aren't going to pass it to anyone and the redirect will be executed from the initialize method, this can be called directly as you suggested.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
Sure-Round: @michxff Can you address this comment?

src/payment/strategies/masterpass/masterpass-payment-strategy.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+                paymentMethodMock.initializationData = {
+                    allowedCardTypes: ['visa', 'amex', 'mastercard'],
+                    checkoutId: 'checkout-id',
+                };
+            });
+
+            it('throws an exception if checkout data is missing', () => {
+                jest.spyOn(store.getState().checkout, 'getCheckout').mockReturnValue(null);
+                const error = 'Unable to proceed because checkout data is unavailable.';
+                expect(() => strategy.initialize(initOptions)).toThrow(error);
```
FranciscoOnt: Instead of expecting the Error's Message, you can expect the error type:
IE:
``` typescript
expect(() => strategy.initialize(initOptions)).toThrow(InvalidArgumentError);
```

src/payment/strategies/masterpass/masterpass-payment-strategy.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+            });
+
+            it('throws an exception if store config is missing', () => {
+                jest.spyOn(store.getState().config, 'getStoreConfig').mockReturnValue(null);
+                const error = 'Unable to proceed because configuration data is unavailable.';
```
FranciscoOnt: If you expect the error type as above, you can remove the unused constants

src/payment/strategies/masterpass/masterpass-payment-strategy.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+                expect(() => strategy.initialize(initOptions)).toThrow(error);
+            });
+
+            it('throws an exception if payment method initialization data is missing', () => {
+                paymentMethodMock.initializationData = null;
+                const error = 'Unable to proceed because payment method data is unavailable or not properly configured.';
+                expect(() => strategy.initialize(initOptions)).toThrow(error);
```
FranciscoOnt: same here

src/payment/strategies/masterpass/masterpass-payment-strategy.spec.ts
FranciscoOnt: you can also use a 'try catch' block to test the error and expect the error type instead of a string
```typescript
            it('throws an exception if checkout data is missing', async () => {
                jest.spyOn(store.getState().checkout, 'getCheckout').mockReturnValue(null);
                try {
                    await strategy.initialize(initOptions);
                } catch (e) {
                    expect(e).toBeInstanceOf(MissingDataError);
                }
            });
```

src/payment/strategies/masterpass/masterpass-payment-strategy.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+    let scriptLoader: MasterpassScriptLoader;
+
+    let initOptions: PaymentInitializeOptions;
+    let paymentMethodMock: PaymentMethod;
+    let stripePaymentMethodMock: PaymentMethod;
+    let masterpassScript: Masterpass;
+
+    beforeEach(() => {
+        initOptions = {
+            methodId: 'masterpass',
```
FranciscoOnt: You are missing the 'masterpass' property of the Initialize options 🍹

FranciscoOnt: Well, you actually need it, unless you want the strategy to throw an exception at line 46

michxff: You're right, I was still working on the tests

michxff:
@davidchin This is ready for another look ♻️

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,153 @@
...
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        if (!options.masterpass) {
+            throw new InvalidArgumentError('Unable to initialize payment because "options.masterpass" argument is not provided.');
+        }
+
+        const walletButton = options.masterpass.walletButton && document.getElementById(options.masterpass.walletButton);
+
+        if (!this._hasPaymentInfo()) {
```
davidchin: Can you remove this check? If `walletButton` is passed, we should always attach an event listener to it.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,153 @@
...
+
+        if (!this._paymentMethod) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        if (!options.masterpass) {
+            throw new InvalidArgumentError('Unable to initialize payment because "options.masterpass" argument is not provided.');
+        }
+
+        const walletButton = options.masterpass.walletButton && document.getElementById(options.masterpass.walletButton);
```
davidchin: We should call `getElementById` after `_masterpassClientSetup`. Otherwise, we won't be able to select it if the element ID is dynamically rendered in the view.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
davidchin: I guess `initializationData` is undocumented, we could probably leave it as it is considering the deadline they're facing. We could leave a comment here to say `TODO: Refactor the API endpoint to return nonce in the right place.`

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,153 @@
...
+
+        if (!storeConfig) {
+            throw new MissingDataError(MissingDataErrorType.MissingCheckoutConfig);
+        }
+
+        if (!this._paymentMethod || !this._paymentMethod.initializationData) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        const payload = this._createMasterpassPayload(this._paymentMethod, checkout, storeConfig);
```
davidchin: I think you need to create the `payload` object as soon as you need it. Otherwise, you might be sending stale data to Masterpass. This is because `payload` is cached inside this closure and might be outdated by the time `_walletCallback` is called. For example, the shopper might apply a coupon code during that time. So, instead of returning a callback here, you should return the `MasterpassClient` instance instead and then use it directly inside `_handleWalletButtonClick`. i.e.:

```ts
initialize() {
    // Some code here
    return this._masterpassClientSetup()
        .then(masterpassClient) => {
            this._masterpassClient = masterpassClient;
        });
}

// And then in 
_handleWalletButtonClick() {
    // Some code here
    if (!this._masterpassClient) {
        return;
    }

    // More code here to get storeConfig, checkout data etc... or move it to `_createMasterpassPayload` so you don't have to do the preparation in this method.
    this._masterpassClient.checkout(this._createMasterpassPayload());
}
```

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -0,0 +1,153 @@
...
+
+        const gateway = this._paymentMethod.initializationData.gateway;
+        const paymentData = this._paymentMethod.initializationData.paymentData;
+
+        if (!gateway) {
+            throw new InvalidArgumentError('Unable to proceed because "paymentMethod.initializationData.gateway" argument is not provided.');
+        }
+
+        if (!paymentData) {
+            throw new InvalidArgumentError('Unable to proceed because "paymentMethod.initializationData.paymentData" argument is not provided.');
```
davidchin: You need to show the wallet here instead of throwing an error. And then finalise the order upon redirection. Nevertheless, I'm happy for you to leave a comment here and we can address this later `TODO: Redirect to Masterpass if nonce has not been generated yet. And then finalise the order when the shopper is redirected back to the checkout page.`

src/payment/strategies/masterpass/masterpass-payment-strategy.spec.ts
```diff
@@ -0,0 +1,295 @@
+import { createClient as createPaymentClient } from '@bigcommerce/bigpay-client';
+import { Action } from '@bigcommerce/data-store';
+import createAction from '@bigcommerce/data-store/lib/create-action';
```
davidchin: `import { createAction, Action } from '@bigcommerce/data-store';` - you can combine with the line above.

Sure-Round: @michxff Let's address this comment it's a quick change.

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
Sure-Round: @michxff Let's add this TODO comment in.

michxff: @Sure-Round added

michxff:
@davidchin @Sure-Round i did all the changes is ready for another look

src/payment/strategies/masterpass/masterpass-payment-strategy.ts
Sure-Round: @michxff Did we address this comment?

michxff:
@Sure-Round done

PR #395, opened 2018-09-05
Billing Address enhancements

capsula4:
- Add status/error selectors for `CheckutService#continueAsGuest`
- Update `CheckutService#continueAsGuest` signature
- Do not overwrite billing info when using `CheckutService#continueAsGuest`

### DIFF
--- a/src/billing/billing-address-action-creator.ts
+++ b/src/billing/billing-address-action-creator.ts
@@ -9,7 +9,7 @@ import { RequestOptions } from '../common/http-request';
 import { GuestCredentials } from '../customer';
 
 import { BillingAddressUpdateRequestBody } from './billing-address';
-import { BillingAddressActionType, UpdateBillingAddressAction } from './billing-address-actions';
+import { BillingAddressActionType, ContinueAsGuestAction, UpdateBillingAddressAction } from './billing-address-actions';
 
 export default class BillingAddressActionCreator {
     constructor(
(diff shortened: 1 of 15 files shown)


src/checkout/checkout-store-status-selector.ts
```diff
@@ -281,7 +282,16 @@ export default class CheckoutStoreStatusSelector {
     }
 
     /**
-     * Checks whether the current customer is updating their shipping address.
+     * Checks whether the shopper is continuing out as a guest.
+     *
+     * @returns True if continuing as guest, otherwise false.
+     */
+    isContinuingAsGuest(): boolean {
+        return this._billingAddress.isUpdating();
```
davidchin: So if the user is just updating their billing address, this method will return `true`? I think we might need to use a different key to track the status if we want to distinguish between the two.

src/billing/billing-address-action-creator.ts
```diff
@@ -15,14 +16,42 @@ export default class BillingAddressActionCreator {
...
+            }
+
+            const billingAddress = state.billingAddress.getBillingAddress();
+
+            let billingAddressRequestBody;
+
+            if (!billingAddress) {
+                billingAddressRequestBody = credentials;
+            } else {
+                const { country, ...existingBillingAddressRequestBody } = billingAddress;
```
davidchin: A question - why do we need to omit the country field?

capsula4: because we send the `countryCode`, `country` is not part of the payload

src/checkout/checkout-store-status-selector.ts
capsula4: Yeah I was thinking about it. The way we queue up things mean we can't really have both things happening at the same time, but yeah I can see how in certain UIs it could be an issue.

src/billing/billing-address-action-creator.ts
davidchin: I see, does it break if we don't omit it?

capsula4: yeah, linting error, we have well defined types for request vs response

PR #401, opened 2018-09-11
feat(checkout-buttons): PAYMENTS-3073 Support credit buttons by implementing funding sources

lord2800:
## What?
Support credit buttons by implementing funding sources.

## Why?
Removes the need to have two separate strategies (to be removed later) to support credit buttons--it's just an option to the Braintree and Paypal button SDKs.

## Testing / Proof
Add a product to your cart, open the cart page, see that with credit enabled you get two side-by-side (by default, changeable with Smart Buttons) buttons. With credit disabled, you get the standard braintree/paypal button.

### DIFF
--- a/src/payment/strategies/paypal/paypal-sdk.ts
+++ b/src/payment/strategies/paypal/paypal-sdk.ts
@@ -1,6 +1,12 @@
 export interface PaypalSDK {
     Button: PaypalButton;
     checkout: PaypalExpressCheckout;
+    FUNDING: PaypalFundingTypeList;
+}
+
+export interface PaypalFundingTypeList {
+    CARD: string;
+    CREDIT: string;
 }
 
 export interface PaypalButton {
@@ -11,8 +17,14 @@ export interface PaypalButtonOptions {
     env?: string;
     commit?: boolean;
     style?: PaypalButtonStyleOptions;
+    funding?: PaypalFundingType;
     payment(): Promise<any>;
     onAuthorize(data: PaypalAuthorizeData): Promise<any>;
+}
+
+export interface PaypalFundingType {
+    allowed?: string[];
+    disallowed?: string[];
 }
 
 export interface PaypalButtonStyleOptions {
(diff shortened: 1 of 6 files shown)


lord2800:
Before this can be merged, the appropriate data needs to be available. There is no reasonable fallback mechanism for this, unfortunately.

src/payment/strategies/paypal/paypal-sdk.ts
```diff
@@ -1,6 +1,12 @@
 export interface PaypalSDK {
     Button: PaypalButton;
     checkout: PaypalExpressCheckout;
+    FUNDING: PaypalFundingTypeList;
+}
+
+export interface PaypalFundingTypeList {
```
icatalina: You might be able to use an enum here and only allow values of this type to go to the `allowed`/`disallowed`.

```ts
// Using a fixed string makes typescript only accept the actual
// enums and not the equivalent string or a number instead.
// Since they are _ambient_ enums (`declare`) they don't get
// compiled, the only act as a definition of the actual value
// in the API.

declare enum TYPES {
    CARD = 'CARD',
    PAYPAL = 'PAYPAL',
}

export interface PaypalFundingTypeList {
	CARD: PaypalFundingTypes.CARD;
	CREDIT: PaypalFundingTypes.CREDIT;
}

export interface PaypalFundingType {
	allowed?: PaypalFundingTypes[];
	disallowed?: PaypalFundingTypes[];
}
```

lord2800: Oh nice, TIL.

lord2800: OK, I played around with this for most of the day and was unable to make the typescript compiler happy. If you're OK with skipping this for now (it sounds like a nice improvement we can do later), then this is more or less ready for review.

lord2800:
There were existing tests for credit buttons, I merely added onto them since the way it works is by passing a new option and asserting that slightly different values are passed to `paypal.Button.render`.

lord2800:
Correct. This is ready for merge when you're ready.

PR #405, opened 2018-09-12
CP-3982 adding product-id to the cart item property for checkout process

bc-krishsenthilraj:
## What?
Added the ProductId as one of the properties to the `internal line item` to have the ProductId as part of the checkout payload while sending the checkout started event to GA.

## Why?
Checkout started event does not have the product id as part of payload. So, we need to add this to the original cart items. 

## Testing / Proof

@bigcommerce/checkout @bigcommerce/cp-dt

### DIFF
--- a/src/cart/internal-line-item.ts
+++ b/src/cart/internal-line-item.ts
@@ -13,6 +13,7 @@ export default interface InternalLineItem {
     quantity: number;
     type: string;
     variantId: number | null;
+    productId?: number;
     addedByPromotion?: boolean;
     sender?: {
         name: string;
--- a/src/cart/internal-carts.mock.ts
+++ b/src/cart/internal-carts.mock.ts
@@ -25,6 +25,7 @@ export function getCart(): InternalCart {
                 integerAmountAfterDiscount: 20000,
                 variantId: 71,
                 addedByPromotion: false,
+                productId: 103,
             },
             {
                 id: '667',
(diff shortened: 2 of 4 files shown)


davidchin:
Hey @bc-krishsenthilraj, there's an issue with the TS compiler. Can you run `npm run lint` and try to identify the issue? Thanks.

src/cart/internal-line-item.ts
```diff
@@ -13,6 +13,7 @@ export default interface InternalLineItem {
     quantity: number;
     type: string;
     variantId: number | null;
+    productId?: number;
```
davidchin: I believe it should be `productId` (without the `?`). If you look at `LineItem`, you can see that `productId` is not an optional property.

bc-krishsenthilraj: When I added it as a `mandatory` property, it throws error for `GiftCertificateItem` since its returning  the `InternalLineItem` object @ https://github.com/bigcommerce/checkout-sdk-js/blob/b14beeea0456e1fb7c39cd445faef3161a84afed/src/cart/map-gift-certificate-to-internal-line-item.ts#L9. 

So I have discussed it with @lord2800 and we decided to add it as a optional property.

davidchin: Ah I see, thanks for explaining. I'm okay with this.

bc-krishsenthilraj: squashed the commits.

bc-ranji:
💚

PR #406, opened 2018-09-13
fix(payment) PAYMENTS-3288 Checking out with PayPal does not send device data

jonathandelefortrie:
## What?

Device data is not being sent as expected when checkout with PayPal, I have removed the transformation applied to the data collector in the getDataCollector method.

## Why?

The `device_info` part of the deviceData should be sent as it is.

## Testing / Proof

Before:
screen shot 2018-09-12 at 4 19 33 pm

After:
screen shot 2018-09-12 at 4 22 43 pm

### DIFF
--- a/src/payment/strategies/braintree/braintree-sdk-creator.spec.ts
+++ b/src/payment/strategies/braintree/braintree-sdk-creator.spec.ts
@@ -6,7 +6,9 @@ import BraintreeSDKCreator from './braintree-sdk-creator';
 import {
     getClientMock,
     getDataCollectorMock,
+    getDeviceDataMock,
     getModuleCreatorMock,
+    getModuleCreatorNewMock,
     getThreeDSecureMock,
     getVisaCheckoutMock,
 } from './braintree.mock';
@@ -112,7 +114,7 @@ describe('Braintree SDK Creator', () => {
 
         beforeEach(() => {
             dataCollectorMock = getDataCollectorMock();
-            dataCollectorCreatorMock = getModuleCreatorMock(dataCollectorMock);
+            dataCollectorCreatorMock = getModuleCreatorNewMock(dataCollectorMock);
             braintreeScriptLoader.loadDataCollector = jest.fn().mockReturnValue(Promise.resolve(dataCollectorCreatorMock));
             braintreeSDKCreator = new BraintreeSDKCreator(braintreeScriptLoader);
             jest.spyOn(braintreeSDKCreator, 'getClient').mockReturnValue(Promise.resolve(clientMock));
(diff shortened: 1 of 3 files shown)


src/payment/strategies/braintree/braintree-sdk-creator.spec.ts
```diff
@@ -135,6 +136,9 @@ describe('Braintree SDK Creator', () => {
         });
 
         it('returns different data collector instance if it is used for PayPal', async () => {
+            braintreeScriptLoader.loadDataCollector = jest.fn(() =>
```
jonathandelefortrie: Hey @danieldelcore @davidchin, I do not feel super comfortable with the fix done is this test and I was wondering if you can provide your opinion on what would be the best thing to do in here.

danieldelcore: This test seems to be trying to assert that `dataCollector.create({ client, kount: true, ...options }))` returns a new instance every time it's called.  But we've actually mocked the `.create` function above (line 116) and it's now responsible for the creation of new instances. We seem to be testing a mock.

Based on that, I'm not entirely sure if we're actually testing anything here. IMO I think we should remove this test. 

The test above, seems to test this scenario well:

```javascript
       it('always returns the same instance of the 3ds client', async () => {
             const dataCollector1 = await braintreeSDKCreator.getDataCollector();
             const dataCollector2 = await braintreeSDKCreator.getDataCollector();
             expect(dataCollector1).toBe(dataCollector2);
             expect(braintreeScriptLoader.loadDataCollector).toHaveBeenCalledTimes(1);
             expect(dataCollectorCreatorMock.create).toHaveBeenCalledTimes(1);
        });
```

davidchin: I already chatted with Jonathan offline. Basically he needs to move the stub to `beforeEach` so it affects every test in the block.

We need to keep the test because it’s testing the caching mechanism. It’s not testing the mock. He needs to update the stub because he can’t test the caching mechanism if the stub returns the same instance no matter what.

icatalina: Actually the stub is already in the beforeEach, but seems like it doesn't return a new version every time we call it.

jonathandelefortrie:
Great! Thanks @davidchin

src/payment/strategies/braintree/braintree-sdk-creator.spec.ts
danieldelcore: Yeah, that was my confusion.

PR #407, opened 2018-09-13
INT-835 Implement Google Pay + Braintree in Payments Section

clopezh:
## What?
Support Google Pay as a Wallet implementing a strategy. Google Pay needs to work with Braintree configuration and load the proper files.

## Why?
I want to be able to open the Google Pay Wallet, select a car and get a nonce from Google Pay using Braintree configuration and after that checkout using Google Pay as a Wallet and Braintree as a Provider.

## Testing / Proof
image

Video: https://drive.google.com/open?id=12WrpSCazOym9DnrjA3tfVXPLW3F4YdE9

## Jira Ticket
https://jira.bigcommerce.com/browse/INT-835

### DIFF
--- a/src/payment/create-payment-strategy-registry.ts
+++ b/src/payment/create-payment-strategy-registry.ts
@@ -21,6 +21,7 @@ import {
     BraintreePaypalPaymentStrategy,
     BraintreeVisaCheckoutPaymentStrategy,
     CreditCardPaymentStrategy,
+    GooglePayPaymentStrategy,
     KlarnaPaymentStrategy,
     LegacyPaymentStrategy,
     NoPaymentDataRequiredPaymentStrategy,
@@ -34,8 +35,15 @@ import {
 } from './strategies';
 import { AfterpayScriptLoader } from './strategies/afterpay';
 import { AmazonPayScriptLoader } from './strategies/amazon-pay';
-import { createBraintreePaymentProcessor, createBraintreeVisaCheckoutPaymentProcessor, VisaCheckoutScriptLoader } from './strategies/braintree';
+import {
+    createBraintreePaymentProcessor,
+    createBraintreeVisaCheckoutPaymentProcessor,
+    BraintreeScriptLoader,
+    BraintreeSDKCreator,
+    VisaCheckoutScriptLoader
+} from './strategies/braintree';
 import { ChasePayPaymentStrategy, ChasePayScriptLoader } from './strategies/chasepay';
+import { GooglePayBraintreeInitializer, GooglePayPaymentProcessor, GooglePayScriptLoader } from './strategies/googlepay';
 import { KlarnaScriptLoader } from './strategies/klarna';
 import { PaypalScriptLoader } from './strategies/paypal';
 import { SquareScriptLoader } from './strategies/square';
@@ -49,6 +57,9 @@ export default function createPaymentStrategyRegistry(
     const registry = new PaymentStrategyRegistry(store, { defaultToken: 'creditcard' });
     const scriptLoader = getScriptLoader();
     const braintreePaymentProcessor = createBraintreePaymentProcessor(scriptLoader);
+    const braintreeScriptLoader = new BraintreeScriptLoader(scriptLoader);
+    const braintreeSdkCreator = new BraintreeSDKCreator(braintreeScriptLoader);
+
     const checkoutRequestSender = new CheckoutRequestSender(requestSender);
     const checkoutValidator = new CheckoutValidator(checkoutRequestSender);
     const orderActionCreator = new OrderActionCreator(
@@ -241,6 +252,25 @@ export default function createPaymentStrategyRegistry(
             requestSender,
             new ChasePayScriptLoader(getScriptLoader()),
             new WepayRiskClient(scriptLoader)
+        )
+    );
+
+    registry.register('googlepaybraintree', () =>
+        new GooglePayPaymentStrategy(
+            store,
+            checkoutActionCreator,
+            paymentMethodActionCreator,
+            paymentStrategyActionCreator,
+            paymentActionCreator,
+            orderActionCreator,
+            new GooglePayPaymentProcessor(
+                store,
+                paymentMethodActionCreator,
+                new GooglePayScriptLoader(scriptLoader),
+                new GooglePayBraintreeInitializer(braintreeSdkCreator),
+                requestSender,
+                new BillingAddressActionCreator(new BillingAddressRequestSender(requestSender))
+            )
         )
     );
 
--- a/src/payment/strategies/braintree/braintree-script-loader.spec.ts
+++ b/src/payment/strategies/braintree/braintree-script-loader.spec.ts
@@ -1,10 +1,21 @@
 import { ScriptLoader } from '@bigcommerce/script-loader';
 
-import { BraintreeClientCreator, BraintreeDataCollector, BraintreeHostWindow, BraintreeModuleCreator, BraintreeThreeDSecure, BraintreeVisaCheckout } from './braintree';
+import { StandardError } from '../../../common/error/errors';
+
+import {
+    BraintreeClientCreator,
+    BraintreeDataCollector,
+    BraintreeHostWindow,
+    BraintreeModuleCreator,
+    BraintreeThreeDSecure,
+    BraintreeVisaCheckout,
+    GooglePayBraintreeSDK,
+} from './braintree';
 import BraintreeScriptLoader from './braintree-script-loader';
 import {
     getClientMock,
     getDataCollectorMock,
+    getGooglePayMock,
     getModuleCreatorMock,
     getThreeDSecureMock,
     getVisaCheckoutMock,
@@ -37,7 +48,7 @@ describe('BraintreeScriptLoader', () => {
 
         it('loads the client', async () => {
             await braintreeScriptLoader.loadClient();
-            expect(scriptLoader.loadScript).toHaveBeenCalledWith('//js.braintreegateway.com/web/3.15.0/js/client.min.js');
+            expect(scriptLoader.loadScript).toHaveBeenCalledWith('//js.braintreegateway.com/web/3.37.0/js/client.min.js');
         });
 
         it('returns the client from the window', async () => {
@@ -62,7 +73,7 @@ describe('BraintreeScriptLoader', () => {
 
         it('loads the ThreeDSecure library', async () => {
             await braintreeScriptLoader.load3DS();
-            expect(scriptLoader.loadScript).toHaveBeenCalledWith('//js.braintreegateway.com/web/3.15.0/js/three-d-secure.min.js');
+            expect(scriptLoader.loadScript).toHaveBeenCalledWith('//js.braintreegateway.com/web/3.37.0/js/three-d-secure.min.js');
         });
 
         it('returns the ThreeDSecure from the window', async () => {
@@ -87,7 +98,7 @@ describe('BraintreeScriptLoader', () => {
 
         it('loads the data collector library', async () => {
             await braintreeScriptLoader.loadDataCollector();
-            expect(scriptLoader.loadScript).toHaveBeenCalledWith('//js.braintreegateway.com/web/3.15.0/js/data-collector.min.js');
+            expect(scriptLoader.loadScript).toHaveBeenCalledWith('//js.braintreegateway.com/web/3.37.0/js/data-collector.min.js');
         });
 
         it('returns the data collector from the window', async () => {
@@ -112,7 +123,7 @@ describe('BraintreeScriptLoader', () => {
 
         it('loads the VisaCheckout library', async () => {
             await braintreeScriptLoader.loadVisaCheckout();
-            expect(scriptLoader.loadScript).toHaveBeenCalledWith('//js.braintreegateway.com/web/3.15.0/js/visa-checkout.min.js');
+            expect(scriptLoader.loadScript).toHaveBeenCalledWith('//js.braintreegateway.com/web/3.37.0/js/visa-checkout.min.js');
         });
 
         it('returns the VisaCheckout from the window', async () => {
@@ -121,4 +132,45 @@ describe('BraintreeScriptLoader', () => {
         });
     });
 
+    describe('#loadGooglePay()', () => {
+        let googlePayMock: BraintreeModuleCreator<GooglePayBraintreeSDK>;
+
+        beforeEach(() => {
+            googlePayMock = getModuleCreatorMock(getGooglePayMock());
+            scriptLoader.loadScript = jest.fn(() => {
+                if (mockWindow.braintree) {
+                    mockWindow.braintree.googlePayment = googlePayMock;
+                }
+
+                return Promise.resolve();
+            });
+        });
+
+        it('loads the GooglePay library', async () => {
+            await braintreeScriptLoader.loadGooglePayment();
+            expect(scriptLoader.loadScript).toHaveBeenCalledWith('//js.braintreegateway.com/web/3.37.0/js/google-payment.min.js');
+        });
+
+        it('returns the GooglePay from the window', async () => {
+            const googlePay = await braintreeScriptLoader.loadGooglePayment();
+            expect(googlePay).toBe(googlePayMock);
+        });
+
+        it('throws when window is not set', async () => {
+            scriptLoader.loadScript = jest.fn(() => {
+                if (mockWindow.braintree) {
+                    mockWindow.braintree.googlePayment = undefined;
+                    mockWindow.braintree = undefined;
+                }
+
+                return Promise.resolve();
+            });
+
+            try {
+                await braintreeScriptLoader.loadGooglePayment();
+            } catch (error) {
+                expect(error).toBeInstanceOf(StandardError);
+            }
+        });
+    });
 });
--- a/src/payment/strategies/braintree/braintree-sdk-creator.spec.ts
+++ b/src/payment/strategies/braintree/braintree-sdk-creator.spec.ts
@@ -1,6 +1,14 @@
 import { NotInitializedError } from '../../../common/error/errors';
+import { getGooglePayBraintreeMock } from '../googlepay/googlepay.mock';
 
-import { BraintreeClient, BraintreeDataCollector, BraintreeModuleCreator, BraintreeThreeDSecure, BraintreeVisaCheckout } from './braintree';
+import {
+    BraintreeClient,
+    BraintreeDataCollector,
+    BraintreeModuleCreator,
+    BraintreeThreeDSecure,
+    BraintreeVisaCheckout,
+    GooglePayBraintreeSDK,
+} from './braintree';
 import BraintreeScriptLoader from './braintree-script-loader';
 import BraintreeSDKCreator from './braintree-sdk-creator';
 import {
@@ -202,6 +210,45 @@ describe('Braintree SDK Creator', () => {
                 .mockImplementation(() => { throw new Error(errorMessage); });
 
             expect(() => braintreeSDKCreator.getVisaCheckout()).toThrow(errorMessage);
+        });
+    });
+
+    describe('#getGooglePaymentComponent()', () => {
+        let googlePayMock: GooglePayBraintreeSDK;
+        let googlePayCreatorMock: BraintreeModuleCreator<GooglePayBraintreeSDK>;
+        let braintreeSDKCreator: BraintreeSDKCreator;
+
+        beforeEach(() => {
+            googlePayMock = getGooglePayBraintreeMock();
+            googlePayCreatorMock = getModuleCreatorMock(googlePayMock);
+            braintreeScriptLoader.loadGooglePayment = jest.fn().mockReturnValue(Promise.resolve(googlePayCreatorMock));
+            braintreeSDKCreator = new BraintreeSDKCreator(braintreeScriptLoader);
+            jest.spyOn(braintreeSDKCreator, 'getClient').mockReturnValue(Promise.resolve(clientMock));
+        });
+
+        it('returns a promise that resolves to the Google Pay client', async () => {
+            const googlePay = await braintreeSDKCreator.getGooglePaymentComponent();
+
+            expect(googlePayCreatorMock.create).toHaveBeenCalledWith({ client: clientMock });
+            expect(googlePay).toBe(googlePayMock);
+        });
+
+        it('always returns the same instance of the Google Pay client', async () => {
+            const googlePay1 = await braintreeSDKCreator.getGooglePaymentComponent();
+            const googlePay2 = await braintreeSDKCreator.getGooglePaymentComponent();
+
+            expect(googlePay1).toBe(googlePay2);
+            expect(braintreeScriptLoader.loadGooglePayment).toHaveBeenCalledTimes(1);
+            expect(googlePayCreatorMock.create).toHaveBeenCalledTimes(1);
+        });
+
+        it('throws if getting the client throws', async () => {
+            const errorMessage = 'some_error';
+
+            jest.spyOn(braintreeSDKCreator, 'getClient')
+                .mockImplementation(() => { throw new Error(errorMessage); });
+
+            expect(() => braintreeSDKCreator.getGooglePaymentComponent()).toThrow(errorMessage);
         });
     });
 
--- a/src/payment/strategies/googlepay/googlepay-braintree-initializer.ts
+++ b/src/payment/strategies/googlepay/googlepay-braintree-initializer.ts
@@ -1,0 +1,84 @@
+import { Checkout } from '../../../checkout';
+import {
+    MissingDataError,
+    MissingDataErrorType
+} from '../../../common/error/errors';
+import PaymentMethod from '../../payment-method';
+import { BraintreeSDKCreator, GooglePayBraintreeSDK } from '../braintree';
+
+import {
+    GooglePaymentData,
+    GooglePayDataRequestV1,
+    GooglePayInitializer,
+    GooglePayPaymentDataRequestV1,
+    TokenizePayload
+} from './googlepay';
+
+export default class GooglePayBraintreeInitializer implements GooglePayInitializer {
+    private _googlePaymentInstance!: GooglePayBraintreeSDK;
+
+    constructor(
+        private _braintreeSDKCreator: BraintreeSDKCreator
+    ) {}
+
+    initialize(
+        checkout: Checkout,
+        paymentMethod: PaymentMethod,
+        hasShippingAddress: boolean
+    ): Promise<GooglePayPaymentDataRequestV1> {
+        if (!paymentMethod.clientToken) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        this._braintreeSDKCreator.initialize(paymentMethod.clientToken);
+
+        return this._braintreeSDKCreator.getGooglePaymentComponent()
+            .then(googleBraintreePaymentInstance => {
+                this._googlePaymentInstance = googleBraintreePaymentInstance;
+
+                return this._createGooglePayPayload(
+                    checkout,
+                    paymentMethod.initializationData.platformToken,
+                    hasShippingAddress
+                );
+            });
+    }
+
+    teardown(): Promise<void> {
+        return this._braintreeSDKCreator.teardown();
+    }
+
+    parseResponse(paymentData: GooglePaymentData): Promise<TokenizePayload> {
+        return this._googlePaymentInstance.parseResponse(paymentData);
+    }
+
+    private _createGooglePayPayload(
+        checkout: Checkout,
+        platformToken: string,
+        hasShippingAddress: boolean
+    ): GooglePayPaymentDataRequestV1 {
+        if (!platformToken) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        const googlePaymentDataRequest: GooglePayDataRequestV1 = {
+            merchantInfo: {
+                authJwt: platformToken,
+            },
+            transactionInfo: {
+                currencyCode: checkout.cart.currency.code,
+                totalPriceStatus: 'FINAL',
+                totalPrice: checkout.grandTotal.toString(),
+            },
+            cardRequirements: {
+                billingAddressRequired: true,
+                billingAddressFormat: 'FULL',
+            },
+            shippingAddressRequired: !hasShippingAddress,
+            emailRequired: true,
+            phoneNumberRequired: true,
+        };
+
+        return this._googlePaymentInstance.createPaymentDataRequest(googlePaymentDataRequest);
+    }
+}
--- a/src/payment/strategies/googlepay/googlepay-payment-processor.spec.ts
+++ b/src/payment/strategies/googlepay/googlepay-payment-processor.spec.ts
@@ -1,0 +1,370 @@
+import {createRequestSender, RequestSender} from '@bigcommerce/request-sender';
+import { createScriptLoader } from '@bigcommerce/script-loader';
+
+import { BillingAddressActionCreator, BillingAddressRequestSender } from '../../../billing';
+import { getCartState } from '../../../cart/carts.mock';
+import { createCheckoutStore, CheckoutStore } from '../../../checkout';
+import { getCheckoutState } from '../../../checkout/checkouts.mock';
+import {
+    MissingDataError,
+    MissingDataErrorType,
+    NotInitializedError,
+    NotInitializedErrorType,
+    StandardError
+} from '../../../common/error/errors';
+import { getConfigState } from '../../../config/configs.mock';
+import { getCustomerState } from '../../../customer/customers.mock';
+import { PaymentMethodActionCreator} from '../../index';
+import PaymentMethod from '../../payment-method';
+import PaymentMethodRequestSender from '../../payment-method-request-sender';
+import { getGooglePay, getPaymentMethodsState } from '../../payment-methods.mock';
+import { BraintreeScriptLoader, BraintreeSDKCreator } from '../braintree';
+
+import {
+    GooglePaymentsError,
+    GooglePayClient,
+    GooglePayInitializer,
+    GooglePayIsReadyToPayResponse,
+    TokenizePayload
+} from './googlepay';
+import GooglePayBraintreeInitializer from './googlepay-braintree-initializer';
+import GooglePayPaymentProcessor from './googlepay-payment-processor';
+import GooglePayScriptLoader from './googlepay-script-loader';
+import {
+    getGooglePaymentDataMock,
+    getGooglePayAddressMock,
+    getGooglePayPaymentDataRequestMock,
+    getGooglePaySDKMock
+} from './googlepay.mock';
+
+describe('GooglePayPaymentProcessor', () => {
+    let processor: GooglePayPaymentProcessor;
+    let paymentMethodActionCreator: PaymentMethodActionCreator;
+    let googlePayScriptLoader: GooglePayScriptLoader;
+    let store: CheckoutStore;
+    let googlePayInitializer: GooglePayInitializer;
+    let billingAddressActionCreator: BillingAddressActionCreator;
+    let clientMock: GooglePayClient;
+    let requestSender: RequestSender;
+
+    beforeEach(() => {
+        const _requestSender: PaymentMethodRequestSender = new PaymentMethodRequestSender(requestSender);
+        const scriptLoader = createScriptLoader();
+        const braintreeScriptLoader = new BraintreeScriptLoader(scriptLoader);
+        const braintreeSdkCreator = new BraintreeSDKCreator(braintreeScriptLoader);
+        billingAddressActionCreator = new BillingAddressActionCreator(new BillingAddressRequestSender(requestSender));
+
+        store = createCheckoutStore({
+            checkout: getCheckoutState(),
+            customer: getCustomerState(),
+            config: getConfigState(),
+            cart: getCartState(),
+            paymentMethods: getPaymentMethodsState(),
+        });
+        paymentMethodActionCreator = new PaymentMethodActionCreator(_requestSender);
+        googlePayScriptLoader = new GooglePayScriptLoader(createScriptLoader());
+        googlePayInitializer = new GooglePayBraintreeInitializer(braintreeSdkCreator);
+        requestSender = createRequestSender();
+
+        processor = new GooglePayPaymentProcessor(
+            store,
+            paymentMethodActionCreator,
+            googlePayScriptLoader,
+            googlePayInitializer,
+            requestSender,
+            billingAddressActionCreator
+        );
+    });
+
+    it('creates an instance of GooglePayPaymentProcessor', () => {
+        expect(processor).toBeInstanceOf(GooglePayPaymentProcessor);
+    });
+
+    describe('#initialize', () => {
+        beforeEach(() => {
+            const googlePayIsReadyToPayResponse = {
+                result: true,
+            };
+            const googlePaymentDataMock = getGooglePaymentDataMock();
+            const googlePaySDK = getGooglePaySDKMock();
+            clientMock = {
+                isReadyToPay: jest.fn(() => Promise.resolve(googlePayIsReadyToPayResponse)),
+                loadPaymentData: jest.fn((a: any) => Promise.resolve(googlePaymentDataMock)),
+                createButton: jest.fn(() => Promise.resolve(new HTMLElement())),
+            };
+            googlePaySDK.payments.api.PaymentsClient = jest.fn(() => clientMock);
+
+            jest.spyOn(googlePayScriptLoader, 'load').mockReturnValue(Promise.resolve(googlePaySDK));
+            jest.spyOn(googlePayInitializer, 'initialize').mockReturnValue(Promise.resolve());
+        });
+
+        it('initializes processor successfully', async () => {
+            jest.spyOn(store, 'dispatch').mockReturnValue(Promise.resolve(store.getState()));
+            jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod').mockReturnValue(getGooglePay());
+            jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod').mockReturnValue(Promise.resolve(store.getState()));
+
+            await processor.initialize('googlepay');
+
+            expect(googlePayScriptLoader.load).toHaveBeenCalled();
+        });
+
+        it('initializes processor without paymentMethod', async () => {
+            jest.spyOn(store, 'dispatch').mockReturnValue(Promise.resolve(store.getState()));
+            jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod').mockReturnValue(undefined);
+            jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod').mockReturnValue(Promise.resolve(store.getState()));
+
+            try {
+                await processor.initialize('googlepay');
+            } catch (error) {
+                expect(error).toBeInstanceOf(MissingDataError);
+                expect(error).toEqual(new MissingDataError(MissingDataErrorType.MissingPaymentMethod));
+            }
+        });
+
+        it('initializes processor without checkout', async () => {
+            jest.spyOn(store, 'dispatch').mockReturnValue(Promise.resolve(store.getState()));
+            jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod').mockReturnValue(getGooglePay());
+            jest.spyOn(store.getState().checkout, 'getCheckout').mockReturnValue(undefined);
+            jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod').mockReturnValue(Promise.resolve(store.getState()));
+
+            try {
+                await processor.initialize('googlepay');
+            } catch (error) {
+                expect(error).toBeInstanceOf(MissingDataError);
+                expect(error).toEqual(new MissingDataError(MissingDataErrorType.MissingCheckout));
+            }
+        });
+
+        it('initializes processor without testMode', async () => {
+            const googlePayMethod: PaymentMethod = getGooglePay();
+            googlePayMethod.config.testMode = undefined;
+            jest.spyOn(store, 'dispatch').mockReturnValue(Promise.resolve(store.getState()));
+            jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod').mockReturnValue(googlePayMethod);
+            jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod').mockReturnValue(Promise.resolve(store.getState()));
+
+            try {
+                await processor.initialize('googlepay');
+            } catch (error) {
+                expect(error).toBeInstanceOf(Error);
+                expect(error).toEqual(new StandardError(new MissingDataError(MissingDataErrorType.MissingPaymentMethod).message));
+            }
+        });
+    });
+
+    describe('#deinitialize', () => {
+        it('deinitializes processor', async () => {
+            jest.spyOn(googlePayInitializer, 'teardown');
+
+            await processor.deinitialize();
+
+            expect(googlePayInitializer.teardown).toHaveBeenCalled();
+        });
+    });
+
+    describe('#createButton', () => {
+        beforeEach(() => {
+            const googlePayIsReadyToPayResponse = {
+                result: true,
+            };
+            const googlePaymentDataMock = getGooglePaymentDataMock();
+            const googlePaySDK = getGooglePaySDKMock();
+            clientMock = {
+                isReadyToPay: jest.fn(() => Promise.resolve(googlePayIsReadyToPayResponse)),
+                loadPaymentData: jest.fn((a: any) => Promise.resolve(googlePaymentDataMock)),
+                createButton: jest.fn(() => Promise.resolve()),
+            };
+            googlePaySDK.payments.api.PaymentsClient = jest.fn(() => clientMock);
+
+            jest.spyOn(googlePayScriptLoader, 'load').mockReturnValue(Promise.resolve(googlePaySDK));
+            jest.spyOn(googlePayInitializer, 'initialize').mockReturnValue(Promise.resolve());
+            jest.spyOn(store, 'dispatch').mockReturnValue(Promise.resolve(store.getState()));
+            jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod').mockReturnValue(getGooglePay());
+            jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod').mockReturnValue(Promise.resolve(store.getState()));
+        });
+
+        it('creates the html button element', async () => {
+            await processor.initialize('googlepay');
+            await processor.createButton(jest.fn(() => {}));
+
+            expect(clientMock.createButton).toHaveBeenCalled();
+        });
+
+        it('throws an error when googlePaymentsClient is not initialized', async () => {
+            try {
+                await processor.createButton(jest.fn(() => {}));
+            } catch (error) {
+                expect(error).toBeInstanceOf(NotInitializedError);
+                expect(error).toEqual(new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized));
+            }
+        });
+    });
+
+    describe('#handleSuccess', () => {
+        let tokenizePayload: TokenizePayload;
+
+        beforeEach(() => {
+            tokenizePayload = {
+                details: {
+                    cardType: 'MasterCard',
+                    lastFour: '4111',
+                    lastTwo: '11',
+                },
+                type: 'AndroidPayCard',
+                nonce: 'nonce',
+                description: '',
+                binData: {},
+            } as TokenizePayload;
+
+            const googlePayIsReadyToPayResponse = {
+                result: true,
+            };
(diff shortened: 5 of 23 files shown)


cbandam:
@clopezh Please add link to JIRA ticket in the description. Also elaborate a little bit more on the `What?` and `Why?` section.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,210 @@
...
+            )
+            .catch((error: Error) => this._handleError(error));
+    }
+
+    deinitialize(options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        return this._googlePayPaymentProcessor.teardown()
+            .then(() => super.deinitialize(options));
+    }
+
+    private _braintreeGooglePayInitializer(paymentsClient: GooglePayClient, googlePaymentInstance: GooglePayBraintreeSDK, button: Element, googlePayOptions: BraintreeGooglePayPaymentInitializeOptions): void {
```
cbandam: Can we wrap this function definition into multiple lines?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,210 @@
...
+        paymentsClient.isReadyToPay(
+            {
+                allowedPaymentMethods: googlePaymentInstance.createPaymentDataRequest().allowedPaymentMethods,
+            }).then( (response: GooglePayIsReadyToPayResponse) => {
+            if (response.result) {
+                if (button) {
+                    button.addEventListener('click', (event: any) => {
+                        event.preventDefault();
+
+                        const paymentDataRequest: GooglePayBraintreePaymentDataRequest = googlePaymentInstance.createPaymentDataRequest(this._getGooglePayPaymentRequest()) as GooglePayBraintreePaymentDataRequest;
```
cbandam: Wrap statement in multi lines please.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,210 @@
...
+
+                        paymentsClient.loadPaymentData(paymentDataRequest)
+                            .then((paymentData: GooglePaymentData) => {
+                                return this._setExternalCheckoutData(paymentData, googlePaymentInstance, googlePayOptions);
+                            }).catch((err: GooglePaymentsError) => {
+                                this._handleError(new Error(err.statusCode));
+                            });
+                    });
+                } else {
+                    this._handleError(new Error('GooglePay button is not ready'));
```
cbandam: `Google Pay` trademark is two words.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,210 @@
...
+                }
+            } else {
+                this._handleError(new Error('GooglePay is not ready to pay'));
```
cbandam: Google Pay trademark is two words.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,210 @@
...
+    private _getGooglePayPaymentRequest(): GooglePayPaymentDataRequest {
+        const state = this._store.getState();
+        const checkout = state.checkout.getCheckout();
+
+        if (!checkout) {
+            throw new MissingDataError(MissingDataErrorType.MissingCheckout);
+        }
+        const googlePaymentDataRequest: GooglePayPaymentDataRequest = {
+            merchantInfo: {
+                merchantId: 'your-merchant-id-from-google',
```
cbandam: How are we planning to set this `Merchant Id`? Is it going to be a value in the Merchant´s config?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,210 @@
...
+                totalPrice: checkout.grandTotal.toString(),
+            },
+            cardRequirements: {
+                // We recommend collecting billing address information, at minimum
+                // billing postal code, and passing that billing postal code with all
+                // Google Pay transactions as a best practice.
+                billingAddressRequired: true,
+                billingAddressFormat: 'FULL',
+            },
+            shippingAddressRequired: true,
```
cbandam: How about `Digital Products` where there is no shipping address?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,210 @@
...
+            emailRequired: true,
+            phoneNumberRequired: true,
+        };
+
+        return googlePaymentDataRequest;
+    }
+
+    private _getGooglePaymentsClient(google: GooglePaySDK): GooglePayClient {
+        const options: GooglePayPaymentOptions = { environment: 'TEST' };
```
cbandam: We need to get this the `isTestMode` value from the payment´s configuration.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,210 @@
...
+        return new google.payments.api.PaymentsClient(options) as GooglePayClient;
+    }
+
+    private _setExternalCheckoutData(paymentData: GooglePaymentData, googlePaymentInstance: GooglePayBraintreeSDK, googlePayOptions: BraintreeGooglePayPaymentInitializeOptions) {
```
cbandam: Wrap into multiple lines please.

src/payment/strategies/googlepay/googlepay.ts
```diff
@@ -0,0 +1,180 @@
...
+    shippingAddress: GooglePayAddress;
+}
+
+export interface GooglePaymentsError {
+    statusCode: string;
+    statusMessage?: string;
+}
+
+/**
+ * A set of options that are required to initialize the Visa Checkout payment
```
cbandam: `Visa Checkout`?

src/payment/strategies/googlepay/googlepay.ts
```diff
@@ -0,0 +1,180 @@
...
+ * method provided by Braintree.
+ *
+ * If the customer chooses to pay with Visa Checkout, they will be asked to
```
cbandam: `Visa Checkout`? Are these comments legit?

src/payment/strategies/googlepay/googlepay-braintree-initializer.ts
```diff
@@ -0,0 +1,72 @@
...
+        this._braintreeSDKCreator.initialize(paymentMethod.clientToken);
+
+        return this._braintreeSDKCreator.getGooglePaymentComponent()
+            .then(googleBraintreePaymentInstance => {
+                this._googlePaymentInstance = googleBraintreePaymentInstance;
+                return this._createGooglePayPayload(googleBraintreePaymentInstance, checkout, paymentMethod.initializationData.platformToken, hasShippingAddress);
+            }).catch((error: Error) => {
+                throw new StandardError(error.message);
+            });
+    }
```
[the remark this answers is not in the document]
clopezh: 👍

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,347 @@
...
+            this._consignmentActionCreator.updateAddress(mapGooglePayAddressToRequestAddress(shippingAddress))
+        ).then(() => this._store.getState());
+    }
+
+    private _synchronizeBillingAddress(billingAddress: GooglePayAddress): Promise<InternalCheckoutSelectors> {
+        if (!this._methodId) {
+            throw new RemoteCheckoutSynchronizationError();
+        }
+
+        return this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(this._methodId))
```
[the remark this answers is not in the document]
clopezh: Yes because we don't save in any local variable de actual billingAddress. and we need to get the `billingAddress.id` to update (if needed) the billing address.

src/payment/strategies/braintree/braintree-script-loader.spec.ts
```diff
@@ -4,11 +4,12 @@ import { BraintreeClientCreator, BraintreeDataCollector, BraintreeHostWindow, Br
 import BraintreeScriptLoader from './braintree-script-loader';
 import {
     getClientMock,
-    getDataCollectorMock,
+    getDataCollectorMock, getGooglePayMock,
```
vmparra: One line per class

src/payment/strategies/braintree/braintree-script-loader.spec.ts
```diff
@@ -4,11 +4,12 @@ import { BraintreeClientCreator, BraintreeDataCollector, BraintreeHostWindow, Br
...
     getModuleCreatorMock,
     getThreeDSecureMock,
     getVisaCheckoutMock,
 } from './braintree.mock';
+import {GooglePayBraintreeSDK} from "../googlepay/googlepay";
```
vmparra: Please add space between curly braces and class name (applies to whole PR)

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,351 @@
+import {RequestSender} from '@bigcommerce/request-sender';
+import Response from '@bigcommerce/request-sender/lib/response';
+
+import {BillingAddressActionCreator} from '../../../billing';
+import {BillingAddressUpdateRequestBody} from '../../../billing/billing-address';
+import CheckoutStore from '../../../checkout/checkout-store';
+import { CheckoutActionCreator } from '../../../checkout/index';
+import InternalCheckoutSelectors from '../../../checkout/internal-checkout-selectors';
+import {
+    InvalidArgumentError, MissingDataError, MissingDataErrorType, NotInitializedErrorType,
```
vmparra: One line per class (Applies for the entire PR)

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,351 @@
...
+        return this._store.dispatch(this._orderActionCreator.submitOrder({ useStoreCredit }, options))
+            .then(() => this._store.dispatch(this._paymentActionCreator.submitPayment(payment)));
+    }
+
+    private _getGooglePaymentsClient(google: GooglePaySDK, testMode: boolean | undefined): GooglePayClient {
+        let environment: EnvironmentType;
+        testMode = true; // TODO: remove when push this code to final review
+        if (testMode === undefined) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        } else {
```
vmparra: No need for the enclosing `else`

src/payment/strategies/index.ts
```diff
@@ -9,6 +9,7 @@ export { default as SagePayPaymentStrategy } from './sage-pay-payment-strategy';
 export { AfterpayPaymentStrategy } from './afterpay';
 export { AmazonPayPaymentStrategy, AmazonPayPaymentInitializeOptions } from './amazon-pay';
 export { BraintreeCreditCardPaymentStrategy, BraintreePaymentInitializeOptions, BraintreePaypalPaymentStrategy, BraintreeVisaCheckoutPaymentStrategy, BraintreeVisaCheckoutPaymentInitializeOptions } from './braintree';
```
vmparra: One line per class please

src/payment/create-payment-strategy-registry.ts
```diff
@@ -240,7 +252,25 @@ export default function createPaymentStrategyRegistry(
...
             requestSender,
             new ChasePayScriptLoader(getScriptLoader()),
-            new WepayRiskClient(scriptLoader)
+            new WepayRiskClient(scriptLoader))
+    );
+
+    registry.register('googlepay', () =>
+        new GooglePayPaymentStrategy(
+            store,
+            new CheckoutActionCreator(
```
vmparra: You can reuse the one declared on line # 81

src/payment/create-payment-strategy-registry.ts
```diff
@@ -240,7 +252,25 @@ export default function createPaymentStrategyRegistry(
...
+                checkoutRequestSender,
+                new ConfigActionCreator(new ConfigRequestSender(requestSender))
+            ),
+            paymentMethodActionCreator,
+            new PaymentStrategyActionCreator(registry, orderActionCreator),
```
vmparra: You can reuse the one created on line # 82

src/payment/create-payment-strategy-registry.ts
```diff
@@ -240,7 +252,25 @@ export default function createPaymentStrategyRegistry(
...
+            paymentActionCreator,
+            orderActionCreator,
+            new GooglePayScriptLoader(scriptLoader),
+            new GooglePayBraintreeInitializer(braintreeSdkCreator),
+            createRequestSender(),
```
vmparra: you can use class property `requestSender`

src/payment/strategies/index.ts
clopezh: This isn't a change in PR, but I'm gonna change it

src/payment/strategies/googlepay/googlepay-braintree-initializer.ts
```diff
@@ -0,0 +1,85 @@
...
+        if (!paymentMethod.clientToken) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        this._braintreeSDKCreator.initialize(paymentMethod.clientToken);
+
+        return this._braintreeSDKCreator.getGooglePaymentComponent()
+            .then(googleBraintreePaymentInstance => {
+                this._googlePaymentInstance = googleBraintreePaymentInstance;
+                return this._createGooglePayPayload(
```
vmparra: Please add a blank line before the `return`

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,363 @@
...
+
+    private _updateShippingAndBillingAddress(paymentSucessPayload: PaymentSuccessPayload): Promise<void> {
+        return Promise.all([
+                this._synchronizeShippingAddress(paymentSucessPayload.shippingAddress),
+                this._synchronizeBillingAddress(paymentSucessPayload.billingAddress),
+            ]).then(() => Promise.resolve());
+    }
+
+    private _synchronizeShippingAddress(shippingAddress: GooglePayAddress): Promise<InternalCheckoutSelectors | void> {
+
```
vmparra: Please remove this line

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,363 @@
...
+                return this._store.dispatch(
+                    this._billingAddressActionCreator.updateAddress(googlePayAddressMapped)
+                );
+            });
+    }
+
+    @bind
+    private _handleWalletButtonClick(event: Event): void {
+
```
vmparra: Please remove this line

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,363 @@
...
+        event.preventDefault();
+
+        this._displayWallet();
+
```
vmparra: Please remove this line

src/payment/strategies/googlepay/googlepay.ts
```diff
@@ -0,0 +1,216 @@
+import { AddressRequestBody } from '../../../address/address';
+import { BillingAddressUpdateRequestBody } from '../../../billing/billing-address';
+import Checkout from '../../../checkout/checkout';
+import PaymentMethod from '../../payment-method';
+import {
```
vmparra: No need to break the import in multiple lines (this applies only when line is too long)

src/payment/strategies/googlepay/googlepay.ts
```diff
@@ -0,0 +1,216 @@
...
+}
+
+/**
+ * A set of options that are required to initialize the GooglePay payment method
+ *
+ * If the customer chooses to pay with GooglePay, they will be asked to
+ * enter their payment details via a modal. You can hook into events emitted by
+ * the modal by providing the callbacks listed below.
+ */
+export interface GooglePayPaymentInitializeOptions {
```
vmparra: Please move this interface to it's own file

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+    GooglePayIsReadyToPayResponse,
+    GooglePayPaymentDataRequestV1,
+    GooglePayPaymentOptions, GooglePaySDK,
+    PaymentSuccessPayload,
+    TokenizePayload
+} from './googlepay';
+import GooglePayPaymentInitializeOptions from './googlepay-initialize-options';
+import GooglePayScriptLoader from './googlepay-script-loader';
+
+export default class GooglePayPaymentStrategy extends PaymentStrategy {
```
icatalina: Since most of the logic in this class is going to be reused in the `CustomerStrategy` and the `CheckoutButtonStrategy`, it might be a good idea to split this class in 2. A payment processor and the strategy.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+
+    private _createOrder(payment: Payment, useStoreCredit?: boolean, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        return this._store.dispatch(this._orderActionCreator.submitOrder({ useStoreCredit }, options))
+            .then(() => this._store.dispatch(this._paymentActionCreator.submitPayment(payment)));
+    }
+
+    private _getGooglePaymentsClient(google: GooglePaySDK, testMode: boolean | undefined): GooglePayClient {
+        let environment: EnvironmentType;
+        testMode = true; // TODO: remove when push this code to final review
+        if (testMode === undefined) {
```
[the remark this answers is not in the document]
clopezh: No actually it's a validation in case of testMode isn't received in `_getGooglePaymentsClient()` method; it can be true, false or undefined

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+                throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+            }
+
+            if (!this._googlePaymentsClient && !this._googlePaymentDataRequest) {
+                throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+            }
+
+            this._googlePaymentsClient.isReadyToPay({
+                allowedPaymentMethods: this._googlePaymentDataRequest.allowedPaymentMethods,
+            }).then( (response: GooglePayIsReadyToPayResponse) => {
```
davidchin: You can omit `GooglePayIsReadyToPayResponse` because when you define the interface for `GooglePaymentsClient`, you already specified that `isReadyToPay` will return `Promise<GooglePayIsReadyToPayResponse>`.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+                if (response) {
```
davidchin: Can `response` ever be null? It seems `response` is not an optional parameter so you shouldn't need a guard statement here.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+                        this._googlePaymentDataRequest = googlePayPaymentDataRequest;
+                    })
+                    .catch((error: Error) => {
+                        this._handleError(error);
+                    });
+            });
+    }
+
+    private _displayWallet(): Promise<InternalCheckoutSelectors> {
+        return new Promise((resolve, reject) => {
```
davidchin: Is there a reason why this promise never gets resolved? I can see `reject` gets called at the bottom of this method. But I can't see `resolve` getting called anywhere. Also, is there a reason why you need to call `new Promise()`? It seems `isReadyToPay` already returns a promise so you should be able to just chain it with other asynchronous calls and return the resulting promise.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+                throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+            }
+
+            this._googlePaymentsClient.isReadyToPay({
+                allowedPaymentMethods: this._googlePaymentDataRequest.allowedPaymentMethods,
+            }).then( (response: GooglePayIsReadyToPayResponse) => {
+                if (response) {
+                    this._googlePaymentsClient.loadPaymentData(this._googlePaymentDataRequest)
+                        .then((paymentData: GooglePaymentData) => {
+                            return this._setExternalCheckoutData(paymentData);
```
davidchin: You can probably flatten the promise chain instead of having it nested.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        if (!testMode) {
+            environment = 'PRODUCTION';
+        } else {
+            environment = 'TEST';
+        }
+
+        const options: GooglePayPaymentOptions = { environment };
```
davidchin: Is there a reason why you need this temporary variable? It can be abbreviated as `return new google.payments.api.PaymentsClient({ environment })`.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+
+        return new google.payments.api.PaymentsClient(options);
+    }
+
+    private _setExternalCheckoutData(paymentData: GooglePaymentData): Promise<void> {
+        return this._googlePayInitializer.parseResponse(paymentData)
+            .then((tokenizePayload: TokenizePayload) => {
+                const paymentSuccessPayload: PaymentSuccessPayload = {
```
davidchin: Is there a reason why you need to declare `PaymentSuccessPayload`? The type can be inferred and if you have a missing property, the compiler will complain when you call `_updateShippingAndBillingAddress` / `_paymentInstrumentSelected` so it's not really necessary to declare it again.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    private _paymentInstrumentSelected(paymentSuccessPayload: PaymentSuccessPayload): Promise<InternalCheckoutSelectors> {
+        if (!this._paymentMethod) {
+            throw new Error('Payment method not initialized');
+        }
+
+        const { id: methodId } = this._paymentMethod;
+
+        return this._store.dispatch(this._paymentStrategyActionCreator.widgetInteraction(() => {
```
davidchin: You need to wrap the `widgetInteraction` around all asynchronous calls, so including `_updateShippingAndBillingAddress`. At the moment, it sits outside. Therefore, it will emit the wrong pending status.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+                    return this._paymentInstrumentSelected(paymentSuccessPayload)
+                        .then(() => onPaymentSelect())
+                        .catch(error => onError(error));
+                });
+            });
+    }
+
+    private _paymentInstrumentSelected(paymentSuccessPayload: PaymentSuccessPayload): Promise<InternalCheckoutSelectors> {
+        if (!this._paymentMethod) {
+            throw new Error('Payment method not initialized');
```
davidchin: We have a special error type of these kind of errors. It is called `NotInitializedError`.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+        }
+
+        const { id: methodId } = this._paymentMethod;
+
+        return this._store.dispatch(this._paymentStrategyActionCreator.widgetInteraction(() => {
+            return this._postForm(paymentSuccessPayload)
+                .then(() => Promise.all([
+                    this._store.dispatch(this._checkoutActionCreator.loadCurrentCheckout()),
```
davidchin: I think you should move these calls inside the method that makes the `/checkout.php` call. Normally, we don't have to make additional calls if we just use the official storefront API endpoints, which automatically give you back the latest data. However, because we're using `/checkout.php` as a hack, we need to make additional calls to hydrate the state.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+                    this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId)),
+                ]));
+        }, { methodId }), { queueId: 'widgetInteraction' });
+    }
+
+    private _postForm(paymentData: PaymentSuccessPayload): Promise<Response<any>> {
```
davidchin: Do you mean `postPaymentData`?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+
+    private _getCardInformation(cardInformation: { cardType: string, lastFour: string }) {
+        return {
+            type: cardInformation.cardType,
+            number: cardInformation.lastFour,
+        };
+    }
+
+    private _handleError(error: Error): never {
+        throw new StandardError(error.message);
```
davidchin: We could probably just throw the error directly instead of having this extra method.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    private _getPayment() {
```
davidchin: Can you define the return type for this method?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+                    throw new MissingDataError(MissingDataErrorType.MissingPayment);
+                }
+
+                const paymentData = {
+                    method: this._methodId,
+                    nonce: paymentMethod.initializationData.nonce,
+                    cardInformation: paymentMethod.initializationData.card_information,
+                };
+
+                this._paymentMethod = paymentMethod;
```
davidchin: We should try to avoid setting anything inside a getter.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+                return {
+                    methodId: this._methodId,
+                    paymentData,
+                };
+            });
+    }
+
+    private _updateShippingAndBillingAddress(paymentSucessPayload: PaymentSuccessPayload): Promise<void> {
+        return Promise.all([
+                this._synchronizeShippingAddress(paymentSucessPayload.shippingAddress),
```
davidchin: Can you please fix the indentation?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+
+        return this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(this._methodId))
+            .then(state => {
+                const remoteBillingAddress = state.billingAddress.getBillingAddress();
+                let googlePayAddressMapped: BillingAddressUpdateRequestBody;
+
+                if (!remoteBillingAddress) {
+                    googlePayAddressMapped = mapGooglePayAddressToRequestAddress(billingAddress) as BillingAddressUpdateRequestBody;
+                } else {
+                    googlePayAddressMapped = mapGooglePayAddressToRequestAddress(billingAddress, remoteBillingAddress.id) as BillingAddressUpdateRequestBody;
```
davidchin: Can we avoid using `as`?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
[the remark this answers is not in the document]
davidchin: The payment method object shouldn't contain any information about the billing address? So why do you need to load it the payment method?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,371 @@
...
+                this._synchronizeBillingAddress(paymentSucessPayload.billingAddress),
+            ]).then(() => Promise.resolve());
+    }
+
+    private _synchronizeShippingAddress(shippingAddress: GooglePayAddress): Promise<InternalCheckoutSelectors | void> {
+        if (!this._methodId) {
+            throw new RemoteCheckoutSynchronizationError();
+        }
+
+        if (!shippingAddress) {
```
davidchin: You shouldn't need a guard statement here because `shippingAddress` is not an optional parameter.

src/payment/strategies/googlepay/googlepay.ts
```diff
@@ -0,0 +1,201 @@
...
+}
+
+export default function mapGooglePayAddressToRequestAddress(address: GooglePayAddress, id?: string): AddressRequestBody | BillingAddressUpdateRequestBody {
+    return {
+        id,
+        firstName: address.name.split(' ').slice(0, -1).join(' '),
+        lastName: address.name.split(' ').slice(-1).join(' '),
+        company: address.companyName,
+        address1: address.address1,
+        address2: address.address2 + address.address3 + address.address4 + address.address5,
```
davidchin: Is it okay to just concatenate the addresses like this? Do you need a separator?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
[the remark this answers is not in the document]
clopezh: Actually payment method object contains `state.billingAddress.getBillingAddress()` and I used to get the `remoteBillingAddress.id` to update the actual billingAddress

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
clopezh: Actually is a validation to know if GooglePay is ready to open the wallet or not, I'll add the else sentence and throw an error to inform that GooglePay is not ready

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
clopezh: I'm gonna change to this:

src/payment/strategies/googlepay/googlepay.ts
clopezh: I concatenate because google pay returns five lines of address (unless I didn't added), and to map with BC Address, I pass address line 1 to address1, and the others address lines as address 2

src/payment/strategies/googlepay/googlepay-payment-strategy.spec.ts
```diff
@@ -0,0 +1,341 @@
...
+            customer: getCustomerState(),
+            config: getConfigState(),
+            cart: getCartState(),
+            paymentMethods: getPaymentMethodsState(),
+        });
+
+        const checkoutRequestSender = new CheckoutRequestSender(createRequestSender());
+        const configRequestSender = new ConfigRequestSender(createRequestSender());
+        const configActionCreator = new ConfigActionCreator(configRequestSender);
+        const _requestSender: PaymentMethodRequestSender = new PaymentMethodRequestSender(requestSender);
```
vmparra: please remove the underscore (due naming conventions this is for private instance members)

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,192 @@
+import { RequestSender } from '@bigcommerce/request-sender';
```
vmparra: Use this

```
import { RequestSender } from '@bigcommerce/request-sender';

import { PaymentStrategy } from '../';
import {
    Payment,
    PaymentActionCreator,
    PaymentInitializeOptions,
    PaymentMethodActionCreator,
    PaymentRequestOptions,
    PaymentStrategyActionCreator
} from '../..';
import { CheckoutActionCreator, CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
import { NotInitializedError } from '../../../common/error/errors';
import {
    InvalidArgumentError,
    MissingDataError,
    MissingDataErrorType,
    NotInitializedErrorType,
} from '../../../common/error/errors';
import { toFormUrlEncoded } from '../../../common/http-request';
import { bindDecorator as bind } from '../../../common/utility';
import { OrderActionCreator, OrderRequestBody } from '../../../order';

import { GooglePayPaymentInitializeOptions, GooglePayPaymentProcessor } from './';
import {
    GooglePaymentData,
    GooglePayAddress,
    GooglePayInitializer,
    PaymentMethodData,
    TokenizePayload,
} from './googlepay';
```

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,192 @@
...
+            },
+            body: toFormUrlEncoded({
+                payment_type: postPaymentData.type,
+                nonce: postPaymentData.nonce,
+                provider: this._methodId,
+                action: 'set_external_checkout',
+                card_information: this._getCardInformation(cardInformation),
+            }),
+        }).then(() => {
+            if (!this._methodId) {
```
vmparra: Do we need to validate this?

```postForm``` it's being called from ```_paymentInstrumentSelected```which validates the same

File: src/payment/strategies/googlepay/googlepay-script-loader.spec.ts [hunk 49]
```diff
@@ -0,0 +1,54 @@
+import { ScriptLoader } from '@bigcommerce/script-loader';
```
vmparra: Use this

```
import { ScriptLoader } from '@bigcommerce/script-loader';

import { StandardError } from '../../../common/error/errors';

import { GooglePayScriptLoader } from './';
import { GooglePayHostWindow, GooglePaySDK } from './googlepay';
import { getGooglePaySDKMock } from './googlepay.mock';```

File: src/payment/strategies/googlepay/googlepay-payment-strategy.spec.ts [hunk 50]
```diff
@@ -0,0 +1,341 @@
+import {createRequestSender, RequestSender} from '@bigcommerce/request-sender';
```
vmparra: use this

````
import { createRequestSender, RequestSender } from '@bigcommerce/request-sender';
import { createScriptLoader } from '@bigcommerce/script-loader';

import { PaymentInitializeOptions, PaymentMethod, PaymentMethodRequestSender, PaymentRequestSender } from '../..';
import { BillingAddressActionCreator, BillingAddressRequestSender } from '../../../billing';
import { getCartState } from '../../../cart/carts.mock';
import {
    createCheckoutStore,
    CheckoutActionCreator,
    CheckoutRequestSender,
    CheckoutStore,
    CheckoutValidator
} from '../../../checkout';
import { getCheckoutState } from '../../../checkout/checkouts.mock';
import { InvalidArgumentError, MissingDataError, MissingDataErrorType } from '../../../common/error/errors';
import { ConfigActionCreator, ConfigRequestSender } from '../../../config';
import { getConfigState } from '../../../config/configs.mock';
import { getCustomerState } from '../../../customer/customers.mock';
import { OrderActionCreator } from '../../../order';
import {
    createPaymentClient,
    createPaymentStrategyRegistry,
    PaymentActionCreator,
    PaymentMethodActionCreator,
    PaymentStrategyActionCreator
} from '../../../payment';
import { getGooglePay, getPaymentMethodsState } from '../../payment-methods.mock';
import { BraintreeScriptLoader, BraintreeSDKCreator } from '../braintree';

import {
    GooglePayBraintreeInitializer,
    GooglePayPaymentProcessor,
    GooglePayPaymentStrategy,
    GooglePayScriptLoader
} from './';
import { GooglePaymentData, GooglePayInitializer, TokenizePayload } from './googlepay';
import { getGoogleOrderRequestBody, getGooglePaymentDataPayload } from './googlepay.mock';

File: src/payment/strategies/googlepay/googlepay-script-loader.ts [hunk 51]
```diff
@@ -0,0 +1,27 @@
+import { ScriptLoader } from '@bigcommerce/script-loader';
+
+import { StandardError } from '../../../common/error/errors/index';
+
+import {
```
vmparra: No need to break it in separate lines

src/payment/strategies/googlepay/googlepay.ts
```diff
@@ -0,0 +1,185 @@
+import { Checkout } from '../../../checkout';
+import PaymentMethod from '../../payment-method';
```
vmparra: When available import from index

src/payment/strategies/braintree/braintree-script-loader.spec.ts
```diff
@@ -1,10 +1,13 @@
 import { ScriptLoader } from '@bigcommerce/script-loader';
 
+import StandardError from '../../../common/error/errors/standard-error';
```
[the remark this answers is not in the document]
clopezh: Actually David Chin told me that is better to import from the top-level not from the index

src/payment/strategies/braintree/braintree-sdk-creator.spec.ts
```diff
@@ -1,4 +1,6 @@
 import { NotInitializedError } from '../../../common/error/errors';
+import { GooglePayBraintreeSDK } from '../googlepay';
+import { getGooglePayBraintreeMock } from '../googlepay/googlepay.mock';
 
 import { BraintreeClient, BraintreeDataCollector, BraintreeModuleCreator, BraintreeThreeDSecure, BraintreeVisaCheckout } from './braintree';
 import BraintreeScriptLoader from './braintree-script-loader';
```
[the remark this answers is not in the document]
clopezh: BraintreeScriptLoader is not in the module file so we don't import from top-level, and is a file that we didn't create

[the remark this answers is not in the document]
clopezh: and actually is not a line edited, I guess if is in master is because is correct there

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,182 @@
...
+            this._walletButton = walletButton;
+            this._walletButton.addEventListener('click', this._handleWalletButtonClick);
+        }
+
+        return this._googlePayPaymentProcessor.initialize(this._methodId)
+            .then(() => super.initialize(options));
+    }
+
+    deinitialize(options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+
```
vmparra: Please remove this line

src/payment/strategies/googlepay/googlepay.ts
```diff
@@ -0,0 +1,185 @@
...
+    cardRequirements: {
+        billingAddressRequired: boolean,
+        billingAddressFormat: AddressFormat,
+    };
+    emailRequired: boolean;
+    phoneNumberRequired: boolean;
+    shippingAddressRequired: boolean;
+}
+
+export interface GooglePayPaymentDataRequestV1 {
```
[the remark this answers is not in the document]
clopezh: Actually the Braintree implementation uses Google Pay v1, and that's the reason to use V1 to create the object, all of these is in initializer which is specifically for Braintree, and the we gonna need to add another object to support Object from Stripe

src/payment/strategies/googlepay/googlepay-braintree-initializer.ts
```diff
@@ -0,0 +1,87 @@
...
+                                    checkout: Checkout,
+                                    platformToken: string,
+                                    hasShippingAddress: boolean
+    ): GooglePayPaymentDataRequestV1 {
+        if (!platformToken) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        const googlePaymentDataRequest: GooglePayDataRequestV1 = {
+            merchantInfo: {
```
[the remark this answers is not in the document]
clopezh: We only need to populate authJwt since that token contains all of these information

src/payment/strategies/googlepay/googlepay-payment-strategy.spec.ts
```diff
@@ -0,0 +1,338 @@
...
+import {
+    GooglePayBraintreeInitializer,
+    GooglePayPaymentProcessor,
+    GooglePayPaymentStrategy,
+    GooglePayScriptLoader
+} from './';
+import { GooglePaymentData, GooglePayInitializer, TokenizePayload } from './googlepay';
+import { getGoogleOrderRequestBody, getGooglePaymentDataPayload } from './googlepay.mock';
+
+
```
vmparra: Remove this line please

src/payment/strategies/googlepay/googlepay.ts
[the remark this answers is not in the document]
cbandam: Can we rename the interface to include the word `Braintree` in the name since this Request Payload is specific for Braintree?

src/payment/strategies/googlepay/googlepay-payment-processor.ts
```diff
@@ -0,0 +1,159 @@
...
+
+                if (!paymentMethod) {
+                    throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+                }
+
+                if (!checkout) {
+                    throw new MissingDataError(MissingDataErrorType.MissingCheckout);
+                }
+
+                const testMode = paymentMethod.config.testMode;
```
vmparra: 🍹  ```const { testMode } = paymentMethod.config;```

src/payment/strategies/googlepay/googlepay.ts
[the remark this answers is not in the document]
cbandam: Also we need to make sure that we use `apiVersion: 2`

src/payment/strategies/googlepay/googlepay.ts
```diff
@@ -0,0 +1,185 @@
...
+    allowedPaymentMethods: string[];
+    apiVersion: number;
```
cbandam: We need to make sure that we use `apiVersion: 2` so that we support JWT platform solution.

src/payment/strategies/googlepay/googlepay.ts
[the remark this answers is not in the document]
clopezh: This will be work on https://jira.bigcommerce.com/browse/INT-982

src/payment/strategies/googlepay/googlepay-braintree-initializer.ts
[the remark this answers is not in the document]
clopezh: This will be work on https://jira.bigcommerce.com/browse/INT-982

src/payment/strategies/googlepay/googlepay.ts
clopezh: This will be work on https://jira.bigcommerce.com/browse/INT-982

src/payment/strategies/googlepay/googlepay-braintree-initializer.ts
```diff
@@ -0,0 +1,87 @@
...
+
+        this._braintreeSDKCreator.initialize(paymentMethod.clientToken);
+
+        return this._braintreeSDKCreator.getGooglePaymentComponent()
+            .then(googleBraintreePaymentInstance => {
+                this._googlePaymentInstance = googleBraintreePaymentInstance;
+
+                return this._createGooglePayPayload(
+                    checkout,
+                    paymentMethod.initializationData.platformToken,
```
metalvegetarianoprogresivo: Can you add a validation for `paymentMethod.initializationData` since you are using it here?

Sure-Round:
@davidchin @icatalina ♻️  This PR is now ready for the Sydney team to review

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,181 @@
...
+        if (!this._methodId) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+        }
+
+        return this._store.dispatch(this._paymentStrategyActionCreator.widgetInteraction(() => {
+            return this._postForm(tokenizePayload, billingAddress);
+        }, { methodId: this._methodId }), { queueId: 'widgetInteraction' });
+    }
+
+    private _postForm(postPaymentData: TokenizePayload, billingAddress: GooglePayAddress): Promise<InternalCheckoutSelectors> {
```
davidchin: I assume `GooglePayPaymentProcessor` is responsible for processing payment data, i.e.: perform any necessary pre-processing and pass the payment data to GooglePay. That's actually what you're doing here in this method. So my question is, why is this method not moved under the payment processor class?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,181 @@
...
+    private _walletButton?: HTMLElement;
+
+    constructor(
+        store: CheckoutStore,
+        private _checkoutActionCreator: CheckoutActionCreator,
+        private _paymentMethodActionCreator: PaymentMethodActionCreator,
+        private _paymentStrategyActionCreator: PaymentStrategyActionCreator,
+        private _paymentActionCreator: PaymentActionCreator,
+        private _orderActionCreator: OrderActionCreator,
+        private _googlePayInitializer: GooglePayInitializer,
```
davidchin: Why do you need to inject `GooglePayInitializer` here? You don't need to because you already have `GooglePayPaymentProcessor`, which acts as a proxy for the former.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    deinitialize(options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        if (this._walletButton) {
+            this._walletButton.removeEventListener('click', this._handleWalletButtonClick);
+        }
+
+        this._walletButton = undefined;
+
+        return Promise.all([
+            this._googlePayInitializer.teardown(),
```
davidchin: Is there a reason why you need to call `teardown` directly? I thought `GooglePayPaymentProcessor#deinitialize ` should handle it already.

src/payment/strategies/googlepay/googlepay-payment-processor.ts
```diff
@@ -0,0 +1,159 @@
...
+    GooglePayClient,
+    GooglePayInitializer,
+    GooglePayPaymentDataRequestV1,
+    GooglePayScriptLoader,
+    GooglePaySDK,
+    TokenizePayload
+} from './';
+
+export default class GooglePayPaymentProcessor {
+    private _googlePaymentsClient!: GooglePayClient;
```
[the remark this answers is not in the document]
clopezh: Yes, if I remove the `!` the variable needs to be initialized, and it is initialized when the strategy initializes.

src/payment/strategies/googlepay/googlepay-payment-processor.ts
```diff
@@ -0,0 +1,159 @@
...
+        return this._googlePaymentsClient.isReadyToPay({
+            allowedPaymentMethods: this._googlePaymentDataRequest.allowedPaymentMethods,
+        }).then( response => {
+            if (response.result) {
+                return this._googlePaymentsClient.loadPaymentData(this._googlePaymentDataRequest)
+                    .then(paymentData => paymentData)
+                    .catch((err: GooglePaymentsError) => {
+                        throw new Error(err.statusCode);
+                    });
+            } else {
```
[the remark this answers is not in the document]
clopezh: Is needed because the method `isReadyTo Pay` is to validate if GooglePay is ready, and it returns a boolean value, if `response.result` is false, we need to throw an exception to notify that, otherwise we try to load the widget and maybe in this part the widget can crash

src/payment/strategies/googlepay/googlepay-payment-processor.ts
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        return this._googlePaymentsClient.isReadyToPay({
+            allowedPaymentMethods: this._googlePaymentDataRequest.allowedPaymentMethods,
+        }).then( response => {
+            if (response.result) {
+                return this._googlePaymentsClient.loadPaymentData(this._googlePaymentDataRequest)
+                    .then(paymentData => paymentData)
+                    .catch((err: GooglePaymentsError) => {
+                        throw new Error(err.statusCode);
```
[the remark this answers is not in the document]
clopezh: GooglePay sends the message in `statusCode` property, that's the reason I returned that value as an error, the `GooglePaymentsError` is only to parse the response

src/payment/strategies/googlepay/googlepay-payment-processor.ts
```diff
@@ -0,0 +1,159 @@
...
+                return Promise.all([
+                    this._googlePayScriptLoader.load(),
+                    this._googlePayInitializer.initialize(checkout, paymentMethod, hasShippingAddress),
+                ])
+                    .then(([googlePay, googlePayPaymentDataRequest]) => {
+                        this._googlePaymentsClient = this._getGooglePaymentsClient(googlePay, testMode);
+                        this._googlePaymentDataRequest = googlePayPaymentDataRequest;
+                    })
+                    .catch((error: Error) => {
+                        throw new StandardError(error.message);
```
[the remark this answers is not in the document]
clopezh: In this case is valid the comment, I'm gonna `throw error`

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,181 @@
...
+import {
+    GooglePaymentData,
+    GooglePayAddress,
+    GooglePayInitializer,
+    PaymentMethodData,
+    TokenizePayload,
+} from './googlepay';
+
+export default class GooglePayPaymentStrategy extends PaymentStrategy {
+    private _googlePayOptions!: GooglePayPaymentInitializeOptions;
```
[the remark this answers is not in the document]
clopezh: The same resposne as previous :

clopezh:
@davidchin ♻️ All the comments are now addressed, Could you please check if they are ok?

Sure-Round:
@davidchin @icatalina ♻️ This PR is ready for another pass.

src/payment/strategies/googlepay/googlepay-payment-processor.ts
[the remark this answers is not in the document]
Sure-Round: @clopezh I think the argument is if all the `else` is doing is `throwing` the exception then why not just let that happen on it's own. e.g., if you look at the below happy path will return the value and un-happy path will throw the exception, it's just not bundled in an `else`


```
            if (response.result) {
                return this._googlePaymentsClient.loadPaymentData(this._googlePaymentDataRequest);
            }

```

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
Sure-Round: @clopezh Can you acknowledge this comment and show the change you made to put it in place.

src/payment/strategies/googlepay/googlepay-payment-processor.spec.ts
```diff
@@ -0,0 +1,368 @@
...
+            const googlePaymentDataMock = getGooglePaymentDataMock();
+            const googlePaySDK = getGooglePaySDKMock();
+            clientMock = {
+                isReadyToPay: jest.fn(() => Promise.resolve(googlePayIsReadyToPayResponse)),
+                loadPaymentData: jest.fn((a: any) => Promise.resolve(googlePaymentDataMock)),
+                createButton: jest.fn(() => Promise.resolve(new HTMLElement())),
+            };
+            googlePaySDK.payments.api.PaymentsClient = jest.fn(() => clientMock);
+
+            spyOn(googlePayScriptLoader, 'load').and.returnValue(Promise.resolve(googlePaySDK));
```
davidchin: Please see my comment regarding the use of `spyOn` vs `jest.spyOn`.

clopezh:
Hello @davidchin & @icatalina as we already talked about the last comments, I already addressed the last ones, Could you please take a look and if possible approve it? ♻️ 

CC @cbandam @vmparra

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -0,0 +1,178 @@
...
+
+        const {
+            onError = () => {},
+            onPaymentSelect = () => {},
+        } = this._googlePayOptions;
+
+        return Promise.resolve(this._getPayment())
+            .then(payment => {
+                if (!payment.paymentData.nonce || !payment.paymentData.cardInformation) {
+                    return this._googlePayPaymentProcessor.displayWallet()
```
davidchin: This block of code seems to be duplicated from `_handleWalletButtonClick`. We need to find a way to keep it DRY. It's not critical, so we could just leave a comment here `TODO: Find a way to share the code with _handleWalletButtonClick 
 method`.

PR #410, opened 2018-09-19
feat(order): CHECKOUT-3563 Add External Source param when creating order

RobinOng:
## What?
As per above.

## Why?
Identify if the request is from external partners

## Testing / Proof
Unit test and manual testing.
screen shot 2018-09-20 at 5 57 39 pm

@bigcommerce/checkout

### DIFF
--- a/src/checkout/create-checkout-service.ts
+++ b/src/checkout/create-checkout-service.ts
@@ -3,7 +3,7 @@ import { createRequestSender } from '@bigcommerce/request-sender';
 import { BillingAddressActionCreator, BillingAddressRequestSender } from '../billing';
 import { getDefaultLogger } from '../common/log';
 import { getEnvironment } from '../common/utility';
-import { ConfigActionCreator, ConfigRequestSender } from '../config';
+import { ConfigActionCreator, ConfigRequestSender, ConfigState } from '../config';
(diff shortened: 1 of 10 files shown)


src/order/order-action-creator.ts
```diff
@@ -115,19 +116,21 @@ export default class OrderActionCreator {
...
                 ...order,
                 customerMessage,
+                externalSource,
             };
         }
 
         return {
             ...order,
             customerMessage,
+            externalSource,
```
capsula4: should there be a test for this new use case?

src/checkout/create-checkout-service.ts
```diff
@@ -43,9 +43,16 @@ export default function createCheckoutService(options?: CheckoutServiceOptions):
         getDefaultLogger().warn('Note that the development build is not optimized. To create a production build, set process.env.NODE_ENV to `production`.');
     }
 
+    const config: ConfigState = {
+        meta: {
+            externalSource: options && options.externalSource,
```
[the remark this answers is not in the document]
RobinOng: Don't think we should set a default value in the sdk because custom checkout that did not set a value will be getting this default value unintentionally.

src/order/order-action-creator.ts
capsula4: as discussed, this has already been covered by the updated mocks.

PR #411, opened 2018-09-20
Currency utilities 

capsula4:
Adds Currency utilities to Checkout SDK

### DIFF
--- a/src/currency/currency-formatter.ts
+++ b/src/currency/currency-formatter.ts
@@ -1,0 +1,72 @@
+import { CurrencyConfig } from './currency';
+
+export default class CurrencyFormatter {
+    private _decimalPlaces: number;
+    private _symbolLocation: string;
+    private _symbol: string;
+    private _thousandsSeparator: string;
+    private _decimalSeparator: string;
+
+    constructor(
+        currencySettings?: CurrencyConfig | any
+    ) {
+        if (!currencySettings) {
+            throw new Error('Currency settings missing');
+        }
+
+        const {
+            decimalPlaces,
+            symbolLocation,
+            symbol,
+            thousandsSeparator,
+            decimalSeparator,
+        } = currencySettings;
+
+        if (
+            typeof symbolLocation !== 'string' ||
+            typeof symbol !== 'string' ||
+            typeof thousandsSeparator !== 'string' ||
+            typeof decimalSeparator !== 'string' ||
+            typeof decimalPlaces !== 'string'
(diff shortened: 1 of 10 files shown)


webpack.config.js
```diff
@@ -10,6 +10,7 @@ const assetConfig = {
     entry: {
         'checkout-sdk': './src/index.ts',
         'checkout-button': './src/checkout-button.ts',
+        'currency-service': './src/currency-service.ts',
```
davidchin: Do we need to spit out a separate file? I think it's fine exporting it as a part of the main package?

src/currency/currency-formatter.ts
```diff
@@ -0,0 +1,72 @@
...
+
+        return amount < 0 ?
+            `-${formattedCurrency}` :
+            formattedCurrency;
+    }
+
+    private _formatNumber(amount: number): string {
+        const positiveAmount = Math.abs(amount);
+        const [ integerAmount, decimalAmount = '' ] = positiveAmount.toString().split('.');
+        const decimalPadding = Array(this._decimalPlaces).fill('0').join('');
```
davidchin: `Array.prototype.fill` is not supported in IE11 😞. Is it possible to find a different way to do the same thing?

src/currency/currency-formatter.ts
```diff
@@ -0,0 +1,72 @@
...
+        this._decimalPlaces = parseInt(decimalPlaces, 10);
+        this._symbolLocation = symbolLocation;
+        this._symbol = symbol;
+        this._thousandsSeparator = thousandsSeparator;
+        this._decimalSeparator = decimalSeparator;
+    }
+
+    format(amount?: number): string {
+        if (typeof amount !== 'number') {
+            throw new Error('Invalid amount provided');
```
davidchin: Maybe throw `InvalidArgument` instead?

src/currency/currency-formatter.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export default class CurrencyFormatter {
+    private _decimalPlaces: number;
+    private _symbolLocation: string;
+    private _symbol: string;
+    private _thousandsSeparator: string;
+    private _decimalSeparator: string;
+
+    constructor(
+        currencySettings?: CurrencyConfig | any
```
davidchin: `currencySettings` seems to be a required property, as the class will throw an error if it is missing. If you want to make the property optional, you should provide some default values.

PR #414, opened 2018-09-21
feat(payments): INT-685 Correctly set up CCAvenue MARS return URL

kevinziegler:
## What?
Add a check to pull correct data for CCAvenue MARS offsite payments.

## Why?
Without this BigPay does not receive a value for `returnUrl` when processing the offsite initialization request.

## Testing / Proof
Manual testing to verify successful payment initialization.

### DIFF
--- a/src/payment/strategies/offsite-payment-strategy.ts
+++ b/src/payment/strategies/offsite-payment-strategy.ts
@@ -1,5 +1,5 @@
 import { CheckoutStore, InternalCheckoutSelectors } from '../../checkout';
-import { OrderActionCreator, OrderRequestBody } from '../../order';
+import { OrderActionCreator, OrderPaymentRequestBody, OrderRequestBody } from '../../order';
 import { PaymentArgumentInvalidError } from '../errors';
 import PaymentActionCreator from '../payment-action-creator';
 import { PaymentRequestOptions } from '../payment-request-options';
@@ -19,7 +19,7 @@ export default class OffsitePaymentStrategy extends PaymentStrategy {
     execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
         const { payment, ...order } = payload;
         const paymentData = payment && payment.paymentData;
-        const orderPayload = payment && payment.gatewayId === 'adyen' ? payload : order;
+        const orderPayload = this._shouldSubmitFullPayload(payment) ? payload : order;
 
         if (!payment || !paymentData) {
             throw new PaymentArgumentInvalidError(['payment.paymentData']);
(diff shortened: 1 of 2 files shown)


src/payment/strategies/offsite-payment-strategy.ts
```diff
@@ -42,4 +42,11 @@ export default class OffsitePaymentStrategy extends PaymentStrategy {
 
         return super.finalize();
     }
+
+    shouldSubmitFullPayload(payment: any): boolean {
+        return payment && (
```
kevinziegler: @davidchin we're still checking the `methodId` here because I haven't yet figured out a way to populate `gatewayId` in this method:

screen shot 2018-09-21 at 3 13 32 pm

However without this check the initialize request fails entirely, like so:

screen shot 2018-09-21 at 3 07 34 pm

src/payment/strategies/offsite-payment-strategy.ts
```diff
@@ -42,4 +42,11 @@ export default class OffsitePaymentStrategy extends PaymentStrategy {
 
         return super.finalize();
     }
+
+    shouldSubmitFullPayload(payment: any): boolean {
```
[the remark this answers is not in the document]
kevinziegler: ♻️

Sure-Round:
@davidchin ♻️ Can you re-review & approve? We tested with the change in BCApp you requested in slack but it still did not click. The only way we have managed to get it running is with the Checkout SDK looking for `payment.methodId` for CCAvenue.

screen shot 2018-09-26 at 1 41 41 pm
.41_pm.png)

src/payment/strategies/offsite-payment-strategy.ts
```diff
@@ -42,4 +42,11 @@ export default class OffsitePaymentStrategy extends PaymentStrategy {
 
         return super.finalize();
     }
+
+    private _shouldSubmitFullPayload(payment: OrderPaymentRequestBody): boolean {
```
davidchin: I think you missed it in my last comment, you also need to change `payment` to `payment?`. That way, you don't need to manually assert the type when you call the method (i.e.: `this._shouldSubmitFullPayload(payment as OrderPaymentRequestBody)`). I usually avoid using `as` as much as possible because you want the type checker to warn you. In this case, `payment` can be undefined, but the warning is suppressed by `as`. Fortunately, you have a guard statement inside the method so you won't have any runtime error. But somebody else can easily remove the guard statement and there won't be any warning from the type checker.

kevinziegler:
♻️ Updated as per @davidchin's commment.  @Sure-Round Would you please re-review/merge when you get the chance?

Sure-Round:
@kevinziegler can you add the `// FIXME: A temporary workaround to support offsite payment methods where their return URL needs to be provided by the core app.` comment david requested

kevinziegler:
♻️ @Sure-Round Comment Added.

PR #416, opened 2018-09-26
CHECKOUT-3475: Add ability to embed checkout in iframe

davidchin:
## What?
* Add the ability to embed the checkout form in an iframe.
* The checkout iframe is seamless - meaning it automatically resizes depending on its content size.
* To use, you need to import the library in both the host and the child window.

**Host window**
```ts
embedCheckout({
    url: 'https://checkout/url',
    container: 'container-id',
    // Optionally, you can add callbacks such as `onLoad` and `onComplete`.
});
```

**Child window (checkout page)**
```ts
const messenger = createEmbeddedCheckoutMessenger({
    parentOrigin: 'https://some/website',
});

messenger.postFrameLoaded(); // Notify when the frame is initially loaded

// Load some data, render the UI

messenger.postLoaded(); // Notify when the checkout form is fully loaded
```

## Why?
* So you can embed the checkout form in other storefront channels, such as Wordpress.

## Testing / Proof
* Unit


src/embedded-checkout/embed-checkout.ts
```diff
@@ -0,0 +1,15 @@
+import EmbeddedCheckout from './embedded-checkout';
+import EmbeddedCheckoutListener from './embedded-checkout-listener';
+import EmbeddedCheckoutOptions from './embedded-checkout-options';
+import parseOrigin from './parse-origin';
+import ResizableIframeCreator from './resizable-iframe-creator';
+
```
capsula4: maybe docblock / usage example? Mention that is alpha stage for now?

src/embedded-checkout/embedded-checkout-events.ts
```diff
@@ -0,0 +1,41 @@
+export enum EmbeddedCheckoutEventType {
+    CheckoutComplete = 'CHECKOUT_COMPLETE',
+    CheckoutError = 'CHECKOUT_ERROR',
+    CheckoutLoaded = 'CHECKOUT_READY',
```
capsula4: can this be `CHECKOUT_LOADED` just to be consistent? (like frame below)

src/embedded-checkout/embedded-checkout-listener.ts
```diff
@@ -0,0 +1,75 @@
...
+    addListener<TType extends keyof EmbeddedCheckoutEventMap>(type: TType, listener: (event: EmbeddedCheckoutEventMap[TType]) => void): void {
+        if (!this._listeners[type]) {
+            this._listeners[type] = [];
+        }
+
+        this._listeners[type].push(listener);
+    }
+
+    removeListener<TType extends keyof EmbeddedCheckoutEventMap>(type: TType, listener: (event: EmbeddedCheckoutEventMap[TType]) => void): void {
+        const index = this._listeners[type] && this._listeners[type].indexOf(listener);
```
capsula4: if `this._listeners[type]` is undefined, would this still try to splice? Maybe just add a test that tries to remove a listener from a type that doesnt exist?

src/embedded-checkout/embedded-checkout.ts
```diff
@@ -0,0 +1,84 @@
+import { IFrameComponent } from 'iframe-resizer';
+
+import { EmbeddedCheckoutEventMap, EmbeddedCheckoutEventType } from './embedded-checkout-events';
+import EmbeddedCheckoutListener from './embedded-checkout-listener';
+import EmbeddedCheckoutOptions from './embedded-checkout-options';
+import ResiableIframeCreator from './resizable-iframe-creator';
```
capsula4: `ResizableIframeCreator`

src/embedded-checkout/embedded-checkout.ts
```diff
@@ -0,0 +1,84 @@
...
+            this.on(EmbeddedCheckoutEventType.FrameLoaded, this._options.onFrameLoad);
+        }
+    }
+
+    attach(): Promise<this> {
+        if (this._isAttached) {
+            return Promise.resolve(this);
+        }
+
+        this._isAttached = true;
```
capsula4: shouldn't this be set on success?  I guess is not a big deal, as if it fails you can always detach and attach again

src/embedded-checkout/iframe-content/embedded-checkout-messenger.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    postLoaded(): void {
+        const message: EmbeddedCheckoutLoadedEvent = {
+            type: EmbeddedCheckoutEventType.CheckoutLoaded,
+        };
+
+        this._postMessage(message);
+    }
+
+    private _postMessage(message: EmbeddedCheckoutEvent): void {
```
capsula4: 🍹  `_postParent(message)` ?

src/embedded-checkout/parse-origin.ts
```diff
@@ -0,0 +1,14 @@
+import { InvalidArgumentError } from '../common/error/errors';
+
+export default function parseOrigin(url: string): string {
+    if (!/^https?:\/\//.test(url)) {
+        throw new InvalidArgumentError('The provided URL must be absolute.');
```
capsula4: what about `http` and `//` ? maybe add coverage for those scenarios?

src/embedded-checkout/resizable-iframe-creator.ts
```diff
@@ -0,0 +1,68 @@
...
+        iframe.src = src;
+        iframe.style.border = 'none';
+        iframe.style.display = 'none';
+        iframe.style.width = '100%';
+
+        container.appendChild(iframe);
+
+        return this._toResizableFrame(iframe, timeout)
+            .catch(error => {
+                container.removeChild(iframe);
```
capsula4: I can see that on error we remove the iframe, so probably `EmbeddedCheckout._isAttached` should be set on success?

Also, is there any reason why you remove it on failure, as opposed to only attach it on success?

davidchin: This is because I have to attach it to the DOM tree, otherwise, the browser doesn't actually load the iframe.

davidchin:
@capsula4 I believe I've addressed all your comments. Could you please have a look again? Thanks! ♻️

src/embedded-checkout/resizable-iframe-creator.ts
capsula4: yeah that was my guess ^

src/embedded-checkout/embed-checkout.ts
```diff
@@ -4,6 +4,27 @@ import EmbeddedCheckoutOptions from './embedded-checkout-options';
...
+ * Embed the checkout form in an iframe.
+ *
+ * Once the iframe is embedded, it will automatically resize according to the
+ * size of the checkout form. It will also notify the parent window when certain
+ * events have occurred. i.e.: when the form is loaded and ready to be used.
+ *
+ * ```js
+ * embedCheckout({
+ *     url: 'https://checkout/url',
+ *     container: 'container-id',
```
capsula4: before seeing the example I thought that container was any selector. Should it be called `containerId`?

davidchin: Good point. I tried to match with the name we use for payment strategies. But I guess this is a separate thing and so it doesn't hurt to use a more appropriate name.

PR #417, opened 2018-09-26
CHECKOUT-3529 Fix createCurrencyService type

capsula4:
## What?
Fix the config to storeConfig type in currency service

## Why?
Because only the storeConfig is available externally.

### DIFF
--- a/src/currency/currency-service.ts
+++ b/src/currency/currency-service.ts
@@ -1,4 +1,4 @@
-import { Config } from '../config';
+import { StoreConfig } from '../config';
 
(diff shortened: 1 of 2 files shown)


src/currency/currency-service.ts
```diff
@@ -13,14 +13,14 @@ export default class CurrencyService {
      * @internal
      */
     constructor(
-        private _config: Config
+        private _config: StoreConfig
```
icatalina: Maybe rename this variable to go along with the name?

```js
private _storeConfig: StoreConfig
```

PR #418, opened 2018-09-27
feat(checkout): INT-951 Implement Masterpass button in cart and quick

cmedinilla:
## What? [INT-951](https://jira.bigcommerce.com/browse/INT-951)
Create checkout-buttons masterpass strategy
&
Hide the shipping address in the customer section as requested in [INT-984](https://jira.bigcommerce.com/browse/INT-984)
## Why?
We need this strategy to implement the cart and quick cart buttons for Masterpass

Due the delivery dates i made only one PR in order to simplify the process

## Testing / Proof
image
image
image
image

### DIFF
--- a/src/checkout-buttons/strategies/masterpass-button-strategy.ts
+++ b/src/checkout-buttons/strategies/masterpass-button-strategy.ts
@@ -1,0 +1,122 @@
+import { CheckoutButtonInitializeOptions, CheckoutButtonOptions } from '../';
+import { CheckoutActionCreator, CheckoutStore } from '../../checkout';
+import {
+    InvalidArgumentError,
+    MissingDataError,
+    MissingDataErrorType,
+    NotInitializedError,
+    NotInitializedErrorType
+} from '../../common/error/errors';
+import { bindDecorator as bind } from '../../common/utility';
+import { Masterpass, MasterpassCheckoutOptions, MasterpassScriptLoader } from '../../payment/strategies/masterpass';
+
+import { CheckoutButtonStrategy } from './';
+
+export default class MasterpassButtonStrategy extends CheckoutButtonStrategy {
+    private _masterpassClient?: Masterpass;
+    private _methodId?: string;
+    private _signInButton?: HTMLElement;
+
+    constructor(
+        private _store: CheckoutStore,
+        private _checkoutActionCreator: CheckoutActionCreator,
+        private _masterpassScriptLoader: MasterpassScriptLoader
+    ) {
+        super();
+    }
+
+    initialize(options: CheckoutButtonInitializeOptions): Promise<void> {
+        const { masterpass: masterpassOptions, methodId } = options;
+
+        if (!masterpassOptions || !methodId) {
+            throw new InvalidArgumentError('Unable to proceed because "options.masterpass" argument is not provided.');
+        }
+
+        if (this._isInitialized) {
+            return super.initialize(options);
+        }
+
+        this._methodId = methodId;
+
+        return this._store.dispatch(this._checkoutActionCreator.loadDefaultCheckout())
+            .then(state => {
+                const paymentMethod = state.paymentMethods.getPaymentMethod(methodId);
+
+                if (!paymentMethod || !paymentMethod.initializationData.checkoutId) {
+                    throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+                }
+
+                return this._masterpassScriptLoader.load(paymentMethod.config.testMode);
+            })
+            .then(masterpass => {
+                this._masterpassClient = masterpass;
+                this._signInButton = this._createSignInButton(masterpassOptions.container);
+
+                return super.initialize(options);
+            });
+    }
+
+    deinitialize(options: CheckoutButtonOptions): Promise<void> {
+        if (!this._isInitialized) {
+            return super.deinitialize(options);
+        }
+
+        if (this._signInButton && this._signInButton.parentNode) {
+            this._signInButton.removeEventListener('click', this._handleWalletButtonClick);
+            this._signInButton.parentNode.removeChild(this._signInButton);
+            this._signInButton = undefined;
+        }
+
+        return super.deinitialize(options);
+    }
+
+    private _createSignInButton(containerId: string): HTMLElement {
+        const buttonContainer = document.querySelector(`#${containerId}`);
+
+        if (!buttonContainer) {
+            throw new Error('Need a container to place the button');
+        }
+
+        const button = document.createElement('input');
+
+        button.type = 'image';
+        button.src = 'https://static.masterpass.com/dyn/img/btn/global/mp_chk_btn_160x037px.svg';
+        buttonContainer.appendChild(button);
+
+        button.addEventListener('click', this._handleWalletButtonClick);
+
+        return button;
+    }
+
+    private _createMasterpassPayload(): MasterpassCheckoutOptions {
+        const state = this._store.getState();
+        const checkout = state.checkout.getCheckout();
+        const paymentMethod = this._methodId ? state.paymentMethods.getPaymentMethod(this._methodId) : null;
+
+        if (!checkout) {
+            throw new MissingDataError(MissingDataErrorType.MissingCheckout);
+        }
+
+        if (!paymentMethod) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        return {
+            checkoutId: paymentMethod.initializationData.checkoutId,
+            allowedCardTypes: paymentMethod.initializationData.allowedCardTypes,
+            amount: checkout.cart.cartAmount.toString(),
+            currency: checkout.cart.currency.code,
+            cartId: checkout.cart.id,
+            suppressShippingAddress: true,
+        };
+    }
+
+    @bind
+    private _handleWalletButtonClick(): void {
+        if (!this._masterpassClient) {
+            throw new NotInitializedError(NotInitializedErrorType.CheckoutButtonNotInitialized);
+        }
+
+        this._masterpassClient.checkout(this._createMasterpassPayload());
+    }
+}
--- a/src/checkout-buttons/checkout-button-options.ts
+++ b/src/checkout-buttons/checkout-button-options.ts
@@ -1,6 +1,6 @@
 import { RequestOptions } from '../common/http-request';
 
-import { BraintreePaypalButtonInitializeOptions } from './strategies';
+import { BraintreePaypalButtonInitializeOptions, MasterpassButtonInitializeOptions } from './strategies';
 
 /**
  * The set of options for configuring the checkout button.
@@ -24,4 +24,5 @@ export interface CheckoutButtonInitializeOptions extends CheckoutButtonOptions {
      * omitted unless you need to support Braintree Credit.
      */
     braintreepaypalcredit?: BraintreePaypalButtonInitializeOptions;
+    masterpass?: MasterpassButtonInitializeOptions;
 }
--- a/src/checkout-buttons/create-checkout-button-registry.ts
+++ b/src/checkout-buttons/create-checkout-button-registry.ts
@@ -6,9 +6,10 @@ import { CheckoutActionCreator, CheckoutRequestSender, CheckoutStore } from '../
 import { Registry } from '../common/registry';
 import { ConfigActionCreator, ConfigRequestSender } from '../config';
 import { BraintreeScriptLoader, BraintreeSDKCreator } from '../payment/strategies/braintree';
+import { MasterpassScriptLoader } from '../payment/strategies/masterpass';
 import { PaypalScriptLoader } from '../payment/strategies/paypal';
 
-import { BraintreePaypalButtonStrategy, CheckoutButtonStrategy } from './strategies';
+import { BraintreePaypalButtonStrategy, CheckoutButtonStrategy, MasterpassButtonStrategy } from './strategies';
 
 export default function createCheckoutButtonRegistry(
     store: CheckoutStore,
@@ -39,6 +40,14 @@ export default function createCheckoutButtonRegistry(
             new PaypalScriptLoader(scriptLoader),
             createFormPoster(),
             true
+        )
+    );
+
+    registry.register('masterpass', () =>
+        new MasterpassButtonStrategy(
+            store,
+            checkoutActionCreator,
+            new MasterpassScriptLoader(scriptLoader)
         )
     );
 
--- a/src/checkout-buttons/index.ts
+++ b/src/checkout-buttons/index.ts
@@ -1,5 +1,6 @@
 export { default as createCheckoutButtonInitializer } from './create-checkout-button-initializer';
-
 export { default as checkoutButtonReducer } from './checkout-button-reducer';
 export { default as CheckoutButtonSelector } from './checkout-button-selector';
 export { default as CheckoutButtonState } from './checkout-button-state';
+
+export { CheckoutButtonOptions, CheckoutButtonInitializeOptions } from './checkout-button-options';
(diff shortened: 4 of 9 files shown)


src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,110 @@
...
+                this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId)),
+                this._store.dispatch(this._checkoutActionCreator.loadDefaultCheckout()),
+            ]
+        );
+
+        return states.then(state => {
+                const statePaymentMethod = state[0];
+                const stateCheckout = state[1];
+
+                if (this._isInitialized) {
```
metalvegetarianoprogresivo: This should be the first thing to check in the initialize, so you don't do unnecessary calls if this is true.

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,110 @@
...
+import {
+    InvalidArgumentError,
+    MissingDataError,
+    MissingDataErrorType
+} from '../../common/error/errors';
+import { PaymentMethod, PaymentMethodActionCreator } from '../../payment';
+import { Masterpass, MasterpassCheckoutOptions, MasterpassScriptLoader } from '../../payment/strategies/masterpass';
+import { CheckoutButtonInitializeOptions, CheckoutButtonOptions } from '../checkout-button-options';
+
+import CheckoutButtonStrategy from './checkout-button-strategy';
```
vmparra: Import from Index

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,110 @@
...
+import { MasterpassButtonInitializeOptions } from './masterpass-button-options';
```
vmparra: Merge with previous import

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,110 @@
+import { Checkout, CheckoutActionCreator, CheckoutStore } from '../../checkout';
+import {
+    InvalidArgumentError,
+    MissingDataError,
+    MissingDataErrorType
+} from '../../common/error/errors';
+import { PaymentMethod, PaymentMethodActionCreator } from '../../payment';
+import { Masterpass, MasterpassCheckoutOptions, MasterpassScriptLoader } from '../../payment/strategies/masterpass';
+import { CheckoutButtonInitializeOptions, CheckoutButtonOptions } from '../checkout-button-options';
```
vmparra: Import from Index

cmedinilla:
@vmparra @metalvegetarianoprogresivo ♻️ INT-910 merged this is rebased a ready for another review

cmedinilla:
Closed #437 and changes added to this PR

cmedinilla:
♻️ @vmparra  All comments was addressed

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,144 @@
+import { CheckoutButtonInitializeOptions, CheckoutButtonOptions } from '../';
+import { Checkout, CheckoutActionCreator, CheckoutStore } from '../../checkout';
+import {
```
vmparra: 🍹 In this case, you can put them in one single line.

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,144 @@
...
+    InvalidArgumentError,
+    MissingDataError,
+    MissingDataErrorType
+} from '../../common/error/errors';
+import { bindDecorator as bind } from '../../common/utility';
+import { PaymentMethod, PaymentMethodActionCreator } from '../../payment';
+import {
```
vmparra: 🍹 Here too.

src/checkout-buttons/strategies/masterpass-button-strategy.ts
cmedinilla: image
i thin is too long here in one line

cmedinilla: i think the line is too long for that:
image

Sure-Round:
@davidchin @icatalina This PR is ready for review ♻️ Of the Integration team PR's this is the top priority one right now.

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,144 @@
...
+            return super.initialize(options);
+        }
+
+        return Promise.all(
+            [
+                this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId)),
+                this._store.dispatch(this._checkoutActionCreator.loadDefaultCheckout()),
+            ]
+        ).then(([statePayment, stateCheckout]) => {
+                this._paymentMethod = statePayment.paymentMethods.getPaymentMethod(methodId);
```
davidchin: You shouldn't have to retain `statePayment` and `stateCheckout`. Both are snapshots of the same state, but captured at different times; so because the checkout action is dispatched later, `stateCheckout` is fresher than `statePayment`. Anyway, my recommendation is to retrieve it directly from `store` so you know for sure you're dealing with the latest data.

```ts

this._paymentMethod = state.paymentMethods.getPaymentMethod(methodId);
this._checkout = state.checkout.getCheckout();
```

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,144 @@
...
+            this._signInButton.parentNode.removeChild(this._signInButton);
+            this._signInButton = undefined;
+        }
+
+        return super.deinitialize(options);
+    }
+
+    private _createSignInButton(masterpassOptions: MasterpassButtonInitializeOptions): void {
+        const { container } = masterpassOptions;
+        const buttoncontainer = document.querySelector(`#${container}`);
```
davidchin: Could you please use camel case for the two words? i.e.: `buttonContainer`.

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,144 @@
...
+        };
+    }
+
+    @bind
+    private _handleWalletButtonClick(): void  {
+        const payload = this._createMasterpassPayload();
+        this.masterpassClient.checkout(payload);
+    }
+
+    private get masterpassClient(): Masterpass {
```
davidchin: Firstly, we actually don't allow computed properties in this project. `get/set` are forbidden because it is very easy to put logic in there that's computationally expensive or induces side effects (for more information, please see this [article](https://github.com/airbnb/javascript#accessors--no-getters-setters)).

Secondly, all private properties and methods need to be prefixed with `_`. This is because, although we're using TypeScript, the source ultimately gets compiled into JS and consumed by JS clients. So we need a way to indicate to the consumers that these properties and methods are not supposed to be used.

If you really want to go with this pattern, you need to name this method as `_getMasterClient`. But often we just inline the guard statement unless there's a real need (i.e.: you need to access the property multiple times).

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,144 @@
...
+        if (!this._masterpassClient) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        return this._masterpassClient;
+    }
+
+    private get paymentMethod(): PaymentMethod {
```
davidchin: Please see my previous comment regarding the use of `get` and the name of the method. If you do want to have a private method for retrieving the `PaymentMethod` object, you should retrieve the data directly from `store` instead. This is because keeping a separate reference is harder to maintain as you always have to remember to update the reference when the object gets stale.

```ts
private _getPaymentMethod(): PaymentMethod {
    const paymentMethod = state.paymentMethods.getPaymentMethod(this._methodId);

    if (!paymentMethod) {
    }

    return paymentMethod;
}
```

That'd also simply your `initialize` method.

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    @bind
+    private _handleWalletButtonClick(): void  {
+        const payload = this._createMasterpassPayload();
+        this.masterpassClient.checkout(payload);
+    }
+
+    private get masterpassClient(): Masterpass {
+        if (!this._masterpassClient) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
```
davidchin: Can you please throw `NotInitializeError`? It is a more appropriate error to throw in this case.

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    private get paymentMethod(): PaymentMethod {
+        if (!this._paymentMethod) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        return this._paymentMethod;
+    }
+
+    private get checkout(): Checkout {
```
davidchin: Please see my previous comment regarding `paymentMethod`.

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,144 @@
...
+        if (!this._checkout) {
+            throw new MissingDataError(MissingDataErrorType.MissingCart);
```
davidchin: You're missing `Checkout` object, not `Cart`.

src/payment/strategies/masterpass/masterpass.ts
```diff
@@ -46,7 +46,6 @@ export interface MasterpassCheckoutOptions {
      * This optional parameter can be used to override the callbackUrl specified in the Merchant Portal.
      */
     callbackUrl?: string;
-
```
davidchin: Unnecessary diff. Could you please revert?

cmedinilla:
@davidchin All comments was addressed only the store state part with the proposed approach is not working
i think is because the checkout doesn't exist where this strategy is called (cart and quick cart section)

My proposal is load only the state checkout to get the latest data and from there load the paymentMethod too

````javascript
        .then(stateCheckout => {
                this._stateCheckout = stateCheckout;

                return this._masterpassScriptLoader.load(this._getPaymentMethod().config.testMode)
                    .then(masterpass => {
                        this._masterpassClient = masterpass;
                        this._createSignInButton(masterpassOptions);
                    });
        })
        .then(() => super.initialize(options));
````

````javascript
    private _getPaymentMethod(): PaymentMethod {
        if (!this._methodId) {
        }

        if (!this._stateCheckout) {
        }

        const paymentMethod = this._stateCheckout.paymentMethods.getPaymentMethod(this._methodId);
        if (!paymentMethod || !paymentMethod.initializationData.checkoutId) {
        }

        return paymentMethod;
    }
````

File: src/checkout-buttons/strategies/masterpass-button-strategy.ts [hunk 15]
```diff
@@ -0,0 +1,149 @@
...
+        if (!masterpassOptions || !methodId) {
+            throw new InvalidArgumentError('Unable to proceed because "options.masterpass" argument is not provided.');
+        }
+        this._methodId = methodId;
+
+        if (this._isInitialized) {
+            return super.initialize(options);
+        }
+
+        return this._store.dispatch(this._checkoutActionCreator.loadDefaultCheckout())
```
davidchin: Hey @cmedinilla, sorry if I didn't explain myself clearly last time, I didn't mean to remove `loadPaymentMethod` and `loadDefaultCheckout`. You still need those. All I'm asking is not to read it from `stateCheckout` and store a separate reference of it. So with what you've done so far, the code could be something like the following:

```ts
initialize(options: CheckoutButtonInitializeOptions): Promise <void> {
    const { masterpass: masterpassOptions, methodId } = options;

    if (!masterpassOptions || !methodId) {
        throw new InvalidArgumentError('Unable to proceed because "options.masterpass" argument is not provided.');
    }

    if (this._isInitialized) {
        return super.initialize(options);
    }

    this._methodId = methodId;

    return Promise.all([
        this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId)),
        this._store.dispatch(this._checkoutActionCreator.loadDefaultCheckout()),
    ])
        .then(() => this._masterpassScriptLoader.load(this._getPaymentMethod().config.testMode))
        .then(masterpass => {
            this._masterpassClient = masterpass;
            
            this._createSignInButton(masterpassOptions);

            return super.initialize(options);
        });
}
```

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,149 @@
...
+
+        return this._masterpassClient;
+    }
+
+    private _getPaymentMethod(): PaymentMethod {
+        if (!this._methodId) {
+            throw new InvalidArgumentError();
+        }
+
+        if (!this._stateCheckout) {
```
davidchin: This check should be removed, as the checkout object shouldn't affect the retrieval of the payment method object.

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,149 @@
...
+        const paymentMethod = this._stateCheckout.paymentMethods.getPaymentMethod(this._methodId);
+        if (!paymentMethod || !paymentMethod.initializationData.checkoutId) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        return paymentMethod;
+    }
+
+    private _getCheckout(): Checkout {
+        if (!this._stateCheckout) {
```
davidchin: This check should be removed as well. The check below is sufficient.

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,149 @@
...
+        if (!this._masterpassClient) {
+            throw new NotInitializedError(NotInitializedErrorType.CheckoutButtonNotInitialized);
+        }
+
+        return this._masterpassClient;
+    }
+
+    private _getPaymentMethod(): PaymentMethod {
+        if (!this._methodId) {
+            throw new InvalidArgumentError();
```
davidchin: In this case, you should throw `NotInitializedError` because first of all this method doesn't accept an argument so it's odd to throw `InvalidArgumentError`, secondly the reason why `methodId` could be undefined is because the strategy is not initialized.

src/checkout-buttons/strategies/masterpass-button-strategy.ts
```diff
@@ -0,0 +1,149 @@
...
+    MasterpassCheckoutOptions,
+    MasterpassScriptLoader
+} from '../../payment/strategies/masterpass';
+
+import { CheckoutButtonStrategy, MasterpassButtonInitializeOptions } from './';
+
+export default class MasterpassButtonStrategy extends CheckoutButtonStrategy {
+    private _masterpassClient?: Masterpass;
+    private _signInButton?: HTMLElement;
+    private _stateCheckout?: InternalCheckoutSelectors;
```
davidchin: Please see my comment above regarding this. It should be removed.

src/checkout-buttons/strategies/masterpass-button-strategy.ts
cmedinilla: I made some changes, now i am not saving the state anymore and i am only using the data to create the payload

cmedinilla:
@davidchin ♻️

Sure-Round:
@cmedinilla If you squash commits I will merge for you.

PR #423, opened 2018-10-02
CHECKOUT-3475: Emit `FRAME_ERROR` event if unable to embed checkout in iframe

davidchin:
## What?
* Emit an error event if unable to embed the checkout form as an iframe.

## Why?
* So you can catch and handle the error in case you can't embed the iframe for whatever reason. i.e.:
```ts
embedCheckout(config)
    .catch(error => {
        window.alert(error.message);
        window.location.assign('/cart');
    });
```
* Because most likely the iframe will be embedded across domains, so you can't listen for error events from the parent frame. Instead, you need to rely on the child frame to tell the parent frame when the checkout form cannot be loaded.

## Testing / Proof
* Unit

### DIFF
--- a/src/embedded-checkout/embedded-checkout-events.ts
+++ b/src/embedded-checkout/embedded-checkout-events.ts
@@ -2,6 +2,7 @@ export enum EmbeddedCheckoutEventType {
     CheckoutComplete = 'CHECKOUT_COMPLETE',
     CheckoutError = 'CHECKOUT_ERROR',
     CheckoutLoaded = 'CHECKOUT_LOADED',
+    FrameError = 'FRAME_ERROR',
     FrameLoaded = 'FRAME_LOADED',
 }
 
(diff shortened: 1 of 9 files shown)


PascalZajac:
This is going to sound stupid but... what sort of errors could occur that would still allow the event to be emitted? Is this just a passthrough mechanism for normal SDK errors being thrown?

davidchin:
Hey @PascalZajac, at the moment, we check if there's a valid cart before rendering the checkout page. We also do other checks, such as checking if the enabled payment methods are supported by UCO. If there's a problem, we redirect the shopper to the cart page. For embedded checkout, instead of doing a redirect, we'll render an error page and emit the `FRAME_ERROR` event as a way to notify the parent frame that something's gone wrong.

PascalZajac:
Understood, I wasn't clear on what sort of error conditions could go wrong and yet still be recoverable and communicable to the parent frame.

davidchin:
No worries. I re-read the PR description and realised it didn't explain the possible error conditions. Sorry about that.

PR #425, opened 2018-10-04
CHECKOUT-3581 Translate Country and State in returned addresses

capsula4:
## What?
Translate `country` and `stateOrProvince` when getting addresses from selectors

## Why?
To provide this functionality out of the box

### DIFF
--- a/src/address/locale-address.ts
+++ b/src/address/locale-address.ts
@@ -1,0 +1,17 @@
+import { Country } from '../geography';
+
+import Address from './address';
+
+export default function localeAddress<T1 extends Address>(address: T1, countries?: Country[]): T1 {
+    const country =  (countries || []).find(({ code }) => code === address.countryCode);
+    const states = !country || !country.subdivisions.length ? [] : country.subdivisions;
+    const state = states.find(({ code }) => code === address.stateOrProvinceCode);
+
+    return Object.assign({},
+        address,
+        {
+            country: country ? country.name : address.country,
+            stateOrProvince: state ? state.name : address.stateOrProvince,
+        }
+    );
+}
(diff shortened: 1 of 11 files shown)


src/address/locale-address.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default function localeAddress<T1 extends Address>(address: T1, countries?: Country[]): T1 {
+    const country =  (countries || []).find(({ code }) => code === address.countryCode);
+    const states = !country || !country.subdivisions.length ? [] : country.subdivisions;
+    const state = states.find(({ code }) => code === address.stateOrProvinceCode);
+
+    return Object.assign({},
+        address,
+        {
+            country: country ? country.name : address.country,
```
capsula4: since this could be a breaking change, maybe expose this new data via new fields? `localedCountry` and `localedStateOrProvince` ?

src/address/locale-address.ts
```diff
@@ -0,0 +1,17 @@
+import { Country } from '../geography';
+
+import Address from './address';
+
+export default function localeAddress<T1 extends Address>(address: T1, countries?: Country[]): T1 {
+    const country =  (countries || []).find(({ code }) => code === address.countryCode);
+    const states = !country || !country.subdivisions.length ? [] : country.subdivisions;
+    const state = states.find(({ code }) => code === address.stateOrProvinceCode);
+
+    return Object.assign({},
```
capsula4: TS wouldn't allow me to deconstruct `address` as I'm using generics

src/checkout/create-internal-checkout-selectors.ts
```diff
@@ -20,26 +20,26 @@ import CheckoutStoreState from './checkout-store-state';
 import InternalCheckoutSelectors from './internal-checkout-selectors';
 
 export default function createInternalCheckoutSelectors(state: CheckoutStoreState, options: CheckoutStoreOptions = {}): InternalCheckoutSelectors {
-    const billingAddress = new BillingAddressSelector(state.billingAddress);
+    const countries = new CountrySelector(state.countries);
```
capsula4: Localing addresses might better be done in external selectors (store selector)

PR #428, opened 2018-10-07
CHECKOUT-3477: Pass styling configuration to child frame

davidchin:
## What?
* Pass styling configuration to child frame.

## Why?
* So the child frame can read the styling configuration and apply styling to itself. i.e.:
```ts
embeddedCheckoutMessenger.receiveStyles(styles => {
    // Create a stylesheet programmatically using `styles`
});
```

## Testing / Proof
* Unit


src/embedded-checkout/iframe-event-poster.ts
```diff
@@ -0,0 +1,14 @@
+export default class IframeEventPoster<TEvent> {
+    constructor(
+        private _targetOrigin: string,
+        private _targetWindow: Window
+    ) {}
+
+    post(event: TEvent): void {
+        if (window === this._targetWindow) {
```
capsula4: should this check be performed at construction time?

davidchin: Hrmm, good point. I guess what I really want is a no-op messenger. So it does nothing if it is not used an iframe.

capsula4: yeah, I was mostly thinking if there could be a way that no checking for it could lead to a bug, so better throw an exception when constructing it.

PR #429, opened 2018-10-10
CHECKOUT-3477: Show loading indicator while waiting for iframe to load

davidchin:
## What?
1. Show a loading indicator while waiting for the iframe to load.
1. Reconfigure the iframe with styling information if it's reloaded again.

## Why?
1. To tell the shopper that something is happening in the background.
2. Otherwise, we'd lose the styling configuration.

## Testing / Proof
Unit

### DIFF
--- a/src/embedded-checkout/embedded-checkout.spec.ts
+++ b/src/embedded-checkout/embedded-checkout.spec.ts
@@ -8,12 +8,14 @@ import { NotEmbeddableError } from './errors';
 import { EmbeddedContentEvent, EmbeddedContentEventType } from './iframe-content/embedded-content-events';
 import IframeEventListener from './iframe-event-listener';
 import IframeEventPoster from './iframe-event-poster';
+import LoadingIndicator from './loading-indicator';
 import ResizableIframeCreator from './resizable-iframe-creator';
 
 describe('EmbeddedCheckout', () => {
     let embeddedCheckout: EmbeddedCheckout;
     let iframe: IFrameComponent;
     let iframeCreator: ResizableIframeCreator;
+    let loadingIndicator: LoadingIndicator;
     let messageListener: IframeEventListener<EmbeddedCheckoutEventMap>;
     let messagePoster: IframeEventPoster<EmbeddedContentEvent>;
     let options: EmbeddedCheckoutOptions;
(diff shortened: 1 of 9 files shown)


src/embedded-checkout/embedded-checkout.spec.ts
```diff
@@ -149,4 +171,50 @@ describe('EmbeddedCheckout', () => {
...
+            options
+        );
+
+        jest.spyOn(messagePoster, 'post');
+
+        await embeddedCheckout.attach();
+
+        messageListener.trigger({ type: EmbeddedCheckoutEventType.FrameLoaded });
+
+        expect(messagePoster.post).toHaveBeenCalledTimes(2);
```
capsula4: should you check the arguments to make sure it reconfigured styles?

src/embedded-checkout/embedded-checkout.spec.ts
```diff
@@ -149,4 +171,50 @@ describe('EmbeddedCheckout', () => {
...
+    });
+
+    it('toggles loading indicator', async () => {
+        await embeddedCheckout.attach();
+
+        expect(loadingIndicator.show).toHaveBeenCalled();
```
capsula4: can you do something like, it toggles loading indicator after checkout has loaded?

```
expect(loadingIndicator.show).toHaveBeenCalled();
expect(loadingIndicator.hide).not.toHaveBeenCalled();

frameLoaded()

expect(loadingIndicator.hide).toHaveBeenCalled();
```

you could also check it hides if frame failed to load.

src/embedded-checkout/loading-indicator.ts
```diff
@@ -0,0 +1,120 @@
...
+
+const DEFAULT_STYLES: LoadingIndicatorStyles = {
+    size: 70,
+    color: '#d9d9d9',
+    backgroundColor: '#ffffff',
+};
+
+const ROTATION_ANIMATION = 'embedded-checkout-loading-indicator-rotation';
+
+export default class LoadingIndicator {
```
capsula4: I have a few doubts with the naming, maybe this could be LoadingOverlay? Indicator to me sounds like just the icon. 🍹 

`_container` could become `_overlay`

src/embedded-checkout/loading-indicator.ts
```diff
@@ -0,0 +1,120 @@
...
+
+            if (!parent) {
+                throw new Error('Unable to attach the loading indicator because the parent ID is not valid.');
+            }
+
+            parent.appendChild(this._container);
+        }
+
+        if (!this._container.parentNode) {
+            throw new Error('Unable to show the loading indicator because it is not attached to the DOM.');
```
capsula4: would this only happen if `parentId` is not provided?

src/embedded-checkout/loading-indicator.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+
+        expect(child.style.visibility).toEqual('visible');
+        expect(child.style.opacity).toEqual('1');
+    });
+
+    it('throws error is parent element does not exist', () => {
+        expect(() => indicator.show('invalid_container')).toThrow();
+    });
+
+    it('hide loading indicator', () => {
```
capsula4: hides

src/embedded-checkout/loading-indicator.ts
davidchin: I understand where you coming from. But in this case, the indicator is in fact the spinner icon that gets shown to the shopper when the page is initially loaded. So I think it is okay to keep it as "indicator"?

src/embedded-checkout/loading-indicator.ts
davidchin: Good question. I might not even need this check. I'll double check.

PR #434, opened 2018-10-15
Update Address type to properly type nullable fields (fixes #420)

### LINKED ISSUE #420: address.address2 and address.countryCode can be null
https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/address/address.ts

Type says it's not optional but in the real world we're seeing null is passed to these fields. 

Can SDK make sure always set empty string or update type definition?

daisy1754:
## What and Why
Address.address2 and Address.countryCode can be null. We had runtime errors caused by these wrong type so this PR defines types that match the reality.

https://github.com/bigcommerce/checkout-sdk-js/issues/420

### DIFF
--- a/src/address/address.ts
+++ b/src/address/address.ts
@@ -7,11 +7,11 @@ export interface AddressRequestBody {
     lastName: string;
     company: string;
     address1: string;
-    address2: string;
+    address2: string | null;
     city: string;
     stateOrProvince: string;
     stateOrProvinceCode: string;
-    countryCode: string;
+    countryCode: string | null;
     postalCode: string;
     phone: string;
     customFields: Array<{


src/address/address.ts
```diff
@@ -7,11 +7,11 @@ export interface AddressRequestBody {
     lastName: string;
     company: string;
     address1: string;
-    address2: string;
+    address2: string | null;
     city: string;
     stateOrProvince: string;
     stateOrProvinceCode: string;
-    countryCode: string;
+    countryCode: string | null;
```
capsula4: This was my bad, I got confused with `country` field, which shouldn't be passed, but that's the response object.

Currently we seem to be accepting `countryCode: null` but it magically defaults to the US when you do that. I think that could be a bug in our APIs.

Most of the fields are not required, so maybe methods accepting addresses should have `Partial<AddressRequestBody>` as type. Would that solve your problems?

capsula4: Right, this looks like a bug in the Server API. Should always return a string.

capsula4: Will investigate and get back to you.

icatalina: If that's a copy/paste of your code, it seems like it might be broken. 

```
return addresses.map(a => a.countryCode).find(e.length > 0);
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

`find` expects a function, `e` is `undefined`, yes. I guess you wanted something like: 

```
return addresses
	.map(a => a.countryCode)
	.find(e => e.length > 0);
```

capsula4: FYI, we have identified places in our API where is returning `null` were a string is expected. Thanks for the effort put into this, feel free to close it if you think your issue has been solved. The fix probably won't be live in all stores until early next week.

snaderiBC: @daisy1754 Keep in mind `customer.addresses` could be an empty array if customer is not signed in.
```customer: {
id: 0,
isGuest: true,
email: "",
firstName: "",
lastName: "",
fullName: "",
addresses: [ ],
storeCredit: 0
}

daisy1754: thank you @capsula4

PR #435, opened 2018-10-15
CHECKOUT-3589 Allow un-assigning items from address

capsula4:
## What?
Allow method `CheckoutService#assignItemsToAddress` to take negative quantities or empty payloads.

## Why?
So items can be un-assigned from addresses.

## Testing / Proof
unit
manual

### DIFF
--- a/src/shipping/consignment-action-creator.ts
+++ b/src/shipping/consignment-action-creator.ts
@@ -1,5 +1,4 @@
 import { createAction, createErrorAction, ThunkAction } from '@bigcommerce/data-store';
-import { find, map } from 'lodash';
 import { Observable } from 'rxjs/Observable';
 import { Observer } from 'rxjs/Observer';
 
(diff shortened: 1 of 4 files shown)


PascalZajac:
What's the rationale for using negative quantities to achieve this behaviour? It seems a bit like overloading to me.

src/shipping/consignment-action-creator.ts
```diff
@@ -287,29 +323,52 @@ export default class ConsignmentActionCreator {
...
+        if (!cart) {
+            throw new MissingDataError(MissingDataErrorType.MissingCart);
+        }
+
+        return this._hydrateLineItems(consignment.lineItemIds, cart).map(existingItem => {
+            const sharedItem = lineItems.find(lineItem => lineItem.itemId === existingItem.itemId);
+
+            return {
+                ...existingItem,
+                quantity: existingItem.quantity - (sharedItem ? sharedItem.quantity : 0),
```
icatalina: 🍹Seems more readable the other way around?

```suggestion
                quantity: sharedItem ? (existingItem.quantity - sharedItem.quantity) : existingItem.quantity,
```

src/shipping/consignment-action-creator.ts
```diff
@@ -287,29 +323,52 @@ export default class ConsignmentActionCreator {
...
         }
 
-        const existingLineItems = map(existingConsignment.lineItemIds, itemId => {
+        return lineItems
+            .concat(this._hydrateLineItems(consignment.lineItemIds, cart))
+            .filter(lineItem => lineItem.quantity > 0);
+    }
+
+    private _hydrateLineItems(lineItemIds: string[], cart: Cart): ConsignmentLineItem[] {
+        return map(lineItemIds, itemId => {
```
icatalina: 🍹I guess if this is a normal array, there's no need for lodash...

```suggestion
        return lineItemIds.map(itemId => {
```

PR #437, opened 2018-10-18
feat(checkout): INT-984 Add a new parameter to hide the shipping address

cmedinilla:
## What? [INT-984](https://jira.bigcommerce.com/browse/INT-984)
Add the parameter suppressShippingAddress as True in the button payload

## Why?
To suppress the shipping address in the masterpass card selection page

## Testing / Proof
...


cmedinilla:
Changes will be addressed in #418

Issue #432, opened 2018-10-11
Multiple Braintree PayPal checkout buttons not appearing correctly on cart page


flyingL123:
There seems to be an issue with the Braintree PayPal integration. I have noticed recently that only one of the three PayPal buttons on our checkout are appearing. Please see below:

The cart page has 3 locations where the button is supposed to appear.

Location 1: Mobile-only top location
Location 2: Desktop-only top location
Location 3: Mobile and Desktop bottom location

The buttons used to appear in all three locations as expected, but recently I've noticed that only Location 1 is appearing, and the other two locations do not show a button.

This is troublesome for us since we need the button to appear in multiple places. For example, here is the mobile layout, which works correctly since the mobile-only top location is the first one in the markup. The PayPal button appears correctly underneath the AmazonPay button:

<img width="378" alt="paypal_location1" src="https://user-images.githubusercontent.com/2486712/46834010-2d06f380-cd78-11e8-8f34-9ebb4aff8fcd.png">

However, there should also be a PayPal button in Location 3 underneath the Amazon button, but as you can see, it does not appear:

<img width="378" alt="paypal_location3" src="https://user-images.githubusercontent.com/2486712/46834796-64769f80-cd7a-11e8-8ff7-47ce5483ed7d.png">

Additionally, on desktop, no PayPal buttons appear since desktop uses Location 2 and Location 3:

<img width="1190" alt="paypal_location2and3" src="https://user-images.githubusercontent.com/2486712/46834043-44de7780-cd78-11e8-9d17-97546f6c1e14.png">

Location 2 and 3 both contain the code to show the button, but they never appear:

<img width="1341" alt="paypal_code" src="https://user-images.githubusercontent.com/2486712/46834096-79523380-cd78-11e8-861f-cf9dbc581814.png">

To reproduce, go to https://www.storeyourboard.com/skateboard-deck-display-floating-mount/ and click "Add to Cart". The issue appears on the cart page as long as an item is in the cart.

In other words, the checkout buttons only appear when an item is in your cart. So make sure there is an item in your cart. Then anytime you visit the cart page, you will see the issue.

I can also see that 3 calls to get the checkout button are being made, but only one is being rendered. [I started by reporting the issue to PayPal](https://github.com/paypal/paypal-checkout/issues/890), but I'm now pretty sure the issue is on the BigCommerce side since the issue does not occur in their testing environment.

[...]

PR #443, opened 2018-10-23
CHECKOUT-3584 Allow rendering checkout buttons more than once

capsula4:
Allows rendering checkout buttons more than once.

### DIFF
--- a/src/checkout-buttons/checkout-button-options.ts
+++ b/src/checkout-buttons/checkout-button-options.ts
@@ -1,6 +1,10 @@
 import { RequestOptions } from '../common/http-request';
 
+<<<<<<< HEAD
 import { BraintreePaypalButtonInitializeOptions, MasterpassButtonInitializeOptions } from './strategies';
+=======
+import { BraintreePaypalButtonInitializeOptions, CheckoutButtonMethodType } from './strategies';
+>>>>>>> fix(checkout-button): CHECKOUT-3584 Allow rendering checkout buttons more than once
 
 /**
  * The set of options for configuring the checkout button.
@@ -9,7 +13,7 @@ export interface CheckoutButtonOptions extends RequestOptions {
     /**
      * The identifier of the payment method.
      */
-    methodId: string;
+    methodId: CheckoutButtonMethodType;
 }
 
 export interface CheckoutButtonInitializeOptions extends CheckoutButtonOptions {
@@ -24,5 +28,13 @@ export interface CheckoutButtonInitializeOptions extends CheckoutButtonOptions {
      * omitted unless you need to support Braintree Credit.
      */
     braintreepaypalcredit?: BraintreePaypalButtonInitializeOptions;
+<<<<<<< HEAD
     masterpass?: MasterpassButtonInitializeOptions;
+=======
+
+    /**
+     * The ID of a container which the checkout button should be inserted.
+     */
+    containerId: string;
+>>>>>>> fix(checkout-button): CHECKOUT-3584 Allow rendering checkout buttons more than once
 }
--- a/src/checkout-buttons/checkout-button-options.ts
+++ b/src/checkout-buttons/checkout-button-options.ts
@@ -1,10 +1,6 @@
 import { RequestOptions } from '../common/http-request';
 
-<<<<<<< HEAD
-import { BraintreePaypalButtonInitializeOptions, MasterpassButtonInitializeOptions } from './strategies';
-=======
 import { BraintreePaypalButtonInitializeOptions, CheckoutButtonMethodType } from './strategies';
->>>>>>> fix(checkout-button): CHECKOUT-3584 Allow rendering checkout buttons more than once
 
 /**
  * The set of options for configuring the checkout button.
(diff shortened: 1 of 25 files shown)


src/checkout-buttons/checkout-button-options.ts
```diff
@@ -24,4 +24,9 @@ export interface CheckoutButtonInitializeOptions extends CheckoutButtonOptions {
      * omitted unless you need to support Braintree Credit.
      */
     braintreepaypalcredit?: BraintreePaypalButtonInitializeOptions;
+
+    /**
+     * The ID of a container which the checkout button should be inserted.
+     */
+    containerId: string;
```
capsula4: Not 100% sure, but I would expect any checkout button to have a `containerId` ?

capsula4: Also, modified this to be `containerId` instead of `container` so it's consistent with embeddable checkout. However I've noticed payment strategies use `container` instead of `containerId`. Thoughts?

src/checkout-buttons/strategies/checkout-button-strategy.ts
```diff
@@ -1,16 +1,16 @@
 import { CheckoutButtonInitializeOptions, CheckoutButtonOptions } from '../checkout-button-options';
 
 export default abstract class CheckoutButtonStrategy {
-    protected _isInitialized = false;
+    protected _isInitialized: {[key: string]: boolean} = {};
```
davidchin: Could you put a space character around the bracket? i.e.: `{ [key: string]: boolean }`

src/checkout-buttons/strategies/checkout-button-method.ts
```diff
@@ -0,0 +1,4 @@
+export enum CheckoutButtonMethod {
```
davidchin: Can you name this enum as `CheckoutButtonMethodType`? So it's clearer that it is not an object?

src/checkout-buttons/checkout-button-options.ts
davidchin: Yeah I guess it does make sense to have it as a common property. `CheckoutButtonInitializer` is still considered to be "alpha" so I guess we _could_ make a breaking change if there's a good reason. We just have to make a corresponding update in the client when we release the change.

src/checkout-buttons/checkout-button-strategy-action-creator.ts
```diff
@@ -19,17 +19,23 @@ export default class CheckoutButtonStrategyActionCreator {
...
-            this._paymentMethodActionCreator.loadPaymentMethod(options.methodId, options),
-            defer(() => this._registry.get(options.methodId).initialize(options)
-                .then(() => createAction(CheckoutButtonActionType.InitializeButtonSucceeded, undefined, meta)))
-        ).pipe(
-            catchError(error => throwErrorAction(CheckoutButtonActionType.InitializeButtonFailed, error, meta))
-        );
+            return concat(
+                of(createAction(CheckoutButtonActionType.InitializeButtonRequested, undefined, meta)),
+                paymentMethod ?
+                    new Promise(resolve => resolve()) :
```
davidchin: I think what you want is `empty()`, which emits nothing.

flyingL123:
Can someone advise if this change has been rolled out? My original issue https://github.com/bigcommerce/checkout-sdk-js/issues/432 was closed but I still have the same problem on our website. Is the fix live yet or is there some delay until the update is actually pushed out?

davidchin:
Hey @flyingL123, we've just deployed the fix few hours ago. However, it may take a few days to roll out to your store, depending on the tier of the store. Unless I'm mistaken, the fix should be promoted to your store (storeyourboard.com) by tomorrow.

flyingL123:
Hi @davidchin and team, I can see this is now live on my store, which is great. However, I am noticing that the PayPal buttons load **much slower** than they used to. The first one (on mobile view) loads quickly, the second one (bottom button on desktop view) load slowly, and the third one (top button on desktop view) loads even slower. You can see them loading one after the other. The third one takes multiple seconds to show up.

You'll notice that the Amazon button do not have that issue. All of them load virtually instantaneously. It seems like this new method of loading button does not work as efficiently. Is there some way to fix this?

You can easily see the problem by visiting storeyourboard.com and adding a product to your cart on desktop. You will see the top PayPal button takes a very long time to load. The one at the bottom of the page loads sooner, but still after a delay.

flyingL123:
@davidchin any feedback on my comment above? There is still an enormous delay loading the same button multiple times.

davidchin:
Hi @flyingL123, thanks for reporting this issue. We'll investigate and find a fix for it as soon as we can.

flyingL123:
@davidchin - I am also finding that when the cart content is updated via ajax, for example, when increasing the number of items in the cart, or removing an item from a cart with more than one item in it, the PayPal button does not get reloaded. The other checkout buttons do, but PayPal does not, it just disappears. You can see this happening on our site [storeyourboard.com](https://www.storeyourboard.com).

flyingL123:
@davidchin what is going on with this? Now it appears the old problem is back and the additional PayPal buttons do not appear at all. Any idea what's going on?

flyingL123:
@davidchin thanks for the explanation, but when do you expect this to be fixed? It is quite frustrating to essentially not have PayPal available on desktop during the busiest shopping time of the year.

PR #447, opened 2018-10-25
feat(payment): PAYMENTS-3071 Support the Paypal checkout button strategy for smart buttons

lord2800:
## What?
Add support for the latest PayPal checkout button SDK.

## Why?
Enables the use of smart buttons for payment options.

## Testing / Proof
With a store that has PayPal Express enabled as a payment method, add an item to the cart via the quick view modal and observe that your buttons are customizable in the theme editor. Additionally, those same customizations apply to the cart page. Finally, you should be able to complete an order via all three flows: paypal button from the quick view modal, paypal button from the cart page, paypal payment option from the checkout page.

Also includes unit tests.

### DIFF
--- a/src/checkout-buttons/strategies/paypal-button-strategy.ts
+++ b/src/checkout-buttons/strategies/paypal-button-strategy.ts
@@ -1,0 +1,132 @@
+import { FormPoster } from '@bigcommerce/form-poster';
+import { pick } from 'lodash';
+
+import { CheckoutStore } from '../../checkout';
+import { InvalidArgumentError, MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType, StandardError } from '../../common/error/errors';
+import { PaymentMethod } from '../../payment';
+import { PaypalActions, PaypalAuthorizeData, PaypalClientToken } from '../../payment/strategies/paypal';
+import { PaypalScriptLoader } from '../../payment/strategies/paypal';
+import { CheckoutButtonInitializeOptions } from '../checkout-button-options';
+
+import CheckoutButtonStrategy from './checkout-button-strategy';
+
+export default class PaypalButtonStrategy extends CheckoutButtonStrategy {
+    private _paymentMethod?: PaymentMethod;
+
+    constructor(
+        private _store: CheckoutStore,
+        private _paypalScriptLoader: PaypalScriptLoader,
+        private _formPoster: FormPoster
+    ) {
+        super();
+    }
+
+    initialize(options: CheckoutButtonInitializeOptions): Promise<void> {
+        if (this._isInitialized[options.containerId]) {
+            return super.initialize(options);
+        }
+
+        const paypalOptions = options.paypal;
+        const state = this._store.getState();
+        const paymentMethod = this._paymentMethod = state.paymentMethods.getPaymentMethod(options.methodId);
+
+        if (!paypalOptions) {
+            throw new InvalidArgumentError();
+        }
+
+        return this._paypalScriptLoader.loadPaypal()
+            .then(paypal => {
+                if (!paymentMethod || !paymentMethod.config.merchantId) {
+                    throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+                }
+
+                const merchantId = paymentMethod.config.merchantId;
+                const env = paymentMethod.config.testMode ? 'sandbox' : 'production';
+                const clientToken: PaypalClientToken = { [env]: paypalOptions.clientId };
+
+                const allowedSources = [];
+                const disallowedSources = [];
+
+                if (paypalOptions.allowCredit) {
+                    allowedSources.push(paypal.FUNDING.CREDIT);
+                } else {
+                    disallowedSources.push(paypal.FUNDING.CREDIT);
+                }
+
+                return paypal.Button.render({
+                    env,
+                    client: clientToken,
+                    commit: paypalOptions.shouldProcessPayment,
+                    funding: {
+                        allowed: allowedSources,
+                        disallowed: disallowedSources,
+                    },
+                    style: {
+                        shape: 'rect',
+                        ...pick(paypalOptions.style, 'layout', 'size', 'color', 'label', 'shape', 'tagline', 'fundingicons'),
+                    },
+                    payment: (data, actions) => this._setupPayment(merchantId, actions, paypalOptions.onPaymentError),
+                    onAuthorize: (data, actions) => this._tokenizePayment(data, actions, paypalOptions.shouldProcessPayment, paypalOptions.onAuthorizeError),
+                }, options.containerId);
+            })
+            .then(() => super.initialize(options));
+    }
+
+    deinitialize(): Promise<void> {
+        if (!Object.keys(this._isInitialized).length) {
+            return super.deinitialize();
+        }
+
+        this._paymentMethod = undefined;
+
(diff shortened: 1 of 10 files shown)


davidchin:
Hey @lord2800, the PR looks good in general, can you add a test to verify that the new property `funding` could be passed to `paypal.Button.render` function? Can you also remove the the `dist` folder? And then I can approve it. Thanks!

lord2800:
Yep, I have a whole lot of tests to write still.

lord2800:
This also needs updating to support smart and credit button options.

src/payment/strategies/paypal/paypal-sdk.ts
```diff
@@ -25,6 +31,73 @@ export interface PaypalButtonStyleOptions {
     fundingicons?: boolean;
 }
 
+export interface PaypalActions {
+    payment: PaypalPaymentActions;
+    request: PaypalRequestActions;
+}
+
+export interface PaypalPaymentActions {
+    create(payload: PaypalCreatePaymentPayload): Promise<any>;
```
davidchin: It'd be nice to define the type for the return value instead of `any`.

src/checkout-buttons/strategies/paypal-button-strategy.ts
```diff
@@ -0,0 +1,134 @@
...
+            throw new InvalidArgumentError();
+        }
+
+        if (!paymentMethod || !paymentMethod.config.merchantId) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        const env = paymentMethod.config.testMode ? 'sandbox' : 'production';
+        const clientToken: PaypalClientToken = {};
+        clientToken[env] = paypalOptions.clientId;
```
davidchin: You can just combine with the line above.

src/checkout-buttons/strategies/paypal-button-strategy.ts
```diff
@@ -0,0 +1,130 @@
...
+                if (onError) {
+                    onError(error);
+                }
+
+                throw error;
+            });
+    }
+
+    private _tokenizePayment(
+        data: any,
```
davidchin: Instead of `any`, can you define a type?

src/checkout-buttons/strategies/paypal-button-strategy.ts
```diff
@@ -0,0 +1,130 @@
...
+        actions?: PaypalActions,
+        shouldProcessPayment?: boolean,
+        onError?: (error: StandardError) => void
+    ): Promise<any> {
```
davidchin: Same here, can you define a type?

src/payment/strategies/paypal/paypal-sdk.ts
lord2800: Yeah, I'm not entirely sure what the type it returns even is. Their docs are ...confusing... at best.

davidchin: Yeah I understand. Just try your best. By the way, this is not a blocker.

src/checkout-buttons/strategies/braintree-paypal-button-strategy.ts
```diff
@@ -57,18 +57,27 @@ export default class BraintreePaypalButtonStrategy extends CheckoutButtonStrateg
                     disallowedSources.push(paypal.FUNDING.CREDIT);
                 }
 
+                let style: PaypalButtonStyleOptions;
+                if (this._offerCredit) {
```
lord2800: This actually needs to be backed out.

lord2800:
No longer WIP.

src/checkout-buttons/strategies/paypal-button-strategy.ts
```diff
@@ -0,0 +1,142 @@
...
+            return super.deinitialize();
+        }
+
+        this._paymentMethod = undefined;
+
+        return super.deinitialize();
+    }
+
+    private _setupPayment(merchantId: string | undefined, actions?: PaypalActions, onError?: (error: StandardError) => void): Promise<string> {
+        return this._store.dispatch(this._checkoutActionCreator.loadDefaultCheckout())
```
davidchin: Do you need to make a call to load the checkout object here? You don't seem to be using it all. For Braintree, we load the checkout object because we need pass the grand total and customer address to Braintree. If PPEC doesn't need to, you probably don't need to make that extra call.

src/checkout-buttons/strategies/paypal-button-strategy.ts
```diff
@@ -0,0 +1,142 @@
...
+            .then(state => {
+                const checkout = state.checkout.getCheckout();
+                const config = state.config.getStoreConfig();
+
+                if (!actions) {
+                    throw new NotInitializedError(NotInitializedErrorType.CheckoutButtonNotInitialized);
+                }
+
+                if (!checkout) {
+                    throw new MissingDataError(MissingDataErrorType.MissingCheckout);
```
davidchin: Can you see my comment above? If you're not using the object, you probably don't need to do a null check here as well.

src/checkout-buttons/strategies/paypal-button-strategy.ts
```diff
@@ -0,0 +1,142 @@
...
+                }
+
+                if (!config) {
+                    throw new MissingDataError(MissingDataErrorType.MissingCheckoutConfig);
```
davidchin: Same here, `config` object is not used in this function. So you can probably omit this check as well.

src/checkout-buttons/strategies/paypal-button-strategy.ts
```diff
@@ -0,0 +1,142 @@
...
+        if (!Object.keys(this._isInitialized).length) {
+            return super.deinitialize();
+        }
+
+        this._paymentMethod = undefined;
+
+        return super.deinitialize();
+    }
+
+    private _setupPayment(merchantId: string | undefined, actions?: PaypalActions, onError?: (error: StandardError) => void): Promise<string> {
```
davidchin: Do you really want to make `merchantId` an optional param? You can if you want to for this private method. But you should move the null check from `initialize` to here, so you're throwing an error if the valid is missing. This is similar to how you handle `actions` - how it is actually a required param but you're just doing the check here instead of `initialize`. Alternatively, you might want to make both params required and do the check from the caller, but I think it might be a little awkward in this case.

src/checkout-buttons/strategies/paypal-button-strategy.ts
```diff
@@ -0,0 +1,142 @@
...
+            throw new NotInitializedError(NotInitializedErrorType.CheckoutButtonNotInitialized);
+        }
+
+        if (!actions) {
+            throw new NotInitializedError(NotInitializedErrorType.CheckoutButtonNotInitialized);
+        }
+
+        const methodId = this._paymentMethod.id;
+
+        return actions.payment.get(data.paymentID || '').then(payload => {
```
davidchin: Does it work if you call `actions.payment.get` without a payment ID? If not, is it better if you don't make that request at all (i.e.: do a null check first)?

src/checkout-buttons/strategies/paypal-button-strategy.ts
lord2800: Remnant from the first iteration where I was using the client-side only version. I'll remove this.

PR #448, opened 2018-10-29
CHECKOUT-3695: Ignore trailing slash and other irrelevant information when comparing event origin

davidchin:
## What?
* When checking if `message` event comes from the expected origin, make sure the configured origin URL is without a trailing slash or other irrelevant information. 

## Why?
* Otherwise, we'll ignore messages coming from `https://bc.wordpress.com` if the origin is configured to be `https://bc.wordpress.com/`.

## Testing / Proof
* Unit

### DIFF
--- a/src/embedded-checkout/iframe-event-listener.spec.ts
+++ b/src/embedded-checkout/iframe-event-listener.spec.ts
@@ -39,6 +39,27 @@ describe('IframeEventListener', () => {
         expect(handleComplete).not.toHaveBeenCalled();
     });
 
+    it('does not respond to event with unrecognized origin', () => {
+        eventEmitter.emit('message', {
+            origin: 'https://foobar.com',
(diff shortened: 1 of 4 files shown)


src/embedded-checkout/iframe-event-listener.spec.ts
```diff
@@ -39,6 +39,27 @@ describe('IframeEventListener', () => {
...
+            origin: 'https://foobar.com',
+            data: {
+                type: EmbeddedCheckoutEventType.CheckoutLoaded,
+            },
+        });
+
+        expect(handleLoaded).not.toHaveBeenCalled();
+    });
+
+    it('handles origin URL with trailing slash', () => {
```
capsula4: `triggers relevant listeners when origin URL has trailing slash` 🍹

PR #452, opened 2018-11-05
fix(checkout-button): PAYMENTS-3071 Use the specified host for the paypal payment creation endpoint

lord2800:
## What?
Use the specified host from `createCheckoutButtonInitializer` options to send the request to create the payment method instead of a relative one.

## Why?
CORS errors abound when using the relative host because of the double redirect.

## Testing / Proof
Unit tested. Also:
screen shot 2018-11-05 at 3 41 12 pm

### DIFF
--- a/src/checkout-buttons/create-checkout-button-registry.ts
+++ b/src/checkout-buttons/create-checkout-button-registry.ts
@@ -22,7 +22,8 @@ import {
 
 export default function createCheckoutButtonRegistry(
     store: CheckoutStore,
-    requestSender: RequestSender
+    requestSender: RequestSender,
+    host?: string
 ): Registry<CheckoutButtonStrategy, CheckoutButtonMethodType> {
     const registry = new Registry<CheckoutButtonStrategy, CheckoutButtonMethodType>();
     const scriptLoader = getScriptLoader();
@@ -92,7 +93,8 @@ export default function createCheckoutButtonRegistry(
         new PaypalButtonStrategy(
             store,
             new PaypalScriptLoader(scriptLoader),
-            formPoster
+            formPoster,
+            host
         )
     );
 
--- a/src/checkout-buttons/strategies/paypal-button-strategy.spec.ts
+++ b/src/checkout-buttons/strategies/paypal-button-strategy.spec.ts
@@ -63,7 +63,11 @@ describe('PaypalButtonStrategy', () => {
         jest.spyOn(paypal.Button, 'render')
             .mockImplementation((options: PaypalButtonOptions) => {
                 eventEmitter.on('payment', () => {
-                    options.payment().catch(() => {});
+                    options.payment({
+                        payerId: 'PAYER_ID',
+                        paymentID: 'PAYMENT_ID',
+                        payerID: 'PAYER_ID',
+                    }, actionsMock).catch(() => {});
                 });
 
                 eventEmitter.on('authorize', () => {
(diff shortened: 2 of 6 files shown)


src/checkout-buttons/create-checkout-button-registry.ts
```diff
@@ -22,7 +22,8 @@ import {
 
 export default function createCheckoutButtonRegistry(
     store: CheckoutStore,
-    requestSender: RequestSender
+    requestSender: RequestSender,
+    host: string | undefined
```
icatalina:
```suggestion
    host?: string
```

src/checkout-buttons/strategies/paypal-button-strategy.ts
```diff
@@ -16,7 +16,8 @@ export default class PaypalButtonStrategy extends CheckoutButtonStrategy {
     constructor(
         private _store: CheckoutStore,
         private _paypalScriptLoader: PaypalScriptLoader,
-        private _formPoster: FormPoster
+        private _formPoster: FormPoster,
+        private _host: string | undefined
```
icatalina: same than before:

```suggestion
        private _host?: string
```

src/checkout-buttons/strategies/paypal-button-strategy.ts
```diff
@@ -87,7 +88,7 @@ export default class PaypalButtonStrategy extends CheckoutButtonStrategy {
             throw new NotInitializedError(NotInitializedErrorType.CheckoutButtonNotInitialized);
         }
 
-        return actions.request.post('/api/storefront/paypal-payment/', { merchantId })
+        return actions.request.post(`${this._host}/api/storefront/paypal-payment/`, { merchantId })
```
icatalina: What happens if host is undefined? `undefined/api/storefront/paypal-payment/`?

src/checkout-buttons/strategies/paypal-button-strategy.spec.ts
```diff
@@ -84,7 +88,8 @@ describe('PaypalButtonStrategy', () => {
         strategy = new PaypalButtonStrategy(
             store,
             paypalScriptLoader,
-            formPoster
+            formPoster,
+            ''
```
icatalina: If we are allowing the value to be `undefined`, all these tests should pass without an empty string.

lord2800: @davidchin so this is an area I'm not entirely sure about. Before my change, the behavior when you didn't pass a `host` was `undefined` since `options && options.host` returns `undefined` when `host` isn't provided. This is now more explicit by the fact that we're capturing it into a variable. WDYT?

davidchin: I think what Ignacio meant was because the last parameter - `host` - is optional, you don't need to make changes to the tests and they will still pass.

src/checkout-buttons/strategies/paypal-button-strategy.ts
```diff
@@ -87,8 +88,12 @@ export default class PaypalButtonStrategy extends CheckoutButtonStrategy {
             throw new NotInitializedError(NotInitializedErrorType.CheckoutButtonNotInitialized);
         }
 
-        return actions.request.post('/api/storefront/paypal-payment/', { merchantId })
-            .then(res => res.id)
+        return actions.request.post(`${this._host || ''}/api/storefront/paypal-payment/`, {
```
davidchin: Do you still need `|| ''` if `host` has a default value of `''`?

lord2800: Nope, good catch.

src/checkout-buttons/strategies/paypal-button-strategy.spec.ts
lord2800: Right, I was meaning overall I guess, rather than specific to this test.

src/checkout-buttons/strategies/paypal-button-strategy.ts
```diff
@@ -87,7 +88,13 @@ export default class PaypalButtonStrategy extends CheckoutButtonStrategy {
             throw new NotInitializedError(NotInitializedErrorType.CheckoutButtonNotInitialized);
         }
 
-        return actions.request.post('/api/storefront/paypal-payment/', { merchantId })
+        return actions.request.post(`${this._host}/api/storefront/paypal-payment/`, {
+            merchantId,
+        }, {
+            headers: {
+                'X-API-INTERNAL': 'This API endpoint is for internal use only and may change in the future',
```
davidchin: Oh I forgot, you can actually import this constant `INTERNAL_USE_ONLY` from `'common/http-request'`

lord2800: Oh, TIL. Will do.

PR #454, opened 2018-11-06
Typescript Error Tests

capsula4:
## What?
Convert `billing` and `error` tests to TS. Also add `code` attribute to `CouponErrorResponseBody` as it's a new API feature.

## Why?
For better TS coverage.


src/common/error/error-response-body.ts
```diff
@@ -1,4 +1,12 @@
 export default interface ErrorResponseBody {
```
davidchin: You're right. At the moment, we need three definitions - Storefront API, Internal API, and Payment API error responses. I think when the current definition was written, we only considered Internal API.

AFAIK, Storefront API error response is in the following format:

```ts
interface StorefrontErrorResponseBody {
    title: string;
    type: string;
    status: number;
    detail?: string;
    instance?: string;
}
```
It may contain additional properties depending on the endpoint, i.e.: `code` from the coupons endpoint.

The internal API error response is the following:

```ts
interface InternalErrorResponseBody {
    title: string;
    status: number;
    type: string;
    code?: int;
    detail?: string;
    instance?: string;
    errors: string[] | { [key: string]: string };
}
```

Payment API error response is the following:

```ts
interface PaymentErrorResponseBody {
    status: string;
    errors: Array<{ code: string, message: string }>;
}
```

So I'm thinking maybe we can do a union type to represent the variants? i.e.: `type ErrorResponseBody = StorefrontErrorResponseBody | InternalErrorResponseBody | PaymentErrorResponseBody`?

src/common/error/error-response-body.ts
```diff
@@ -1,4 +1,12 @@
...
+    status?: number | string;
     type?: string;
-    errors?: Array<{ code: string }>;
+    detail?: string;
+    errors?: ErrorListResponseBody;
 }
+
+// todo: create different errorLists, payments use objects, order use strings
+export type ErrorListResponseBody =
+    Array<{ code: string; message?: string; } | null | string>;
```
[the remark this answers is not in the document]
capsula4: just to accomodate for a test, not sure if it's a real scenario but since there is a test for it...
screen shot 2018-11-08 at 2 54 00 pm

src/common/error/error-response-body.ts
capsula4: yeah I was definitely thinking in doing that, but also thought in doing that change along with changes to RequestError, which at the moment seems to handle way too many scenarios (error message handling)

src/coupon/index.ts
```diff
@@ -2,6 +2,7 @@ export * from './coupon-actions';
 export * from './gift-certificate-actions';
 
 export { default as Coupon } from './coupon';
+export { default as CouponErrorResponseBody } from './errors/coupon-error-response-body';
```
davidchin: Can we have a `coupon/errors/index.ts` file? And export `CouponErrorResponseBody` from there?

src/common/error/errors/request-error.ts
```diff
@@ -1,32 +1,42 @@
...
     headers: {},
     status: 0,
     statusText: '',
 };
 
-export default class RequestError extends StandardError {
-    body: any;
+// todo: this request error accepts a TPayload, but there should be subclasses for different request errors
+// with well defined payloads (e.g. Internal Order API, Payment (BigPay) and Storefront API).
+export default class RequestError<TPayload extends ErrorResponseBody = any> extends StandardError {
```
davidchin: Is it possible to make the default type `ErrorResponseBody` instead of `any`?

capsula4:
@davidchin  Thanks for the review. I ended up refactoring the `RequestErrorFactory` to return different RequestError subclasses to handle different payloads appropriately.

src/common/error/request-error-factory.ts
```diff
@@ -27,15 +27,40 @@ export default class RequestErrorFactory {
...
-        return error && typeof error === 'object' && error.code ?
-            error.code :
-            'default';
+        return error ? error.code : 'default';
+    }
+
+    private _isStorefrontErrorResponseBody(
+        errorResponse: ErrorResponseBody
+    ): errorResponse is StorefrontErrorResponseBody {
+        const storefrontErrorResponse = errorResponse as StorefrontErrorResponseBody;
```
davidchin: I could be wrong, but I think you probably don't need to cast it as `StorefrontErrorResponseBody`.

src/common/error/error-response-body.ts
```diff
@@ -17,7 +18,7 @@ export interface InternalErrorResponseBody {
     code?: number;
     detail?: string;
     instance?: string;
-    errors: string[] | { [key: string]: string };
```
capsula4: couldn't find an example of internal API returning `errors: { [key: string]: string };`

PR #455, opened 2018-11-07
CHECKOUT-2934 Convert utility tests to TS

capsula4:
## What?
Convert tests to TS
* utility
* locale
* form
* customer

## Why?
For better TS coverage


src/common/utility/create-freeze-proxy.spec.ts
```diff
@@ -1,6 +1,8 @@
 import createFreezeProxy, { createFreezeProxies } from './create-freeze-proxy';
 
 class Foobar {
+    _data: any;
```
davidchin: Maybe add `protected` keyword?

PR #463, opened 2018-11-08
fix(payment): INT-1080 Round totalPrice to 2 decimals

vmparra:
## What?
As per described

## Why?
In production environment we are getting this error

image

## Testing / Proof
...

@bigcommerce/checkout

### DIFF
--- a/src/payment/strategies/googlepay/googlepay-braintree-initializer.ts
+++ b/src/payment/strategies/googlepay/googlepay-braintree-initializer.ts
@@ -1,3 +1,5 @@
+import { round } from 'lodash';
+
 import { Checkout } from '../../../checkout';
 import { MissingDataError, MissingDataErrorType } from '../../../common/error/errors';
 import PaymentMethod from '../../payment-method';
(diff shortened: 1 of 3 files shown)


src/payment/strategies/googlepay/googlepay-braintree-initializer.ts
```diff
@@ -79,7 +81,7 @@ export default class GooglePayBraintreeInitializer implements GooglePayInitializ
             transactionInfo: {
                 currencyCode: checkout.cart.currency.code,
                 totalPriceStatus: 'FINAL',
-                totalPrice: checkout.grandTotal.toString(),
+                totalPrice: round(checkout.grandTotal, 2).toString(),
```
davidchin: Can I ask, why do you need to manually do the rounding? The grand total amount should already be rounded according to the current settings configured by the merchant.

src/payment/strategies/googlepay/googlepay-braintree-initializer.ts
```diff
@@ -79,7 +79,7 @@ export default class GooglePayBraintreeInitializer implements GooglePayInitializ
             transactionInfo: {
                 currencyCode: checkout.cart.currency.code,
                 totalPriceStatus: 'FINAL',
-                totalPrice: checkout.grandTotal.toString(),
+                totalPrice: checkout.grandTotal.toFixed(2),
```
davidchin: Can you do `round(checkout.grandTotal, 2).toFixed(2)`? Because some currencies have more than 2 decimal places, so we want to round the number first before padding the number with 0.

vmparra:
@davidchin ♻️ Changes have been addressed. Can you please re-review?

PR #466, opened 2018-11-13
CHECKOUT-3706: Switch to different height calculation method if `contentId` is provided.

davidchin:
## What?
* Switch to a different method to calculate the height of the iframe if `bodyId` is passed.

## Why?
* You might want to exclude certain elements from the calculation. For example, you could have an element that's fixed to the bottom of the page. That could cause a problem if the calculation method is set to `lowestElement`. This is because when the page content shrinks, the fixed element will remain in the same position because its position is relative to the iframe rather than the content. Also, because it is the lowest element on the page, therefore the iframe itself won't reduce its height.

## Testing / Proof
* Unit

### DIFF
--- a/src/embedded-checkout/iframe-content/iframe-embedded-checkout-messenger.spec.ts
+++ b/src/embedded-checkout/iframe-content/iframe-embedded-checkout-messenger.spec.ts
@@ -101,4 +101,21 @@ describe('EmbeddedCheckoutMessenger', () => {
 
         expect(handler).toHaveBeenCalledWith(styles);
     });
+
+    it('invokes message callbacks if registered', () => {
+        const handler = jest.fn();
+
+        messenger = new IframeEmbeddedCheckoutMessenger(
+            messageListener,
+            messagePoster,
+            { [EmbeddedCheckoutEventType.FrameLoaded]: handler }
+        );
+
+        messenger.postFrameLoaded({ contentId: 'foobar' });
+
+        expect(handler).toHaveBeenCalledWith({
+            type: EmbeddedCheckoutEventType.FrameLoaded,
+            payload: { contentId: 'foobar' },
+        });
+    });
 });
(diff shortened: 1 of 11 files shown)


src/embedded-checkout/iframe-content/embedded-content-event-handler.ts
```diff
@@ -0,0 +1,23 @@
+import { EmbeddedCheckoutEvent, EmbeddedCheckoutEventType, EmbeddedCheckoutFrameLoadedEvent } from '../embedded-checkout-events';
+
+export default class EmbeddedContentEventHandler {
+    handle(message: EmbeddedCheckoutEvent): void {
```
capsula4: I was wondering if instead of having a `EmbeddedContentEventHandler`, we should be able to register different handlers against different event types? 🍹

src/embedded-checkout/iframe-content/iframe-embedded-checkout-messenger.spec.ts
```diff
@@ -4,13 +4,15 @@ import { EmbeddedCheckoutEvent, EmbeddedCheckoutEventType } from '../embedded-ch
...
 
+import EmbeddedContentEventHandler from './embedded-content-event-handler';
 import { EmbeddedContentEventMap, EmbeddedContentEventType } from './embedded-content-events';
 import IframeEmbeddedCheckoutMessenger from './iframe-embedded-checkout-messenger';
 
 describe('EmbeddedCheckoutMessenger', () => {
     let messenger: IframeEmbeddedCheckoutMessenger;
     let messageListener: IframeEventListener<EmbeddedContentEventMap>;
     let messagePoster: IframeEventPoster<EmbeddedCheckoutEvent>;
+    let messageHandler: EmbeddedContentEventHandler;
```
capsula4: maybe you could be passing a `messageHandlerRepository` ? 🍹

PR #467, opened 2018-11-13
CHECKOUT-2934 Convert payment instrument tests to TypeScript

capsula4:
## What?
Convert payment instrument tests to TypeScript

## Why?
For 100% TS support


src/payment/bigpay-response.ts
```diff
@@ -0,0 +1,6 @@
+export default interface BigpayResponse<T = any> {
```
davidchin: Can we name it as `PaymentResponse` instead of using an internal name i.e.: `BigPay`?

PR #473, opened 2018-11-15
fix(checkout-button): PAYMENTS-3071 Use the specified host for paypal payment creation

lord2800:
## What?
Use the specified host for paypal payment creation and provide the cart id to it along with the internal-only header.

## Why?
The endpoint used to rely on the session to retrieve the current cart, which is a bad idea.

## Testing / Proof
Unit tested.

### DIFF
--- a/src/checkout-buttons/strategies/paypal/paypal-button-strategy.ts
+++ b/src/checkout-buttons/strategies/paypal/paypal-button-strategy.ts
@@ -1,13 +1,13 @@
 import { FormPoster } from '@bigcommerce/form-poster';
 import { pick } from 'lodash';
 
-import { CheckoutStore } from '../../../checkout';
+import { CheckoutActionCreator, CheckoutStore } from '../../../checkout';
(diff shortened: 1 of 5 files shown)


src/checkout-buttons/strategies/paypal/paypal-button-strategy.ts
```diff
@@ -34,12 +36,17 @@ export default class PaypalButtonStrategy extends CheckoutButtonStrategy {
             throw new InvalidArgumentError();
         }
 
-        return this._paypalScriptLoader.loadPaypal()
-            .then(paypal => {
+        return Promise.all([
+            this._paypalScriptLoader.loadPaypal(),
+            this._store.dispatch(this._checkoutActionCreator.loadDefaultCheckout()),
```
davidchin: It seems you only need the cart information inside the `payment` callback. Is it possible for you to make that call inside the callback instead? This way, we can render the button faster. And we only make the additional network work when the shopper clicks on the PayPal button.

lord2800: Yep, I was thinking the same thing after I left. I'll do that shortly.

PR #475, opened 2018-11-18
CHECKOUT-3462: Add `cacheAction` decorator for caching asynchronous actions

davidchin:
## What?
* Add `cacheAction` decorator for caching asynchronous actions.
* Example:
```ts
class FooActions {
    loadFoo(id: number, options?: ActionOptions): Observable<Action<Foo>> {
        // Do some stuff and return an observable action
    }
}

const actions = new FooActions();

actions.loadFoo(1, { useCache: true });
actions.loadFoo(1, { useCache: true }); // The second call will use the cached value of the first call
actions.loadFoo(2, { useCache: true }); // This call will not use the cached value because the parameters are different
```

## Why?
* Previously, if an action is executed twice in parallel, it won't be able to return cached data because the cache is not available at the time of execution. Now, with this change, it can because the decorator internally caches the action and replays the cached value for the second call as soon as it becomes available.

## Testing / Proof
* Unit


src/checkout-buttons/checkout-button-strategy-action-creator.spec.ts
```diff
@@ -55,29 +45,19 @@ describe('CheckoutButtonStrategyActionCreator', () => {
...
-        await Observable.from(strategyActionCreator.initialize(options)(store))
+        await strategyActionCreator.initialize(options)
             .toArray()
             .toPromise();
 
         expect(paymentMethodActionCreator.loadPaymentMethod)
-            .not.toBeCalled();
-    });
-
-    it('does not load payment method if already loaded required payment method', async () => {
```
icatalina: It seems like this is still a valid test to keep. Or at least cover the case of not reloading the payment method if it is already loaded/loading.

PR #480, opened 2018-11-21
fix(payments): INT-1119 Modify approach in Checkout Button Strategies to consume unique ids

metalvegetarianoprogresivo:
## What? [INT-1119](https://jira.bigcommerce.com/browse/INT-1119)
This fixes the issue created by now using the unique ids on the templates for GooglePay and MasterPass.

## Why?
As per bigcommerce/bigcommerce#27267 we need to adjust both Masterpass & Google Pay checkout button strategies to consume unique ids

## Testing / Proof
image
image
image
image

@bigcommerce/checkout

### DIFF
--- a/src/checkout-buttons/strategies/googlepay/googlepay-button-strategy.ts
+++ b/src/checkout-buttons/strategies/googlepay/googlepay-button-strategy.ts
@@ -56,7 +56,7 @@ export default class GooglePayButtonStrategy extends CheckoutButtonStrategy {
     }
 
     private _createSignInButton(containerId: string): HTMLElement {
-        const container = document.querySelector(`#${containerId}`);
+        const container = document.getElementById(containerId);
 
         if (!container) {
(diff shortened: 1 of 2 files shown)


src/checkout-buttons/strategies/googlepay/googlepay-button-strategy.ts
```diff
@@ -56,7 +56,7 @@ export default class GooglePayButtonStrategy extends CheckoutButtonStrategy {
     }
 
     private _createSignInButton(containerId: string): HTMLElement {
-        const container = document.querySelector(`#${containerId}`);
+        const container = document.getElementById(`${containerId}`);
```
davidchin: Can you remove the string interpolation? Just `document.getElementById(containerId)` is fine.

metalvegetarianoprogresivo:
@davidchin Just updated the PR to addressing your comments. Can you review it, please?

PR #499, opened 2018-12-10
CHECKOUT-3797 Surface transaction_declined error message

icatalina:
## What?
- As per title

## Why?
- We need to surface these error messages for shoppers to understand what the problem is

## Testing / Proof
- Unit / Functional

### Manually tested:
- [x] transaction_declined error message gets exposed for consumers


src/common/error/errors/map-from-payment-error-response.ts
```diff
@@ -30,3 +30,7 @@ function joinErrors(errors: Array<{ code: string, message?: string }>): string |
         return result;
     }, []).join(' ');
 }
+
+export const ALTERNATIVE_PAYMENT_ERROR_CODES = [
```
PascalZajac: When we say "alternative", alternative to _what_?

icatalina: to `payment`...

PascalZajac: 😕 I still don't get it. As in, I would expect this variable to imply the presence of `PAYMENT_ERROR_CODES` somewhere and this is an alternative set of reasons something fails, but I don't see anything like that (either here or where this variable is consumed).

icatalina: I'll add `payment` as a `PAYMENT_ERROR_CODE` and remove the `ALTERNATIVE_` prefix, it looks like this can cause some confusion.

icatalina: Can you have a new look? I've added a `fixup` commit.

patel-biren:
💚

PR #502, opened 2018-12-11
CHECKOUT-3714 Convert payment strategies to enum

capsula4:
## What?
Use enums for PaymentStrategies

## Why?
For more robust code / stricter types

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -295,7 +295,7 @@
     },
     "@types/query-string": {
       "version": "5.1.0",
-      "resolved": "http://registry.npmjs.org/@types/query-string/-/query-string-5.1.0.tgz",
+      "resolved": "https://registry.npmjs.org/@types/query-string/-/query-string-5.1.0.tgz",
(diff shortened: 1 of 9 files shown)


src/payment/strategies/braintree/visacheckout.mock.ts
```diff
@@ -49,7 +49,7 @@ export function getPaymentSuccessPayload(): VisaCheckoutPaymentSuccessPayload {
             countryCode: 'ES',
             postalCode: '2007',
         },
-        paymentMethodType: 'TOKEN',
```
davidchin: This change seems to be a mistake?

src/payment/strategies/braintree/visacheckout.ts
```diff
@@ -115,7 +115,7 @@ export interface VisaCheckoutPaymentSuccessPayload {
         countryCode: string;
         postalCode: string;
     };
-    paymentMethodType: 'PAN' | 'TOKEN';
```
davidchin: This change seems to be a mistake?

PR #503, opened 2018-12-13
fix(payment): INT-1079 Add taxTotal in map to internal order

metalvegetarianoprogresivo:
## What? [INT-1079](https://jira.bigcommerce.com/browse/INT-1079)
This PR adds the object `taxTotal` to the `map-to-internal-order`(also we add the interfaces for `order` and `internal-order` and their corresponding mocks), so once the information reach `bigpay-client`, it will catch the `taxTotal` and map it correctly. Currently `checkout-sdk-js` is not sending `taxTotal` as part of the payload received by `bigpay-client` and this creates a bug in `bigpay`.

## Why?
There was an issue where the transaction_rbits were not generated because `taxTotal` was not sent to `bigpay-client` and this creates a bug in `bigpay`. This fixes the issue on `checkout-sdk-js`.

## Testing / Proof
image
image
[Video of a purchase using Chasepay+Wepay](https://drive.google.com/file/d/1GxGy0fU_u4amneDBVmfT6HUQ8L0dIUvf/view?usp=sharing)

This PR depends on bigcommerce/bigcommerce#27647

@vmparra @clopezh @cbandam  @bigcommerce/checkout

### DIFF
--- a/src/order/internal-order.ts
+++ b/src/order/internal-order.ts
@@ -33,6 +33,10 @@ export default interface InternalOrder {
         amount: number;
     };
     taxes: Array<{ name: string, amount: number }>;
+    taxTotal: {
+        amount: number;
+        integerAmount: number;
+    };
     handling: {
         amount: number;
         integerAmount: number;
--- a/src/order/internal-orders.mock.ts
+++ b/src/order/internal-orders.mock.ts
@@ -170,6 +170,10 @@ export function getCompleteOrder(): InternalOrder {
                 amount: 3,
             },
         ],
+        taxTotal: {
+            amount: 0,
+            integerAmount: 0,
+        },
         handling: {
             amount: 8,
             integerAmount: 800,
--- a/src/order/internal-orders.mock.ts
+++ b/src/order/internal-orders.mock.ts
@@ -171,8 +171,8 @@ export function getCompleteOrder(): InternalOrder {
             },
         ],
         taxTotal: {
-            amount: 0,
-            integerAmount: 0,
+            amount: 3,
+            integerAmount: 300,
         },
         handling: {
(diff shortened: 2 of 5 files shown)


icatalina:
Hi @metalvegetarianoprogresivo!

I know this is a WIP, but the title/description of this PR makes no sense on the context of this repo. There's no reference to bigpay, wepay or transctions rbits anywhere in this PR.

The title in the PR should describe what you are changing in this repo. In this case, the change seems to be adding TaxTotal values to the internal order mapper.

You can explain in more detail the reasons why you are doing the change within the _why_ block.

metalvegetarianoprogresivo:
@icatalina got it, i'll update the tittle.

metalvegetarianoprogresivo:
@icatalina @serroba @Sure-Round 
I just pushed a [commit](https://github.com/bigcommerce/checkout-sdk-js/pull/503/commits/36f6ed5d381d964ad6710c53d659a659f068e3e9), that tackles the approach [commented here](https://github.com/bigcommerce/bigcommerce/pull/27647#issuecomment-447839251), adding up the `taxes` to obtain the `taxTotal` without touching the API. If you decide to use this approach we can close bigcommerce/bigcommerce#27647. Can you look at it?

icatalina:
We don't know if that'll be the final approach, or if there are issues doing the addition in the FE. It might be better if we wait until the other PR gets resolved.

metalvegetarianoprogresivo:
We just addressed the comment on bigcommerce/bigcommerce#27647 about the documentation and testing.

Sure-Round:
@icatalina I'm merging this since the sibling PR is now in all 3 tiers.

PR #507, opened 2018-12-27
CHECKOUT-3798 Throw CheckoutNotAvailable if the response is in the 400 range

icatalina:
## What?
- As per title

## Why?
- We need to distinguish between errors (5xx) and missing checkouts

## Testing / Proof
- Unit


src/config/config.ts
```diff
@@ -58,6 +58,7 @@ export interface StoreLinks {
     checkoutLink: string;
     createAccountLink: string;
     forgotPasswordLink: string;
+    loginLink: string;
```
PascalZajac: Was this missed in Robin's PR?

icatalina: This is new. I need to know where to redirect the user when they logout.

icatalina:
The experience will depend on the implementation. For our checkout the experience will be a redirect to the login page (no error modal), hence the `loginLink` url.

PR #516, opened 2019-01-29
CHECKOUT-3890: Rehydrate shipping options after applying coupon

davidchin:
## What?
* Include available shipping options when applying or removing coupons.

## Why?
* We should rehydrate shipping options because coupons may affect shipping quotes.

## Testing / Proof
* Unit

### DIFF
--- a/src/common/http-request/join-includes.ts
+++ b/src/common/http-request/join-includes.ts
@@ -1,0 +1,5 @@
+import { uniq } from 'lodash';
+
+export default function joinIncludes(includes: string[]): string {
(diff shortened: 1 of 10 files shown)


src/common/http-request/join-includes.ts
```diff
@@ -0,0 +1,5 @@
+import { uniq } from 'lodash';
+
+export default function joinIncludes(includes: string[]): string {
+    return uniq(includes).join(',');
```
icatalina: Nice 🙂

PR #527, opened 2019-02-22
PAYMENTS-4034 catch error when nonce return 400, squarev2

jonathandelefortrie:
## What?
catch error when the nonce request fail

## Why?
the nonce return an error for wrong card details entered, the error gets lost and the submit button stuck in spinning state.

## Testing / Proof
Coming...

### DIFF
--- a/src/payment/strategies/square/square-payment-strategy.ts
+++ b/src/payment/strategies/square/square-payment-strategy.ts
@@ -1,5 +1,6 @@
 import { RequestSender, Response } from '@bigcommerce/request-sender';
 import { omit } from 'lodash';
+import { noop } from 'rxjs';
 
 import { CheckoutActionCreator, CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
 import {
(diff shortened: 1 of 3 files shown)


src/payment/strategies/square/square-payment-strategy.ts
```diff
@@ -190,8 +190,10 @@ export default class SquarePaymentStrategy implements PaymentStrategy {
             throw new StandardError();
         }
 
-        if (errors && this._squareOptions && this._squareOptions.onError) {
-            this._squareOptions.onError(errors);
+        if (errors) {
+            if (this._squareOptions && this._squareOptions.onError) {
+                this._squareOptions.onError(errors);
+            }
             this._deferredRequestNonce.reject(errors);
```
capsula4: I would prefer to always reject it even if `errors` is empty, just to avoid a promise ended up being on hold. But this is still an improvement on what we currently have. Maybe cover this bugfix by adding a new unit test?

jonathandelefortrie: hey @capsula4 , thanks for the feedback, I'm going to delegate that PR to you as I'm closing my laptop in few minutes.

bc-anna:
🙇 thanks for jumping onto this one to close it out @capsula4 @davidchin

PR #529, opened 2019-02-27
feat(payments): INT-1266 payment strategy for affirm. 

dbarriga:
## What? [INT-1266](https://jira.bigcommerce.com/browse/INT-1266l)
Create affirm payment strategy.

## Why?
In order to accept Affirm as a payment method, the strategy needs to be created and set to communicate stores and Affirm. 

## Siblings PR
[BCAPP#28859](https://github.com/bigcommerce/bigcommerce/pull/28859)
[BIGPAY#1554](https://github.com/bigcommerce/bigpay/pull/1554)
[NG-CHEKOUT#1108](https://github.com/bigcommerce-labs/ng-checkout/pull/1108)

## Testing / Proof
- [ ] QE Approval

screen shot 2019-02-27 at 12 49 26 pm
screen shot 2019-02-27 at 12 49 43 pm
screen shot 2019-02-27 at 12 54 26 pm


src/payment/strategies/affirm/affirm-payment-strategy.ts
```diff
@@ -0,0 +1,249 @@
...
+        const publicApiKey = paymentMethod.initializationData.publicKey;
+
+        affirmJs(publicApiKey, this._getScriptURI(testMode));
+
+        return Promise.resolve(this._store.getState());
+    }
+
+    execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        const paymentId = payload.payment && payload.payment.methodId;
+        const useStoreCredit = !!payload.useStoreCredit;
```
vmparra: Please avoid coercion.

dbarriga:
@vmparra changes are ready.

src/payment/strategies/affirm/affirm-payment-strategy.ts
```diff
@@ -0,0 +1,273 @@
...
+
+        if (!paymentId) {
+            throw new PaymentArgumentInvalidError(['payment.methodId']);
+        }
+
+        if (!config) {
+            throw new MissingDataError(MissingDataErrorType.MissingCheckoutConfig);
+        }
+
+        return this._store.dispatch(this._remoteCheckoutActionCreator.initializePayment(paymentId, { useStoreCredit }))
```
davidchin: I guess the reason why you have to post `useStoreCredit` via the remote endpoint is so that you can persist that setting the session and retrieve it back when you're redirected back. As I said, we need to reconsider this approach. It's merely a temporary workaround for AfterPay and we should aim to not replicate it for Affirm. Plus there are other reasons.

src/payment/strategies/affirm/affirm-payment-strategy.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+            await new Promise(resolve => process.nextTick(resolve));
+        });
+
+        it('notifies store credit usage to remote checkout service', () => {
+            expect(remoteCheckoutActionCreator.initializePayment).toHaveBeenCalledWith( paymentMethod.id, { useStoreCredit: false });
+            expect(store.dispatch).toHaveBeenCalledWith(initializePaymentAction);
+        });
+
+        it('does not resolve if execution is successful', () => {
+            expect(successHandler).not.toHaveBeenCalled();
```
davidchin: `successHandler` also doesn't get called when there's a failure. So I'm not sure this is the correct verification.

src/payment/strategies/affirm/affirm-payment-strategy.ts
```diff
@@ -0,0 +1,273 @@
...
+            .then(affirm.checkout(this._initializeCheckout(useStoreCredit)))
+            .then(affirm.checkout.open())
+            .then(affirm.ui.ready(
+                () => {
+                    affirm.ui.error.on('close', () => {
+                        window.location.href = `${config.links.checkoutLink}.php?action=set_external_checkout&provider=affirm&status=cancelled`;
```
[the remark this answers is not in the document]
davidchin: Also, I suppose Affirm presents some custom UI form to the shopper and asks them to interact with it before redirecting them elsewhere? If so, and if there could be a point of failure after the `initialize` promise is resolved, we should provide an error callback so there's a way for us to inform the client / app about the failure so it can decide how to handle it.

src/payment/strategies/affirm/affirm-payment-strategy.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+
+        payload = merge({}, getOrderRequestBody(), {
+            payment: {
+                methodId: paymentMethod.id,
+                gatewayId: paymentMethod.gateway,
+            },
+        });
+    });
+
+    describe('#initialize()', () => {
```
davidchin: There's a whole bunch of logic around preparing the data required for Affirm script to initialize - in fact most of the code is for that purpose. How come we don't see any tests around it?

src/payment/strategies/affirm/affirm-payment-strategy.ts
```diff
@@ -0,0 +1,273 @@
...
+                const orderPayload = {
+                    useStoreCredit: affirm.settings.useStoreCredit,
+                };
+
+                const paymentPayload = {
+                    methodId: payment.providerId,
+                    paymentData: { nonce: paymentMethod.initializationData.token },
+                };
+
+                return this._store.dispatch(this._orderActionCreator.submitOrder(orderPayload, options))
```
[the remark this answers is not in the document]
snaderiBC: The disadvantage of going to 3rd party site (like Affirm in this case) with Cart instead of Order is the cart amount could change. It doesn't make sense to clear payment first and then create an order for that order.
Keep in mind that we finalise payment when BigPay sends an asynchronous notification request to BC to confirm the payment has been cleared.
As David mentioned above we should avoid creating these special cases.

src/payment/strategies/affirm/affirm-payment-strategy.ts
```diff
@@ -0,0 +1,273 @@
...
+        if (!paymentMethod ) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        const { testMode } = paymentMethod.config;
+        const { publicKey } = paymentMethod.initializationData;
+
+        affirmJs(publicKey, this._getScriptURI(testMode));
+
+        return Promise.resolve(this._store.getState());
```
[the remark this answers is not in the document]
dbarriga: Hey David, I've been reading affirm documentation and, it does not appear this is an async function. It does not provide any callback nor manage asynchronously in any of their examples, is more of an initialization that works on `window` object.

michxff:
Hi @davidchin, thank you for the feedback, after checking the `PaymentMethod` interface it seems that the publicKey should be on the `clientToken` field and the authorization token should be on the `nonce `field, what do you think if we use this approach instead of the `initializationData`

dbarriga:
Closing pull request due to not using redirect flow in Affirm and ticket removal.

PR #533, opened 2019-03-06
CHECKOUT-3852 Expose proper error type for Coupon/GiftCertificates

capsula4:
## What?
As per title

## Why?
So consumers can access more information like why coupon failed

@bigcommerce/checkout

### DIFF
--- a/src/checkout/checkout-store-error-selector.ts
+++ b/src/checkout/checkout-store-error-selector.ts
@@ -1,5 +1,6 @@
 import { BillingAddressSelector } from '../billing';
 import { CartSelector } from '../cart';
+import { RequestError } from '../common/error/errors';
 import { selector } from '../common/selector';
 import { ConfigSelector } from '../config';
 import { CouponSelector, GiftCertificateSelector } from '../coupon';
@@ -330,7 +331,7 @@ export default class CheckoutStoreErrorSelector {
      *
      * @returns The error object if unable to apply, otherwise undefined.
      */
-    getApplyCouponError(): Error | undefined {
+    getApplyCouponError(): RequestError | undefined {
         return this._coupons.getApplyError();
     }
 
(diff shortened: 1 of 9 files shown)


src/coupon/coupon-selector.ts
```diff
@@ -13,11 +14,11 @@ export default class CouponSelector {
         return this._coupon.data;
     }
 
-    getRemoveError(): Error | undefined {
+    getRemoveError(): RequestError | undefined {
```
davidchin: Do we need to specify the error response type?

src/checkout/checkout-store-error-selector.ts
```diff
@@ -330,7 +332,7 @@ export default class CheckoutStoreErrorSelector {
      *
      * @returns The error object if unable to apply, otherwise undefined.
      */
-    getApplyCouponError(): Error | undefined {
+    getApplyCouponError(): RequestError<StorefrontErrorResponseBody> | undefined {
```
davidchin: Correct me if I'm wrong, but I think these getters return more than just `RequestError`. Any error that's thrown inside an observable action will be surfaced through here. So we might still need to have `Error` or `CustomError`.

capsula4: at the moment this error is generated in the couponActionCreator, when the promise from couponRequestSender fails, so I think it's safe to assume it's RequestError in this case.

src/coupon/coupon-selector.ts
capsula4: Actually I think we might not want to. The generic type is the body of the original error response, but part of the work I did was to standardise how we return RequestErrors from the 3 different APIs that we interact with (internal, payments, storefront), so we shouldn't really need to access the original API payload (even though it's available)

PR #535, opened 2019-03-08
CHECKOUT-3967 Round properly when using CurrencyService

capsula4:
## What?
Make currency service round instead of truncate

## Why?
It's a more accurate behaviour

## Testing / Proof
Unit

@bigcommerce/checkout

### DIFF
--- a/src/currency/currency-formatter.ts
+++ b/src/currency/currency-formatter.ts
@@ -57,22 +57,16 @@ export default class CurrencyFormatter {
 
     private _formatNumber(amount: number): string {
         const positiveAmount = Math.abs(amount);
-        const [ integerAmount, decimalAmount = '' ] = positiveAmount.toString().split('.');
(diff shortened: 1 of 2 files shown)


src/currency/currency-formatter.ts
```diff
@@ -57,22 +57,16 @@ export default class CurrencyFormatter {
 
     private _formatNumber(amount: number): string {
         const positiveAmount = Math.abs(amount);
-        const [ integerAmount, decimalAmount = '' ] = positiveAmount.toString().split('.');
+        const [ integerAmount, decimalAmount = '' ] = (positiveAmount.toFixed(this._decimalPlaces)).toString().split('.');
```
davidchin: I remember `toFixed` is actually not a reliable way to round decimals. For example `859.385.toFixed(2)` returns `859.38` instead of `859.39`. I found this [StackOverflow post](https://stackoverflow.com/questions/10015027/javascript-tofixed-not-rounding) which might help you to overcome this issue.

PR #536, opened 2019-03-11
feat(checkout): INT-1245 Checkout Using Zip

FranciscoOnt:
[INT-1245](https://jira.bigcommerce.com/browse/INT-1245)

## What?
Create the Zip Payment strategy along with the required interfaces and tests

## Why?
A payment strategy is required to support zip on the checkout

## Testing / Proof
Pending Deploy(#)
ZIp lightbox demo gif

Coverage: 
Screen Shot 2019-03-22 at 1 05 01 PM

@cmedinilla 
@bigcommerce/checkout @bigcommerce/payments @bigcommerce/integrations

### DIFF
--- a/src/payment/strategies/zip/zip-payment-strategy.spec.ts
+++ b/src/payment/strategies/zip/zip-payment-strategy.spec.ts
@@ -1,0 +1,228 @@
+import { createClient as createPaymentClient } from '@bigcommerce/bigpay-client';
+import { createAction, Action } from '@bigcommerce/data-store';
+import { createRequestSender, RequestSender } from '@bigcommerce/request-sender';
+import { createScriptLoader, ScriptLoader } from '@bigcommerce/script-loader';
+import { of, Observable } from 'rxjs';
+
+import { getCartState } from '../../../cart/carts.mock';
+import { createCheckoutStore, CheckoutRequestSender, CheckoutStore, CheckoutValidator } from '../../../checkout';
+import { getCheckoutState } from '../../../checkout/checkouts.mock';
+import { NotInitializedError } from '../../../common/error/errors';
+import { getConfigState } from '../../../config/configs.mock';
+import { getCustomerState } from '../../../customer/customers.mock';
+import { OrderActionCreator, OrderActionType, OrderRequestBody } from '../../../order';
+import { OrderFinalizationNotRequiredError } from '../../../order/errors';
+import { getOrderRequestBody } from '../../../order/internal-orders.mock';
+import { PaymentMethod, PaymentMethodActionCreator } from '../../../payment';
+import { getPaymentMethodsState, getZip } from '../../../payment/payment-methods.mock';
+import { getZipScriptMock } from '../../../payment/strategies/zip/zip.mock';
+import { PaymentMethodCancelledError, PaymentMethodInvalidError } from '../../errors';
+import { PaymentRequestSender } from '../../index';
+import PaymentActionCreator from '../../payment-action-creator';
+import { PaymentActionType } from '../../payment-actions';
+import PaymentMethodRequestSender from '../../payment-method-request-sender';
+import { PaymentInitializeOptions } from '../../payment-request-options';
+import PaymentStrategy from '../payment-strategy';
+
+import { Zip } from './zip';
+import ZipPaymentStrategy from './zip-payment-strategy';
+import ZipScriptLoader from './zip-script-loader';
+
+describe('ZipPaymentStrategy', () => {
+
+    let orderActionCreator: OrderActionCreator;
+    let paymentActionCreator: PaymentActionCreator;
+    let paymentMethodActionCreator: PaymentMethodActionCreator;
+    let paymentMethodMock: PaymentMethod;
+    let requestSender: RequestSender;
+    let store: CheckoutStore;
+    let scriptLoader: ScriptLoader;
+    let strategy: PaymentStrategy;
+    let submitOrderAction: Observable<Action>;
+    let submitPaymentAction: Observable<Action>;
+    let zipClient: Zip;
+    let zipScriptLoader: ZipScriptLoader;
+
+    beforeEach(() => {
+        paymentMethodMock = getZip();
+        scriptLoader = createScriptLoader();
+
+        store = createCheckoutStore({
+            checkout: getCheckoutState(),
+            customer: getCustomerState(),
+            config: getConfigState(),
+            cart: getCartState(),
+            paymentMethods: getPaymentMethodsState(),
+        });
+
+        zipClient = getZipScriptMock('approved');
+        zipScriptLoader = new ZipScriptLoader(scriptLoader);
+        requestSender = createRequestSender();
+
+        const paymentClient = createPaymentClient(store);
+        const paymentMethodRequestSender = new PaymentMethodRequestSender(requestSender);
+        const paymentRequestSender = new PaymentRequestSender(createPaymentClient());
+
+        orderActionCreator = new OrderActionCreator(paymentClient, new CheckoutValidator(new CheckoutRequestSender(createRequestSender())));
+        paymentActionCreator = new PaymentActionCreator(paymentRequestSender, orderActionCreator);
+        paymentMethodActionCreator = new PaymentMethodActionCreator(paymentMethodRequestSender);
+        submitOrderAction = of(createAction(OrderActionType.SubmitOrderRequested));
+        submitPaymentAction = of(createAction(PaymentActionType.SubmitPaymentRequested));
+
+        jest.spyOn(orderActionCreator, 'submitOrder')
+            .mockReturnValue(submitOrderAction);
+        jest.spyOn(paymentActionCreator, 'submitPayment')
+            .mockReturnValue(submitPaymentAction);
+        jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod')
+            .mockResolvedValue(store.getState());
+        jest.spyOn(store, 'dispatch')
+            .mockResolvedValue(store.getState());
+        jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod')
+            .mockReturnValue(paymentMethodMock);
+        jest.spyOn(zipClient.Checkout, 'init');
+        jest.spyOn(zipScriptLoader, 'load')
+            .mockResolvedValue(zipClient);
+
+        strategy = new ZipPaymentStrategy(
+            store,
+            orderActionCreator,
+            paymentActionCreator,
+            paymentMethodActionCreator,
+            zipScriptLoader
+        );
+    });
+
+    describe('#initialize()', () => {
+        let zipOptions: PaymentInitializeOptions;
+
+        beforeEach(() => {
+            zipOptions = { methodId: 'zip' };
+        });
+
+        it('initializes the strategy successfully', async () => {
+            await strategy.initialize(zipOptions);
+
+            expect(zipScriptLoader.load).toHaveBeenCalled();
+        });
+    });
+
+    describe('#execute()', () => {
+        let zipOptions: PaymentInitializeOptions;
+        let orderRequestBody: OrderRequestBody;
+
+        beforeEach(async () => {
+            orderRequestBody = {
+                ...getOrderRequestBody(),
+                payment: {
+                    methodId: 'zip',
+                },
+            };
+            zipOptions = { methodId: 'zip' };
+
+            await strategy.initialize(zipOptions);
+        });
+
+        it('executes the strategy successfully and submits the payment', async () => {
+            const { payment, ...order } = orderRequestBody;
+            const expectedPayment = {
+                methodId: 'zip',
+                paymentData: {
+                    nonce: 'checkoutId',
+                },
+            };
+            const zipInitPayload = {
+                onComplete: expect.any(Function),
+                onCheckout: expect.any(Function),
+            };
+            await strategy.execute(orderRequestBody, zipOptions);
+
+            expect(orderActionCreator.submitOrder).toHaveBeenCalledWith(order, zipOptions);
+            expect(paymentActionCreator.submitPayment).toHaveBeenCalledWith(expectedPayment);
+            expect(store.dispatch).toHaveBeenCalledWith(submitOrderAction);
+            expect(store.dispatch).toHaveBeenCalledWith(submitPaymentAction);
+            expect(zipClient.Checkout.init).toHaveBeenCalledWith(zipInitPayload);
+        });
+
+        it('fails to execute the strategy if the zip client was not loaded', async () => {
+            jest.spyOn(zipScriptLoader, 'load')
+                .mockResolvedValue(undefined);
+            await strategy.initialize(zipOptions);
+
+            try {
+                await strategy.execute(orderRequestBody, zipOptions);
+            } catch (error) {
+                expect(error).toBeInstanceOf(NotInitializedError);
+                expect(orderActionCreator.submitOrder).not.toHaveBeenCalled();
+                expect(paymentActionCreator.submitPayment).not.toHaveBeenCalled();
+                expect(store.dispatch).not.toHaveBeenCalledWith(submitOrderAction);
+                expect(store.dispatch).not.toHaveBeenCalledWith(submitPaymentAction);
+                expect(zipClient.Checkout.init).not.toHaveBeenCalled();
+            }
+        });
+
+        it('fails to execute the strategy if lightbox fails to provide a checkoutId', async () => {
+            const noIdZipClient = getZipScriptMock('noCheckoutId');
+            jest.spyOn(zipScriptLoader, 'load')
+                .mockResolvedValue(noIdZipClient);
+            await strategy.initialize(zipOptions);
+
+            try {
+                await strategy.execute(orderRequestBody, zipOptions);
+            } catch (error) {
+                expect(error).toBeInstanceOf(PaymentMethodInvalidError);
+                expect(paymentActionCreator.submitPayment).not.toHaveBeenCalled();
+                expect(store.dispatch).not.toHaveBeenCalledWith(submitPaymentAction);
+            }
+        });
+
+        it('cancels the strategy execution if the lightbox is closed', async () => {
+            const cancelledZipClient = getZipScriptMock('cancelled');
+            jest.spyOn(zipScriptLoader, 'load')
+                .mockResolvedValue(cancelledZipClient);
+            await strategy.initialize(zipOptions);
+
+            try {
+                await strategy.execute(orderRequestBody, zipOptions);
+            } catch (error) {
+                expect(error).toBeInstanceOf(PaymentMethodCancelledError);
+                expect(paymentActionCreator.submitPayment).not.toHaveBeenCalled();
+                expect(store.dispatch).not.toHaveBeenCalledWith(submitPaymentAction);
+            }
+        });
+
+        it('fails to execute the strategy if the response state provided by the lightbox is neither approved or cancelled', async () => {
+            const errorZipClient = getZipScriptMock('error');
+            jest.spyOn(zipScriptLoader, 'load')
+                .mockResolvedValue(errorZipClient);
+            await strategy.initialize(zipOptions);
+
+            try {
+                await strategy.execute(orderRequestBody, zipOptions);
+            } catch (error) {
+                expect(error).toBeInstanceOf(PaymentMethodInvalidError);
+                expect(paymentActionCreator.submitPayment).not.toHaveBeenCalled();
+                expect(store.dispatch).not.toHaveBeenCalledWith(submitPaymentAction);
+            }
+        });
+    });
+
+    describe('#deinitialize()', () => {
+        it('deinitializes strategy', async () => {
+            const zipOptions: PaymentInitializeOptions = { methodId: 'zip' };
+            await strategy.initialize(zipOptions);
+            await strategy.deinitialize();
+
+            expect(await strategy.deinitialize()).toEqual(store.getState());
+        });
+    });
+
+    describe('#finalize()', () => {
+        it('throws error to inform that order finalization is not required', async () => {
+            try {
+                await strategy.finalize();
+            } catch (error) {
+                expect(error).toBeInstanceOf(OrderFinalizationNotRequiredError);
+            }
+        });
+    });
+});
--- a/src/payment/strategies/zip/zip-payment-strategy.ts
+++ b/src/payment/strategies/zip/zip-payment-strategy.ts
@@ -1,0 +1,106 @@
+import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
+import {
+    InvalidArgumentError,
+    MissingDataError,
+    MissingDataErrorType,
+    NotInitializedError,
+    NotInitializedErrorType,
+} from '../../../common/error/errors';
+import { OrderActionCreator, OrderRequestBody } from '../../../order';
+import { OrderFinalizationNotRequiredError } from '../../../order/errors';
+import { PaymentMethodCancelledError } from '../../errors';
+import PaymentMethodInvalidError from '../../errors/payment-method-invalid-error';
+import PaymentActionCreator from '../../payment-action-creator';
+import PaymentMethod from '../../payment-method';
+import PaymentMethodActionCreator from '../../payment-method-action-creator';
+import { PaymentInitializeOptions, PaymentRequestOptions } from '../../payment-request-options';
+import PaymentStrategy from '../payment-strategy';
+
+import { Zip, ZipModalEvent } from './zip';
+import ZipScriptLoader from './zip-script-loader';
+
+export default class ZipPaymentStrategy implements PaymentStrategy {
+    private _paymentMethod?: PaymentMethod;
+    private _zipClient?: Zip;
+
+    constructor(
+        private _store: CheckoutStore,
+        private _orderActionCreator: OrderActionCreator,
+        private _paymentActionCreator: PaymentActionCreator,
+        private _paymentMethodActionCreator: PaymentMethodActionCreator,
+        private _zipScriptLoader: ZipScriptLoader
+    ) { }
+
+    initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
+        return this._zipScriptLoader.load()
+            .then(zip => {
+                this._zipClient = zip;
+
+                return this._store.getState();
+            });
+    }
+
+    deinitialize(options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        this._paymentMethod = undefined;
+        this._zipClient = undefined;
+
+        return Promise.resolve(this._store.getState());
+    }
+
+    execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        const { payment, ...order } = payload;
+        const { _zipClient: zipClient } = this;
+
+        if (!payment) {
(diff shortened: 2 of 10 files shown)


src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -0,0 +1,145 @@
...
+    @bind
+    private _handleResponse(response: ZipResponse): void {
+        if (response.state === ZipModalEvent.CancelCheckout) {
+            this._lightboxEvents$.next({ type: ZipModalEvent.CancelCheckout });
+
+            return;
+        }
+        const { state, checkoutId } = response;
+
+        this._requestSender.post('checkout.php', {
```
cbandam: Please double check with checkout team if this is a valid way to set checkout data.

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -0,0 +1,145 @@
...
+
+        return this._displayWallet();
+    }
+
+    finalize(options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        return Promise.reject(new OrderFinalizationNotRequiredError());
+    }
+
+    private _displayWallet(): Promise<InternalCheckoutSelectors> {
+        return this._store.dispatch(this._paymentStrategyActionCreator.widgetInteraction(() => {
```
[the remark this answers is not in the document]
FranciscoOnt: Widget Interaction is used to display the `loading` indicator when pressing the button and wait for an event to cancel the action and re-enable the `place order`  button allowing to select another payment method.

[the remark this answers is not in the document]
davidchin: This is actually not needed. As long as the returning promise of `execute` doesn't get resolved, the client app can use `isSubmittingOrder` flag to determine whether or not to block the UI. `widgetInteraction` is only needed if we need to trigger UI outside of the normal flow.

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -0,0 +1,159 @@
...
+            });
+    }
+
+    @bind
+    private _handleResponse(response: ZipResponse): void {
+        if (response.state === ZipModalEvent.CancelCheckout) {
+            this._lightboxEvents$.next({ type: ZipModalEvent.CancelCheckout });
+
+            return;
+        } else if (response.state === ZipModalEvent.CheckoutApproved) {
```
davidchin: If there's a `return` statement above, you don't actually need to write `else` because if the condition is true above, the code below won't get executed anyway.

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+            submitOrderAction = of(createAction(OrderActionType.SubmitOrderRequested));
+            orderActionCreator.submitOrder = jest.fn(() => submitOrderAction);
+
+            await strategy.initialize(zipOptions);
+        });
+
+        it('expect to not call the orderActionCreator', async () => {
+            await strategy.deinitialize(zipOptions);
+
+            expect(orderActionCreator.submitOrder).not.toHaveBeenCalled();
```
davidchin: Not sure if this test is necessary.

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+            };
+            await strategy.execute(orderRequestBody, zipOptions);
+
+            expect(zipClient.Checkout.init).toHaveBeenCalledWith(expectedPayload);
+        });
+
+        it('completes the checkout and submit the order', async () => {
+            await strategy.execute(orderRequestBody, zipOptions);
+
+            expect(orderActionCreator.submitOrder).toBeCalled();
```
davidchin: It might be a good idea to verify the payload you're passing.

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+            expect(paymentActionCreator.submitPayment).toBeCalled();
```
davidchin: Same here.

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+
+        it('cancels the checkout if the lightbox is closed', async () => {
+            zipClient = getZipScriptMock(false);
+            jest.spyOn(zipScriptLoader, 'load')
+                .mockReturnValueOnce(Promise.resolve(zipClient));
+            await strategy.initialize(zipOptions);
+            await strategy.execute(orderRequestBody, zipOptions);
+
+            expect(orderActionCreator.submitOrder).not.toBeCalled();
+            expect(paymentActionCreator.submitPayment).not.toBeCalled();
```
davidchin: Should probably also verify that `PaymentMethodCancelledError` is returned / thrown.

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+            await strategy.execute(orderRequestBody, zipOptions);
+
+            expect(store.getState).toHaveBeenCalled();
+        });
+
+        it('succesfully dispatches widget Interaction', async () => {
+            await strategy.execute(orderRequestBody, zipOptions);
+
+            expect(paymentMethodActionCreator.loadPaymentMethod).toHaveBeenCalledWith('zip');
+            expect(store.getState().paymentMethods.getPaymentMethod).toHaveBeenCalledWith('zip');
```
davidchin: I'm not sure if any of the assertions in this test verifies that widget interaction action is successfully dispatched. Anyway, it doesn't matter as the call is not needed as I explained earlier.

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+        it('expect to get the payment method', async () => {
+            await strategy.execute(orderRequestBody, zipOptions);
+
+            expect(store.getState().paymentMethods.getPaymentMethod).toHaveBeenCalledWith(zipOptions.methodId);
+        });
+
+        it('returns the payment information', async () => {
+            await strategy.execute(orderRequestBody, zipOptions);
+
+            expect(store.getState).toHaveBeenCalled();
```
davidchin: It does not verify that the return value of the method you're testing. Also, the expected value shouldn't be "payment information" as written in the test description. Rather, it should return the current state.

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+            orderRequestBody = getOrderRequestBody();
+            zipOptions = { methodId: 'zip' };
+
+            await strategy.initialize(zipOptions);
+        });
+
+        it('dispatch widget interaction when execute is called', async () => {
+            await strategy.execute(orderRequestBody, zipOptions);
+
+            expect(store.dispatch).toHaveBeenCalled();
```
davidchin: You can be dispatching anything. Therefore this test is not really useful.

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+        });
+
+        it('expect to get the payment method', async () => {
+            await strategy.execute(orderRequestBody, zipOptions);
+
+            expect(store.getState().paymentMethods.getPaymentMethod).toHaveBeenCalledWith(zipOptions.methodId);
```
davidchin: Calling a getter that has no side effect is only an implementation detail - therefore there's no point of testing it. You should aim to either test the subject's interaction with its collaborators, or its observable behaviour (i.e.: return values of its methods etc...).

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -0,0 +1,159 @@
...
+    private _fetchPayload(resolve: (response: ZipPostResponse) => void): void {
+        this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(this._methodId))
+            .then(state => {
+                const paymentMethod = state.paymentMethods.getPaymentMethod(this._methodId);
+
+                if (!paymentMethod || !paymentMethod.clientToken) {
+                    throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+                }
+
+                resolve(JSON.parse(paymentMethod.clientToken));
```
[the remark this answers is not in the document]
cmedinilla: We are returning a JSON parseable string because, the zip token creation return 2 values a checkout id and a URI, we need these 2 values to can display the lightbox

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -0,0 +1,159 @@
...
+            methodId: this._methodId,
+            paymentData: { nonce: this._zipLightboxResponse.checkoutId } as NonceInstrument,
+        };
+
+        return this._store.dispatch(this._orderActionCreator.submitOrder(order, options))
+            .then(() => this._store.dispatch(this._paymentActionCreator.submitPayment(paymentPayload)));
+    }
+
+    @bind
+    private _fetchPayload(resolve: (response: ZipPostResponse) => void): void {
```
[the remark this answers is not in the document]
FranciscoOnt: `resolve` is part of a promise passed to the method by the lightbox, so when it resolves the lightbox can be opened with the object passed to resolve function.

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -0,0 +1,159 @@
...
+    private _submitOrder(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        const order = { useStoreCredit: payload.useStoreCredit };
+
+        if (!this._zipLightboxResponse) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        const paymentPayload = {
+            methodId: this._methodId,
+            paymentData: { nonce: this._zipLightboxResponse.checkoutId } as NonceInstrument,
```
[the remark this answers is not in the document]
FranciscoOnt: Under normal circumstances, at this point it should not be undefined, but `checkoutId` is defined as `string?` because the Lightbox response can be `state: 'cancelled'` only and completely omit the `checkoutId` on that case.

src/payment/strategies/zip/zip-payment-strategy.ts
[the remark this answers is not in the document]
davidchin: That's what I thought as well. It passes a callback so we can perform any necessary async actions inside `onCheckout` callback. Do you have the API doc for the Zip JS client? I can't seem to find it.

src/payment/strategies/zip/zip-payment-strategy.ts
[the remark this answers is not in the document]
davidchin: I see, you're using the same interface for both success and cancelled state. I guess can be worked around by having a simple `if` statement checking the presence of `checkoutId`.

src/payment/strategies/zip/zip-payment-strategy.ts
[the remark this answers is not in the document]
davidchin: I understand that Zip might have a special requirement. But it just feels like a shortcut / snowflake to me as `clientToken` field isn't supposed to contain a JSON parsable string. Anyway, I'll defer the decision to the payments team. cc @bc-anna @IamDavidovich.

src/payment/strategies/zip/zip-payment-strategy.ts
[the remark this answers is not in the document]
cmedinilla: https://zip-online.api-docs.io/v1/api-specification/javascript-lightbox

FranciscoOnt:
@davidchin , thanks for your feedback 👍 , it's now ready to re-review ♻️

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -0,0 +1,119 @@
...
+
+                this._zipClient.Checkout.init({
+                    onComplete: ({ checkoutId, state }) => {
+                        if (state === ZipModalEvent.CancelCheckout) {
+                            return reject(new PaymentMethodCancelledError());
+                        }
+
+                        if (state === ZipModalEvent.CheckoutApproved) {
+                            if (!this._methodId) {
+                                return reject(new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized));
```
davidchin: I would probably just get the `methodId` from `payload` instead of storing it as an instance variable. So the whole thing could be written like this

```ts
const { payment, ...order } = payload;
const { _zipClient: zipClient } = this;

if (!payment) {
    throw new PaymentArgumentInvalidError();
}

// NOTE: I've moved the check here because at this point we can already determine if payment method is initialised or not.
if (!zipClient) {
}

return this._store.dispatch(this._orderActionCreator.submitOrder(order, options))
    .then(() => new Promise((resolve, reject) => {
        zipClient.Checkout.init({
            onComplete: ({ checkoutId, state }) => {
                if (state === ZipModalEvent.CheckoutApproved && checkoutId) {
                    return resolve(checkoutId);
                }

                if (state === ZipModalEvent.CancelCheckout) {
                    return reject(new PaymentMethodCancelledError());
                }

                return reject(new PaymentMethodInvalidError());
            },

            onCheckout: () => {
                // Pass the clientToken to ZipPay here
            },
        });
    })
    // NOTE: I've moved the call here to reduce code nesting.
    .then(nonce => this._store.dispatch(this._paymentActionCreator.submitPayment({
        ...payment,
        paymentData: { nonce },
    }));
```

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+                expect(store.dispatch).not.toHaveBeenCalledWith(submitPaymentAction);
+            }
+        });
+    });
+
+    describe('#deinitialize()', () => {
+        let zipOptions: PaymentInitializeOptions;
+        let submitOrderAction: Observable<Action>;
+
+        beforeEach(async () => {
```
davidchin: Do you really need a `beforeEach` block when there's only one test?

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+        it('fails to execute the strategy if the zip client was not loaded', async () => {
+            const { payment, ...order } = orderRequestBody;
+            jest.spyOn(zipScriptLoader, 'load')
+                .mockReturnValueOnce(Promise.resolve(undefined));
+            await strategy.initialize(zipOptions);
+            try {
+                await strategy.execute(orderRequestBody, zipOptions);
+            } catch (error) {
+                expect(error).toBeInstanceOf(NotInitializedError);
+                expect(orderActionCreator.submitOrder).toHaveBeenCalledWith(order, zipOptions);
```
davidchin: We should not even attempt to submit the order if ZipClient is not defined.

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+                expect(zipScriptLoader.load).not.toHaveBeenCalled();
+            }
+        });
+    });
+
+    describe('#execute()', () => {
+        let zipOptions: PaymentInitializeOptions;
+        let orderRequestBody: OrderRequestBody;
+        const zipInitPayload =  {
+                    onComplete: expect.any(Function),
```
davidchin: Could you please fix the indentation?

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+        beforeEach(async () => {
+            orderRequestBody = getOrderRequestBody();
+            zipOptions = { methodId: 'zip' };
+
+            await strategy.initialize(zipOptions);
+        });
+
+        it('executes the strategy successfully and submits the payment', async () => {
+            const { payment, ...order } = orderRequestBody;
+            const expected = {
```
davidchin: `expected` what? We're asserting several things here so it's good to be specific about what you're expecting.

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+        });
+
+        it('initializes the strategy successfully', async () => {
+            await strategy.initialize(zipOptions);
+
+            expect(paymentMethodActionCreator.loadPaymentMethod).toHaveBeenCalledWith(zipOptions.methodId);
+            expect(zipScriptLoader.load).toHaveBeenCalled();
+        });
+
+        it('does not initialize the strategy if method id is not provided', async () => {
```
davidchin: `methodId` is always defined (as you can see in the TS definition) so you don't really need a test for it.

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+            paymentActionCreator,
+            paymentMethodActionCreator,
+            zipScriptLoader
+        );
+
+        jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod');
+    });
+
+    it('creates an instance of ZipPaymentStrategy', () => {
+        expect(strategy).toBeInstanceOf(ZipPaymentStrategy);
```
davidchin: I don't know what you're trying to test here because you've actually constructed the object yourself. So this is always going to be true.

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+    let store: CheckoutStore;
+    let scriptLoader: ScriptLoader;
+    let strategy: PaymentStrategy;
+    let submitOrderAction: Observable<Action>;
+    let submitPaymentAction: Observable<Action>;
+    let zipClient: Zip;
+    let zipScriptLoader: ZipScriptLoader;
+
+    beforeEach(() => {
+        paymentMethodMock = { ...getZip() };
```
davidchin: Just `getZip()`

src/payment/strategies/zip/zip-payment-strategy.ts
FranciscoOnt: I copied the zipClient null check before `submitOrder`, to it can't be submitted  if the client is not loaded.
But it seems that the arrow function of  the `dispatch`causes the linter to ignore the null check from before (see image) , so I had to duplicate the code as a workaround, should I leave that change or is there something else I can do?
Screen Shot 2019-03-25 at 10 26 27 AM

davidchin: Have you tried what I suggest above, which is to hold a direct reference to the client instance and use it within the method?

FranciscoOnt:
@davidchin , I  made the changes you suggested, could you please re-review? ♻️

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,256 @@
...
+                expect(error).toBeInstanceOf(PaymentMethodCancelledError);
+                expect(paymentActionCreator.submitPayment).not.toHaveBeenCalled();
+                expect(store.dispatch).not.toHaveBeenCalledWith(submitPaymentAction);
+            }
+        });
+
+        it('fails to execute the strategy if lightbox response state is not handled by the strategy', async () => {
+            const erroredZipClient = getZipScriptMock('error');
+            jest.spyOn(zipScriptLoader, 'load')
+                .mockReturnValueOnce(Promise.resolve(erroredZipClient));
```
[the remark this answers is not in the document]
FranciscoOnt: It does not fail to load the client, it resolves a regular zip client but the lightbox response on `onComplete` response  is `state: 'error'` instead of the regular `approved` or `cancelled`.
It should be more clear after I reword this test's name.

FranciscoOnt:
@davidchin , thanks for your feedback, I made all the changes you suggested 👍 , could your please re-review?

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+
+        jest.spyOn(orderActionCreator, 'submitOrder')
+            .mockReturnValue(submitOrderAction);
+        jest.spyOn(paymentActionCreator, 'submitPayment')
+            .mockReturnValue(submitPaymentAction);
+        jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod')
+            .mockResolvedValue(store.getState());
+        jest.spyOn(store, 'dispatch')
+            .mockResolvedValue(store.getState());
+        jest.spyOn(store, 'getState');
```
davidchin: I don't think this is necessary.

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -0,0 +1,104 @@
...
+                        if (!this._paymentMethod || !this._paymentMethod.clientToken) {
+                            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+                        }
+
+                        openModal(JSON.parse(this._paymentMethod.clientToken));
+                    },
+                });
+            })
+            .then(nonce => this._store.dispatch(this._paymentActionCreator.submitPayment({
+                    methodId: payment.methodId,
```
davidchin: Please fix the indentation. i.e.: 

```ts
.then(nonce => 
    this._store.dispatch(this._paymentActionCreator.submitPayment({
        methodId: payment.methodId,
        paymentData: { nonce },
    }))
));
```

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+            }
+        });
+
+        it('does not initialize the strategy if client token is not provided', async () => {
+            jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod')
+                .mockReturnValue({
+                    ...paymentMethodMock,
+                    clientToken: null,
+                });
+            try {
```
davidchin: Line break before `try`. 💅

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+                expect(error).toBeInstanceOf(MissingDataError);
+                expect(zipScriptLoader.load).not.toHaveBeenCalled();
+            }
+        });
+    });
+
+    describe('#execute()', () => {
+        let zipOptions: PaymentInitializeOptions;
+        let orderRequestBody: OrderRequestBody;
+        beforeEach(async () => {
```
davidchin: Line break before `beforeEach`. 💅

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+            expect(store.dispatch).toHaveBeenCalledWith(submitOrderAction);
+            expect(store.dispatch).toHaveBeenCalledWith(submitPaymentAction);
+            expect(zipClient.Checkout.init).toHaveBeenCalledWith(zipInitPayload);
+        });
+
+        it('fails to execute the strategy if the zip client was not loaded', async () => {
+            jest.spyOn(zipScriptLoader, 'load')
+                .mockResolvedValue(undefined);
+            await strategy.initialize(zipOptions);
+            try {
```
davidchin: Line break before `try`. 💅

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+                expect(zipClient.Checkout.init).not.toHaveBeenCalled();
+            }
+        });
+
+        it('fails to execute the strategy if lightbox fails to provide a checkoutId', async () => {
+            const noIdZipClient = getZipScriptMock('noCheckoutId');
+            jest.spyOn(zipScriptLoader, 'load')
+                .mockResolvedValue(noIdZipClient);
+            await strategy.initialize(zipOptions);
+            try {
```
davidchin: Line break before `try`. 💅

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+                expect(store.dispatch).not.toHaveBeenCalledWith(submitPaymentAction);
+            }
+        });
+
+        it('cancels the strategy execution if the lightbox is closed', async () => {
+            const cancelledZipClient = getZipScriptMock('cancelled');
+            jest.spyOn(zipScriptLoader, 'load')
+                .mockResolvedValue(cancelledZipClient);
+            await strategy.initialize(zipOptions);
+            try {
```
davidchin: Line break before `try`. 💅

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+                expect(store.dispatch).not.toHaveBeenCalledWith(submitPaymentAction);
+            }
+        });
+
+        it('fails to execute the strategy if the response state provided by the lightbox is neither approved or cancelled', async () => {
+            const errorZipClient = getZipScriptMock('error');
+            jest.spyOn(zipScriptLoader, 'load')
+                .mockResolvedValue(errorZipClient);
+            await strategy.initialize(zipOptions);
+            try {
```
davidchin: Line break before `try`. 💅

src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+            await strategy.initialize(zipOptions);
+
+            expect(paymentMethodActionCreator.loadPaymentMethod).toHaveBeenCalledWith(zipOptions.methodId);
+            expect(zipScriptLoader.load).toHaveBeenCalled();
+        });
+
+        it('does not initialize the strategy if payment method is not provided', async () => {
+            jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod')
+                .mockReturnValueOnce(undefined);
+            try {
```
davidchin: Line break before `try`. 💅

FranciscoOnt:
@davidchin , I made the changes and squashed the commits, thanks! 👍

FranciscoOnt:
Hi @davidchin, sorry for the inconvenience, our team found a bug in the implementation that caused the discounts applied by coupons not to be submitted to Zip and we had to move the `loadPaymentMethod` call to the execute method to prevent this bad behavior and ensure the discounts are sent properly, could you please review the changes?
I will make any corrections ASAP.

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+
+        return Promise.resolve(this._store.getState());
+    }
+
+    execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        const { payment, ...order } = payload;
+        const { _zipClient: zipClient } = this;
+
+        if (!payment) {
+            throw new InvalidArgumentError('Unable to submit payment because "payload.payment.methodId" argument is not provided.');
```
icatalina: This error message doesn't reflect the actual error. You are checking for the whole `payment` object to be present:

```suggestion
            throw new InvalidArgumentError('Unable to submit payment because "payload.payment" argument is not provided.');
```

FranciscoOnt:
@icatalina , thanks for the review, I made the change you suggested and squashed the commits.

PR #554, opened 2019-04-06
feat(payment): INT-1398 Add shipping and billing address before Klarna authorization

josesonora:
## What?
I added billing and shipping address to the payload of the `load` method for Klarna.

## Why?
According to Legal Privacy and Europe laws, Klarna must not receive PII data from BC until shopper selects option at checkout or authorizes so we no longer send this info during the `create_session` call instead of that we update the session with the info before open the Klarna's widget 

## Testing / Proof
Video: https://drive.google.com/open?id=1UWqsZfH08c3LSx9RqqZDDZ4Sn5ARWL7_

image

image

Sibling PRs:
[BigPay INT-1398](https://github.com/bigcommerce/bigpay/pull/1585)

### DIFF
--- a/src/payment/strategies/klarna/klarna-credit.ts
+++ b/src/payment/strategies/klarna/klarna-credit.ts
@@ -1,7 +1,8 @@
 export default interface KlarnaCredit {
     authorize(params: any, callback: (res: KlarnaAuthorizationResponse) => void): void;
     init(params: KlarnaInitParams): void;
-    load(params: KlarnaLoadParams, callback?: (res: KlarnaLoadResponse) => void): void;
+    load(params: KlarnaLoadParams, callback: (res: KlarnaLoadResponse) => void): void;
+    load(params: KlarnaLoadParams, data: KlarnaUpdateSessionParams, callback: (res: KlarnaLoadResponse) => void): void;
 }
 
 export interface KlarnaInitParams {
@@ -10,6 +11,10 @@ export interface KlarnaInitParams {
 
 export interface KlarnaLoadParams {
     container: string;
+    payment_method_category?: string;
+    payment_method_categories?: string;
+    instance_id?: string;
+    preferred_payment_method?: string;
 }
 
 export interface KlarnaLoadResponse {
@@ -26,4 +31,20 @@ export interface KlarnaAuthorizationResponse {
     error?: {
         invalid_fields: string[];
     };
+}
+
+export type KlarnaUpdateSessionParams = Partial<{
+    billing_address: KlarnaAddress;
+    shipping_address: KlarnaAddress;
+}>;
+
+export interface KlarnaAddress {
+    street_address: string;
+    city: string;
+    country: string;
+    given_name: string;
+    family_name: string;
+    postal_code: string;
+    region: string;
+    email?: string;
 }
--- a/src/payment/strategies/klarna/klarna-payment-strategy.spec.ts
+++ b/src/payment/strategies/klarna/klarna-payment-strategy.spec.ts
@@ -13,7 +13,6 @@ import {
 import { OrderActionCreator, OrderActionType, OrderRequestBody, OrderRequestSender } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
 import { getOrderRequestBody } from '../../../order/internal-orders.mock';
-import { getKlarna, getPaymentMethodsState } from '../../../payment/payment-methods.mock';
 import { RemoteCheckoutActionCreator, RemoteCheckoutActionType, RemoteCheckoutRequestSender } from '../../../remote-checkout';
 import { PaymentMethodCancelledError, PaymentMethodInvalidError } from '../../errors';
 import PaymentMethod from '../../payment-method';
@@ -21,9 +20,14 @@ import PaymentMethodActionCreator from '../../payment-method-action-creator';
 import { PaymentMethodActionType } from '../../payment-method-actions';
 import PaymentMethodRequestSender from '../../payment-method-request-sender';
 
+import { getCheckoutStoreState } from '../../../checkout/checkouts.mock';
+import { MissingDataError } from '../../../common/error/errors';
+import { getKlarna } from '../../payment-methods.mock';
+
 import KlarnaCredit from './klarna-credit';
 import KlarnaPaymentStrategy from './klarna-payment-strategy';
 import KlarnaScriptLoader from './klarna-script-loader';
+import { getEUBillingAddress, getKlarnaUpdateSessionParams } from './klarna.mock';
 
 describe('KlarnaPaymentStrategy', () => {
     let initializePaymentAction: Observable<Action>;
@@ -38,11 +42,15 @@ describe('KlarnaPaymentStrategy', () => {
     let submitOrderAction: Observable<Action>;
     let store: CheckoutStore;
     let strategy: KlarnaPaymentStrategy;
+    let paymentMethodMock: PaymentMethod;
 
     beforeEach(() => {
-        store = createCheckoutStore({
-            paymentMethods: getPaymentMethodsState(),
-        });
+        paymentMethodMock = { ...getKlarna(), clientToken: 'foo' };
+        store = createCheckoutStore(getCheckoutStoreState());
+
+        jest.spyOn(store, 'dispatch').mockReturnValue(Promise.resolve(store.getState()));
+        jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod').mockReturnValue(paymentMethodMock);
+
         orderActionCreator = new OrderActionCreator(
             new OrderRequestSender(createRequestSender()),
             new CheckoutValidator(new CheckoutRequestSender(createRequestSender()))
@@ -63,7 +71,7 @@ describe('KlarnaPaymentStrategy', () => {
         klarnaCredit = {
             authorize: jest.fn((params, callback) => callback({ approved: true, authorization_token: 'bar' })),
             init: jest.fn(() => {}),
-            load: jest.fn((options, callback) => callback({ show_form: true })),
+            load: jest.fn((options, data, callback) => callback({ show_form: true })),
         };
 
         paymentMethod = getKlarna();
@@ -118,11 +126,51 @@ describe('KlarnaPaymentStrategy', () => {
 
         it('loads widget', () => {
             expect(klarnaCredit.init).toHaveBeenCalledWith({ client_token: 'foo' });
+            expect(klarnaCredit.load)
+                .toHaveBeenCalledWith({ container: '#container' }, {}, expect.any(Function));
             expect(klarnaCredit.load).toHaveBeenCalledTimes(1);
+        });
+
+        it('loads widget in EU', async () => {
+            store = store = createCheckoutStore({
+                ...getCheckoutStoreState(),
+                billingAddress: { data: getEUBillingAddress(), errors: {}, statuses: {} },
+            });
+            strategy = new KlarnaPaymentStrategy(
+                store,
+                orderActionCreator,
+                paymentMethodActionCreator,
+                remoteCheckoutActionCreator,
+                scriptLoader
+            );
+            jest.spyOn(store, 'dispatch').mockReturnValue(Promise.resolve(store.getState()));
+            jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod').mockReturnValue(paymentMethodMock);
+
+            await strategy.initialize({ methodId: paymentMethod.id, klarna: { container: '#container', onLoad } });
+
+            expect(klarnaCredit.load)
+                .toHaveBeenCalledWith({ container: '#container' }, getKlarnaUpdateSessionParams(), expect.any(Function));
         });
 
         it('triggers callback with response', () => {
             expect(onLoad).toHaveBeenCalledWith({ show_form: true });
+        });
+
+        it('throws error if required data is not loaded', async () => {
+            store = createCheckoutStore();
+            strategy = new KlarnaPaymentStrategy(
+                store,
+                orderActionCreator,
+                paymentMethodActionCreator,
+                remoteCheckoutActionCreator,
+                scriptLoader
+            );
+
+            try {
+                await strategy.initialize({ methodId: paymentMethod.id, klarna: { container: '#container', onLoad } });
+            } catch (error) {
+                expect(error).toBeInstanceOf(MissingDataError);
+            }
         });
     });
 
--- a/src/payment/strategies/klarna/klarna-payment-strategy.ts
+++ b/src/payment/strategies/klarna/klarna-payment-strategy.ts
@@ -1,3 +1,7 @@
+import { includes } from 'lodash';
+
+import { Address } from '../../../address';
+import BillingAddress from '../../../billing/billing-address';
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
 import {
     InvalidArgumentError,
@@ -15,12 +19,13 @@ import PaymentMethodActionCreator from '../../payment-method-action-creator';
 import { PaymentInitializeOptions, PaymentRequestOptions } from '../../payment-request-options';
 import PaymentStrategy from '../payment-strategy';
 
-import KlarnaCredit, { KlarnaLoadResponse } from './klarna-credit';
+import KlarnaCredit, { KlarnaAddress, KlarnaLoadResponse, KlarnaUpdateSessionParams } from './klarna-credit';
 import KlarnaScriptLoader from './klarna-script-loader';
 
 export default class KlarnaPaymentStrategy implements PaymentStrategy {
     private _klarnaCredit?: KlarnaCredit;
     private _unsubscribe?: (() => void);
+    private _supportedEUCountries = ['AT', 'DE', 'DK', 'FI', 'GB', 'NL', 'NO', 'SE', 'CH'];
 
     constructor(
         private _store: CheckoutStore,
@@ -96,18 +101,26 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
         return this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId))
             .then(state => new Promise<KlarnaLoadResponse>((resolve, reject) => {
                 const paymentMethod = state.paymentMethods.getPaymentMethod(methodId);
+                const billingAddress = state.billingAddress.getBillingAddress();
+                const shippingAddress = state.shippingAddress.getShippingAddress();
 
                 if (!paymentMethod) {
                     throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+                }
+
+                if (!billingAddress) {
+                    throw new MissingDataError(MissingDataErrorType.MissingBillingAddress);
                 }
 
                 if (!this._klarnaCredit || !paymentMethod.clientToken) {
                     throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
                 }
 
+                const updateSessionData = this._getUpdateSessionData(billingAddress, shippingAddress);
+
                 this._klarnaCredit.init({ client_token: paymentMethod.clientToken });
 
-                this._klarnaCredit.load({ container }, response => {
+                this._klarnaCredit.load({ container }, updateSessionData, response => {
                     if (onLoad) {
                         onLoad(response);
                     }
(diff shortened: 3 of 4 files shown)


src/payment/strategies/klarna/klarna-credit.ts
```diff
@@ -1,7 +1,7 @@
 export default interface KlarnaCredit {
     authorize(params: any, callback: (res: KlarnaAuthorizationResponse) => void): void;
     init(params: KlarnaInitParams): void;
-    load(params: KlarnaLoadParams, callback?: (res: KlarnaLoadResponse) => void): void;
+    load(params: KlarnaLoadParams, data?: object, callback?: (res: KlarnaLoadResponse) => void): void;
```
cbandam: Create type for Klarna data parameter.

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -96,18 +96,49 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
...
 
                 if (!paymentMethod) {
                     throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
                 }
 
+                if (!billingAddress) {
+                    throw new MissingDataError(MissingDataErrorType.MissingBillingAddress);
+                }
+
+                if (!shippingAddress) {
```
cbandam: Shipping address is optional. For example, Digital Products don´t require a shipping address.

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -96,18 +96,49 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
...
+                }
+
                 if (!this._klarnaCredit || !paymentMethod.clientToken) {
                     throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
                 }
 
                 this._klarnaCredit.init({ client_token: paymentMethod.clientToken });
 
-                this._klarnaCredit.load({ container }, response => {
+                this._klarnaCredit.load({ container }, {
```
cbandam: Should we need to add validation so that PII data is only sent when country is EU?

src/payment/strategies/klarna/klarna-payment-strategy.spec.ts
```diff
@@ -117,13 +123,52 @@ describe('KlarnaPaymentStrategy', () => {
...
             expect(klarnaCredit.init).toHaveBeenCalledWith({ client_token: 'foo' });
+            expect(klarnaCredit.load).toHaveBeenCalledWith({ container: '#container' }, { billing_address: billingAddress, shipping_address: shippingAddress }, expect.any(Function) );
             expect(klarnaCredit.load).toHaveBeenCalledTimes(1);
         });
 
         it('triggers callback with response', () => {
             expect(onLoad).toHaveBeenCalledWith({ show_form: true });
         });
+
+        it('throws error if required data is not loaded', async () => {
```
cbandam: Not sure if this is the best way to verify error throws. Have you tried with jest's `toThrowError`?

clopezh: I guess in case of this test, is good to go with Sonora's approach since `strategy.initialize` is a promise and needs to be resolved (Maybe we can check if toThrowError supports promise response

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -96,18 +96,49 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
...
+                    billing_address: {
```
clopezh: You can take all this object into a separate method to reduce lines in single method i.e. `billing_address: mapKlarnaBillingAddress(billingAddress) ` also this comment apply to shipping_address

src/common/error/errors/missing-data-error.ts
```diff
@@ -2,6 +2,7 @@ import StandardError from './standard-error';
 
 export enum MissingDataErrorType {
     MissingBillingAddress,
+    MissingShippingAddress,
```
cbandam: I don´t see that we are using this exception anywhere, do we still need it?

src/payment/strategies/klarna/klarna-credit.ts
```diff
@@ -27,3 +27,26 @@ export interface KlarnaAuthorizationResponse {
...
+export interface KlarnaUpdateSessionParams {
+    billing_address: {
+        street_address: string,
+        city: string,
+        country: string,
+        given_name: string,
+        family_name: string,
+        postal_code: string,
+        region: string,
+        email?: string,
```
cbandam: 🍹 You could create a separate type calle `KlarnaAddress` and use this type to define `billing_address` and `shipping_address`

src/common/error/errors/missing-data-error.ts
josesonora: No, I forgot remove it, I'll do that 👍

cbandam:
@josesonora can you specify sibling PRs?

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -121,6 +132,35 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
...
+        data.billing_address = this.__mapToKlarnaAddress(billingAddress, billingAddress.email);
+
+        if (shippingAddress) {
+            data.shipping_address = this.__mapToKlarnaAddress(shippingAddress, billingAddress.email);
+        }
+
+        return data;
+    }
+
+    private __mapToKlarnaAddress(address: BillingAddress | Address, email?: string): KlarnaAddress {
```
clopezh: only use one underscore in the name of this method please

JoseBobadilla20:
💚 Tested

src/payment/strategies/klarna/klarna-credit.ts
```diff
@@ -27,3 +27,19 @@ export interface KlarnaAuthorizationResponse {
         invalid_fields: string[];
     };
 }
+
+export interface KlarnaUpdateSessionParams {
```
icatalina: This doesn't seem right, considering the actual definition of the method, all these properties seem optional. Which looks like the reason we have the `| {}` in the above definition.

A more accurate type description could be:
```ts
export default interface KlarnaCredit {  
	// ...
	load(params: KlarnaLoadParams, callback: (res: KlarnaLoadResponse) => void): void; 
	load(params: KlarnaLoadParams, data: KlarnaUpdateSessionParams, callback: (res: => void): void; 
	// ...
}

export type KlarnaUpdateSessionParams = Partial<{
    billing_address: KlarnaAddress;
    shipping_address: KlarnaAddress;
}>
```

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -121,6 +132,35 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
...
+        if (billingAddress.countryCode === 'US') {
+            return {};
+        }
+
+        const data = {} as KlarnaUpdateSessionParams;
+
+        data.billing_address = this._mapToKlarnaAddress(billingAddress, billingAddress.email);
+
+        if (shippingAddress) {
+            data.shipping_address = this._mapToKlarnaAddress(shippingAddress, billingAddress.email);
```
icatalina: Also, since email is optional, why are we using the billing email address to fill the shipping one? shouldn't just be left blank?

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -121,6 +132,35 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
             }));
     }
 
+    private _getUpdateSessionData(billingAddress: BillingAddress, shippingAddress?: Address): KlarnaUpdateSessionParams | {} {
+        if (billingAddress.countryCode === 'US') {
```
icatalina: I wonder, why are we doing this conditionally for non-US countries? Can't we use the same logic for all of them and just send the data at this stage for both US and the rest?

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -121,6 +132,35 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
...
+            return {};
+        }
+
+        const data = {} as KlarnaUpdateSessionParams;
```
icatalina: Avoid using `as`. I'm guessing it was done this way because `shipping_address` is optional, you can just define `data` with the mandatory fields and append the optional ones if required.

```ts
const data: KlarnaUpdateSessionParams = {
    billing_address: this._mapToKlarnaAddress(billingAddress, billingAddress.email),
};

if (shippingAddress) {
    data.shipping_address = this._mapToKlarnaAddress(shippingAddress, billingAddress.email);
}

return data;
```

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -121,6 +132,35 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
             }));
     }
 
+    private _getUpdateSessionData(billingAddress: BillingAddress, shippingAddress?: Address): KlarnaUpdateSessionParams | {} {
```
icatalina: Same as before, if `KlarnaUpdateSessionParams` can be completely empty, the type should reflect that.

src/payment/strategies/klarna/klarna-payment-strategy.ts
josesonora: We can do that but if we remove this conditional, data from US stores will be override with the same data so I think that is unnecessary

src/payment/strategies/klarna/klarna-payment-strategy.ts
josesonora: Klarna's implementation in BigPay is doing the same thing (adding the billing email to the shipping data), so if I don't follow same process Klarna will throw a mismatch error

src/payment/strategies/klarna/klarna-credit.ts
```diff
@@ -1,7 +1,7 @@
 export default interface KlarnaCredit {
     authorize(params: any, callback: (res: KlarnaAuthorizationResponse) => void): void;
     init(params: KlarnaInitParams): void;
-    load(params: KlarnaLoadParams, callback?: (res: KlarnaLoadResponse) => void): void;
+    load(params: KlarnaLoadParams, data: KlarnaUpdateSessionParams | {}, callback?: (res: KlarnaLoadResponse) => void): void;
```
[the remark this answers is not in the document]
josesonora: @icatalina Thanks for your feedback, I addressed all of your comments but I have a doubt  regarding this comment `Can we make the types match that?`, this is  what I did to address it: https://github.com/bigcommerce/checkout-sdk-js/pull/554/commits/703843233a53b177751ab6c7185d67fba788b238, but I don't know if that is the what you're expecting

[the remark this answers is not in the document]
icatalina: All I was expecting was the double definition of `load` with and without `data`. But thanks for also typing `KlarnaLoadParams`.

src/payment/strategies/klarna/klarna-payment-strategy.ts
icatalina: fair enough

src/payment/strategies/klarna/klarna-payment-strategy.ts
icatalina: My suggestion was to remove this and the code in bcapp that fills in the data for US merchants.

src/payment/strategies/klarna/klarna-payment-strategy.spec.ts
```diff
@@ -118,12 +126,49 @@ describe('KlarnaPaymentStrategy', () => {
...
+                remoteCheckoutActionCreator,
+                scriptLoader
+            );
+            jest.spyOn(store, 'dispatch').mockReturnValue(Promise.resolve(store.getState()));
+            jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod').mockReturnValue(paymentMethodMock);
+
+            await strategy.initialize({ methodId: paymentMethod.id, klarna: { container: '#container', onLoad } });
+
+            expect(klarnaCredit.load)
+                .toHaveBeenCalledWith({ container: '#container' }, getKlarnaUpdateSessionParams(), expect.any(Function) );
```
icatalina:
```suggestion
                .toHaveBeenCalledWith({ container: '#container' }, getKlarnaUpdateSessionParams(), expect.any(Function));
```

src/payment/strategies/klarna/klarna-payment-strategy.spec.ts
```diff
@@ -118,12 +126,49 @@ describe('KlarnaPaymentStrategy', () => {
 
         it('loads widget', () => {
             expect(klarnaCredit.init).toHaveBeenCalledWith({ client_token: 'foo' });
+            expect(klarnaCredit.load)
+                .toHaveBeenCalledWith({ container: '#container' }, {}, expect.any(Function) );
```
icatalina:
```suggestion
                .toHaveBeenCalledWith({ container: '#container' }, {}, expect.any(Function));
```

src/payment/strategies/klarna/klarna-payment-strategy.ts
josesonora: Nevertheless with the suggested approach we wouldn't be meting the acceptance criteria of this other User Story https://jira.bigcommerce.com/browse/INT-1402 which explicitly requires to send the PII data at the earliest point (at create session) in US stores. However what we can do is to change this validation to verify only EU countries instead of countries that are different to US

cbandam: Hi @icatalina, we reviewed this approach with Klarna. They said that for non-EU countries, it is better to still send the shoppers personal info on `create_session` because Klarna runs a prescreening on their end that helps to determine which payment methods are available from the beginning.

Sure-Round:
@icatalina  can we get a ♻️ ? Per your request on standardizing on a single approach when reviewed with Klarna they said that for non-EU countries, it is better to still send the shoppers personal info on create_session because Klarna runs a prescreening on their end that helps to determine which payment methods are available from the beginning. So there will end up being some logic that bifurcates along the country parameter no matter what.

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -15,12 +18,13 @@ import PaymentMethodActionCreator from '../../payment-method-action-creator';
...
 import PaymentStrategy from '../payment-strategy';
 
-import KlarnaCredit, { KlarnaLoadResponse } from './klarna-credit';
+import KlarnaCredit, { KlarnaAddress, KlarnaLoadResponse, KlarnaUpdateSessionParams } from './klarna-credit';
 import KlarnaScriptLoader from './klarna-script-loader';
 
 export default class KlarnaPaymentStrategy implements PaymentStrategy {
     private _klarnaCredit?: KlarnaCredit;
     private _unsubscribe?: (() => void);
+    private _supportedEUCountries = ['AT', 'DE', 'DK', 'FI', 'GB', 'NL', 'NO', 'SE', 'CH'];
```
icatalina: I understand why we do this, but is there a sibling PR in the BE with a list of countries? or the BE is driven by `!= US`?

src/payment/strategies/klarna/klarna-payment-strategy.spec.ts
```diff
@@ -118,12 +126,49 @@ describe('KlarnaPaymentStrategy', () => {
 
         it('loads widget', () => {
             expect(klarnaCredit.init).toHaveBeenCalledWith({ client_token: 'foo' });
+            expect(klarnaCredit.load)
+                .toHaveBeenCalledWith({ container: '#container' }, {}, expect.any(Function));
             expect(klarnaCredit.load).toHaveBeenCalledTimes(1);
         });
 
+        it('loads widget in EU', async () => {
+            store = createCheckoutStore(getCheckoutEUStoreState());
```
icatalina: Since this is only used once, there's no need to add the `getCheckoutEUStoreState` method. Just overwrite the part of the state that you care about: 

```suggestion
            store = createCheckoutStore({
            	...getCheckoutStoreState(),
            	billingAddress: { data: getEUBillingAddress(), errors: {}, state: {} },
            });
```

src/payment/strategies/klarna/klarna-payment-strategy.spec.ts
```diff
@@ -118,12 +126,49 @@ describe('KlarnaPaymentStrategy', () => {
...
+                store = createCheckoutStore();
+                strategy = new KlarnaPaymentStrategy(
+                    store,
+                    orderActionCreator,
+                    paymentMethodActionCreator,
+                    remoteCheckoutActionCreator,
+                    scriptLoader
+                );
+
+                await strategy.initialize({ methodId: paymentMethod.id, klarna: { container: '#container', onLoad } });
```
icatalina: You only need the `try` for `initialize`

```ts
try {
	await strategy.initialize({ methodId: paymentMethod.id, klarna: { container: '#container', onLoad } });
} catch (error) {
    // ...
```

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -121,6 +133,35 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
             }));
     }
 
+    private _getUpdateSessionData(billingAddress: BillingAddress, shippingAddress?: Address): KlarnaUpdateSessionParams {
+        if (this._supportedEUCountries.indexOf(billingAddress.countryCode) === -1) {
```
icatalina: Please, use `lodash#includes` it is easier to understand what's going on.

```suggestion
        if (includes(this._supportedEUCountries, billingAddress.countryCode)) {
```

You will need to import it.

src/payment/strategies/klarna/klarna-payment-strategy.ts
josesonora: BigPay is using same list => [INT-1398](https://github.com/bigcommerce/bigpay/pull/1585/files#diff-f3036719369d49b5c9e7d5d3c4b3213cR6)

josesonora:
@icatalina Please continue reviewing ♻️

josesonora:
@icatalina Hi, Could you take a look again to this PR please?

PascalZajac:
Ignacio is currently on leave so we'll have to get @davidchin or @capsula4 to look at this in the meantime.

src/payment/strategies/klarna/klarna-credit.ts
```diff
@@ -27,3 +32,19 @@ export interface KlarnaAuthorizationResponse {
         invalid_fields: string[];
     };
 }
+
+export type KlarnaUpdateSessionParams = Partial<{
```
capsula4: I don't think defining a type as a partial is recommended. What you had was probably better. 

What you could do instead is make `_getUpdateSessionData` return `KlarnaUpdateSessionParams | undefined` and update
```javascript
load(params: KlarnaLoadParams, data: KlarnaUpdateSessionParams, callback: (res: KlarnaLoadResponse) => void): void;
```
to 
```
load(loadParams: KlarnaLoadParams, updateParams: KlarnaUpdateSessionParams | undefined, callback: (res: KlarnaLoadResponse) => void): void;
```

src/payment/strategies/klarna/klarna-payment-strategy.spec.ts
capsula4: not a blocker, but a better way to test for exceptions is using 

```javascript
expect(() => await strategy.initialize({ methodId: paymentMethod.id, klarna: { container: '#container', onLoad } }))
    .toThrowError(MissingDataError)
```

src/payment/strategies/klarna/klarna-credit.ts
josesonora: My first approach was `load(loadParams: KlarnaLoadParams, updateParams: KlarnaUpdateSessionParams | {}, callback: (res: KlarnaLoadResponse) => void): void;` but I changed it as suggested by Ignacio

Sure-Round:
@josesonora please squash your commits and we will merge tomorrow

josesonora:
@Sure-Round Done 👍

src/payment/strategies/klarna/klarna-credit.ts
capsula4: yeah, what you could have done is define the argument as partial, not the type itself, e.g:

```javascript
load(loadParams: KlarnaLoadParams, updateParams: Partial<KlarnaUpdateSessionParams>, callback: (res: KlarnaLoadResponse) => void): void
```

Usually is not advised to define the type itself as a partial it might be harder to make it reusable in the future. The stricter your types are the better.

PR #555, opened 2019-04-18
Update Readme: Customer signin actions require named parameters

fresh5447:
## What?
Update to Readme

## Why?
Described incorrect parameters required for `signInCustomer`

Update name of method to sign in guest from `signInCustomer` to `continueAsGuest`

Update `selectShippingOption` to no longer accept `address.id` as a paremeter.

## Testing / Proof
Could only get it to work using named parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ console.log(state.data.getConfig());
 Before you can collect other checkout information from the customer, you should first ask them to sign in. Once they are signed in, the checkout state will be populated with their personal details, such as their addresses.
 
 ```js
-const state = await service.signInCustomer('foo@bar.com', 'password123');
+const state = await service.signInCustomer({email: 'foo@bar.com', password: 'password123'});
 
 console.log(state.data.getCustomer());
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -133,15 +133,15 @@ console.log(state.data.getConfig());
 Before you can collect other checkout information from the customer, you should first ask them to sign in. Once they are signed in, the checkout state will be populated with their personal details, such as their addresses.
 
 ```js
-const state = await service.signInCustomer('foo@bar.com', 'password123');
+const state = await service.signInCustomer({email: 'foo@bar.com', password: 'password123'});
```
davidchin: Thanks @fresh5447. Yes, you're right, the example needs to be updated! 😄Can you please add a space after `{` and before `}`?

README.md
```diff
@@ -133,15 +133,15 @@ console.log(state.data.getConfig());
...
 
 console.log(state.data.getCustomer());
 ```
 
 Alternatively, you can ask the customer to continue as a guest.
 
 ```js
-const state = await service.signInCustomer('foo@bar.com');
+const state = await service.signInCustomer({email: 'foo@bar.com'});
```
davidchin: While you're at it, can you please change the method from `signInCustomer` to `continueAsGuest`? Thanks!

README.md
fresh5447: No problem - @davidchin  change made.

I also just committed one more change. I was only able to get the method `selectShippingOption` by passing it *only* the selected option ID. So I assume the docs are outdated here as well. Furthermore, it doesn't appear an address has an id when you query `getShippingAddress`

PR #560, opened 2019-05-08
feat(checkout): INT-1520 Pass useStoreCredit flag when initialize payment

cmedinilla:
## What? [INT-1520](https://jira.bigcommerce.com/browse/INT-1520)
Initialize payment sending the useStoreCredit flag

## Why?
In order to can update the total amount according to that

## Testing / Proof
image

https://drive.google.com/open?id=1A8p5EifpG3RRWN3HQjP86JiWQRdkHu2l

### DIFF
--- a/src/payment/strategies/zip/zip-payment-strategy.ts
+++ b/src/payment/strategies/zip/zip-payment-strategy.ts
@@ -8,6 +8,7 @@ import {
 } from '../../../common/error/errors';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
+import { RemoteCheckoutActionCreator } from '../../../remote-checkout';
 import { PaymentMethodCancelledError } from '../../errors';
 import PaymentMethodInvalidError from '../../errors/payment-method-invalid-error';
 import PaymentActionCreator from '../../payment-action-creator';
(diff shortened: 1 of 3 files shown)


JoseBobadilla20:
💚 Tested

Sure-Round:
@cmedinilla Can you attach a video of the expected behavior?

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -59,7 +62,10 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
             throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
         }
 
-        return this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(payment.methodId))
+        return this._store.dispatch(this._remoteCheckoutActionCreator.initializePayment(payment.methodId, { useStoreCredit }))
```
davidchin: I assume you need to do this because the client token generation requires the order amount with store credit applied? Is it possible for you to just create the order first? Unless it's not really possible, you should always try to create the order first before forwarding order details to a hosted payment method. If you do that, you don't really need to do that extra call because the order amount will be final with the store credit applied.

cmedinilla:
@Sure-Round https://drive.google.com/open?id=1A8p5EifpG3RRWN3HQjP86JiWQRdkHu2l

cmedinilla:
@davidchin I changed the order of the actions but i am still need send the flag to can added to the grand total

PR #564, opened 2019-05-13
feat(payment): INT-1247 Checkout using Zip Registration Referred

FranciscoOnt:
## What?
Added additional handlers for `declined` and `referred` responses from the zip Lightbox. 
Referred flow sends a request to clear the cart and save the externalId and then redirects to order confirmation page.

## Why?
Zip referred registration flow requires the order to seem purchased and redirect to order confirmation page.

## Testing / Proof
[Pending Deploy]
Screen Shot 2019-05-13 at 1 31 07 PM

### DIFF
--- a/src/payment/strategies/zip/zip-payment-strategy.spec.ts
+++ b/src/payment/strategies/zip/zip-payment-strategy.spec.ts
@@ -76,6 +76,8 @@ describe('ZipPaymentStrategy', () => {
 
         jest.spyOn(remoteCheckoutActionCreator, 'initializePayment')
             .mockReturnValue(submitOrderAction);
+        jest.spyOn(requestSender, 'sendRequest')
+            .mockReturnValue(Promise.resolve());
         jest.spyOn(orderActionCreator, 'submitOrder')
             .mockReturnValue(submitOrderAction);
         jest.spyOn(paymentActionCreator, 'submitPayment')
@@ -96,7 +98,8 @@ describe('ZipPaymentStrategy', () => {
             paymentActionCreator,
             paymentMethodActionCreator,
             remoteCheckoutActionCreator,
-            zipScriptLoader
+            zipScriptLoader,
+            requestSender
         );
     });
 
@@ -228,6 +231,17 @@ describe('ZipPaymentStrategy', () => {
                 expect(paymentActionCreator.submitPayment).not.toHaveBeenCalled();
                 expect(store.dispatch).not.toHaveBeenCalledWith(submitPaymentAction);
             }
+        });
+
+        it('continues to order confirmation if the registration is referred', async () => {
+            const referredZipClient = getZipScriptMock('referred');
+            const requestUrl = '/api/storefront/payment/zip/save-external-id';
+            jest.spyOn(zipScriptLoader, 'load')
+                .mockResolvedValue(referredZipClient);
+            await strategy.initialize(zipOptions);
+
+            await strategy.execute(orderRequestBody, zipOptions);
+            expect(requestSender.sendRequest).toHaveBeenCalledWith(requestUrl, expect.any(Object));
         });
     });
 
--- a/src/payment/strategies/zip/zip-payment-strategy.ts
+++ b/src/payment/strategies/zip/zip-payment-strategy.ts
@@ -1,3 +1,5 @@
+import { RequestSender, Response } from '@bigcommerce/request-sender';
+
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
 import {
     InvalidArgumentError,
@@ -6,6 +8,7 @@ import {
     NotInitializedError,
     NotInitializedErrorType,
 } from '../../../common/error/errors';
+import { ContentType, INTERNAL_USE_ONLY } from '../../../common/http-request';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
 import { RemoteCheckoutActionCreator } from '../../../remote-checkout';
@@ -29,7 +32,8 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
         private _paymentActionCreator: PaymentActionCreator,
         private _paymentMethodActionCreator: PaymentMethodActionCreator,
         private _remoteCheckoutActionCreator: RemoteCheckoutActionCreator,
-        private _zipScriptLoader: ZipScriptLoader
+        private _zipScriptLoader: ZipScriptLoader,
+        private _requestSender: RequestSender
     ) { }
 
     initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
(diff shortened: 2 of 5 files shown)


src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -1,5 +1,7 @@
 import { createClient as createPaymentClient } from '@bigcommerce/bigpay-client';
 import { createAction, Action } from '@bigcommerce/data-store';
+import { createFormPoster } from '@bigcommerce/form-poster';
+import { FormPoster } from '@bigcommerce/form-poster';
```
metalvegetarianoprogresivo: Unify this with the previous line.

metalvegetarianoprogresivo: Also, check the travis build.

cmedinilla:
In your commit payments must be in singular "payment"

cmedinilla:
In your commit message payments must be in singular "payment"

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -1,3 +1,6 @@
+import { FormPoster } from '@bigcommerce/form-poster';
+import { RequestSender, Response } from '@bigcommerce/request-sender';
+
```
cmedinilla: Why this extra break line?

FranciscoOnt: It's to separate the imports that are in `node_modules` from the rest of imports.

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -91,16 +109,48 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
...
+                checkoutId,
+            }),
+            method: 'post',
+        };
+
+        return this._requestSender.sendRequest(url, options);
+    }
+
+    private _goToOrderConfirmation() {
+        this._formPoster.postForm('/checkout/order-confirmation', {
```
davidchin: What are you trying to do here? Why are you doing HTML post to the order confirmation page? Do you just want to do a redirect? Also, we normally don't do any redirection to the order confirmation in the SDK because we let the client decide what they want to do immediately after a successful order submission. Having said that, we do have an exception for redirect payment methods where the callback happens directly from the provider / BigPay. But Zip seems to follow in-context flow instead. Therefore, resolving the promise seems to be a better solution, otherwise it'd be quite difficult to document.

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -91,16 +109,48 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
...
                     ));
             });
     }
 
     finalize(options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
         return Promise.reject(new OrderFinalizationNotRequiredError());
     }
+
+    private _prepareForReferredRegistration(checkoutId: string): Promise<Response> {
+        const url = `checkout.php?action=zip_referred_registration`;
```
davidchin: Don't use `` if you are not doing any string interpolation.

davidchin: Also, what type of endpoint is that? It's not a versioned storefront or internal API endpoint. And it's not a legacy endpoint (i.e.: `set_external_checkout`). Do you just want to post some arbitrary data to the server? I believe there's a better way to achieve it (i.e.: via provider-specific endpoints).

src/payment/strategies/zip/zip-payment-strategy.ts
FranciscoOnt: Yes, after the registration has entered the referred state Zip wants the shopper to be redirected to an order confirmation page ([docs](https://zip-online.api-docs.io/v1/checkout-and-registration/new-account-registration-flow-zip-money)) , how is the correct way of doing this?

FranciscoOnt:
@davidchin , I've made some changes as you suggested and removed `isNonceLike` usage, could you please re-review ♻️

src/payment/strategies/zip/zip-payment-strategy.ts
davidchin: Yep, just resolve the promise, you don't need to do the redirect here. The checkout app will perform the redirection, just like all the other payment methods.

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -69,13 +76,22 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
...
                                     return reject(new PaymentMethodCancelledError());
                                 }
 
+                                if (state === ZipModalEvent.CheckoutDeclined) {
+                                    return reject(new Error(this._zipDeclinedMessage));
+                                }
+
+                                if (state === ZipModalEvent.CheckoutReferred && checkoutId) {
+                                    return this._prepareForReferredRegistration(checkoutId)
+                                        .then(() => resolve(undefined));
```
davidchin: Just `resolve()` 🍺

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -91,16 +107,48 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
...
-                            methodId: payment.methodId,
-                            paymentData: { nonce },
-                        }))
+                    .then(nonce => {
+                            if (nonce !== undefined) {
+                                return this._store.dispatch(this._paymentActionCreator.submitPayment({
+                                    methodId: payment.methodId,
+                                    paymentData: { nonce },
+                                }));
+                            } else {
```
davidchin: You can remove this redundant `else`. 🍺

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -91,16 +107,48 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
...
     }
+
+    private _prepareForReferredRegistration(checkoutId: string): Promise<Response> {
+        const url = '/internalapi/v1/payments/zip';
+        const options = {
+            headers: {
+                Accept: 'text/html',
+                'Content-Type': 'application/x-www-form-urlencoded; charset=UTF-8',
+            },
+            body: toFormUrlEncoded({
```
davidchin: As I said above, you shouldn't need to encode this if you're sending the request body as JSON.

FranciscoOnt: Hi @davidchin , I tried to do this, but I was unable to access `externalId` in the post params, can you give some pointers on how to send this request correctly without using `toFormUrlEncoded `?

FranciscoOnt:
Hi @davidchin, I've made some changes to math the new approach, could you please re-review? ♻️

FranciscoOnt:
Hi @davidchin , I've made the changes as you suggested ♻️

davidchin:
I think everything looks good. But I can't approve it yet because there's an issue with the proposed endpoint. Once that gets resolved, I'll approve this PR 😄

Sure-Round:
@davidchin ♻️ I believe the endpoint that was blocking this has been addressed and merged. 29918 was the PR.

Sure-Round:
@FranciscoOnt please squash

FranciscoOnt:
@Sure-Round , I have squashed the commits

PR #567, opened 2019-05-14
feat(payments): PAYMENTS-4228 includes currency code for vaulted instrument

PharaohAsh:
## What?
Includes currency code to the shopper instrument calls.
* List Instruments
* TrustedShippingAddress

The currency code is based on the shopper's selected currency on the storefront. If the selected shopper currency code is not transactional, fallback to the store's default currency.
TODO: Bigpay client JS version bump

## Why?
Prepping for multi-currency for payments.

## Depends On
https://github.com/bigcommerce/bigcommerce/pull/30019

## Testing / Proof
Updated specs.

Passes currency code on {{trusted_shipping_address}} endpoint (amongst others) and uses store currency if shopper selected currency not transactional:
Screen Shot 2019-05-21 at 3 44 53 pm

### DIFF
--- a/src/checkout/checkout-service.ts
+++ b/src/checkout/checkout-service.ts
@@ -1020,7 +1020,7 @@ export default class CheckoutService {
      * `CheckoutStoreSelector#getInstruments`.
      *
      * ```js
-     * const state = service.deleteInstrument('123', 'USD');
+     * const state = service.deleteInstrument('123');
      *
      * console.log(state.data.getInstruments());
      * ```
--- a/src/checkout/checkout-service.ts
+++ b/src/checkout/checkout-service.ts
@@ -1014,22 +1014,23 @@ export default class CheckoutService {
     }
 
     /**
-     * Deletes a payment instrument by an id.
+     * Deletes a payment instrument by an id and currency.
      *
      * Once an instrument gets removed, it can no longer be retrieved using
      * `CheckoutStoreSelector#getInstruments`.
      *
      * ```js
-     * const state = service.deleteInstrument('123');
+     * const state = service.deleteInstrument('123', 'USD');
      *
      * console.log(state.data.getInstruments());
      * ```
      *
      * @param instrumentId - The identifier of the payment instrument to delete.
+     * @param currencyCode - The currency code as selected by the shopper.
      * @returns A promise that resolves to the current state.
      */
-    deleteInstrument(instrumentId: string): Promise<CheckoutSelectors> {
-        const action = this._instrumentActionCreator.deleteInstrument(instrumentId);
+    deleteInstrument(instrumentId: string, currencyCode: string): Promise<CheckoutSelectors> {
+        const action = this._instrumentActionCreator.deleteInstrument(instrumentId, currencyCode);
 
         return this._dispatch(action);
     }
--- a/src/checkout/checkout-service.ts
+++ b/src/checkout/checkout-service.ts
@@ -1014,7 +1014,7 @@ export default class CheckoutService {
     }
 
     /**
-     * Deletes a payment instrument by an id and currency.
+     * Deletes a payment instrument by an id.
      *
      * Once an instrument gets removed, it can no longer be retrieved using
      * `CheckoutStoreSelector#getInstruments`.
@@ -1026,11 +1026,10 @@ export default class CheckoutService {
      * ```
      *
      * @param instrumentId - The identifier of the payment instrument to delete.
-     * @param currencyCode - The currency code as selected by the shopper.
      * @returns A promise that resolves to the current state.
      */
-    deleteInstrument(instrumentId: string, currencyCode: string): Promise<CheckoutSelectors> {
-        const action = this._instrumentActionCreator.deleteInstrument(instrumentId, currencyCode);
+    deleteInstrument(instrumentId: string): Promise<CheckoutSelectors> {
+        const action = this._instrumentActionCreator.deleteInstrument(instrumentId);
 
         return this._dispatch(action);
     }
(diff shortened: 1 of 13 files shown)


src/checkout/checkout-service.ts
```diff
@@ -1014,22 +1014,23 @@ export default class CheckoutService {
     }
 
     /**
-     * Deletes a payment instrument by an id.
+     * Deletes a payment instrument by an id and currency.
```
capsula4: maybe `currency code` for extra clarity?

src/checkout/checkout-service.ts
```diff
@@ -1014,22 +1014,23 @@ export default class CheckoutService {
...
      * console.log(state.data.getInstruments());
      * ```
      *
      * @param instrumentId - The identifier of the payment instrument to delete.
+     * @param currencyCode - The currency code as selected by the shopper.
      * @returns A promise that resolves to the current state.
      */
-    deleteInstrument(instrumentId: string): Promise<CheckoutSelectors> {
-        const action = this._instrumentActionCreator.deleteInstrument(instrumentId);
+    deleteInstrument(instrumentId: string, currencyCode: string): Promise<CheckoutSelectors> {
```
capsula4: Is the currency code strictly necessary to delete an instrument? If so, this seems like a breaking change/

PharaohAsh: The future potential is we'll eventually be moving everything to require a currency code. I'm happy to keep it as optional for now though.

PharaohAsh: @capsula4 I do have a question though:
```
const { code } = config.shopperCurrency;
```
If I make the currency optional, does that imply this can be undefined? Or am I going about it the wrong way?

capsula4: @PharaohAsh if we will eventually move things to require a currency code, that could break custom checkouts. If we plan to actually make currency code a requirement in the API is not enough by fixing the SDK as partners have to manually migrate to a newer version, or they could even be integrating against Bigpay API directly. We might need to communicate this and give them time to migrate. I would prefer we keep it optional. 

On the other hand `shopperCurrency` should always be defined. If the shopper hasn't selected manually any currency, it should match the default currency. Keep in mind that shopper currency includes display currencies which I guess are not eligible for a parameter for deleting an instrument.

PharaohAsh: @capsula4  Ah, right. That makes sense. I didn't know about the custom checkouts... that's complex indeed.

In regards to display currencies vs transactional currencies: theoretically, we should not be storing an instrument against a display currency and hence there should not be a case where you will be listing instruments for that particular display currency. In turn you wouldn't get a case where you'd delete an instrument under a display currency either.

Which is actually a very good point you've raised. Is there a way to determine what is and isn't a transactional currency?

PharaohAsh: @capsula4 It looks like I'll have to decide the currency type on BCApp and prevent users from adding cards on display currencies.

src/payment/instrument/instrument-action-creator.ts
```diff
@@ -125,10 +125,12 @@ export default class InstrumentActionCreator {
 
         const { customerId } = cart;
         const { storeId } = config.storeProfile;
+        const { code } = config.shopperCurrency;
```
capsula4: The active shopper currency does not necessarily have to match the currency that the stored instrument was created with. Also, `shopperCurrency` could be a display currency.

However, I'm still thinking that if the instrument ID is unique, we shouldn't need the currency code to delete it. If it's a 1-1 relationship, I would encourage to drop the requirement for a currency code when deleting an instrument.

If you still think that we need to pass currency code to delete an instrument, probably the correct way would involve modifying the instruments endpoint so each instrument returns the currency linked with each instrument.

src/checkout/checkout-service.ts
```diff
@@ -1020,7 +1020,7 @@ export default class CheckoutService {
      * `CheckoutStoreSelector#getInstruments`.
      *
      * ```js
-     * const state = service.deleteInstrument('123');
+     * const state = service.deleteInstrument('123', 'USD');
```
capsula4: no longer need this

src/payment/instrument/instrument-action-creator.ts
```diff
@@ -131,4 +131,28 @@ export default class InstrumentActionCreator {
             storeId,
         };
     }
+
+    private _getSessionContextWithCurrency(store: ReadableCheckoutStore): SessionContext {
```
capsula4: I think it should be okay to, instead of creating a new method, just modify `_getSessionContext` to include the currency? They look pretty much like the same method to me

src/payment/instrument/instrument-action-creator.spec.ts
```diff
@@ -73,7 +77,23 @@ describe('InstrumentActionCreator', () => {
 
             expect(instrumentRequestSender.getVaultAccessToken).toHaveBeenCalled();
             expect(instrumentRequestSender.loadInstruments).toHaveBeenCalledWith(
-                { storeId, customerId, authToken: vaultAccessToken },
+                { storeId, customerId, currencyCode, authToken: vaultAccessToken },
+                shippingAddress
+            );
+        });
+
+        it('sends a request to get a list of instruments when shopper currency is not transactional', async () => {
```
capsula4: `sends a request to get a list of instruments using default currency when shopper currency is not transactional`

src/payment/instrument/instrument-action-creator.spec.ts
```diff
@@ -73,7 +77,23 @@ describe('InstrumentActionCreator', () => {
 
             expect(instrumentRequestSender.getVaultAccessToken).toHaveBeenCalled();
             expect(instrumentRequestSender.loadInstruments).toHaveBeenCalledWith(
-                { storeId, customerId, authToken: vaultAccessToken },
+                { storeId, customerId, currencyCode, authToken: vaultAccessToken },
```
capsula4: `it('sends a request to get a list of instruments using shopper currency')`

PR #570, opened 2019-05-17
INT-1450 Add support for Stripe V3 + 3DS using payment Intents

clopezh:
[INT-1450](https://jira.bigcommerce.com/browse/INT-1450)

## What?
Add support for Stripe V3 + 3DS that can handle stripe validations with stripe.js (handleCardPayment action)

## Why?
To support 3DS on Stripe is needed to add a new strategy that use Stripe.js elements and actions to handle 3DS flow and confirm a payment intent that previously was created from back-end

## Testing / Proof
image

## Sibling PRs
- [BCAPP](https://github.com/bigcommerce/bigcommerce/pull/29991)
- [NGCheckout](https://github.com/bigcommerce-labs/ng-checkout/pull/1227)
- [NGPayments](https://github.com/bigcommerce-labs/ng-payments/pull/799)
- [BigPay](https://github.com/bigcommerce/bigpay/pull/1649)

**Ping** @bigcommerce/intersys-integrations @bigcommerce/payments  @bigcommerce/checkout

### DIFF
--- a/src/payment/strategies/stripev3/stripev3-initialize-options.ts
+++ b/src/payment/strategies/stripev3/stripev3-initialize-options.ts
@@ -1,0 +1,20 @@
+import { StripeStyleProps } from './stripev3';
+
+/**
+ * A set of options that are required to initialize the Stripe payment method.
+ *
+ * Once Stripe payment is initialized, credit card form fields, provided by the
+ * payment provider as iframes, will be inserted into the current page. These
+ * options provide a location and styling for each of the form fields.
+ */
+export default interface StripeV3PaymentInitializeOptions {
+    /**
+     * The location to insert the credit card number form field.
+     */
+    containerId: string;
+
+    /**
+     * The set of CSS styles to apply to all form fields.
+     */
+    style?: StripeStyleProps;
+}
--- a/src/payment/strategies/stripev3/stripev3-payment-strategy.spec.ts
+++ b/src/payment/strategies/stripev3/stripev3-payment-strategy.spec.ts
@@ -1,0 +1,252 @@
+import { createRequestSender } from '@bigcommerce/request-sender';
+import { createScriptLoader } from '@bigcommerce/script-loader';
+
+import { getCartState } from '../../../cart/carts.mock';
+import {
+    createCheckoutStore,
+    CheckoutRequestSender,
+    CheckoutValidator
+} from '../../../checkout';
+import CheckoutStore from '../../../checkout/checkout-store';
+import { getCheckoutState } from '../../../checkout/checkouts.mock';
+import {
+    InvalidArgumentError,
+    MissingDataError,
+    StandardError
+} from '../../../common/error/errors';
+import { getConfigState } from '../../../config/configs.mock';
+import { getCustomerState } from '../../../customer/customers.mock';
+import { OrderFinalizationNotRequiredError } from '../../../order/errors';
+import OrderActionCreator from '../../../order/order-action-creator';
+import {
+    createPaymentClient,
+    PaymentInitializeOptions,
+    PaymentMethodRequestSender,
+    PaymentRequestSender
+} from '../../../payment';
+import { PaymentArgumentInvalidError } from '../../errors';
+import PaymentActionCreator from '../../payment-action-creator';
+import PaymentMethodActionCreator from '../../payment-method-action-creator';
+import { getPaymentMethodsState, getStripeV3 } from '../../payment-methods.mock';
+
+import { StripeResponse } from './stripev3';
+import StripeV3PaymentStrategy from './stripev3-payment-strategy';
+import StripeV3ScriptLoader from './stripev3-script-loader';
+import {
+    getStripeV3HandleCardResponse,
+    getStripeV3InitializeOptionsMock,
+    getStripeV3JsMock,
+    getStripeV3OrderRequestBodyMock
+} from './stripev3.mock';
+
+describe('StripeV3PaymentStrategy', () => {
+    let store: CheckoutStore;
+    let strategy: StripeV3PaymentStrategy;
+    let paymentMethodActionCreator: PaymentMethodActionCreator;
+    let paymentActionCreator: PaymentActionCreator;
+    let orderActionCreator: OrderActionCreator;
+    let stripeScriptLoader: StripeV3ScriptLoader;
+
+    beforeEach(() => {
+        store = createCheckoutStore({
+            checkout: getCheckoutState(),
+            customer: getCustomerState(),
+            config: getConfigState(),
+            cart: getCartState(),
+            paymentMethods: getPaymentMethodsState(),
+        });
+
+        const requestSender = createRequestSender();
+        const paymentMethodRequestSender: PaymentMethodRequestSender = new PaymentMethodRequestSender(requestSender);
+        const paymentClient = createPaymentClient(store);
+        const scriptLoader = createScriptLoader();
+
+        paymentMethodActionCreator = new PaymentMethodActionCreator(paymentMethodRequestSender);
+        paymentActionCreator = new PaymentActionCreator(new PaymentRequestSender(paymentClient), orderActionCreator);
+        orderActionCreator = new OrderActionCreator(
+            paymentClient,
+            new CheckoutValidator(
+                new CheckoutRequestSender(requestSender)
+            )
+        );
+        stripeScriptLoader = new StripeV3ScriptLoader(scriptLoader);
+
+        strategy = new StripeV3PaymentStrategy(
+            store,
+            paymentMethodActionCreator,
+            paymentActionCreator,
+            orderActionCreator,
+            stripeScriptLoader
+        );
+    });
+
+    describe('#initialize()', () => {
+        const stripeV3JsMock = getStripeV3JsMock();
+        let stripeV3Options: PaymentInitializeOptions;
+
+        beforeEach(() => {
+            stripeV3Options = getStripeV3InitializeOptionsMock();
+
+            jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod').mockReturnValue(getStripeV3());
+        });
+
+        it('loads stripe v3 script', async () => {
+            jest.spyOn(stripeScriptLoader, 'load').mockReturnValue(Promise.resolve(stripeV3JsMock));
+
+            const promise =  strategy.initialize(stripeV3Options);
+
+            expect(stripeScriptLoader.load).toHaveBeenCalled();
+
+            return expect(promise).resolves.toBe(store.getState());
+        });
+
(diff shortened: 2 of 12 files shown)


src/payment/strategies/stripev3/stripev3-script-loader.spec.ts
```diff
@@ -0,0 +1,59 @@
+import { ScriptLoader } from '@bigcommerce/script-loader';
+
+import { StandardError } from '../../../common/error/errors';
+
+import {StripeHostWindow, StripeV3JsOptions} from './stripev3';
```
FranciscoOnt: space between `{` and the import names

src/payment/strategies/stripev3/stripev3-script-loader.ts
```diff
@@ -0,0 +1,26 @@
+import { ScriptLoader } from '@bigcommerce/script-loader';
+
+import {StandardError} from '../../../common/error/errors';
```
FranciscoOnt: space between `{` and the import names

src/payment/strategies/stripev3/stripev3.mock.ts
```diff
@@ -0,0 +1,62 @@
+import OrderRequestBody from '../../../order/order-request-body';
+import {PaymentInitializeOptions} from '../../payment-request-options';
```
FranciscoOnt: space between { and the import names

metalvegetarianoprogresivo:
Can you add the link to the jira ticket to the description?

clopezh:
@FranciscoOnt All changes done! ♻️ 
@metalvegetarianoprogresivo Jira ticket added to PR description ♻️

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -0,0 +1,129 @@
...
+                }
+
+                return this._getStripeJs().handleCardPayment(
+                    paymentMethod.clientToken, this._getCardElement(), {}
+                ).then((stripeResponse: StripeResponse) => {
+                    if (stripeResponse.error) {
+                        throw new StandardError(stripeResponse.error.message);
+                    } else {
+                        if (!stripeResponse.paymentIntent.id) {
+                            throw new PaymentArgumentInvalidError(['paymentIntent.id']);
```
davidchin: `PaymentArgumentInvalidError` is only appropriate when the SDK user forgets to pass in a required parameter. So it shouldn't be used here as `paymentIntent.id` is not something passed in by the SDK user.

I had a look at the doc, it seems `id` would only be undefined if `error` is defined. So you could probably combine the check with the above. i.e.:

```ts
if (response.error || !response.paymentIntent.id) {
    throw new StandardError(response.error && response.error.message);
}
```

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -0,0 +1,129 @@
...
+                        };
+
+                        return this._store.dispatch(this._orderActionCreator.submitOrder(order, options))
+                            .then(() =>
+                                this._store.dispatch(this._paymentActionCreator.submitPayment(paymentPayload))
+                            );
+                    }
+                });
+            })
+            .catch((error: Error) => { throw new StandardError(error.message); });
```
davidchin: You don't actually need to do this unless you want to throw a specific error type.

src/payment/strategies/stripev3/stripev3-script-loader.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { StripeHostWindow, StripeV3Js } from './stripev3';
+
+export default class StripeV3ScriptLoader {
+    constructor(
+        private _scriptLoader: ScriptLoader,
+        private _window: StripeHostWindow = window
+    ) {}
+
+    load(stripePublishableKey: string): Promise<StripeV3Js> {
```
davidchin: Just `publishableKey` is clear enough 🍺

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -0,0 +1,129 @@
...
+
+                return Promise.resolve(this._store.getState());
+            });
+    }
+
+    execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        const { payment, ...order } = payload;
+
+        if (!payment) {
+            throw new PaymentArgumentInvalidError(['payment.paymentData']);
```
davidchin: Just `throw new PaymentArgumentInvalidError(['payment']);` as `paymentData` is not required.

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -0,0 +1,129 @@
...
+
+        return this._stripeScriptLoader.load(paymentMethod.initializationData.stripePublishableKey)
+            .then((stripeJs: StripeV3Js) => {
+                this.stripeJs = stripeJs;
+                const elements = this.stripeJs.elements();
+                this.cardElement = elements.create('card', {
+                    style: DOMElement.elementProps,
+                });
+
+                if (!this.cardElement) {
```
davidchin: `create` method actually always returns `StripeCardElement` so you don't actually need to do a null check here. If you just define a variable like this, you should be able to use it directly.

```ts
const cardElement = elements.create('card', {
    style: stripeOptions.style,
});

cardElement.mount(`#${stripeOptions.containerId}`);
```

If you are planning to use `cardElement` in another method, you can store it as an instance member afterwards. i.e.: 

```ts
this._cardElement = cardElement;
```

src/payment/strategies/stripev3/stripev3-initialize-options.ts
```diff
@@ -0,0 +1,20 @@
...
+export default interface StripeV3PaymentInitializeOptions {
+    /**
+     * The location to insert the credit card number form field.
+     */
+    containerId: string;
+
+    /**
+     * The set of CSS styles to apply to all form fields.
+     */
+    elementProps?: StripeStyleProps;
```
davidchin: `style` is a better name for the property.

src/payment/strategies/stripev3/stripev3-payment-strategy.spec.ts
```diff
@@ -0,0 +1,260 @@
...
+                    testMode: true,
+                },
+                type: 'PAYMENT_TYPE_API',
+            };
+
+            jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod').mockReturnValue(stripeV3);
+
+            const response = strategy.execute(getStripeV3OrderRequestBodyMock());
+
+            expect(response).rejects.toBeInstanceOf(StandardError);
```
davidchin: If you want to use `rejects` or `resolves`, you'll need to add `return` before the `expect`. i.e.:

```ts
return expect(response).rejects.toBeInstanceOf(StandardError);
```

Otherwise, the test won't wait for the promise to reject / resolve and just pass itself.

clopezh:
Hello @davidchin Thank you for your feedback, I already addressed all the comments, Could you please take a look and If possible approve it? ♻️

src/payment/strategies/stripev3/stripev3-payment-strategy.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+
+import { getCartState } from '../../../cart/carts.mock';
+import {
+    createCheckoutStore,
+    CheckoutRequestSender,
+    CheckoutValidator
+} from '../../../checkout';
+import CheckoutStore from '../../../checkout/checkout-store';
+import { getCheckoutState } from '../../../checkout/checkouts.mock';
+import {InvalidArgumentError, MissingDataError, StandardError} from '../../../common/error/errors';
```
davidchin: Space around `{}` 💅

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -0,0 +1,123 @@
...
+} from '../../../common/error/errors';
+import { OrderActionCreator, OrderRequestBody } from '../../../order';
+import { OrderFinalizationNotRequiredError } from '../../../order/errors';
+import { PaymentArgumentInvalidError } from '../../errors';
+import PaymentActionCreator from '../../payment-action-creator';
+import PaymentMethodActionCreator from '../../payment-method-action-creator';
+import { PaymentInitializeOptions, PaymentRequestOptions } from '../../payment-request-options';
+import PaymentStrategy from '../payment-strategy';
+
+import {
```
davidchin: Can you update the import statements to be:

```ts
import { StripeCardElement, StripeV3Client } from './stripev3';
import StripeV3ScriptLoader from './stripev3-script-loader';
```

💅

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    deinitialize(options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        this._getCardElement().unmount();
+
+        return Promise.resolve(this._store.getState());
+    }
+
+    private _getStripeJs(): StripeV3Client {
+        if (!this._stripeV3Client) {
+            throw new InvalidArgumentError('Unable to initialize payment because "stripeJs" argument is not provided.');
```
davidchin: You should throw `NotInitializedError` error here because the reason why it is undefined it's because `initialize` is not called.

PS: Side note, we actually don't recommend you to create private getters like this. This is because you want to throw different errors depending on the caller. i.e.: If you're trying to access the property in `execute`, you want to throw `NotInitializedError`, but if you're trying to access it in `initialize`, you'd want to throw a different error. I understand you try to avoid doing a check time you access the property. But the more important thing is to make sure we're throwing the correct error.

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -0,0 +1,123 @@
...
+                });
+            });
+    }
+
+    finalize(options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        return Promise.reject(new OrderFinalizationNotRequiredError());
+    }
+
+    deinitialize(options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        this._getCardElement().unmount();
```
davidchin: You should catch `NotInitializedError` because `unmount` action is actually not critical so we shouldn't halt the execution of the app. That again is one of the reasons why we want you guys to do an explicit check instead of using convenient getters, because here you can just return early instead of having to do a `try/catch`.

clopezh:
@davidchin Thank you again for your feedback! all is done! Could you please take a look and if possible approve it? ♻️

PR #573, opened 2019-05-22
feat(payment): INT-1540 Zip Feedback, declined handler

FranciscoOnt:
## What?
Add a handler for a `cancelled order` case

## Why?
Its possible for an Zip account application to be declined instantly and it needs a handler to show an error message [as seen here](https://zip-online.api-docs.io/v1/checkout-and-registration/checkout-with-registration-decline).

## Testing / Proof
[Demo Video](https://drive.google.com/open?id=1PFETFnO7fkXJaJ_SLmZawarAd5ChYfLO)
Screen Shot 2019-05-22 at 3 38 17 PM

### DIFF
--- a/src/payment/strategies/zip/zip-payment-strategy.ts
+++ b/src/payment/strategies/zip/zip-payment-strategy.ts
@@ -9,8 +9,7 @@ import {
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
 import { RemoteCheckoutActionCreator } from '../../../remote-checkout';
-import { PaymentMethodCancelledError } from '../../errors';
-import PaymentMethodInvalidError from '../../errors/payment-method-invalid-error';
+import { PaymentMethodCancelledError, PaymentMethodDeclinedError, PaymentMethodInvalidError } from '../../errors';
 import PaymentActionCreator from '../../payment-action-creator';
 import PaymentMethod from '../../payment-method';
 import PaymentMethodActionCreator from '../../payment-method-action-creator';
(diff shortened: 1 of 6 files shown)


src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -80,6 +82,10 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
                                     return resolve(checkoutId);
                                 }
 
+                                if (state === ZipModalEvent.CheckoutDeclined) {
+                                    return reject(new StandardError(this._zipDeclinedMessage));
```
vmparra: There's no need for a constant here as this is the only place were it's been used.

FranciscoOnt:
Done @vmparra ! 👍

davidbarocio:
💚Tested

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -80,6 +81,10 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
                                     return resolve(checkoutId);
                                 }
 
+                                if (state === ZipModalEvent.CheckoutDeclined) {
+                                    return reject(new StandardError('Unfortunately your application was declined. Please select another payment method.'));
```
davidchin: It might be a good idea to define a more specific error type. i.e.: `PaymentDeclinedError` instead of generic `StandardError`. This way, the SDK user can simply check the `type` and react to it. Otherwise, they will need to check the error message itself, which is not very useful.

PS: I think currently `StandardError` is getting abused so I'm thinking about making it an abstract class instead in the future.

Sure-Round:
@FranciscoOnt Please rebase and squash and we will merge tomorrow

FranciscoOnt:
@Sure-Round , I've rebased and squashed the commits

PR #574, opened 2019-05-22
feat(payment): INT-1562 Add billing and shipping data in auth instead of load method calls in Klarna

josesonora:
## What? [INT-1562](https://jira.bigcommerce.com/browse/INT-1562)
Add the billing and shipping data during the call to authorize method in Klarna SDK

## Why?
We need to meet the PII requirements for EU countries. According to Klarna´s requirements for EU countries, PII data has to be sent during the `authorize` transaction, after the shopper has explicitly selected Klarna as payment method. Currently in the scenario when Klarna is the only available payment method, it is auto selected and the the PII data is sent on widget load.

## Testing / Proof
[Video](https://drive.google.com/open?id=1WRyrva0HINmS4Iq0ZbnwMpulxOluyosU)

image

image

### DIFF
--- a/src/payment/strategies/klarna/klarna-credit.ts
+++ b/src/payment/strategies/klarna/klarna-credit.ts
@@ -1,8 +1,8 @@
 export default interface KlarnaCredit {
-    authorize(params: any, callback: (res: KlarnaAuthorizationResponse) => void): void;
+    authorize(params: any, data: KlarnaUpdateSessionParams, callback: (res: KlarnaAuthorizationResponse) => void): void;
+    authorize(data: KlarnaUpdateSessionParams, callback: (res: KlarnaAuthorizationResponse) => void): void;
     init(params: KlarnaInitParams): void;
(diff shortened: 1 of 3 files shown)


src/payment/strategies/klarna/klarna-credit.ts
```diff
@@ -1,8 +1,8 @@
 export default interface KlarnaCredit {
-    authorize(params: any, callback: (res: KlarnaAuthorizationResponse) => void): void;
+    authorize(params: any, data: KlarnaUpdateSessionParams, callback: (res: KlarnaAuthorizationResponse) => void): void;
```
Sure-Round: Do we have duplicate lines here?

josesonora: @Sure-Round It's an overloaded method, I'm following same recommendation that Catalina gave me [here](https://github.com/bigcommerce/checkout-sdk-js/pull/554#discussion_r275182376) and I consider that this is a similar case

PR #576, opened 2019-05-23
INT-1500: Fix integer conversion rounding error

davidchin:
## What?
* Use `round` instead of `floor`

## Why?
* To fix potential rounding issue when converting numbers into integers.

## Testing / Proof
* Unit


davidchin:
cc @michxff @snaderiBC

PR #600, opened 2019-05-31
fix(payments): INT-1573 Checkout load error when non-supported Klarna country is selected

BC-JMolina:
## What?
 - Remove reject(new PaymentMethodInvalidError())

## Why?
In the scenario when the merchant does not support the shopper's billing country based on their merchant agreement with Klarna the checkout page should not throw an error, instead it should just show the widget showing that the country is not supported.

## Testing / Proof
Klarna-INT-1573

## Sibling PRs
[BCApp INT-1573](https://github.com/bigcommerce/bigcommerce/pull/30174)


cbandam:
@BC-JMolina please add testing proof

PR #602, opened 2019-06-03
feat(order): CHECKOUT-2530 Add spam protection for order creation

RobinOng:
## What?
As per above.

## Why?
To prevent spam/bot from creating orders.

## Testing / Proof
Added unit tests.


src/checkout/checkout-service.ts
```diff
@@ -1052,6 +1054,25 @@ export default class CheckoutService {
         return this._dispatch(action);
     }
 
+    /**
+     * Initializes the spam protection for order creation.
+     *
+     * The order creation will fail if spam protection is enabled but
```
capsula4: this is the documentation the developer gets, maybe explain a bit more what the spam protection is about?

src/checkout/checkout-service.ts
```diff
@@ -1052,6 +1054,25 @@ export default class CheckoutService {
...
+     * is not completed.
+     *
+     * ```js
+     * await service.initializeOrderSpamProtection('spamProtectionContainer');
+     * ```
+     *
+     * @param containerId - The container ID where the spam protection should be initialized.
```
capsula4: `The container ID where the spam protection should be rendered.`  ?

src/checkout/checkout-service.ts
```diff
@@ -1052,6 +1054,25 @@ export default class CheckoutService {
...
+     * @returns A promise that resolves to the current state.
+     */
+    initializeOrderSpamProtection(containerId: string): Promise<CheckoutSelectors> {
```
capsula4: I wonder if we actually need a registry of different type of spam protection strategies, considering this is a library and might get extended, same way we do with shipping strategies.

`initializeSpamProtection(options: spamProtectionOptions)`

Also, checkout SDK is all about checkout to order, so I would drop the word `order` from the method name.

Since this is a public library, it's better to tackle this issues early as refactoring later would cause breaking changes.

RobinOng: Initially I had it as `initializeSpamProtection` without the word `order`. The reason why I included it is because I am afraid that developers might misunderstand and thought to put it in other places and expect it to work, for example putting it to prevent people for spamming the login. Happy to change it back to `initializeSpamProtection` since the documentation should be enough to explain this.

src/checkout/checkout-service.ts
RobinOng: 👍

src/checkout/checkout-service.ts
RobinOng: I will abstract it to `spamProtectionOptions`

src/checkout/checkout-service.ts
RobinOng: 👌 Will add more details on this.

src/checkout/checkout-service.spec.ts
```diff
@@ -970,4 +978,21 @@ describe('CheckoutService', () => {
                 .toHaveBeenCalledWith(error);
         });
     });
+
+    describe('#initializeOrderSpamProtection()', () => {
```
capsula4: `#initializeSpamProtection`

src/checkout/checkout-service.ts
```diff
@@ -1052,6 +1054,26 @@ export default class CheckoutService {
...
+     *
+     * With spam protection enabled, the customer has to be verified as
+     * a human. The order creation will fail if spam protection
+     * is enabled but verification fails.
+     *
+     * ```js
+     * await service.initializeSpamProtection('spamProtectionContainer');
+     * ```
+     *
+     * @param containerId - The container ID where the spam protection should be rendered.
```
capsula4: doesn't match implementation

src/order/order-action-creator.ts
```diff
@@ -68,14 +68,25 @@ export default class OrderActionCreator {
...
+
+                if (!storeConfig) {
+                    throw new MissingDataError(MissingDataErrorType.MissingCheckoutConfig);
+                }
 
                 if (!checkout) {
                     throw new MissingDataError(MissingDataErrorType.MissingCheckout);
                 }
 
+                if (storeConfig.checkoutSettings.isSpamProtectionEnabled && !spamProtectionToken) {
```
capsula4: consider having a const for better readability

```javascript
const { isSpamProtectionEnabled } = storeConfig.checkoutSettings;

if (isSpamProtectionEnabled && !spamProtectionToken) {
```

src/order/order-action-creator.ts
```diff
@@ -68,14 +68,25 @@ export default class OrderActionCreator {
...
+                    throw new MissingDataError(MissingDataErrorType.MissingSpamProtection);
+                }
+
                 return from(
                     this._checkoutValidator.validate(checkout, options)
-                        .then(() => this._orderRequestSender.submitOrder(this._mapToOrderRequestBody(payload, checkout.customerMessage, externalSource), options))
+                        .then(() => this._orderRequestSender.submitOrder(this._mapToOrderRequestBody(payload, checkout.customerMessage, externalSource, spamProtectionToken), options))
```
capsula4: this line it's already a bit long, might be good to split the arguments into new lines
```javascript
.then(() => this._orderRequestSender.submitOrder(this._mapToOrderRequestBody(
  payload,
  checkout.customerMessage,
  externalSource,
  spamProtectionToken,
), options))
```

src/common/error/errors/missing-data-error.ts
```diff
@@ -57,6 +58,9 @@ function getErrorMessage(type: MissingDataErrorType): string {
     case MissingDataErrorType.MissingShippingAddress:
         return 'Unable to proceed because shipping address data is unavailable.';
 
+    case MissingDataErrorType.MissingSpamProtection:
+            return 'Unable to proceed because spam protection is not completed.';
```
capsula4: indentation off. Maybe this should be `Unable to proceed because spam protection token is not provided`

Since this is for developers, it could be that the challenge was rendered, but the developer is not passing it in the code itself.

src/common/error/errors/missing-data-error.ts
```diff
@@ -57,6 +58,9 @@ function getErrorMessage(type: MissingDataErrorType): string {
     case MissingDataErrorType.MissingShippingAddress:
         return 'Unable to proceed because shipping address data is unavailable.';
 
+    case MissingDataErrorType.MissingSpamProtection:
```
capsula4: `MissingSpamProtectionToken`

src/order/spam-protection/spam-protection-actions.ts
```diff
@@ -0,0 +1,24 @@
+import { Action } from '@bigcommerce/data-store';
+
+export enum SpamProtectionActionType {
+    InitializeFailed = 'SPAM_PROTECTION_INITIALIZE_FAILED',
+    InitializeRequested = 'SPAM_PROTECTION_INITIALIZE_REQUESTED',
+    Completed = 'SPAM_PROTECTION_COMPLETED',
```
capsula4: Maybe for consistency with how we name actions, this should be:
`InitializeSucceeded = 'SPAM_PROTECTION_INITIALIZE_SUCCEEDED'`

src/order/spam-protection/spam-protection-actions.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export type SpamProtectionAction =
+    InitializeRequestedAction |
+    InitializeFailedAction |
+    CompletedAction;
```
capsula4: same as above, replace all `completed` with `succeeded`, e.g.`InitializeSucceededAction`

src/checkout/checkout-service.ts
```diff
@@ -1052,6 +1054,26 @@ export default class CheckoutService {
...
 
+    /**
+     * Initializes the spam protection for order creation.
+     *
+     * With spam protection enabled, the customer has to be verified as
+     * a human. The order creation will fail if spam protection
+     * is enabled but verification fails.
+     *
+     * ```js
+     * await service.initializeSpamProtection('spamProtectionContainer');
```
capsula4: this needs updating too

src/order/spam-protection/spam-protection-action-creator.ts
```diff
@@ -0,0 +1,40 @@
...
+                throw new MissingDataError(MissingDataErrorType.MissingCheckoutConfig);
+            }
+
+            observer.next(createAction(SpamProtectionActionType.InitializeRequested, undefined));
+
+            const recaptchaSitekey = config.storeConfig.checkoutSettings.googleRecaptchaSitekey;
+
+            return this._googleRecaptcha.render(containerId, recaptchaSitekey)
+                .then(key => {
+                    observer.next(createAction(SpamProtectionActionType.Completed, key));
```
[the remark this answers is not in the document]
davidchin: By the way, besides the "complete" action, you might also want to trigger an "expired" action as well.

PR #608, opened 2019-06-07
docs(checkout): Update readme to clarify continueAsGuest

jamiekpac:
## What?
Update README to reduce confusion around continueAsGuest method.

## Why?
The confusion comes about because the guest data is separated from the customer object in some ways but not in others. The customer object has a reference to isGuest, but doesn't also get utilised to store the guests email, the guest email is actually stored against the cart object when it's set with this method.

Logging getCustomer() after setting continueAsGuest({ email: 'foo@bar.com' }) can cause confusion as the returned data will have the email displayed as an empty string and the isGuest flag set to true.

Would it make sense to mention that the getCart() email will be updated to include the guest user email, and the getCustomer() isGuest will be updated to indicate the user is a guest so consumers of the API know where the separation lies?

## Testing / Proof
n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ Alternatively, you can ask the customer to continue as a guest.
 ```js
 const state = await service.continueAsGuest({ email: 'foo@bar.com' });
 
-console.log(state.data.getCustomer());
+console.log(state.data.getCart());
 ```
 
 ### Set shipping details


README.md
```diff
@@ -143,7 +143,7 @@ Alternatively, you can ask the customer to continue as a guest.
 ```js
 const state = await service.continueAsGuest({ email: 'foo@bar.com' });
 
-console.log(state.data.getCustomer());
+console.log(state.data.getCart());
```
capsula4: I think the email is stored as part of the billing address, not the cart?

jamiekpac: I could be wrong but I think your question highlights that this is somewhat confusing to consumers of the SDK. The only place 'email' is referenced in the README is the 'Sign in customer' section.

From what we've seen in our implementation the customer/guest is set as the first step in the checkout process (before the billing or shipping sections), if the user logs in as a customer the email for the user is against getCustomer(), if the user opts to checkout as a guest the email for the user is against getCart().

If the relevant email for the submitOrder() will always be against the getBillingAddress() rather than getCart or getCustomer after calling signInCustomer() or continueAsGuest() in the first step of the checkout process it might be worth noting in the README.

I think you have the right approach by trying to keep the SDK as a declarative API but it might be a little too obscure for some of the key methods.

capsula4: You are right, if the user opts to continue as a guest, then the email won't be part of the Customer object. However, we don't store any `email` information as part of the `Cart` object https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/cart/cart.ts 

You can actually access the email in the `Checkout` object, inside the `billingAddress` property https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/checkout/checkout.ts, which you can access also using the `getBillingAddress()` method. Let me know if I'm missing something.

jamiekpac: When I log the object returned with `data.getCart()` I'm seeing something like this, which includes the email that has been applied with either method (continueAsGuest or signInCustomer) is this unexpected behaviour?

`{
    "id": "33a78ad1-5ad2-49b7-9e9c-d214f9a598a1",
    "customerId": 0,
    "email": "my-cart-email-here-either-customer-or-guest@email.com",
    "currency": {
        "name": "Australian Dollars",
        "code": "AUD",
        "symbol": "$",
        "decimalPlaces": 2
    },
    "isTaxIncluded": true,
    "baseAmount": 69.95,
    "discountAmount": 0,
    "cartAmount": 69.95,
    "coupons": [],
    "discounts": [
        {
            "id": "{{hidden}}",
            "discountedAmount": 0
        }
    ],
    "lineItems": {
        "physicalItems": [
            {
                "id": "{{hidden}}",
                "parentId": null,
                "variantId": 4284,
                "productId": 489,
                "sku": "SM91WB1C_BLK_2XL",
                "name": "Combat Camo Short Black",
                "url": "{{hidden}}",
                "quantity": 1,
                "brand": "SUPERDRY SPORT",
                "isTaxable": true,
                "imageUrl": "{{hidden}}",
                "discounts": [],
                "discountAmount": 0,
                "couponAmount": 0,
                "listPrice": 69.95,
                "salePrice": 69.95,
                "extendedListPrice": 69.95,
                "extendedSalePrice": 69.95,
                "isShippingRequired": true,
                "giftWrapping": null,
                "addedByPromotion": false,
                "options": [
                    {
                        "name": "Size",
                        "nameId": 850,
                        "value": "2XL",
                        "valueId": 2798
                    }
                ]
            }
        ],
        "digitalItems": [],
        "giftCertificates": [],
        "customItems": []
    },
    "createdTime": "2019-06-11T03:32:41+00:00",
    "updatedTime": "2019-06-11T04:08:21+00:00"
}`

capsula4: You are right, our TS file is not up to date with what's returned by the API. Cart payload contains the email, so probably it should be added here https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/cart/cart.ts

capsula4:
closed in favour of https://github.com/bigcommerce/checkout-sdk-js/pull/635

PR #615, opened 2019-06-25
CHECKOUT-4209: Throw `OrderFinalizationNotRequiredError` if payment method is no longer available for shopper

davidchin:
## What?
* Throw `OrderFinalizationNotRequiredError` if the selected payment method for an order is no longer available for the shopper.

## Why?
* It's a more suitable error to throw because order finalisation is actually not required if the payment method is no longer available for the shopper. That would allow shoppers to carry on as they can choose an alternative payment method to complete their order.
* This is required to fix an AfterPay bug, because the payment method has a special flow which could prevent the shopper from continuing unless they reset their checkout state.

## Testing / Proof
* Unit

### DIFF
--- a/src/payment/payment-strategy-action-creator.spec.ts
+++ b/src/payment/payment-strategy-action-creator.spec.ts
@@ -455,6 +455,26 @@ describe('PaymentStrategyActionCreator', () => {
                 expect(action.payload).toBeInstanceOf(OrderFinalizationNotRequiredError);
             }
         });
+
+        it('returns rejected promise if payment method referenced in order object no longer exists', async () => {
+            store = createCheckoutStore({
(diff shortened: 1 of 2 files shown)


src/payment/payment-strategy-action-creator.spec.ts
```diff
@@ -455,6 +455,26 @@ describe('PaymentStrategyActionCreator', () => {
                 expect(action.payload).toBeInstanceOf(OrderFinalizationNotRequiredError);
             }
         });
+
+        it('returns rejected promise if payment method is no longer available', async () => {
```
capsula4: `returns rejected promise if payment method referenced in order no longer exists` ?

PR #619, opened 2019-07-01
feat(payment): INT-1608 Modify Paypal Payments Pro's strategy for 3DS

josesonora:
## What? [INT-1608](https://jira.bigcommerce.com/browse/INT-1608)
Transform the existing Cybersource's stategy to be the Cardinal's one and also modify the existing Paypal's strategy to be extended from Cardinal's strategy

## Why?
Existing Cybersource's strategy contains all of the logic for Cardinal flow and it is totally transparent for other strategies who want to use it so we renamed it to Cardinal strategy and also extended Paypal Payments Pro's strategy from it

## Testing / Proof
[video | authorize](https://drive.google.com/open?id=14Wr6-X3cgo5V8Q9c8opUeyWY9ilQA3ew)
[video | purchase](https://drive.google.com/open?id=1LcC67x4A-PIeLTnap_V6FvHPKfETrg8W)

## Sibling PRs
[bigpay](https://github.com/bigcommerce/bigpay/pull/1711)
[bcapp](https://github.com/bigcommerce/bigcommerce/pull/30475)
[bigpay](https://github.com/bigcommerce/bigpay/pull/1724)
[ng-payments](https://github.com/bigcommerce-labs/ng-payments/pull/831)
[ng-checkout](https://github.com/bigcommerce-labs/ng-checkout/pull/1353)

### DIFF
--- a/src/payment/strategies/cardinal/cardinal-client.ts
+++ b/src/payment/strategies/cardinal/cardinal-client.ts
@@ -19,6 +19,7 @@ import {
     CardinalPaymentBrand,
     CardinalScriptLoader,
     CardinalSignatureValidationErrors,
+    CardinalSignatureVerification,
     CardinalSDK,
     CardinalTriggerEvents,
     CardinalValidatedAction,
@@ -38,29 +39,25 @@ export interface CardinalOrderData {
 
 export default class CardinalClient {
     private _sdk?: Promise<CardinalSDK>;
-    private _clientToken?: string;
 
     constructor(
         private _scriptLoader: CardinalScriptLoader
     ) {}
 
-    initialize(testMode?: boolean): Promise<void> {
+    initialize(provider: string, testMode?: boolean): Promise<void> {
         if (!this._sdk) {
-            this._sdk = this._scriptLoader.load(testMode);
+            this._sdk = this._scriptLoader.load(provider, testMode);
         }
 
         return this._sdk.then(() => {});
     }
 
     configure(clientToken: string): Promise<void> {
-        if (this._clientToken) { return Promise.resolve(); }
-
         return this._getClientSDK()
             .then(client => new Promise<void>((resolve, reject) => {
                 client.on(CardinalEventType.SetupCompleted, () => {
                     client.off(CardinalEventType.SetupCompleted);
                     client.off(CardinalEventType.Validated);
-                    this._clientToken = clientToken;
 
                     resolve();
                 });
@@ -74,6 +71,7 @@ export default class CardinalClient {
                             if (includes(CardinalSignatureValidationErrors, data.ErrorNumber)) {
                                 reject(new MissingDataError(MissingDataErrorType.MissingPaymentMethod));
                             }
+                            reject(new StandardError(data.ErrorDescription));
                             break;
                     }
                 });
@@ -81,7 +79,7 @@ export default class CardinalClient {
                 client.setup(CardinalInitializationType.Init, {
                     jwt: clientToken,
                 });
-        }));
+            }));
     }
 
     runBinProcess(binNumber: string): Promise<void> {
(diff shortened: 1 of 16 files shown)


src/payment/strategies/credit-card/credit-card-cardinal-payment-strategy.ts
```diff
@@ -17,15 +17,15 @@ import PaymentStrategy from '../payment-strategy';
 
 import { CardinalClient, CardinalOrderData } from './index';
 
-export default class CyberSourcePaymentStrategy implements PaymentStrategy {
-    private _paymentMethod?: PaymentMethod;
+export default class CreditCardCardinalPaymentStrategy implements PaymentStrategy {
```
davidchin: I think a better abstraction is to use composition rather than inheritance. Now you're making `PaypalProPaymentStrategy` inheriting a strategy that's supposed to be a strategy built specifically for CyberSource. We usually only use inheritance when a sub-class is a type of the parent class. In this case, it's not the right abstraction to use. You can tell by looking at the `execute` method of `PaypalProPaymentStrategy`, where half of the method is overridden because PayPal requires a different behaviour. 

I think what you really want is to share code between the two strategies - you want them to have the capability of triggering 3DS flow using the Cardinal client. You can do that by simply moving the shared code related to Cardinal into a separate class and inject instances of that class into the aforementioned strategies.

michxff:
thanks @davidchin it's ready for another look ♻️

josesonora:
@davidchin done 👍, please continue reviewing ♻️

src/payment/strategies/cardinal/cardinal-three-d-secure-flow.ts
```diff
@@ -0,0 +1,122 @@
...
+        return this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId))
+            .then(state => {
+                const method = state.paymentMethods.getPaymentMethod(methodId);
+
+                if (!method || !method.config) {
+                    throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+                }
+
+                return this._cardinalClient.initialize(method.config.testMode)
+                    .then(() => {
```
davidchin: Can you reduce the extra level of nesting? i.e.:

```ts
return this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId))
    .then(state => {
        const method = state.paymentMethods.getPaymentMethod(methodId);

        if (!method || !method.config) {
        }

        return this._cardinalClient.initialize(method.config.testMode);
    })
    .then(() => {
        if (!method.clientToken) {
        }

        this._clientToken = method.clientToken;

        return this._cardinalClient.configure(this._clientToken);
    });
```

src/payment/strategies/cardinal/cardinal-three-d-secure-flow.ts
```diff
@@ -0,0 +1,122 @@
...
+                    });
+            });
+    }
+
+    start(payment: Payment): Promise<InternalCheckoutSelectors> {
+        if (!payment.paymentData) {
+            throw new MissingDataError(MissingDataErrorType.MissingPayment);
+        }
+
+        const paymentData = payment.paymentData as CreditCardInstrument;
```
davidchin: Use `isCreditCardLike` to do type checking instead of coercing the type. i.e.:

```ts
if (!isCreditCardLike(payment.paymentData)) {
}
```

src/payment/strategies/cardinal/cardinal-three-d-secure-flow.ts
```diff
@@ -0,0 +1,122 @@
...
+                            ...paymentData,
+                            threeDSecure,
+                        },
+                    }))
+                );
+            });
+    }
+
+    private _getOrderData(paymentData: CreditCardInstrument): CardinalOrderData {
+        const billingAddress = this._store.getState().billingAddress.getBillingAddress();
```
davidchin: You can call `getState()` once and store it in a variable. i.e.:

```ts
```

src/payment/strategies/cardinal/cardinal-three-d-secure-flow.ts
```diff
@@ -0,0 +1,122 @@
...
+            .then(() => {
+                if (!this._clientToken) {
+                    throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+                }
+
+                payment = {
+                    ...payment,
+                    paymentData: {
+                        ...paymentData,
+                        threeDSecure: { token: this._clientToken },
```
davidchin: How come we're posting the client token back to BigPay? Doesn't the client token originally come from BigPay?

src/payment/strategies/paypal/paypal-pro-payment-strategy.spec.ts
```diff
@@ -1,127 +1,220 @@
...
 import { PaymentActionType, SubmitPaymentAction } from '../../payment-actions';
+import PaymentMethod from '../../payment-method';
+import PaymentMethodActionCreator from '../../payment-method-action-creator';
+import PaymentMethodRequestSender from '../../payment-method-request-sender';
+import { getPaypal } from '../../payment-methods.mock';
 import PaymentRequestSender from '../../payment-request-sender';
+import { CardinalClient, CardinalScriptLoader, CardinalThreeDSecureFlow } from '../cardinal';
 
-import PaypalProPaymentStrategy from './paypal-pro-payment-strategy';
+import { PaypalProPaymentStrategy } from './index';
```
davidchin: `import PaypalProPaymentStrategy from './paypal-pro-payment-strategy';`

src/payment/strategies/cardinal/cardinal-three-d-secure-flow.ts
josesonora: Yes, you're right but we need to send the client token back to `bigpay` due to it contains a unique `ReferenceID` that is not stored in `bigpay` and is formed dynamically each time the client token is generated

josesonora:
@davidchin thank you for your comments, they had been addressed, please continue reviewing ♻️

src/payment/strategies/cardinal/cardinal-client.ts
```diff
@@ -96,22 +97,28 @@ export default class CardinalClient {
                 return new Promise<ThreeDSecureToken>((resolve, reject) => {
                     client.on(CardinalEventType.Validated, (data: CardinalValidatedData, jwt: string) => {
                         client.off(CardinalEventType.Validated);
-                        switch (data.ActionCode) {
-                            case CardinalValidatedAction.Success:
-                                resolve({ token: jwt });
-                                break;
-                            case CardinalValidatedAction.NoAction:
-                                if (data.ErrorNumber > 0) {
+                        if (data.ActionCode) {
```
davidchin: You can reduce the level of nesting by using a guard statement. i.e.:

```ts
if (!data.ActionCode) {
    return resolve({ token: jwt });
}

switch (data.ActionCode) {
    // cases go here
}
```

src/payment/strategies/cardinal/cardinal-script-loader.ts
```diff
@@ -13,9 +13,11 @@ export default class CardinalScriptLoader {
         private _window: CardinalWindow = window
     ) {}
 
-    load(testMode?: boolean): Promise<CardinalSDK> {
+    load(provider: string, testMode?: boolean): Promise<CardinalSDK> {
+        const url = testMode ? SDK_TEST_URL : SDK_PROD_URL;
+
         return this._scriptLoader
-            .loadScript(testMode ? SDK_TEST_URL : SDK_PROD_URL)
+            .loadScript(url + '?v=' + provider)
```
josesonora: Hi @davidchin, I made this small change after your approval due to an issue that QA found testing with `Cybersource` and `Paypal` with `3DS` enabled at the same time, it was caused because the `loadScript` method returns the same instance of the `Cardinal SDK` and we used it to call same methods but for different providers so now we're adding the provider name as a versioner in order to get a different instance for each provider. What do you think about this? Is it ok for you?

PascalZajac:
@josesonora can you please rebase this so we can release it.

PR #620, opened 2019-07-01
INT-1577 Support Stored Credit Card (vaulting) for Stripe V3

clopezh:
[INT-1577](https://jira.bigcommerce.com/browse/INT-1577)
## What?
Add support for Stripe V3 + Vault that can handle stripe validations with stripe.js (createPaymentMethod & handleCardPayment actions)

## Why?
To support Vaulting on Stripe V3 is needed to modify the existing strategy that use Stripe.js elements and actions to handle 3DS flow and confirm a payment intent that previously was created from back-end (which used a new end point created on BigPay). Also a change into ng-checkout was required to support instruments into `HostedWidgetPaymentMethod`

## Testing / Proof
image

image

image

ezgif com-video-to-gif

## Sibling PRs
- [BigPay-Client-JS](https://github.com/bigcommerce/bigpay-client-js/pull/79)
- [NGCheckout](https://github.com/bigcommerce-labs/ng-checkout/pull/1282)
- [BigPay](https://github.com/bigcommerce/bigpay/pull/1730)

@bigcommerce/checkout @bigcommerce/intersys-integrations

### DIFF
--- a/src/order/order-request-body.ts
+++ b/src/order/order-request-body.ts
@@ -1,4 +1,4 @@
-import { CreditCardInstrument, VaultedInstrument } from '../payment';
+import { CreditCardInstrument, HostedInstrument, VaultedInstrument } from '../payment';
 
 /**
  * An object that contains the information required for submitting an order.
@@ -39,5 +39,5 @@ export interface OrderPaymentRequestBody {
      * An object that contains the details of a credit card or vaulted payment
      * instrument.
      */
-    paymentData?: CreditCardInstrument | VaultedInstrument;
+    paymentData?: CreditCardInstrument | VaultedInstrument | HostedInstrument;
 }
--- a/src/payment/payment-action-creator.ts
+++ b/src/payment/payment-action-creator.ts
@@ -1,36 +1,28 @@
 import { createAction, ThunkAction } from '@bigcommerce/data-store';
-import { pick } from 'lodash';
 import { concat, from, of } from 'rxjs';
 import { catchError, switchMap } from 'rxjs/operators';
 
-import { mapToInternalAddress } from '../address';
-import { mapToInternalCart } from '../cart';
 import { InternalCheckoutSelectors } from '../checkout';
 import { throwErrorAction } from '../common/error';
-import { StandardError } from '../common/error/errors';
-import { mapToInternalCustomer } from '../customer';
-import { mapToInternalOrder, OrderActionCreator } from '../order';
-import { mapToInternalShippingOption } from '../shipping';
+import { OrderActionCreator } from '../order';
 
-import isVaultedInstrument from './is-vaulted-instrument';
 import Payment from './payment';
 import { InitializeOffsitePaymentAction, PaymentActionType, SubmitPaymentAction } from './payment-actions';
-import PaymentMethod from './payment-method';
-import PaymentMethodSelector from './payment-method-selector';
-import PaymentRequestBody from './payment-request-body';
 import PaymentRequestSender from './payment-request-sender';
+import PaymentRequestTransformer from './payment-request-transformer';
 
 export default class PaymentActionCreator {
     constructor(
         private _paymentRequestSender: PaymentRequestSender,
-        private _orderActionCreator: OrderActionCreator
+        private _orderActionCreator: OrderActionCreator,
+        private _paymentRequestTransformer: PaymentRequestTransformer
     ) {}
 
     submitPayment(payment: Payment): ThunkAction<SubmitPaymentAction, InternalCheckoutSelectors> {
         return store => concat(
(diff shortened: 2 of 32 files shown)


src/payment/strategies/stripev3/stripev3-payment-strategy.spec.ts
```diff
@@ -24,15 +26,20 @@ import {
     PaymentMethodRequestSender,
     PaymentRequestSender
 } from '../../../payment';
-import { PaymentArgumentInvalidError } from '../../errors';
+import { getShippingAddress } from '../../../shipping/shipping-addresses.mock';
+import {PaymentArgumentInvalidError} from '../../errors';
```
metalvegetarianoprogresivo: Add spaces between the curly bracket.
```suggestion
import { PaymentArgumentInvalidError } from '../../errors';
```

src/payment/payment-action-creator.ts
```diff
@@ -60,6 +64,23 @@ export default class PaymentActionCreator {
         };
     }
 
+    generatePaymentIntent(
```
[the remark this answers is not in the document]
clopezh: Reason behind why we added this new endpoint is because existing one handles ClientToken only, on this one we handle PaymentIntents which maps a different set of data (CustomerId, BillingAddress, CurrencyCode, GrandTotal, etc.), on the other hand, we designed this endpoint to communicate directly SDK and BigPay because we there’s no need to store the token in memory.

src/order/order-request-body.ts
```diff
@@ -18,6 +18,12 @@ export default interface OrderRequestBody {
      * works if the customer has previously signed in.
      */
     useStoreCredit?: boolean;
+
+    /**
+     * If true, apply to save the credit card as an instrument. It only works if a hosted field needs
+     * to be vaulted.
+     */
+    shouldSaveInstrument?: boolean;
```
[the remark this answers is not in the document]
clopezh: `paymentData` type is of 1 of 2 subtypes `CreditCardInstument` or `VaultedInstrument`, our case is neither one. If we decide to handle ours as a `CreditCardInstrument` we still have to pass the rest of cc info which we don’t have access as this is handled by Stripe library. That is why we had to create this new property as part of `OrderRequestBody`.
We believe it is a better place to have this property in here just like `useStoreCredit` flag.

clopezh:
Hey @davidchin I addressed some of your comments and also I answer some of your questions, Could you please take a look? ♻️ Thank you!!

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -70,39 +72,58 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
...
+                            throw new StandardError(stripePaymentMethod.error && stripePaymentMethod.error.message);
+                        }
+
+                        const stripeCardPaymentOptions = {
+                            ...this._mapStripeCardPaymentOptions(shouldSaveInstrument),
+                            payment_method: stripePaymentMethod.paymentMethod.id,
+                        };
+
+                        return this._getStripeJs().handleCardPayment(paymentIntent, stripeCardPaymentOptions)
+                            .then(stripeResponse => {
```
davidchin: This level of nesting can be removed by chaining with the outer promise instead. i.e.:

```ts
return this._getStripeJs().createPaymentMethod('card', this._cardElement, this._mapStripePaymentMethodOptions())
    .then(stripePaymentMethod => {
        // Do what you have to do with `stripePaymentMethod`
        return this._getStripeJs().handleCardPayment(paymentIntent, stripeCardPaymentOptions);
    })
    .then(stripeResponse => {
        // Do what you have to do with `stripePaymentMethod`
    });
```

clopezh:
Hey @davidchin I already addressed the last comments, and also I added extra unit testing to increase the percentage of coverage, Could you please take a look and if possible approve it? Thank you! ♻️

src/payment/payment-request-transformer.ts
```diff
@@ -0,0 +1,90 @@
...
+import { mapToInternalShippingOption } from '../shipping';
+
+import isVaultedInstrument from './is-vaulted-instrument';
+import Payment from './payment';
+import PaymentMethod from './payment-method';
+import PaymentMethodSelector from './payment-method-selector';
+import PaymentRequestBody from './payment-request-body';
+
+export default class PaymentRequestTransformer {
+    static transform(payment: Payment, checkoutState: InternalCheckoutSelectors): PaymentRequestBody {
```
davidchin: We prefer not to use `static` methods as it's not possible to mock them. Can you make this an instance method and inject the instance into places where you need it?

clopezh:
@davidchin Comment addressed!! 👍 Could you please take a look and approve it? ♻️

PR #623, opened 2019-07-05
feat(payment): INT-1650 Adding IIN field to vaulted instrument strategy

ric-sanchez:
## What?
Adding BIN Number for vaulted credit cards in the form of the issuer_identification_number (iin), already stored in the BigPay database.

## Why?
This iin value should be received by Checkout SDK strategy so it can be used in the Cardinal's BIN Detection step.

## Testing / Proof
image

### DIFF
--- a/src/payment/payment.ts
+++ b/src/payment/payment.ts
@@ -4,7 +4,7 @@ export default interface Payment {
     paymentData?: PaymentInstrument & PaymentInstrumentMeta;
 }
 
-export type PaymentInstrument = CreditCardInstrument | NonceInstrument | VaultedInstrument | CryptogramInstrument | HostedInstrument;
+export type PaymentInstrument = CreditCardInstrument | NonceInstrument | VaultedInstrument | CryptogramInstrument | HostedInstrument | ThreeDSVaultedInstrument;
 
 export interface PaymentInstrumentMeta {
     deviceSessionId?: string;
@@ -33,6 +33,11 @@ export interface VaultedInstrument {
     instrumentId: string;
     ccCvv?: string;
     ccNumber?: string;
+}
+
+export interface ThreeDSVaultedInstrument extends VaultedInstrument {
+    iin?: string;
+    threeDSecure?: ThreeDSecure | ThreeDSecureToken;
 }
 
 export interface CryptogramInstrument {
--- a/src/payment/payments.mock.ts
+++ b/src/payment/payments.mock.ts
@@ -10,7 +10,7 @@ import { getConsignments } from '../shipping/consignments.mock';
 import { getFlatRateOption } from '../shipping/internal-shipping-options.mock';
 import { getShippingAddress } from '../shipping/shipping-addresses.mock';
 
-import Payment, { CreditCardInstrument } from './payment';
+import Payment, { CreditCardInstrument, ThreeDSVaultedInstrument, VaultedInstrument } from './payment';
 import { getAuthorizenet, getPaymentMethodsMeta } from './payment-methods.mock';
 import PaymentRequestBody from './payment-request-body';
 import PaymentResponseBody from './payment-response-body';
@@ -32,6 +32,12 @@ export function getCreditCardInstrument(): CreditCardInstrument {
         ccName: 'BigCommerce',
         ccNumber: '4111111111111111',
         ccCvv: '123',
+    };
+}
+
+export function getVaultedInstrument(): VaultedInstrument {
+    return {
+        instrumentId: '123',
     };
 }
 
--- a/src/payment/strategies/cybersource/cardinal.ts
+++ b/src/payment/strategies/cybersource/cardinal.ts
@@ -1,7 +1,3 @@
-import Address from '../../../address/address';
-import BillingAddress from '../../../billing/billing-address';
-import { CreditCardInstrument } from '../../payment';
-
 export const CardinalSignatureValidationErrors = [100004, 1010, 1011, 1020];
 
 export interface CardinalSDK {
@@ -42,15 +38,6 @@ export interface CardinalSetupCompletedData {
 export interface CardinalModuleState {
     loaded: boolean;
     module: string;
-}
-
-export interface CardinalOrderData {
-    billingAddress: BillingAddress;
-    shippingAddress?: Address;
-    currencyCode: string;
-    id: string;
-    amount: number;
-    paymentData: CreditCardInstrument;
 }
 
 export enum CardinalInitializationType {
@@ -109,8 +96,8 @@ export interface CardinalConsumer {
     Email1?: string;
     Email2?: string;
     ShippingAddress?: CardinalAddress;
-    BillingAddress: CardinalAddress;
-    Account: CardinalAccount;
+    BillingAddress?: CardinalAddress;
+    Account?: CardinalAccount;
 }
 
 export interface CardinalAccount {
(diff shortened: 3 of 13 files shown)


icatalina:
Hi @ric-sanchez! is this value always available in the response?

ric-sanchez:
> Hi @ric-sanchez! is this value always available in the response?

Yes, @icatalina, this value is always available for vaulted instruments

src/payment/payments.mock.ts
```diff
@@ -35,6 +35,15 @@ export function getCreditCardInstrument(): CreditCardInstrument {
     };
 }
 
+export function getVaultedInstrument(): VaultedInstrument {
+    return {
+        ccCvv: undefined,
```
icatalina: We can just omit `ccCvv`

src/payment/payments.mock.ts
```diff
@@ -35,6 +35,15 @@ export function getCreditCardInstrument(): CreditCardInstrument {
...
+        ccNumber: '',
```
icatalina: Why the empty string? Since the value is typed as optional, it should either be `undefine` or the actual value. Unless that's what you are seeing in the API in which case we might need to adjust the types.

src/payment/strategies/cybersource/cardinal.ts
```diff
@@ -110,7 +112,7 @@ export interface CardinalConsumer {
     Email2?: string;
     ShippingAddress?: CardinalAddress;
     BillingAddress: CardinalAddress;
-    Account: CardinalAccount;
+    Account?: CardinalAccount;
```
icatalina: It seems like all values are optional, any reason why we keep `billingAddress` as mandatory?

image

src/payment/strategies/cybersource/cybersource-payment-strategy.spec.ts
```diff
@@ -368,6 +368,24 @@ describe('CyberSourcePaymentStrategy', () => {
...
+                    payment: {
+                        methodId: paymentMethodMock.id,
+                        gatewayId: paymentMethodMock.gateway,
+                        paymentData: getVaultedInstrument(),
+                    },
+                });
+
+                await strategy.execute(payload);
+
+                expect(cardinalClient.getThreeDSecureData).toHaveBeenCalled();
```
icatalina: How is this proof of the `iin` being used?

src/payment/strategies/cybersource/cybersource-payment-strategy.ts
```diff
@@ -79,7 +86,7 @@ export default class CyberSourcePaymentStrategy implements PaymentStrategy {
         const paymentData = payment.paymentData as CreditCardInstrument;
 
         return this._cardinalClient.configure(clientToken)
-            .then(() => this._cardinalClient.runBindProcess(paymentData.ccNumber))
+            .then(() => this._cardinalClient.runBindProcess(this._getBinNumber(paymentData)))
```
icatalina: If this can take either a `ccNumber` or a `iin`, could you update the variable name inside `runBindProcess` to read something like `binNumber` instead of `ccNumber`?

In fact, I'm not sure if this method should be called `runBindProcess` or `runBinProcess`... 😕

src/payment/strategies/cybersource/cybersource-payment-strategy.ts
```diff
@@ -112,7 +119,11 @@ export default class CyberSourcePaymentStrategy implements PaymentStrategy {
             );
     }
 
-    private _getOrderData(paymentData: CreditCardInstrument): CardinalOrderData {
+    private _getBinNumber(payment: CardinalSupportedPaymentInstrument): string {
+        return isVaultedInstrument(payment) ? payment.iin : payment.ccNumber;
+    }
+
+    private _getOrderData(paymentData: CardinalSupportedPaymentInstrument): CardinalOrderData {
```
icatalina: I know u haven't changed this as part of this PR, but would you mind changing it to a single call to `getState`?

```ts

const shippingAddress = state.shippingAddress.getShippingAddress();
const checkout = state.checkout.getCheckout();
const order = state.order.getOrder();
```

src/payment/strategies/cybersource/cardinal.ts
```diff
@@ -1,9 +1,11 @@
 import Address from '../../../address/address';
 import BillingAddress from '../../../billing/billing-address';
-import { CreditCardInstrument } from '../../payment';
+import { CreditCardInstrument, VaultedInstrument } from '../../payment';
 
 export const CardinalSignatureValidationErrors = [100004, 1010, 1011, 1020];
 
+export type CardinalSupportedPaymentInstrument = CreditCardInstrument | VaultedInstrument;
```
icatalina: This file should only contain types that are specific to the cardinal sdk. `CardinalSupportedPaymentInstrument` is a responsibility of the `cardinal-client`, so the type should probably live there.

src/payment/strategies/cybersource/cardinal.ts
```diff
@@ -50,7 +52,7 @@ export interface CardinalOrderData {
     currencyCode: string;
     id: string;
     amount: number;
-    paymentData: CreditCardInstrument;
+    paymentData?: CreditCardInstrument;
```
icatalina: Same than the type above, this is a responsibility of the `cardinal-client` not the `cardinal-sdk`. Could we move this whole type (`CardinalOrderData`) to the `cardinal-client` file?

src/payment/strategies/cybersource/cybersource-payment-strategy.ts
ric-sanchez: Yeah, you're right. `runBinProcess` is the correct one

src/payment/strategies/cybersource/cybersource-payment-strategy.spec.ts
ric-sanchez: The description of the test may not be accurate, the idea is the use of VaultedInstrument as PaymentData. I'll update the test name to reflect that...

src/payment/payment.ts
```diff
@@ -33,6 +33,8 @@ export interface VaultedInstrument {
     instrumentId: string;
     ccCvv?: string;
     ccNumber?: string;
+    iin?: string;
```
icatalina: I had a quick discussion with @davidchin. This type is public, by adding these properties here, we hint the user they can pass them in, which can cause confusion as we are really not using the copies within the payload.

A better approach will be to create a new interface for the request bigpay, and keep this one free of those properties.

```ts
export type PaymentInstrument = CreditCardInstrument | NonceInstrument | VaultedInstrument | CryptogramInstrument | ThreeDSVaultedInstrument;

// ...

export interface ThreeDSVaultedInstrument extends VaultedInstrument {
    iin?: string;
    threeDSecure?: ThreeDSecure | ThreeDSecureToken;
}
```

src/payment/strategies/cybersource/cybersource-payment-strategy.spec.ts
ric-sanchez: Method `getThreeDSecureData` is called only when the strategy successfully place the order, and in the scenario of a Vaulted Instrument being passed as paymentData, the assertion of `getThreeDSecureData` is the proper one

TomA-R:
Hey @ric-sanchez, is this ready to be re-reviewed? It's usually a good idea to ping (@githubname) the reviewers so they're notified, because it's often hard to track all the open PRs that require review.

ric-sanchez:
> Hey @ric-sanchez, is this ready to be re-reviewed? It's usually a good idea to ping (@ githubname) the reviewers so they're notified, because it's often hard to track all the open PRs that require review (or know when they're ready for re-review)

Sure, @TomA-R. Tagging @icatalina for him to review again 👍

src/payment/payment.ts
icatalina: Thanks heaps for this.

icatalina:
forgot to ping you @ric-sanchez

ric-sanchez:
@icatalina @PascalZajac , Ready to merge!

Issue #618, opened 2019-07-01
Braintree Credit Card and Braintree PayPal Credit Card payment errors


cherupil:
> @capsula4 thanks for your quick reply - I viewed the api requests from the HAR file - it is a little confusing to me since it seems like the posting of the order is fine, as well as the get request for the completed order data:

![2018-10-12_1028](https://user-images.githubusercontent.com/5169882/46878778-bb778580-ce09-11e8-9d6b-1c330e77ed9a.png)

 

> I have double checked and the order is not posting to the store, when i refresh the page the order is still in the cart. One thing I did notice was the post data from showing a blank customer message:

![00000270](https://user-images.githubusercontent.com/5169882/46878674-7e12f800-ce09-11e8-91ce-c5b4d9be5726.png)

> I am not exactly sure why this is showing as I have not added any place for a customer message, and I couldnt find it being required in the docs anywhere

_Originally posted by @PeteyRev in https://github.com/bigcommerce/checkout-sdk-js/issues/430#issuecomment-429365869_

I'm experiencing this same issue, I've done as @PeteyRev has suggested and changed `service` from a Vuex store property to data on the main app component, the results are no different. Maybe @PeteyRev can provide more insight as to the solution for this or @davidchin given that this has the 'investigate' tag on it now. I'm able to successfully use the test payment gateway with the same code, but braintree credit cards and braintree paypal result in this exact error. The orders are created as 'Pending'.

<img width="799" alt="Screen Shot 2019-06-30 at 11 44 46 PM" src="https://user-images.githubusercontent.com/8054842/60409355-4b747f80-9b91-11e9-89b0-b6b6f0f04a63.png">
<img width="1322" alt="Screen Shot 2019-06-30 at 11 45 04 PM" src="https://user-images.githubusercontent.com/8054842/60409356-4b747f80-9b91-11e9-89d6-94c8c05898a7.png">
<img width="680" alt="Screen Shot 2019-06-30 at 11 45 52 PM" src="https://user-images.githubusercontent.com/8054842/60409357-4b747f80-9b91-11e9-8970-0904c879d1a4.png">

cherupil:
@icatalina I've installed the update after your pull request, and both using the cli as well as on production we are encountering the same error. Any thoughts here?

[...]

Issue #430, opened 2018-10-11
Issue with credit card payments


PeteyRev:
Hi, I have had success with testing the payment gateway, but when trying anything else (Braintree,Authorize) I keep getting this same error https://screencast.com/t/aLrMObkKGtY. I am pretty positive im doing this according to the docs, can you give any insight on this error? https://screencast.com/t/YrYxkMkMs

capsula4:
Code looks okay to me, but it's hard to see what could be failing. Maybe try to debug, find where it's blowing up, and see what's causing the exception?

We use the SDK in production and those payment methods work. Do you have a HAR files includes network requests? API responses might help us track down the issue.

PeteyRev:
@capsula4 thanks for your quick reply - I viewed the api requests from the HAR file - it is a little confusing to me since it seems like the posting of the order is fine, as well as the get request for the completed order data:
![2018-10-12_1028](https://user-images.githubusercontent.com/5169882/46878778-bb778580-ce09-11e8-9d6b-1c330e77ed9a.png)

 I have double checked and the order is not posting to the store, when i refresh the page the order is still in the cart. One thing I did notice was the post data from showing a blank customer message:
![00000270](https://user-images.githubusercontent.com/5169882/46878674-7e12f800-ce09-11e8-91ce-c5b4d9be5726.png)

I am not exactly sure why this is showing as I have not added any place for a customer message, and I couldnt find it being required in the docs anywhere

PeteyRev:
@capsula4 dont want to be that guy, but I am still struggling with this. I even used this react sdk example you provided into my own project and am still getting a error - I realize this isnt a production ready version but it still should still work right?

capsula4:
@PeteyRev I'm really sorry I wasn't able to get back to you, this slipped through my things to do. The Checkout SDK React Example is just a quick example of how to use the SDK and might have some issues with specific payment providers.

Back to your original question, the order seems to be created successfully. Most orders are creating as pending as the payment still needs to be manually processed through the control panel. The funds are usually just captured.

The error you are seeing is hard to debug without having the actual code. I would try to use Chrome debugger and trace the error and see what's the part that is actually blowing up. Seems like an array is expected but null is returned.

PeteyRev:
@capsula4 The example sdk doesn't work at all - not relative to payment methods. Does anyone have an actual live example or a working checkout? I have yet to find any one using this.

agrohs:
[...]

PR #632, opened 2019-07-12
CHECKOUT-4254 Make sure changes to the public object don't affect the internal copies

icatalina:
## What?
- As per title

## Why?
- The internal state of the `checkout-sdk` should be shield from changes to the exposed objects.

## Testing / Proof
- Functional / Unit

Closes: #430 
Closes: #618 

### DIFF
--- a/src/common/utility/clone-decorator.ts
+++ b/src/common/utility/clone-decorator.ts
@@ -1,0 +1,68 @@
+import { cloneDeep, memoize } from 'lodash';
+
+export default function cloneDecorator<T extends Method>(target: object, key: string, descriptor: TypedPropertyDescriptor<T>): TypedPropertyDescriptor<T>;
+export default function cloneDecorator<T extends Constructor<object>>(target: T): T;
+export default function cloneDecorator(target: any, key?: any, descriptor?: any): any {
+    if (!key || !descriptor) {
+        return cloneClassDecorator(target);
+    }
+
+    return cloneMethodDecorator(target, key, descriptor);
+}
+
+export function cloneClassDecorator<T extends Constructor<object>>(target: T): T {
+    const decoratedTarget = class extends target {};
+
+    Object.getOwnPropertyNames(target.prototype)
+        .forEach(key => {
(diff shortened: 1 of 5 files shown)


src/common/utility/clone-decorator.ts
```diff
@@ -0,0 +1,61 @@
...
+
+            if (!descriptor || key === 'constructor') {
+                return;
+            }
+
+            Object.defineProperty(
+                decoratedTarget.prototype,
+                key,
+                cloneMethodDecorator(target.prototype, key, descriptor)
+                );
```
davidchin: Indentation 💅

src/common/utility/clone-decorator.ts
```diff
@@ -0,0 +1,61 @@
...
+    return decoratedTarget;
+}
+
+export function cloneMethodDecorator<T extends Method>(target: object, key: string, descriptor: TypedPropertyDescriptor<T>): TypedPropertyDescriptor<T> {
+    if (typeof descriptor.value !== 'function') {
+        return descriptor;
+    }
+
+    let method = descriptor.value;
+    const memoizedCloneDeep = memoize(cloneDeep);
```
davidchin: We'll need to pass in a cache key resolver here because, by default, Lodash's memoize only uses the first parameter as its cache key. For example, if a memoized function accepts two parameters, and you call it twice (the first argument is the same but the second argument is different), you will get the cached result for the second call. That's not what you want as the computed result might be different.

We have a `CacheKeyResolver` already so maybe you can just use that?

icatalina: But `cloneDeep` only takes one param.

image

davidchin: Ah yep, sorry I misread. I thought you're memoizing the decorated method. Ok, yes you're right, what you have should be fine and you don't need to pass in a cache key resolver.

PS: What you might want to consider is to replace the default `memoize.Cache` with a `WeakMap`? That would help with the garbage collection of objects. Have to check with the browser support though (it seems we're okay?).

PR #646, opened 2019-07-29
CHECKOUT-4263 Support for applying/removing store credit

capsula4:
## What?
Add support for applying/removing Store Credit via API endpoint

## Why?
So store credit can be applied at any time, not just when submitting an order

## Testing / Proof
unit

@bigcommerce/checkout


src/store-credit/store-credit-action-creator.ts
```diff
@@ -0,0 +1,39 @@
...
+import { ApplyStoreCreditAction, StoreCreditAction, StoreCreditActionType } from './store-credit-actions';
+import StoreCreditRequestSender from './store-credit-request-sender';
+
+export default class StoreCreditActionCreator {
+    constructor(
+        private _storeCreditRequestSender: StoreCreditRequestSender
+    ) {}
+
+    applyStoreCredit(useStoreCredit: boolean, options?: RequestOptions): ThunkAction<StoreCreditAction, InternalCheckoutSelectors> {
+        return store => Observable.create(async (observer: Observer<ApplyStoreCreditAction>) => {
```
davidchin: This comment is completely optional as it's simply a different way of writing the same thing. But lately, we've been using observable operators + factories instead of creating observables directly using `new` or `Observable.create`. For example, your code would look something like this instead: 

```ts
return store => (
    concat(
        of(createAction(StoreCreditActionType.ApplyStoreCreditRequested)),
        defer(async () => {
            const { checkout: { getCheckout } } = store.getState();

            if (!checkout) {
            }

            const { body } = await useStoreCredit ?
                this._storeCreditRequestSender.applyStoreCredit(checkout.id, options)
                this._storeCreditRequestSender.removeStoreCredit(checkout.id, options);
            
            return createAction(StoreCreditActionType.ApplyStoreCreditSucceeded, body);
        })
    ).pipe(
        catchError(error => throwErrorAction(StoreCreditActionType.ApplyStoreCreditFailed, error))
    );
);
```

In your case, it doesn't really make a difference. But in some other cases where we have to chain or combine two or more actions, operators make it easier to achieve (as you can just insert another observable in `concat`). 🍺

src/remote-checkout/remote-checkout-reducer.ts
```diff
@@ -26,7 +26,7 @@ export default function remoteCheckoutReducer(
 }
 
 function dataReducer(
-    data: RemoteCheckout = DEFAULT_STATE.data,
+    data: RemoteCheckout,
```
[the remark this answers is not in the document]
capsula4: it was giving a linting error and it's also not required, the default state is passed when the main combined reducer is created. No other reducer does this.

src/payment/strategies/afterpay/afterpay-payment-strategy.ts
```diff
@@ -50,61 +48,57 @@ export default class AfterpayPaymentStrategy implements PaymentStrategy {
...
+    async execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
         const paymentId = payload.payment && payload.payment.gatewayId;
 
         if (!paymentId) {
             throw new PaymentArgumentInvalidError(['payment.gatewayId']);
         }
 
-        const useStoreCredit = !!payload.useStoreCredit;
-        const state = this._store.getState();
+        let state = this._store.getState();
```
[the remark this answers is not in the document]
capsula4: I'm accessing the state before loading the payment method, so I don't think I can use `const` ?

src/store-credit/store-credit-reducer.ts
```diff
@@ -0,0 +1,57 @@
...
+    errors: StoreCreditErrorsState = {},
+    action: StoreCreditAction
+): StoreCreditErrorsState {
+    switch (action.type) {
+    case StoreCreditActionType.ApplyStoreCreditRequested:
+    case StoreCreditActionType.ApplyStoreCreditSucceeded:
+        return { ...errors, applyError: undefined };
+
+    case StoreCreditActionType.ApplyStoreCreditFailed:
+        return { ...errors, applyError: action.payload };
```
davidchin: Can you use the new reducer helper? i.e.: `objectSet(errors, 'applyError', action.payload)`

src/store-credit/store-credit-reducer.ts
```diff
@@ -0,0 +1,57 @@
...
+    const reducer = combineReducers<StoreCreditState>({
+        errors: composeReducers(errorsReducer, clearErrorReducer),
+        statuses: statusesReducer,
+    });
+
+    return reducer(state, action);
+}
+
+function errorsReducer(
+    errors: StoreCreditErrorsState = {},
```
davidchin: Can you provide the default value defined in `DEFAULT_STATE`? i.e.: `DEFAULT_STATE.errors`.

src/checkout/checkout-selector.ts
```diff
@@ -12,6 +12,7 @@ import CheckoutState, { DEFAULT_STATE } from './checkout-state';
 export default interface CheckoutSelector {
     getCheckout(): Checkout | undefined;
     getGrandTotal(useStoreCredit?: boolean): number | undefined;
```
davidchin: Is it possible to remove this getter and replace it with `getOutstandingBalance`? Also, to make it backwards compatible, modify `getOutstandingBalance` so it accepts `useStoreCredit`. If `useStoreCredit` flag is not passed, return `checkout.outstandingBalance`. Otherwise, return what `getGrandTotal` is currently returning.

src/checkout/checkout-selector.ts
```diff
@@ -83,20 +82,7 @@ export function createCheckoutSelectorFactory(): CheckoutSelectorFactory {
...
-        getCheckout,
-        getCheckout => () => {
-            const checkout = getCheckout();
-
-            if (!checkout) {
-                return;
-            }
-
-            return checkout.outstandingBalance;
+            return useStoreCredit ? Math.max(grandTotal - storeCredit, 0) : checkout.outstandingBalance;
```
davidchin: Can you make it so that if `useStoreCredit` is not passed (`undefined`), return `checkout.outstandingBalance`?

PR #650, opened 2019-07-30
CHECKOUT-4272: Add dependencies needed for refactoring selectors

davidchin:
## What?
Add `reselect` dependency.

## Why?
This library can help us write more efficient selectors. Currently, we have selectors as objects; and we create them whenever there's a change in the state. But not all of the methods depend on the piece of state that's changed. i.e.:

```ts
@bind
class Selector {
    constructor(private state: { a: string, b: string }) {}

    getA() {
        return this.state.a;
    }

    getB(message: string) {
        return { value: this.state.b, message };
    }
}
```

In the example above, 'getA' depends on `state.a` and `getB` depends on `state.b`. When `state` changes, there is no way of knowing whether or not the output of `getA` or `getB` has changed, besides actually calling the methods and comparing the output with the previous result. 

Ideally, what we want is to produce a new `getA` and `getB` if `state.a` or `state.b` changes respectively. This allows us to pass the functions as values. This is useful in React as you can pass the getters as props into a component, and the component can detect whether there's a need to re-render by doing a strict equal comparison.

If we write the getters as separate functions, we can achieve what we want. `reselect` is a tool that could help us achieve what we want faster i.e.:

```ts
const getA = createSelector(
    state => state.a,
    a => () => a
);

const getB = createSelector(
    state => state.b,
    b => memoize((message: string) => { value: b, message })
);
```

Now we can explicitly specify the state dependency of `getA` and `getB`. And we only get new getter functions when the relevant state changes. Also, in the case of `getB`, I can ensure that we only produce new objects when depended inputs (`state.b` and `message`) are different. This means that not only I can pass getters into components, I can also pass in the outputs of these getters and React can easily figure out whether or not to re-render.

## Testing / Proof
Unit

### DIFF
--- a/src/common/selector/create-shallow-equal-selector.spec.ts
+++ b/src/common/selector/create-shallow-equal-selector.spec.ts
@@ -1,0 +1,60 @@
+import createShallowEqualSelector from './create-shallow-equal-selector';
+
+describe('createShallowEqualSelector()', () => {
+    interface TestState {
+        items: Array<{ id: number }>;
+        messages: string[];
+    }
+
+    it('creates selector that does shallow comparison instead of strict comparison', () => {
+        const combiner = jest.fn((messages: string[]) => messages.join(', '));
+        const selector = createShallowEqualSelector(
+            (state: TestState) => state.messages,
+            combiner
+        );
+
+        const result = selector({ messages: ['a', 'b'], items: [] });
+        const result2 = selector({ messages: ['a', 'b'], items: [] });
+
+        expect(result)
+            .toEqual('a, b');
+        expect(result2)
+            .toEqual('a, b');
+        expect(combiner)
+            .toHaveBeenCalledTimes(1);
+
+        const result3 = selector({ messages: ['b', 'c'], items: [] });
+
+        expect(result3)
+            .toEqual('b, c');
+        expect(combiner)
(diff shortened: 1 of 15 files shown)


src/common/selector/create-shallow-equal-selector.spec.ts
```diff
@@ -0,0 +1,60 @@
+import createShallowEqualSelector from './create-shallow-equal-selector';
+
+describe('createShallowEqualSelector()', () => {
+    interface TestState {
+        items: Array<{ id: number }>;
+        messages: string[];
+    }
+
+    it('creates selector that does shallow comparison instead of strict comparison', () => {
```
capsula4: this seems to be testing the memoization?

capsula4: wouldn't the test behave the same if you were doing strict comparison?

davidchin: Good question. It would be different because if you look at the following:

```ts
const result = selector({ messages: ['a', 'b'], items: [] });
const result2 = selector({ messages: ['a', 'b'], items: [] });
```

I'm actually passing in a new array instance for each call. Therefore, it won't hit the cache if the selector is configured to use strict comparison.

PR #652, opened 2019-08-01
CHECKOUT-4272: Avoid doing another round of transformation for subscription filters

davidchin:
## What?
* Avoid doing another round of transformation for subscription filters. This is done by introducing a projection object, which is responsible for listening to changes to the underlying data store and applying transformation required for external usage.

## Why?
* We only need to transform the data for the subscribers once. Currently, we do it more than once - one for the subscriber, and one for each filter that's passed into the `subscribe` method. That affects performance. Also, that prevents us from applying one-time memorization on the transformation function. This is because if you call the transformation function inside the filters, both current and previous state will be passed to it, which effectively expires the one-time memorization cache.

## Testing / Proof
* Unit


davidchin:
Good idea. I guess I could but I would need to spend a little bit more time on it. Maybe I'll do it later.

PR #663, opened 2019-08-05
CHECKOUT-4272: Optimise instrument selector and reducer

davidchin:
## What?
* Refactor `InstrumentSelector` to return new getters only when there are changes to relevant data.
* Update `instrumentReducer` to transform state only when it is necessary.

## Why?
Please refer to my previous PRs. They are related to this one.

## Testing / Proof
Unit

### DIFF
--- a/src/payment/instrument/instrument-selector.ts
+++ b/src/payment/instrument/instrument-selector.ts
@@ -1,40 +1,75 @@
+import { createSelector } from '../../common/selector';
+import { memoizeOne } from '../../common/utility';
+
 import Instrument from './instrument';
-import InstrumentState, { InstrumentMeta } from './instrument-state';
+import InstrumentState, { DEFAULT_STATE, InstrumentMeta } from './instrument-state';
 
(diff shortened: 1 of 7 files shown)


src/payment/instrument/instrument-selector.ts
```diff
@@ -1,40 +1,75 @@
...
+    const getLoadError = createSelector(
+        (state: InstrumentState) => state.errors.loadError,
+        loadError => () => loadError
+    );
 
-    getLoadError(): Error | undefined {
-        return this._instruments.errors && this._instruments.errors.loadError;
-    }
+    const getDeleteError = createSelector(
+        (state: InstrumentState) => state.errors.failedInstrument,
```
capsula4: unrelated to this PR, but maybe we can rename for consistency?  `deleteInstrumentId`

davidchin: Nice pickup. I'll address it separately :)

PR #668, opened 2019-08-05
CHECKOUT-4272: Optimise checkout external selectors

davidchin:
## What?
Refactor `CheckoutSelectors` to return new getters only when there are changes to relevant data.

## Why?
Please refer to my previous PRs. They are related to this one.

## Testing / Proof
Unit

### DIFF
--- a/src/checkout/checkout-service.spec.ts
+++ b/src/checkout/checkout-service.spec.ts
@@ -262,7 +262,7 @@ describe('CheckoutService', () => {
         it('returns state', () => {
             expect(checkoutService.getState()).toEqual(expect.objectContaining({
                 data: expect.any(Object),
-                errors: expect.any(CheckoutStoreErrorSelector),
+                errors: expect.any(Object),
                 statuses: expect.any(Object),
             }));
         });
--- a/src/checkout/checkout-service.spec.ts
+++ b/src/checkout/checkout-service.spec.ts
@@ -263,7 +263,7 @@ describe('CheckoutService', () => {
             expect(checkoutService.getState()).toEqual(expect.objectContaining({
                 data: expect.any(Object),
                 errors: expect.any(CheckoutStoreErrorSelector),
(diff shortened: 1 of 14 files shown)


src/checkout/checkout-service.spec.ts
```diff
@@ -294,7 +294,8 @@ describe('CheckoutService', () => {
 
             await checkoutService.loadCheckout();
 
-            expect(filter).toHaveBeenCalledWith(checkoutService.getState());
+            expect(Object.keys((filter as jest.Mock).mock.calls[0][0]))
```
icatalina: I learnt not long ago `ts-jest` provides a helper function for not having to do `as jest.Mock`:

https://github.com/kulshekhar/ts-jest/blob/master/docs/user/test-helpers.md#mocked

src/checkout/checkout-store-error-selector.ts
```diff
@@ -322,71 +193,125 @@ export default class CheckoutStoreErrorSelector {
...
+            getApplyCouponError: state.coupons.getApplyError,
+            getRemoveCouponError: state.coupons.getRemoveError,
+            getApplyGiftCertificateError: state.giftCertificates.getApplyError,
+            getRemoveGiftCertificateError: state.giftCertificates.getRemoveError,
+            getLoadInstrumentsError: state.instruments.getLoadError,
+            getDeleteInstrumentError: state.instruments.getDeleteError,
+            getLoadConfigError: state.config.getLoadError,
+        };
+
+        return {
```
icatalina: How come these 2 are not combined in a single return?

davidchin: This is because I have to pass `selector` into `getError`

src/checkout/checkout-service.spec.ts
davidchin: Ooo, this is nice 😄

PR #670, opened 2019-08-05
INT-1759 Using same order reference ID when an error occurs

rlopz:
[INT-1759](https://jira.bigcommerce.com/browse/INT-1759)
## What?
We started retrieving the orderReferenceId from order to use the same value when InvalidPaymentMethod or MFA-Abandoned occurs we do not want to generate another orderReferenceId in this cases

## Why?
We do not want to hit the amazon endpoint for generating another orderReferenceId because we already have one

## Siblings PR
[INT-1744 BCAPP](https://github.com/bigcommerce/bigcommerce/pull/31095)

## Testing / Proof
Screen Shot 2019-08-26 at 12 45 39 PM
Screen Shot 2019-08-26 at 12 47 01 PM

### DIFF
--- a/src/order/order.ts
+++ b/src/order/order.ts
@@ -14,6 +14,7 @@ export default interface Order {
     customerId: number;
     customerMessage: string;
     discountAmount: number;
+    handlingCostTotal: number;
     hasDigitalItems: boolean;
     isComplete: boolean;
     isDownloadable: boolean;
@@ -22,13 +23,12 @@ export default interface Order {
     orderAmount: number;
     orderAmountAsInteger: number;
     orderId: number;
+    payments?: OrderPayments;
     shippingCostTotal: number;
     shippingCostBeforeDiscount: number;
-    handlingCostTotal: number;
+    status: string;
     taxes: Tax[];
     taxTotal: number;
-    payments?: OrderPayments;
-    status: string;
 }
 
 export type OrderPayments = Array<GatewayOrderPayment | GiftCertificateOrderPayment>;
--- a/src/order/orders.mock.ts
+++ b/src/order/orders.mock.ts
@@ -22,9 +22,9 @@ export function getOrder(): Order {
         customerCanBeCreated: true,
         customerId: 0,
         discountAmount: 10,
+        handlingCostTotal: 8,
         hasDigitalItems: false,
         isComplete: true,
-        status: 'ORDER_STATUS_AWAITING_FULFILLMENT',
         isDownloadable: false,
         isTaxIncluded: false,
         lineItems: {
@@ -37,16 +37,6 @@ export function getOrder(): Order {
             ],
             customItems: [],
         },
-         taxes: [
-            {
-                name: 'Tax',
-                amount: 3,
-            },
-        ],
-        taxTotal: 3,
-        shippingCostTotal: 15,
-        shippingCostBeforeDiscount: 20,
-        handlingCostTotal: 8,
         orderAmount: 190,
         orderAmountAsInteger: 19000,
         orderId: 295,
@@ -54,6 +44,16 @@ export function getOrder(): Order {
             getGatewayOrderPayment(),
             getGiftCertificateOrderPayment(),
         ],
+        shippingCostTotal: 15,
+        shippingCostBeforeDiscount: 20,
+        status: 'ORDER_STATUS_AWAITING_FULFILLMENT',
+        taxes: [
+            {
+                name: 'Tax',
+                amount: 3,
+            },
+        ],
+        taxTotal: 3,
     };
 }
 
--- a/src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.spec.ts
+++ b/src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.spec.ts
@@ -76,6 +76,7 @@ describe('AmazonPayPaymentStrategy', () => {
             walletSpy(options);
 
             options.onReady(orderReference);
+            options.onPaymentSelect(orderReference);
         }
 
         bind(id: string) {
@@ -469,26 +470,10 @@ describe('AmazonPayPaymentStrategy', () => {
     });
 
     describe('When 3ds is enabled', () => {
+        const amazon3ds = getAmazonPay();
+        const payload = getOrderRequestBody();
         const paymentMethodsState = {
-            data: [{
-                id: 'amazon',
-                logoUrl: '',
-                method: 'widget',
-                supportedCards: [],
-                config: {
-                    displayName: 'AmazonPay',
-                    is3dsEnabled: true,
-                    merchantId: '0c173620-beb6-4421-99ef-03dc71a60685',
-                    testMode: false,
-                },
-                type: 'PAYMENT_TYPE_API',
-                initializationData: {
-                    clientId: '087eccf4-7f68-4384-b0a9-5f2fd6b0d344',
-                    region: 'US',
-                    redirectUrl: '/remote-checkout/amazon/redirect',
-                    tokenPrefix: 'ABCD|',
-                },
-            }],
+            data: [amazon3ds],
             meta: {
                 geoCountryCode: 'AU',
                 deviceSessionId: 'a37230e9a8e4ea2d7765e2f3e19f7b1d',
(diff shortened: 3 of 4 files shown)


rlopz:
unit testing will be added after this approach is approved

src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.ts
```diff
@@ -138,10 +138,17 @@ export default class AmazonPayPaymentStrategy implements PaymentStrategy {
...
+        const order = state.order.getOrder();
+
+        return order ? order.providerTransactionId : undefined;
+    }
+
     private _createWallet(options: AmazonPayPaymentInitializeOptions): Promise<AmazonPayWallet> {
         return new Promise((resolve, reject) => {
             const { container, onError = noop, onPaymentSelect = noop, onReady = noop } = options;
-            const referenceId = this._getOrderReferenceId();
+            const referenceId = this._getOrderReferenceId() || this._getOrderReferenceIdFromOrder();
```
metalvegetarianoprogresivo: What happens if both ar `undefined`?

src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.ts
```diff
@@ -256,7 +263,8 @@ export default class AmazonPayPaymentStrategy implements PaymentStrategy {
                         .then(() => {
                             confirmationFlow.success();
 
-                            return new Promise<never>(() => {});
+                            return new Promise<never>(() => {
+                            });
```
metalvegetarianoprogresivo: There's no point on splitting this in two lines.

src/order/order.ts
```diff
@@ -29,6 +29,7 @@ export default interface Order {
     taxTotal: number;
     payments?: OrderPayments;
     status: string;
+    providerTransactionId: string;
```
metalvegetarianoprogresivo: Can you sort this alphabetically?

src/order/orders.mock.ts
```diff
@@ -25,6 +25,7 @@ export function getOrder(): Order {
         hasDigitalItems: false,
         isComplete: true,
         status: 'ORDER_STATUS_AWAITING_FULFILLMENT',
+        providerTransactionId: 'providerTransactionId',
```
metalvegetarianoprogresivo: Can you sort this alphabetically?

src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.ts
rlopz: if `this._getOrderReferenceId()` comes undefined another `orderReferenceId` will be generated

rlopz:
@metalvegetarianoprogresivo I have addressed your comments, could you please take another look? ♻️

src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.ts
metalvegetarianoprogresivo: Okay, sound good.

rlopz:
> Hi @rlopz, thanks for your PR, can you write a test to verify your change?
Hi @davidchin, the unit test that I made on the last PR related with 3ds on amazon is already covering this new code. 
Screen Shot 2019-08-19 at 11 23 41 AM
Screen Shot 2019-08-19 at 11 23 27 AM

rlopz:
Hi @davidchin, I just change some code in order to follow the suggestion from Sydney team, could you please take another look ♻️

src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.ts
```diff
@@ -92,6 +94,10 @@ export default class AmazonPayPaymentStrategy implements PaymentStrategy {
             throw new InvalidArgumentError('Unable to proceed because "payload.payment.methodId" argument is not provided.');
         }
 
+        if (!this._isPaymentMethodSelected) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
```
davidchin: Are you sure this is the right error to throw? As far I know, `_isPaymentMethodSelected` would be false if the shopper hasn't selected a credit card option, or we are not able to save the billing address provided by Amazon, not purely because the developer using this library has forgotten to `initialize` the payment method. You can either create a new custom error type that's more specific or use an existing one such as `PaymentMethodFailedError`.

src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.ts
```diff
@@ -167,7 +178,10 @@ export default class AmazonPayPaymentStrategy implements PaymentStrategy {
                 },
                 onPaymentSelect: orderReference => {
                     this._synchronizeBillingAddress()
-                        .then(() => onPaymentSelect(orderReference))
+                        .then(() => {
+                            this._isPaymentMethodSelected = true;
```
davidchin: I believe `onPaymentSelect` gets called each time the shopper selects a new credit card option. Does that mean once it's set to `true`, there's a chance that something could fail later and the flag should be unset?

src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.ts
```diff
@@ -138,10 +144,15 @@ export default class AmazonPayPaymentStrategy implements PaymentStrategy {
         return amazon ? amazon.referenceId : undefined;
     }
 
+    private _getOrderReferenceIdFromInitializationData(): string | undefined {
+
```
davidchin: Unnecessary line break. 🍹

src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.spec.ts
```diff
@@ -470,26 +471,10 @@ describe('AmazonPayPaymentStrategy', () => {
     });
 
     describe('When 3ds is enabled', () => {
```
davidchin: I still can't see a test that verifies you're using the order reference from the `initializationData` as a fallback.

src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.ts
rlopz: Hi @davidchin thanks for the observation, I did not found the `PaymentMethodFailedError` but one that I think works perfectly for this case which is: `MissingPaymentMethod`

src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.ts
rlopz: Yes you are right `onPaymentSelect` is called each time the shopper selects a new credit card option. Actually this flag was added in order to be able to identify when a credit card is not selected. This scenario occurs when the wallet widget is loaded after a purchase failed with an `InvalidPaymentMethod` code

rlopz:
Hi @davidchin, I have addressed your comments ♻️, could you please take another look? also we have detected that exceptions are failing on execute methods not only for `Amazon` even we tested with `Authorizenet` and same error is happening, for example when `PaymentNotInitialized` is threw a spinner remains loading

src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.ts
```diff
@@ -85,11 +87,15 @@ export default class AmazonPayPaymentStrategy implements PaymentStrategy {
         const sellerId = this._getMerchantId();
 
         if (!referenceId || !sellerId) {
-            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+            return Promise.reject(new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized));
```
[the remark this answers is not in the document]
rlopz: Done! Thank you

rlopz:
Hi @davidchin I addressed your comments. Could you review it again?

PascalZajac:
@rlopz the commit message and PR title mention different things, can you please amend them appropriately? Then we'll merge this. Please ensure they reflect _what_ changed, not ticket titles.

rlopz:
> @rlopz the commit message and PR title mention different things, can you please amend them appropriately? Then we'll merge this. Please ensure they reflect _what_ changed, not ticket titles.

Hi @PascalZajac I just have updated both texts, could you please take another look? ♻️

PascalZajac:
@rlopz one last request, can you change the scope in the commit message from `checkout` to `payment`?

PR #673, opened 2019-08-06
CHECKOUT-4272: Use another data store as projection instead of plain observable so you can still notify subscribers when subscription filters are applied

davidchin:
## What?
* Use another instance of `DataStore` as the projection instead of a plain observable. Otherwise, `notifyState` method doesn't work if subscribers having subscription filters applied.
* Fix `unsubscribe` function. At the moment, it is broken because it is not bound to the subscription object.

## Why?
These are regression bugs.

## Testing / Proof
Unit


davidchin:
This PR depends on this https://github.com/bigcommerce/data-store-js/pull/25

PR #678, opened 2019-08-12
fix(payment): INT-1782 Add payment method tracking for Affirm & Zip

michxff:
## What? [INT-1782](https://jira.bigcommerce.com/browse/INT-1782)

- Add payment method tracking for Affirm & Zip.

## Why?

When an order is created and his payment was not finished (incomplete orders) we need to keep track which payment method was used.

## Sibling PRs

https://github.com/bigcommerce/bigcommerce/pull/31194

## Testing / Proof

https://drive.google.com/open?id=1JRdHRaEWkIxE8LajF874d9jlTjvNgujs


cbandam:
@michxff please check commit message, there is a typo in  `fffix`

michxff:
closing due we should not update payment method for an unpaid order

PR #679, opened 2019-08-13
CHECKOUT-4272: Only set up event listeners for iframe resizer when it is in use

davidchin:
## What?
Only set up event listeners for the iframe resizer when it is in use.

## Why?
Otherwise, the event listeners (window resize etc...) will take up CPU resources and affect performance.

## Testing / Proof
CircleCI


src/common/iframe/iframe-resizer.spec.ts
```diff
@@ -0,0 +1,20 @@
+
```
icatalina: empty line?

PR #683, opened 2019-08-15
feat(checkout): CHECKOUT-4294 Require province based on country

bdav87:
## What?
Requiring province/state based on specific countries, rather than requiring them just because a list of states was returned in association with a country.

## Why?
Some countries that return states  in checkout do not actually require states in address. For example, Germany and Switzerland both display states when selected, but this is not a requirement for addressing packages in those countries. More often, the building number, street, and postal code are sufficient.

I used this site as a reference for address requirements: https://www.bitboost.com/ref/international-address-formats.html#Formats

The list of countries requiring state/province can be updated over time if any country's postal requirements change.

## Testing / Proof
I forked checkout and checkout SDK, and referenced my SDK branch within checkout.
Within my BC VM I was able to verify that state/province is required in the shipping address form based on whether the ISO2 value of the country selected matches any of the countries in the `countriesRequiringProvince` array.

More importantly, I get the same set of shipping quotes for states like Germany and Switzerland when I exclude the state. I have FedEx, USPS, and UPS enabled with all options selected, sending from an origin in Austin, Texas, USA.

Germany, no state selected:
image

Germany, state selected:
image

### DIFF
--- a/src/form/form-selector.ts
+++ b/src/form/form-selector.ts
@@ -92,7 +92,7 @@ export function createFormSelectorFactory(): FormSelectorFactory {
             ...field,
             name: 'stateOrProvinceCode',
             options: { items },
-            required: true,
+            required: requireProvince(country),
             type: 'array',
             fieldType: 'dropdown',
             itemtype: 'string',
@@ -107,6 +107,18 @@ export function createFormSelectorFactory(): FormSelectorFactory {
         }
 
         return { ...field, required: Boolean(hasPostalCodes) };
+    }
+
+    function requireProvince(country?: Country): boolean {
+        const countriesRequiringProvince = [
+            'US',
+            'CA',
+            'AU',
+            'MX',
+            'MY',
+        ];
+
+        return country ? countriesRequiringProvince.indexOf(country.code) !== -1 : false;
     }
 
     return memoizeOne((


PascalZajac:
🤔 I'm not sure whether this change belongs in the SDK or in the API layer - if we validate payloads differently depending on the checkout source, I can see that leading to confusion / duplication.

bdav87:
> 🤔 I'm not sure whether this change belongs in the SDK or in the API layer - if we validate payloads differently depending on the checkout source, I can see that leading to confusion / duplication.

I have some thoughts about that, though I'm conflicted on what the best approach may be and welcome feedback from @bigcommerce/checkout and any other stakeholders.

We currently have state/province form validation duplicated across other areas of BigCommerce on the frontend - In Cornerstone on the [cart page ](https://github.com/bigcommerce/cornerstone/blob/master/assets/js/theme/cart/shipping-estimator.js#L58), as well as the deprecated Blueprint checkout and cart pages. 

In checkout, there is no enforcement of this specific form field requirement from the back end; I can submit orders with the state/province absent on my fork of checkout. Since the SDK is open source, and checkout will be open source in the near future, a developer would have the option of determining which countries they want to enforce these requirements on if they fork both repos.

However...

On the storefront, the cart page shipping estimator does enforce states/provinces based on their existence in the back end. The cart page specifically uses the V1 remote API when requesting quotes; I believe this is also consumed by stencil-utils. 

If we get stencil-utils to use the storefront APIs across the board for all cart functionality, then enforcing this at the API level would be simpler and also help avoid duplication in the back end. Ultimately we still would need to modify the SDK and our themes to adapt to the new validation.

The approach in this PR is the most direct and simple way of allowing shoppers internationally to complete checkout without being forced to do something that may be unfamiliar, like selecting province in checkout. But is it scalable, or even the _right_ thing to do?

bdav87:
@bigcommerce/checkout Considering we don't have API validation for province requirements, would an implementation at the SDK layer be acceptable? At the moment, the SDK only checks for the existence of provinces to determine if they're required, rather than setting a requirement per individual country.

I would think we could adjust the validation later, both in the SDK and in the API if necessary. I'd appreciate any thoughts around this PR or any potential solution we could pursue.

icatalina:
Hey @bdav87!

Before we can introduce a change like this we need to validate all payment and shipping providers to make sure none of them requires country or state to operate correctly.

Could you elaborate a bit more on why these countries? I'm pretty sure most countries where you have a postal code won't require a province or state to be selected.

Also, as Pascal already mentioned, the current list of countries/states is managed at the API level, it will make sense to append this kind of information to the same API and enhance the validation we do in the _backend_.

I know at the moment we can create 3rd party checkouts with no state, but maybe instead of embracing that behavior, we should actually make the validation consistent in the API and start requiring states for those countries that have them. Ultimately this is more of a product than an engineering decision tho.

bdav87:
Hi @icatalina,

Thank you for looking this over! I agree we probably should review payment gateways and shipping providers completely to check if a province is necessary. TL;DR, this should probably be validated in the backend first, and it'd be helpful to have Product weigh in, especially since we have a larger internationalization project in place currently.

The reason I picked the specific countries listed in the PR is around research I did on address formats and expectations around including a state/province on mail. I made some assumptions around countries I was familiar with, then double checked by Googling for their address requirements.

I spent more time making sure that European countries could still get quotes, since the associated JIRA was raised because of merchant feedback around checkout friction for their European shoppers.

Like you said, most countries with postal codes don't seem to enforce this requirement, and in my limited testing it was possible to get shipping quotes from USPS, UPS, and FedEx with the province excluded.

PR #687, opened 2019-08-22
fix(payment): INT-1845 handle payment method cancelled for Affirm

michxff:
## What? [INT-1845](https://jira.bigcommerce.com/browse/INT-1845)

- Release inventory for incomplete orders.

## Why?

- We need to return the products back to the inventory when the payment was not completed.

## Testing / Proof

WIP


src/payment/strategies/affirm/affirm-payment-strategy.ts
```diff
@@ -91,6 +90,13 @@ export default class AffirmPaymentStrategy implements PaymentStrategy {
                     paymentData: { nonce: result.checkout_token },
                 };
 
+                if (result.checkout_token === 'payment_canceled') {
+                    return this._store.dispatch(
```
davidchin: The indentation is wrong, as the `catch` applies to the `dispatch` call rather than `submitPayment`. It should be something like this:

```ts
return this._store.dispatch(this._paymentActionCreator.submitPayment(paymentPayload))
    .catch(() => {
        throw new PaymentMethodCancelledError();
    });
```

src/payment/strategies/affirm/affirm-payment-strategy.ts
```diff
@@ -91,6 +90,13 @@ export default class AffirmPaymentStrategy implements PaymentStrategy {
...
+                        this._paymentActionCreator.submitPayment(paymentPayload)).catch(() => {
```
davidchin: I'm not sure what you're trying to do here TBH. You're submitting the same payload to our Payment service when `checkout_token` is `payment_canceled` versus when it's not. I suppose you want to forward the token to the Payment service no matter what the value is so the service can make additional calls to the payment provider. In that case you can simply combine this call with the call below (as the only difference is the rejected value needs to be different when `checkout_token` is `payment_canceled`).

michxff: The approach has the following use cases:

1. When the shopper closes the modal without finishing the payment process I'm setting the payment token as `payment_canceled` so that the request to the payment service can be sent and the order and inventory will be updated (similar to non-modal payment providers), for this use case I'm not showing an error modal.

2. When we have a valid token, the error modal should be displayed if an error occurred.

If I combine the `submitPayment` calls to have a common `catch` somehow, I'll need to determine whether the reject comes from the 'payment_canceled' or the one with a valid token and re-throw the error, or even could come from the affirm initialization calls (first `.then`).

please let me know whether still make sense or if you have some recommendation to have a common `catch`

michxff:
thanks for your review @davidchin I updated the code to avoid duplicated submitPayment statements, it's ready for another look

PR #689, opened 2019-08-27
INT-1696 [SDK] Support Stored Credit Card (vaulting)

vmparra:
## What?
[INT-1696](https://jira.bigcommerce.com/browse/INT-1696)

## Why?
...

## Testing / Proof
image


src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -0,0 +1,784 @@
...
+
+            await strategy.initialize(options);
+            await strategy.execute(getOrderRequestBody());
+
+            expect(paymentActionCreator.submitPayment).toHaveBeenCalledTimes(2);
+            expect(paymentActionCreator.submitPayment).toHaveBeenLastCalledWith({
+                methodId: 'authorizenet',
+                paymentData: {
+                    nonce: JSON.stringify({
+                        threeDS2Token: 'token' ,
```
metalvegetarianoprogresivo: Remove this space before the comma.
```suggestion
                        threeDS2Token: 'token',
```

erikbc:
💚 Tested 
Test results described in the following file
https://docs.google.com/spreadsheets/d/16whNNEQBRwgACXfIyVhnaaykN3q_RXrGKXQubdl6Wpw/edit#gid=0

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -0,0 +1,289 @@
...
+                            this._store.dispatch(this._paymentActionCreator.submitPayment(payment)))
+                        .catch(error => {
+                            if (!(error instanceof RequestError) || !some(error.body.errors, { code: 'three_d_secure_required' })) {
+                                return Promise.reject(error);
+                            }
+
+                            if (error.body.three_ds_result.result_code === ResultCode.ChallengeShopper) {
+                                return this._handle3DS2Challenge(error.body.three_ds_result, payment.methodId)
+                                    .then((payment: Payment) =>
+                                        this._store.dispatch(this._paymentActionCreator.submitPayment(payment)));
```
davidchin: The closing parenthesis of `then` should be on the next line 🍺

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -0,0 +1,289 @@
...
+                            }
+
+                            return Promise.reject(error);
+                        });
+                }
+
+                if (error.body.three_ds_result.result_code === ResultCode.ChallengeShopper) {
+
```
davidchin: Redundant line break. 🍺

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -0,0 +1,289 @@
...
+                });
+            });
+    }
+
+    finalize(): Promise<InternalCheckoutSelectors> {
+        return Promise.reject(new OrderFinalizationNotRequiredError());
+    }
+
+    deinitialize(): Promise<InternalCheckoutSelectors> {
+        if (this._adyenOldComponent) {
```
davidchin: Please see my previous comment about this. You only need to keep the `adyenComponent.unmount` part.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -0,0 +1,289 @@
...
+
+        if (!threeDS2ChallengeWidgetSize) {
+            return '01';
+        }
+
+        return threeDS2ChallengeWidgetSize;
+    }
+
+    private _handle3DS2Challenge(resultObject: ThreeDS2Result, paymentMethodId: string): Promise<Payment> {
+
```
davidchin: Redundant linebreak 🍺

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -0,0 +1,289 @@
...
+                            },
+                        };
+
+                        resolve(paymentPayload);
+                    },
+                    onError: (error: AdyenError) => reject(error),
+                    size: this._getThreeDS2ChallengeWidgetSize(),
+                });
+
+            threeDS2Component.mount(`#${this._getAdyenV2PaymentInitializeOptions().containerId}-3ds`);
```
davidchin: I think it's fine to provide a default value of the container. But you should also allow the client app to pass in a container ID for the 3DS component. The reason for this is because the client app wouldn't know they need to have a container with the `-3ds` suffix. If you make this an option, it'd be documented and the client app would know they need to have such container in the DOM in advance.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -0,0 +1,289 @@
...
+                    onComplete: (challengeData: ThreeDS2OnComplete) => {
+                        const challengePaymentPayload = {
+                            ...challengeData.data,
+                            paymentData: resultObject.payment_data,
+                        };
+
+                        const paymentPayload = {
+                            methodId: paymentMethodId,
+                            paymentData: {
+                                nonce: JSON.stringify(challengePaymentPayload, null, 2),
```
davidchin: If you don't need to add additional whitespace to the JSON output, I prefer you leaving out the 2nd and 3rd argument.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -0,0 +1,289 @@
...
+                    onComplete: (fingerprintData: ThreeDS2OnComplete) => {
+                        const fingerprintPaymentPayload = {
+                            ...fingerprintData.data,
+                            paymentData: resultObject.payment_data,
+                        };
+
+                        const paymentPayload = {
+                            methodId: paymentMethodId,
+                            paymentData: {
+                                nonce: JSON.stringify(fingerprintPaymentPayload, null, 2),
```
davidchin: If you don't need to add additional whitespace to the JSON output, I prefer you leaving out the 2nd and 3rd argument.

src/payment/strategies/adyenv2/adyenv2-initialize-options.ts
```diff
@@ -0,0 +1,25 @@
...
+export default interface AdyenV2PaymentInitializeOptions {
+    /**
+     * The location to insert the Adyen component.
+     */
+    containerId: string;
+
+    /**
+     * Optional. Overwriting the default options
+     */
+    options?: CardComponentOptions;
```
davidchin: I think you'll need to be a little bit more explicit here to prevent problems in the future. As far as I know, there are two types of components you could potentially create via this strategy - the credit card component and the 3DS widget. This field is for the card component. And the field below (`threeDS2ChallengeWidgetSize`) is for the 3DS widget. For future-proofing, I suggest you to be explicit about the two options - have a `cardOptions` and `threeDS2Options`.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -0,0 +1,289 @@
...
+            paymentMethodsResponse: paymentMethod.initializationData.paymentMethodsResponse,
+        };
+
+        return this._adyenV2ScriptLoader.load(configuration)
+            .then(adyenCheckout => {
+                this._adyenCheckout = adyenCheckout;
+
+                const adyenComponent = this._adyenCheckout.create(paymentMethod.method, {
+                        ...adyenv2.options,
+                        onChange: (state: AdyenCardState) => {
```
davidchin: If you allow the client app to be able to override every option Adyen JS allows, what you have won't work properly as it will override the `onChange` option that's passed in. What you need to do is to call `options.onChange` (if it exists) here so it doesn't get completely overridden. Otherwise, if you don't want the client app to be able to specify an `onChange` callback, you'll need to omit it from the public interface.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -0,0 +1,289 @@
...
+        }
+
+        return threeDS2ChallengeWidgetSize;
+    }
+
+    private _handle3DS2Challenge(resultObject: ThreeDS2Result, paymentMethodId: string): Promise<Payment> {
+
+        return new Promise((resolve, reject) => {
+            if (!this._adyenCheckout) {
+                throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
```
davidchin: This might not be the most suitable error to throw. If you look at the error message, it says "Unable to proceed because payment method data is unavailable or not properly configured." This is not entirely true. I think a more suitable error to throw is `NotInitializedError` because one of the main reasons why `_adyenCheckout` is undefined is because the Adyen JS client is not initialized.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -0,0 +1,289 @@
...
+
+            threeDS2Component.mount(`#${this._getAdyenV2PaymentInitializeOptions().containerId}-3ds`);
+        });
+    }
+
+    private _handle3DS2Fingerprint(resultObject: ThreeDS2Result, paymentMethodId: string): Promise<Payment> {
+
+        return new Promise((resolve, reject) => {
+            if (!this._adyenCheckout) {
+                throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
```
davidchin: This might not be the most suitable error to throw. If you look at the error message, it says "Unable to proceed because payment method data is unavailable or not properly configured." This is not entirely true. I think a more suitable error to throw is `NotInitializedError` because one of the main reasons why `_adyenCheckout` is undefined is because the Adyen JS client is not initialized.

src/payment/strategies/adyenv2/adyenv2-script-loader.ts
```diff
@@ -0,0 +1,60 @@
...
+    constructor(
+        private _scriptLoader: ScriptLoader,
+        private _window: AdyenHostWindow = window
+    ) {}
+
+    load(configuration: AdyenConfiguration): Promise<AdyenCheckout> {
+        return this._loadStylesheet(
+            `https://checkoutshopper-${configuration.environment}.adyen.com/checkoutshopper/sdk/3.0.0/adyen.css`
+        ).then(() => {
+            return this._scriptLoader
```
davidchin: I believe you should be able to load both assets in parallel instead of in series. This would speed up the initialisation time. i.e.:

```ts
return Promise.all([
    this._loadStylesheet(`https://checkoutshopper-${configuration.environment}.adyen.com/checkoutshopper/sdk/3.0.0/adyen.css`),
    this._scriptLoader(`https://checkoutshopper-${configuration.environment}.adyen.com/checkoutshopper/sdk/3.0.0/adyen.js`),
])
    .then(() => {
        if (!this._window.AdyenCheckout) {
        }

        return new this._window.AdyenCheckout(configuration);
    })
    .catch(() => {
    });
```

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -0,0 +1,289 @@
...
+                const adyenComponent = this._adyenCheckout.create(paymentMethod.method, {
+                        ...adyenv2.options,
+                        onChange: (state: AdyenCardState) => {
+                            this._updateStateContainer(state);
+                        },
+                    });
+
+                adyenComponent.mount(`#${adyenv2.containerId}`);
+                if (this._adyenComponent) {
+                    this._adyenOldComponent = this._adyenComponent;
```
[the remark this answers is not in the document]
clopezh: Ok, the reason that drove us to do this is because Adyen supports multiple APMs i.e. `scheme` & `bcmc`, and whem we switch from scheme to bcmc, initialize method (for bcmc) is called before deinitialize (for scheme) and that induce to unmount the existing component in the right moment (bcmc in this case which is already mounted) and the payment section keep as blank. Do you know how to keep this process async to support switch providers those use the same component and call first deinitialize (to unmount the first component) and after that initialize (to mount the last component) ??

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -0,0 +1,289 @@
...
+                this._adyenComponent = adyenComponent;
+
+                return Promise.resolve(this._store.getState());
+            });
+    }
+
+    execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        const { payment, ...order } = payload;
+        const paymentData = payment && payment.paymentData;
+        const shouldSaveInstrument = Boolean(paymentData && (paymentData as HostedInstrument).shouldSaveInstrument);
```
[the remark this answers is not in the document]
clopezh: You're correct I forgot that we made a change in UCO to don't need this anymore

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
clopezh: I already post my reason to keep old components

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
clopezh: Yes actually this is by Adyen requirement, to mount a new component in the same page (now is handled by a modal (I already have this in this branch => https://github.com/clopezh/checkout-sdk-js/tree/INT-1901) so in this case which is the process to document it ?

clopezh: Do we just need to add a comment in this line?

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
clopezh: I'll omit onChange callback, since the only behavior of this is to get the data from adyen component so it is not needed to override

clopezh:
@davidchin  I almost addressed most of the comments and in other I kept an answer, Could you please take a look? ♻️

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
davidchin: What I meant is to have both `containerId` (which refers to the default credit card component) and `threeDS2ContainerId`. And if `threeDS2ContainerId` is not provided, you can fallback to `${containerId}-3ds`.

davidchin:
Thanks @clopezh for the update, I've replied to your comments :)

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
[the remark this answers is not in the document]
clopezh: scheme & bcmc is using different containers, but the same instace as I already see in the method you linked, let me try doing that change and see if the flow works

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
clopezh: Correct I'll added

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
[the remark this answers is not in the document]
clopezh: I finally did a success test with this line: 
`[paymentMethod.gateway, paymentMethod.id].filter(Boolean).join('-')`

clopezh:
@davidchin I already did all of your requests, could you please take a look and if possible approve it? ♻️

vmparra:
@davidchin Latest round of comments addresses, PR rebased and squashed.

icatalina:
hi @vmparra! 

can we please fill in the PR description? (I know it was merged already)

PR #692, opened 2019-08-29
CHECKOUT-4367: Stop throwing generic `StandardError` type

davidchin:
## What?
Stop throwing generic `StandardError` type.

## Why?
* It's supposed to be a base class, intended for other custom error types to inherit from. But it has been used directly. It is not preferred because it's quite generic which doesn't give you more information than a regular `Error` type. This change makes it `abstract` so it's not possible to construct it directly.
* In order to do so, I've replaced all the usage of `StandardError` with more specific error types.

## Testing / Proof
Unit

### DIFF
--- a/src/common/error/errors/standard-error.ts
+++ b/src/common/error/errors/standard-error.ts
@@ -3,6 +3,7 @@ import { setPrototypeOf } from '../../utility';
 import CustomError from './custom-error';
 
 export default class StandardError extends Error implements CustomError {
+    name = 'StandardError';
     type = 'standard';
 
     constructor(message?: string) {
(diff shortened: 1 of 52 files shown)


src/common/error/errors/standard-error.ts
```diff
@@ -2,7 +2,12 @@ import { setPrototypeOf } from '../../utility';
 
 import CustomError from './custom-error';
 
-export default class StandardError extends Error implements CustomError {
+/**
+ * This error type should not be constructed directly. It is a base class for
+ * all custom errors thrown in this library.
+ */
+export default abstract class StandardError extends Error implements CustomError {
```
icatalina: Nice 👍

PR #694, opened 2019-09-03
INT-1743 [Feedback - Post-Launch] PaymentMethodNotAllowed – The constraint on ORO are not handled

rlopz:
[INT-1743](https://jira.bigcommerce.com/browse/INT-1743)
## What?
We are handling the PaymentMethodNotAllowed error

## Why?
we want to be able to handle this error inside bigcommerce page

## Testing / Proof
PaymentMethodNotAllowed

## Siblings PRs
[INT-1743 BCAPP](https://github.com/bigcommerce/bigcommerce/pull/31545)

### DIFF
--- a/src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.spec.ts
+++ b/src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.spec.ts
@@ -38,6 +38,7 @@ import {
 } from '../../../remote-checkout';
 import { getRemoteCheckoutState, getRemoteCheckoutStateData } from '../../../remote-checkout/remote-checkout.mock';
 import { getConsignmentsState } from '../../../shipping/consignments.mock';
+import PaymentMethodNotAllowedError from '../../errors/payment-method-not-allowed-error';
 import PaymentMethod from '../../payment-method';
 import { getAmazonPay, getPaymentMethodsState } from '../../payment-methods.mock';
 import { PaymentInitializeOptions } from '../../payment-request-options';
@@ -76,6 +77,7 @@ describe('AmazonPayPaymentStrategy', () => {
             walletSpy(options);
 
             options.onReady(orderReference);
+            options.onPaymentSelect(orderReference);
         }
 
         bind(id: string) {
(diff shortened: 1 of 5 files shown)


src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.spec.ts
```diff
@@ -468,7 +470,26 @@ describe('AmazonPayPaymentStrategy', () => {
         }
     });
 
-    describe('When 3ds is enabled', () => {
+    describe( '#initialize', () => {
```
clopezh:
```suggestion
    describe('#initialize', () => {
```
🍹

src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.spec.ts
```diff
@@ -468,7 +470,26 @@ describe('AmazonPayPaymentStrategy', () => {
...
+                    address: {},
+                    paymentMethodNotAllowed: true,
+                },
+            };
+
+            jest.spyOn(store, 'getState');
+            jest.spyOn(store.getState().remoteCheckout, 'getCheckout')
+                .mockReturnValue(remoteCheckout);
+
+            expect(strategy.initialize({ methodId: paymentMethod.id, amazon: { container: 'wallet' } }))
```
clopezh: As this expect is returning a Promise, please return the last expect in test: 
```suggestion
            return expect(strategy.initialize({ methodId: paymentMethod.id, amazon: { container: 'wallet' } }))
```

src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.spec.ts
```diff
@@ -468,7 +470,26 @@ describe('AmazonPayPaymentStrategy', () => {
...
+                .resolves.toThrow(PaymentMethodNotAllowedError);
+        });
+    });
+
+    describe('#execute When 3ds is enabled', () => {
```
clopezh: Your description of a `describe` needs to be each public method, in this case `execute`, `when 3ds is enabled` is a more detailed tests, so I suggest to remove from here and add it in each test that applies 🍹

src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.spec.ts
rlopz: If add the return the assert fails

rlopz:
@clopezh I have addressed your comments, could you please take another look ♻️

JoseBobadilla20:
💚 Tested 
QA Details: I checked the correct functionality of the credit cards

src/payment/strategies/amazon-pay/amazon-pay-payment-strategy.spec.ts
davidchin: You'll need to put a `return`. But instead of using `resolves.toThrow`, you'll need to use `rejects.toThrow`: https://jestjs.io/docs/en/expect#rejects

rodna-bc:
This PR will be closed if the last changes in BCAPP pr are approved. Because with the refactor in BCapp this PR is not necessary anymore.

PR #695, opened 2019-09-04
CHECKOUT-4388: Mark index file of Embedded Checkout as file with side effect

davidchin:
## What?
Mark the index file of Embedded Checkout as a file with side effects.

## Why?
* Otherwise, `iframeResizer.contentWindow`, which is used in the index file, gets stripped out.
* It appears the file has been stripped out since the beginning. But the problem doesn't affect anything because the file is also exported in the index file of `iframeResizer` package. When the package gets imported, it also imports `iframeResizer.contentWindow`, which performs some side effect calls (i.e.: setting up event listeners). But since the change introduced by #679, we no longer require `iframeResizer` inside the iframe, which means `iframeResizer.contentWindow` from `iframeResizer` also doesn't get required and executed.
* The workaround for this problem is simply to require `iframeResizer.contentWindow` in `createEmbeddedCheckoutMessenger`, which is always called inside the iframe. That way, Webpack won't strip out the package because it will not consider it unused; and we will only set up the event listeners (the side effects of `iframeResizer.contentWindow`) when Embedded Checkout is active.

## Testing / Proof
Manual
Screen Shot 2019-09-04 at 4 18 40 pm

### DIFF
--- a/src/embedded-checkout/iframe-content/create-embedded-checkout-messenger.ts
+++ b/src/embedded-checkout/iframe-content/create-embedded-checkout-messenger.ts
@@ -1,3 +1,4 @@
+import { setupContentWindowForIframeResizer } from '../../common/iframe';
 import { EmbeddedCheckoutEvent, EmbeddedCheckoutEventType } from '../embedded-checkout-events';
 import IframeEventListener from '../iframe-event-listener';
 import IframeEventPoster from '../iframe-event-poster';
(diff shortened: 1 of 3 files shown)


src/embedded-checkout/iframe-content/create-embedded-checkout-messenger.ts
```diff
@@ -35,6 +35,8 @@ import NoopEmbeddedCheckoutMessenger from './noop-embedded-checkout-messenger';
  * @returns - An instance of `EmbeddedCheckoutMessenger`
  */
 export default function createEmbeddedCheckoutMessenger(options: EmbeddedCheckoutMessengerOptions): EmbeddedCheckoutMessenger {
+    require('iframe-resizer/js/iframeResizer.contentWindow');
```
icatalina: Might be good adding a comment in here.

PR #703, opened 2019-09-13
fix(shopper): CHECKOUT-4415 Add Customer Group to Customer object in payments payload

Tharaae:
## What?
Fixing ticket CHECKOUT-4415: After CHECKOUT-4405 (API and checkoutSDK to support Customer Group) has been merged, Checkout SDK needs to be updated to retrieve the customer group as part of the customer information by requesting `customer.customerGroup`

e.g. https://my-dev-store-574230549.store.bcdev/api/storefront/checkout/08efed69-92fd-41c5-ba58-60a9aa0313e7?include=customer.customerGroup

## Why?
Mainly to fix the issue PAYMENTS-4691: CyberSource Field 90 (Customer Group) Does Not Get Passed on UCO

## Testing / Proof
Unit Testing
Screenshots

Screen Shot 2019-09-13 at 5 43 47 pm

Screen Shot 2019-09-13 at 5 47 24 pm

ping @bigcommerce-labs/payments
ping @bigcommerce/checkout 
ping @bigcommerce/payments

### DIFF
--- a/src/customer/map-to-internal-customer.ts
+++ b/src/customer/map-to-internal-customer.ts
@@ -22,5 +22,6 @@ export default function mapToInternalCustomer(customer: Customer, billingAddress
         firstName,
         lastName,
         name: customer.fullName || [firstName, lastName].join(' '),
+        customerGroupName: customer.customerGroup.name,
     };
 }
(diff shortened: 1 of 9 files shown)


src/customer/map-to-internal-customer.ts
```diff
@@ -22,5 +22,7 @@ export default function mapToInternalCustomer(customer: Customer, billingAddress
         firstName,
         lastName,
         name: customer.fullName || [firstName, lastName].join(' '),
+        customerGroupId: customer.customerGroup.id,
```
icatalina: Could you check if we need customer group id?

PR #705, opened 2019-09-16
CHECKOUT-4403 Upgrade to latest Typescript

capsula4:
## What?
Upgrade to latest TS

## Why?
We are one major version behind. This way we can use latest features

## Testing / Proof
linter

### DIFF
--- a/src/address/map-to-internal-address.ts
+++ b/src/address/map-to-internal-address.ts
@@ -10,7 +10,10 @@ import InternalAddress from './internal-address';
  * the transition period as we are moving to adopt the new storefront API object
  * schema.
  */
-export default function mapToInternalAddress(address: Address | BillingAddress, consignments?: Consignment[]): InternalAddress {
+export default function mapToInternalAddress(
+    address: Address | BillingAddress,
+    consignments?: Consignment[]
+): InternalAddress<any> {
     let addressId;
 
     if (isBillingAddressLike(address)) {
--- a/src/embedded-checkout/iframe-content/iframe-embedded-checkout-messenger.ts
+++ b/src/embedded-checkout/iframe-content/iframe-embedded-checkout-messenger.ts
@@ -113,7 +113,7 @@ export default class IframeEmbeddedCheckoutMessenger implements EmbeddedCheckout
                 const handler = this._messageHandlers[key as keyof EmbeddedCheckoutEventMap];
 
                 if (handler) {
-                    handler.call(null, message);
+                    (handler as (event: EmbeddedCheckoutEvent) => void).call(null, message);
                 }
             });
     }
--- a/src/embedded-checkout/iframe-event-listener.ts
+++ b/src/embedded-checkout/iframe-event-listener.ts
@@ -73,7 +73,9 @@ export default class IframeEventListener<TEventMap extends IframeEventMap<keyof 
 
     @bind
     private _handleMessage(event: MessageEvent): void {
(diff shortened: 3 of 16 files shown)


src/embedded-checkout/iframe-content/iframe-embedded-checkout-messenger.ts
```diff
@@ -113,7 +113,7 @@ export default class IframeEmbeddedCheckoutMessenger implements EmbeddedCheckout
                 const handler = this._messageHandlers[key as keyof EmbeddedCheckoutEventMap];
 
                 if (handler) {
-                    handler.call(null, message);
+                    (handler as any).call(null, message);
```
capsula4: I tried to avoid using any in all these cases but the solution didn't seem straight forward. Looked like a types refactor would be required to avoid `any`.

davidchin: Hrmm yeah I'm not sure what's going on. Anyway, it might slightly better if you write `as (event: EmbeddedCheckoutEvent) => void` instead.

src/embedded-checkout/iframe-event-listener.ts
```diff
@@ -44,7 +44,7 @@ export default class IframeEventListener<TEventMap extends IframeEventMap<keyof
             this._listeners[type] = listeners = [];
         }
 
-        listeners.push(listener);
+        listeners.push(listener as any);
```
davidchin: I think you can avoid this by re-defining the `EventListeners` type, i.e.:

```ts
type EventListeners<TEventMap> = {
    [key in keyof TEventMap]?: Array<(event: TEventMap[key]) => void>;
};
```

src/embedded-checkout/iframe-event-listener.ts
```diff
@@ -77,7 +77,7 @@ export default class IframeEventListener<TEventMap extends IframeEventMap<keyof
             return;
         }
 
-        this.trigger(event.data);
+        this.trigger(event.data as any);
```
davidchin: I think you're getting an error because the `data` property of `MessageEvent` is untyped. You can get around this by specifying a type for it in the line above, i.e.:

```ts
isIframeEvent(event.data as TEventMap[keyof TEventMap], event.data.type as keyof TEventMap)
```

But in order to do this, you'll also need to fix the type in `isIframeEvent` and `IframeEvent`, i.e.:

```ts
function isIframeEvent<TEvent extends IframeEvent<TType>, TType extends string | number | symbol> {
    // ...
}
```

And
```ts
interface IframeEvent<TType extends string | number | symbol = string, TPayload = any> {
    // ...
}
```

src/address/map-to-internal-address.ts
```diff
@@ -10,7 +10,7 @@ import InternalAddress from './internal-address';
  * the transition period as we are moving to adopt the new storefront API object
  * schema.
  */
-export default function mapToInternalAddress(address: Address | BillingAddress, consignments?: Consignment[]): InternalAddress {
+export default function mapToInternalAddress(address: Address | BillingAddress, consignments?: Consignment[]): InternalAddress<any> {
```
davidchin: Maybe it's better to say `InternalAddress<string | number>`?

capsula4: If I change it to `string | number` it causes like 50 errors, might be good to keep it as `any` ?

I tried setting `<T extends string | number>`  but linter was still complaining

PR #708, opened 2019-09-17
CHECKOUT-4428: Build and publish release from CircleCI

davidchin:
## What?
Build and publish new releases from CircleCI.

## Why?
So we don't have to create releases locally.

## Testing / Proof
CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,44 +1,138 @@
+aliases:
+  - &node_executor
+      executor:
+        name: node/node
+        node-version: "10"
+
+  - &release_filter
+      branches:
+        only:
+          - master
+
+  - &pull_request_filter
+      branches:
+        ignore:
+          - master
+
 version: 2.1
 
 orbs:
   ci: bigcommerce/internal@volatile
   node: bigcommerce/internal-node@volatile
+  checkout-sdk-js:
+    commands:
+      install_dependencies:
+        steps:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -1,44 +1,138 @@
...
 orbs:
   ci: bigcommerce/internal@volatile
   node: bigcommerce/internal-node@volatile
+  checkout-sdk-js:
+    commands:
+      install_dependencies:
+        steps:
+          - restore_cache:
+              keys:
+                - checkout-sdk-js
```
capsula4: for the cache key, might it be better to use a dynamic key, like 
`checkout-sdk-js-{{ checksum "package-lock.json" }}`

davidchin: I think we only need a dynamic key if we want to cache the local `node_modules` folder and use `npm install` command to install our dependencies. The dynamic key can ensure that when we change the `package-lock.json` file, we could rebuild the `node_modules` folder to avoid any inconsistency due to the updates in the dependency tree.

But if we use `npm ci` and cache the `~/.npm` folder, we don't need the dynamic cache key because the command always rebuilds the `node_modules` folder. And it can grab the cached dependencies from `~/.npm` folder when it rebuilds the `node_modules` folder. If we use a dynamic key, we won't be able to use the cache when `package-lock.json` changes.

PR #709, opened 2019-09-18
CHECKOUT-4415: Check if `customerGroup` is defined before accessing it

davidchin:
## What?
Check if `customerGroup` is defined before accessing it.

## Why?
It can be undefined (i.e.: a guest shopper always doesn't belong to a group; also a returning shopper may not belong to a group).

## Testing / Proof
Manual


icatalina:
Thanks 🙇

PR #710, opened 2019-09-19
feat(payment): INT-1901 Use modal to handle 3DS

vmparra:
## What?
3DS 2.0 should show in a modal and you should not be able to do anything else. As well the loading icon keeps showing see screenshot.

## Why?
AdyenV2 Feedback

## Testing / Proof
image

NewAdyen3DSModal

## Sibling PR
[UCO](https://github.com/bigcommerce/checkout-js/pull/120)

### DIFF
--- a/src/payment/strategies/adyenv2/adyenv2-initialize-options.ts
+++ b/src/payment/strategies/adyenv2/adyenv2-initialize-options.ts
@@ -1,6 +1,6 @@
 import Omit from '../../../common/types/omit';
 
-import { CreditCardComponentOptions, ThreeDS2ComponentOptions } from './adyenv2';
+import { AdyenCreditCardComponentOptions, AdyenThreeDS2Options } from './adyenv2';
 
 /**
  * A set of options that are required to initialize the AdyenV2 payment method.
@@ -21,12 +21,12 @@ export default interface AdyenV2PaymentInitializeOptions {
     threeDS2ContainerId: string;
 
     /**
-     * Optional. Overwriting the default options
+     * ThreeDS2Options
      */
-    options?: Omit<CreditCardComponentOptions, 'onChange'>;
+    threeDS2Options: AdyenThreeDS2Options;
 
     /**
-     * Optional. Contains all three ds 2 options
+     * Optional. Overwriting the default options
      */
-    threeDS2Options?: ThreeDS2ComponentOptions;
+    options?: Omit<AdyenCreditCardComponentOptions, 'onChange'>;
 }
--- a/src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
+++ b/src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
@@ -5,7 +5,8 @@ import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
 import {
     InvalidArgumentError,
     MissingDataError,
-    MissingDataErrorType, NotInitializedErrorType,
+    MissingDataErrorType,
+    NotInitializedErrorType,
     RequestError
 } from '../../../common/error/errors';
 import NotInitializedError from '../../../common/error/errors/not-initialized-error';
@@ -124,8 +125,13 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
                 if (error.body.three_ds_result.result_code === ResultCode.IdentifyShopper) {
                     return this._handle3DS2Fingerprint(error.body.three_ds_result, payment.methodId)
                         .then((payment: Payment) =>
-                            this._store.dispatch(this._paymentActionCreator.submitPayment(payment))
-                        )
+                            this._store.dispatch(this._paymentActionCreator.submitPayment({
+                                ...payment,
+                                paymentData: {
+                                    ...payment.paymentData,
+                                    shouldSaveInstrument,
+                                },
+                            })))
                         .catch(error => {
                             if (!(error instanceof RequestError) || !some(error.body.errors, { code: 'three_d_secure_required' })) {
                                 return Promise.reject(error);
@@ -134,7 +140,13 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
                             if (error.body.three_ds_result.result_code === ResultCode.ChallengeShopper) {
                                 return this._handle3DS2Challenge(error.body.three_ds_result, payment.methodId)
                                     .then((payment: Payment) =>
-                                        this._store.dispatch(this._paymentActionCreator.submitPayment(payment))
+                                        this._store.dispatch(this._paymentActionCreator.submitPayment({
+                                            ...payment,
+                                            paymentData: {
+                                                ...payment.paymentData,
+                                                shouldSaveInstrument,
+                                            },
+                                        }))
                                     );
                             }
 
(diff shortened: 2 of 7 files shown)


erikbc:
💚 Tested

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -226,23 +226,36 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
 
             const threeDS2Container = this._getAdyenV2PaymentInitializeOptions().threeDS2ContainerId;
 
             threeDS2Component.mount(`#${threeDS2Container}`);
+            on3DSLoading();
         });
     }
 
     private _handle3DS2Fingerprint(resultObject: ThreeDS2Result, paymentMethodId: string): Promise<Payment> {
+
```
clopezh: This line is not needed since the return is the first line of the method

src/payment/strategies/adyenv2/adyenv2-script-loader.ts
```diff
@@ -21,16 +21,16 @@ export default class AdyenV2ScriptLoader {
             this._loadStylesheet(`https://checkoutshopper-${configuration.environment}.adyen.com/checkoutshopper/sdk/3.0.0/adyen.css`),
             this._scriptLoader.loadScript(`https://checkoutshopper-${configuration.environment}.adyen.com/checkoutshopper/sdk/3.0.0/adyen.js`),
         ])
-        .then(() => {
-            if (!this._window.AdyenCheckout) {
+            .then(() => {
```
clopezh: Keep the correct indentation for this promise: 
```
return Promise.all([
            this._loadStylesheet(`https://checkoutshopper-${configuration.environment}.adyen.com/checkoutshopper/sdk/3.0.0/adyen.css`),
            this._scriptLoader.loadScript(`https://checkoutshopper-${configuration.environment}.adyen.com/checkoutshopper/sdk/3.0.0/adyen.js`),
        ])
        .then(() => {...
```

src/payment/strategies/adyenv2/adyenv2-initialize-options.ts
```diff
@@ -26,7 +26,11 @@ export default interface AdyenV2PaymentInitializeOptions {
     options?: Omit<CreditCardComponentOptions, 'onChange'>;
 
     /**
-     * Optional. Contains all three ds 2 options
+     * Specify Three3DSChallenge Widget Size
      */
-    threeDS2Options?: ThreeDS2ComponentOptions;
```
davidchin: I don't think you can make this change anymore as this would be a breaking change. That is why you should never ever merge WIP PRs into master. Also, ideally, the name of the interface should be prefixed with `Adyen` because in the end all these interfaces are exported under a single TS interface file (https://github.com/bigcommerce/checkout-sdk-js/blob/master/dist/checkout-sdk.d.ts). This might cause a name collision in the future. So I suggest you mark `ThreeDS2ComponentOptions` as deprecated and create another one called `AdyenV2ThreeDS2ComponentOptions` and declare `threeDS2Options` field as `AdyenV2ThreeDS2ComponentOptions | ThreeDS2ComponentOptions`. This comment also applies to `CreditCardComponentOptions `

src/payment/strategies/adyenv2/adyenv2-initialize-options.ts
```diff
@@ -26,7 +26,11 @@ export default interface AdyenV2PaymentInitializeOptions {
...
+    threeDS2ChallengeWidgetSize?: string;
+
+    on3DSLoading?(): void;
```
davidchin: These options could fold under the `threeDS2Options`.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -210,6 +222,12 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
                 throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
             }
 
+            const { on3DSComplete, on3DSLoading } = this._getAdyenV2PaymentInitializeOptions();
+
+            if (!on3DSComplete || !on3DSLoading) {
+                throw new InvalidArgumentError('"options.adyenv2" argument was not provided during initialization.');
```
davidchin: The error message is not accurate. You're throwing an error because the callbacks are not provided. Also, all these really required callbacks? These are marked as optional. So if they are optional, you should just do a null check before using them or provide a fallback function (i.e.: `noop`) before calling them.

src/payment/strategies/adyenv2/adyenv2-initialize-options.ts
vmparra: Hi David, thank you for your concern. I believe the experiment hasn't been rolled out yet and dependencies such BigPay have not yet been merged either.

davidchin: Not 100% sure what you meant by "Adyen's feedback" because the API of the SDK and the API of Adyen JS client are two separate things. We design how the public interface should look like for the SDK. And the consumer of the SDK would provide values according to that interface. Similarly, Adyen designs how their public interface should look like for their client library. And the SDK, as the consumer of the client library, would provide values according to their interface. Sometimes the interfaces might be the same, sometimes they might not, which means the values need to be mapped.

You could essentially do the same thing in a backward compatible way, i.e.:

```ts
threeDS2Options: {
    widgetSize?: string;
    onLoading?(cancel: () => void): void;
    onComplete?(): void;
}
```

Regarding the use of experiments, imagine this: version 1.2.3 contains your WIP change which you think it wouldn't matter because the WIP change wouldn't be executed unless you turn on the experiment. Then one day everything is ready and you turn on the experiment. That would effectively turn on the WIP change, because although our checkout is always on the latest version of SDK, other third party checkouts might still be using version 1.2.3. That is why you should never merge a WIP PR and you should never change the interface of a method after it is merged to master.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -226,14 +240,21 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
+                        onComplete();
+
+                        reject(error);
+                    },
                     size: this._getThreeDS2ChallengeWidgetSize(),
                 });
 
             const threeDS2Container = this._getAdyenV2PaymentInitializeOptions().threeDS2ContainerId;
 
+            onLoading(reject);
```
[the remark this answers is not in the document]
davidchin: I'm not sure what your intention is. But it seems this callback intends to pass a function, which can cancel the 3DS flow if it's called. I don't think it's doing that at the moment as you're simply passing the `reject` function of the returning promise. That does not cancel or unmount the 3DS component. I believe you also need to call `unmount` on it.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -226,14 +240,21 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
                             },
                         };
 
+                        onComplete();
+
                         resolve(paymentPayload);
                     },
-                    onError: (error: AdyenError) => reject(error),
+                    onError: (error: AdyenError) => {
+                        onComplete();
```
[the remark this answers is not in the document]
vmparra: I think you're right, we'll remove this callback since modal is hidden when submit button gets clicked.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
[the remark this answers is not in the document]
vmparra: > I'm not sure what your intention is. But it seems this callback intends to pass a function, which can cancel the 3DS flow if it's called. I don't think it's doing that at the moment as you're simply passing the `reject` function of the returning promise. That does not cancel or unmount the 3DS component. I believe you also need to call `unmount` on it.

Hi David, I believe **cancel** works (and unmounts by modal, we've verified this happens directly on DOM) in both cases, on first fingerprint check and also on challenge check. 

### Fingerprint
Cancel On Fingerprint

### Challenge
Cancel On Challenge

PR #712, opened 2019-09-23
INT-1956 Disable or enable checkout button depending if a payment method is selected or not

rlopz:
[INT-1956](https://jira.bigcommerce.com/browse/INT-1956)
## What?
Adding the ability to disable or enable the checkout button if a payment method is selected or not

## Why?
Because we want to prevent the user to click the checkout button if a payment method is not selected

## Testing / Proof
INT-1956

## Siblings PRs
[INT-1956 SDK](https://github.com/bigcommerce/checkout-sdk-js/pull/712)

### DIFF
--- a/src/payment/strategies/amazon-pay/amazon-pay-payment-initialize-options.ts
+++ b/src/payment/strategies/amazon-pay/amazon-pay-payment-initialize-options.ts
@@ -1,4 +1,5 @@
 import { StandardError } from '../../../common/error/errors';
+import PaymentMethod from '../../payment-method';
 
 import AmazonPayOrderReference from './amazon-pay-order-reference';
 import AmazonPayWidgetError from './amazon-pay-widget-error';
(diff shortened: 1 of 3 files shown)


capsula4:
Hey @rlopz, thanks for looking into this.

After looking at the code, it seems like you don't need these SDK changes to achieve what you need. `disableSubmit` seems to behave pretty much as a callback. I can see we already have callbacks in place for both scenarios.

In Checkout JS you could replace o `disableSubmit(false)` to be executed inside of `onPaymentSelected()`, and `disableSubmit(true)` inside `onError()`. The other bit that you are passing, `this._paymentMethod` I think you should be able to infer that information in Checkout JS? Let me know what are your thoughts.

rlopz:
> Hey @rlopz, thanks for looking into this.
> 
> After looking at the code, it seems like you don't need these SDK changes to achieve what you need. `disableSubmit` seems to behave pretty much as a callback. I can see we already have callbacks in place for both scenarios.
> 
> In Checkout JS you could replace o `disableSubmit(false)` to be executed inside of `onPaymentSelected()`, and `disableSubmit(true)` inside `onError()`. The other bit that you are passing, `this._paymentMethod` I think you should be able to infer that information in Checkout JS? Let me know what are your thoughts.

Hi @capsula4, after your comments I was looking into the code and you are totally right so I will closed this PR and make the necessary changes in the other one [INT-1956](https://github.com/bigcommerce/checkout-js/pull/121)

PR #720, opened 2019-10-04
fix(order): CHECKOUT-4450 Handle recaptcha challenge not finish loading on slow connection

RobinOng:
## What?
As per above.

## Why?
Previously, when the shopper is on slow connection, the recaptcha challenge window might not finish loading when the shopper place an order. This will cause the checkout-sdk to throw an error and the UI will stop working, but there is no indication to the shopper on that is happening.

With this fix, we will handle this situtation gracefully. We will retry multiple times within a 7 second timeframe when the user place an order. If after 7 seconds the recaptcha challenge window has yet to finish loading, we will throw an error, but not block the UI. Shopper can then retry by trying to place order again.

## Testing / Proof
Added unit tests and manual testing.

Place an order on slow 3g connection:
Screen Shot 2019-10-04 at 10 39 48 am

### DIFF
--- a/src/order/spam-protection/google-recaptcha.ts
+++ b/src/order/spam-protection/google-recaptcha.ts
@@ -1,9 +1,14 @@
-import { Observable, Subject } from 'rxjs';
+import { defer, of, throwError, Observable, Subject } from 'rxjs';
+import { catchError, delay, retryWhen, switchMap } from 'rxjs/operators';
 
 import { MutationObserverFactory } from '../../common/dom/mutation-observer';
 import { NotInitializedError, NotInitializedErrorType } from '../../common/error/errors';
 
-import { SpamProtectionFailedError, SpamProtectionNotCompletedError } from './errors';
+import {
+    SpamProtectionFailedError,
+    SpamProtectionNotCompletedError,
+    SpamProtectionNotLoadedError,
+} from './errors';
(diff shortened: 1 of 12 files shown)


src/order/spam-protection/google-recaptcha.ts
```diff
@@ -47,34 +51,52 @@ export default class GoogleRecaptcha {
...
+        const timeout = window.setTimeout(() => {
+            window.clearInterval(interval);
 
-        this._recaptcha.execute();
+            event$.next({
+                error: new SpamProtectionNotLoadedError(),
+            });
+        }, 7000);
+
+        const interval = window.setInterval(() => {
```
davidchin: I think it'd work using `setTimeout` and `setInterval`. But you could potentially just use `retryWhen` operator to handle this scenario. i.e.:

```ts
const maxRetries = 70;
const retryInterval = 100;

return defer(() => {
    const element = document.querySelector('iframe[src*="bframe"]');

    return element ?
        of(element) :
        throwError(new SpamProtectionNotLoadedError());
})
    .pipe(
        retryWhen(errors => errors.pipe(
            delay(retryInterval),
            switchMap((error, index) =>
                index < maxRetries ? of(error) : throwError(error)
            )
        )),
        switchMap(element => {
            this._watchRecaptchaChallengeWindow(event$, element);
            recaptcha.execute();

            return event$;
        }),
        catchError(error => of({ error }))
    );
```

PR #723, opened 2019-10-10
fix(checkout): CHECKOUT-3365 Update cart and checkout state when shipping options are loaded

ankurbigcomm:
## What?
Update the Cart summary on the checkout page if the user's cart is updated on a different tab or window. 

Existing Behavior: 
User changes quantity in a different tab (cart page)
User is prompted with an error on checkout page but quantity and price is not updated.
User selects the shipping address and the update to the cart summary is reflected.

New Behavior: 
User Changes quantity in a different tab(cart page)
User is prompted with an error on checkout page and the quantity is updated keeping the previous shipping address.
If user updates the shipping address the cart summary is changed accordingly else remains the same.

## Why?
The error message provides a feedback to the shopper that the cart has been updated. However the updated summary is displayed only after he/she confirms the shipping address. 
This change will provide a more updated version of the cart summary and would only change if the user changes his shipping address.

## Testing / Proof
The cart summary is updated as soon as the error is displayed.

Screen Shot 2019-10-11 at 10 12 02 am

### DIFF
--- a/src/cart/cart-reducer.spec.ts
+++ b/src/cart/cart-reducer.spec.ts
@@ -103,6 +103,17 @@ describe('cartReducer()', () => {
         }));
     });
 
+    it('returns new data when shipping options are loaded', () => {
+        const action = {
+            type: ConsignmentActionType.LoadShippingOptionsSucceeded,
+            payload: getCheckout(),
+        };
+
+        expect(cartReducer(initialState, action)).toEqual(expect.objectContaining({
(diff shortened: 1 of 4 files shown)


ankurbigcomm:
> Changes look good, might be good to add unit test coverage for this scenario

Added Unit test cases. Since the new action did pretty much the same thing as other actions in the switch block have just written test to see if the reducer step is actually executed.

PR #725, opened 2019-10-14
CHECKOUT-4485: Use ESLint to fix inconsistency in import statements

davidchin:
## What?
Use ESLint to fix inconsistency in import statements.

## Why?
So these issues can be picked up automatically.

## Testing / Proof
CircleCI


src/common/utility/array-replace.ts
```diff
@@ -8,6 +8,7 @@ export interface ArrayReplaceOptions {
     matchObject(objectA: any, objectB: any): boolean;
 }
 
+/* eslint-disable import/export */
```
icatalina: I'm assuming the issue here is with the multiple `export default`, can we define the function multiple times and export once?

```ts
function arrayReplace<T>(currentArray: T[] | undefined, newArray: T[], options?: ArrayReplaceOptions): T[];
function arrayReplace<T>(currentArray: T[], newArray?: T[], options?: ArrayReplaceOptions): undefined;
function arrayReplace<T>(currentArray?: T[], newArray?: T[], options?: ArrayReplaceOptions): T[] | undefined;

export default arrayReplace;
```

PR #726, opened 2019-10-14
feat(payment): INT-1990 Resolve browser info for Adyen v2

clopezh:
## What?
BrowserInfo values are now sent from frontEnd to backEnd, this variable send this info: 
```
{
      userAgent,
      acceptHeader,
      language,
      colorDepth,
      screenHeight,
      screenWidth,
      timeZoneOffset,
      javaEnabled
}
```

## Why?
Adyen V2 needs browserInfo values direct from the browser to the backend to handle 3DS validations

## Testing / Proof
image
image

### DIFF
--- a/src/payment/is-vaulted-instrument.ts
+++ b/src/payment/is-vaulted-instrument.ts
@@ -1,4 +1,4 @@
-import { HostedVaultedInstrument, PaymentInstrument, VaultedInstrument } from './payment';
+import { FormattedPayload, FormattedVaultedInstrument, HostedVaultedInstrument, PaymentInstrument, VaultedInstrument } from './payment';
 
 export default function isVaultedInstrument(instrument: PaymentInstrument): instrument is VaultedInstrument {
     return Boolean((instrument as VaultedInstrument).instrumentId);
@@ -10,4 +10,15 @@ export function isHostedVaultedInstrument(instrument: PaymentInstrument): instru
         !instrument.hasOwnProperty('ccNumber') &&
         !instrument.hasOwnProperty('ccCvv')
     );
+}
+
+export function isFormattedVaultedInstrument(instrument: PaymentInstrument): instrument is FormattedPayload<FormattedVaultedInstrument> {
+    const formattedInstrument = (instrument as FormattedPayload<FormattedVaultedInstrument>).formattedPayload;
+
+    if (!formattedInstrument) {
+        return false;
+    }
+
+    return typeof formattedInstrument.bigpay_token === 'string' ||
+        Boolean(formattedInstrument.bigpay_token && formattedInstrument.bigpay_token.token);
 }
--- a/src/payment/payment-request-transformer.ts
+++ b/src/payment/payment-request-transformer.ts
@@ -10,7 +10,7 @@ import { CardExpiryFormatter, CardNumberFormatter, HostedInputValues } from '../
 import { mapToInternalOrder } from '../order';
 import { mapToInternalShippingOption } from '../shipping';
 
-import isVaultedInstrument from './is-vaulted-instrument';
+import isVaultedInstrument, { isFormattedVaultedInstrument } from './is-vaulted-instrument';
 import Payment, { CreditCardInstrument, HostedCreditCardInstrument, HostedVaultedInstrument, VaultedInstrument } from './payment';
 import PaymentMethod from './payment-method';
 import PaymentRequestBody from './payment-request-body';
@@ -35,7 +35,8 @@ export default class PaymentRequestTransformer {
         const orderMeta = checkoutState.order.getOrderMeta();
         const internalCustomer = customer && billingAddress && mapToInternalCustomer(customer, billingAddress);
 
-        const authToken = instrumentMeta && payment.paymentData && isVaultedInstrument(payment.paymentData) ?
+        const authToken = instrumentMeta && payment.paymentData &&
+            (isVaultedInstrument(payment.paymentData) || isFormattedVaultedInstrument(payment.paymentData)) ?
             `${checkoutState.payment.getPaymentToken()}, ${instrumentMeta.vaultAccessToken}` :
             checkoutState.payment.getPaymentToken();
 
--- a/src/payment/payment.ts
+++ b/src/payment/payment.ts
@@ -1,3 +1,4 @@
+import { BrowserInfo } from '../common/browser-info';
 import { Omit } from '../common/types';
 
 export default interface Payment {
@@ -9,7 +10,7 @@ export default interface Payment {
 export type PaymentInstrument = (
     CreditCardInstrument |
     CryptogramInstrument |
-    FormattedPayload<PaypalInstrument | FormattedHostedInstrument | FormattedVaultedInstrument> |
+    FormattedPayload<AdyenV2Instrument | PaypalInstrument | FormattedHostedInstrument | FormattedVaultedInstrument> |
     HostedCreditCardInstrument |
     HostedInstrument |
     NonceInstrument |
@@ -38,6 +39,8 @@ export interface CreditCardInstrument {
 export type HostedCreditCardInstrument = Omit<CreditCardInstrument, 'ccExpiry' | 'ccName' | 'ccNumber' | 'ccCvv'>;
 
 export type HostedVaultedInstrument = Omit<VaultedInstrument, 'ccNumber' | 'ccCvv'>;
+
+export type AdyenV2Instrument = AdyenV2Token | AdyenV2Card;
 
 export interface NonceInstrument {
     nonce: string;
@@ -95,12 +98,29 @@ export interface PaypalInstrument {
     };
 }
 
+interface AdyenV2Token extends FormattedVaultedInstrument {
+    browser_info: BrowserInfo;
+    credit_card_token?: void;
+}
+
+interface AdyenV2Card {
+    browser_info: BrowserInfo;
+    credit_card_token: {
+        token: string;
+    };
+    bigpay_token?: void;
+}
+
 export interface FormattedHostedInstrument {
     vault_payment_instrument: boolean | null;
 }
 
 export interface FormattedVaultedInstrument {
-    bigpay_token: string | null;
+    bigpay_token: {
+        credit_card_number_confirmation?: string;
+        verification_value?: string;
+        token: string;
+    } | string | null;
 }
 
 export interface FormattedPayload<T> {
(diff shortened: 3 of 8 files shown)


src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -87,7 +87,18 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
         return this._store.dispatch(this._orderActionCreator.submitOrder(order, options))
             .then(() => {
                 if (paymentData && isVaultedInstrument(paymentData)) {
-                    return this._store.dispatch(this._paymentActionCreator.submitPayment({...payment, paymentData}));
+                    const paymentPayload = {
+                        methodId: payment.methodId,
+                        gatewayId: payment.gatewayId,
+                        paymentData: {
+                            nonce:  JSON.stringify({
```
icatalina: Instead of passing everything in the nonce, we can now use the `formattedPayload` key and bigypay client will just send that payload to bigpay:

```ts
paymentData: {
	formattedPayload: {
		nonce: paymentData.nonce,
		browser_info: { 
 			user_agent,
            accept_header,
            language,
            color_depth,
            screen_height,
            screen_width,
            time_zone_offset,
            java_enabled,
		},
	}
}
```

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -289,9 +300,23 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
+
+    private _getBrowserInfo(): BrowserInfoRequest {
+        return {
+            userAgent: navigator.userAgent,
+            acceptHeader: 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,/;q=0.8',
+            language: navigator.language,
+            colorDepth: screen.colorDepth,
+            screenHeight: screen.height,
+            screenWidth: screen.width,
+            timeZoneOffset: new Date().getTimezoneOffset(),
```
hx: According to [Adyen's docs](https://docs.adyen.com/checkout/3d-secure/native-3ds2/api-reference#browserinfo) this should be a string, but `getTimezoneOffset` returns a number.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -289,9 +300,23 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
 
             this._stateContainer = JSON.stringify(state);
         }
     }
+
+    private _getBrowserInfo(): BrowserInfoRequest {
+        return {
+            userAgent: navigator.userAgent,
+            acceptHeader: 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,/;q=0.8',
+            language: navigator.language,
```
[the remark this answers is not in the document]
clopezh: Actually in SDK is using a type called `NavigatorLanguage` and only contains language and languages, so Can I use this => `navigator.languages[0]` ?

ref: 
image

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -289,9 +300,23 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
+            colorDepth: screen.colorDepth,
+            screenHeight: screen.height,
+            screenWidth: screen.width,
+            timeZoneOffset: new Date().getTimezoneOffset(),
+            javaEnabled: navigator.javaEnabled(),
```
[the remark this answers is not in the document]
clopezh: `navigator.javaEnabled()` is supported in Chrome, Safari, Mozilla, Edge... So I guess we can use that

clopezh:
@icatalina I already did the changes, Could you please take a look??

Also @hx I gave u some comments in yours, and other are resolved now, Could you please take a look? 

♻️

clopezh:
This PR is temporary flagged, since this change depends on INT-1902:  once touch TIER 3  then will need to update this one.

clopezh:
This PR is already unblocked, Could @bigcommerce/checkout  review it again? ♻️

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -315,4 +325,15 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
     private _updateAdyenComponentState(newState: AdyenCardState) {
         this._adyenComponentState = newState;
     }
+
+    private _getBrowserInfo(): BrowserInfoRequest {
+        return {
+            language: navigator.language || navigator.languages[0],
```
davidchin: As mentioned by @hx, you'll need call `navigator.userLanguage` for IE11 users. I don't think `navigator.languages` is supported by IE as well. You will get a compiler error if you write `userLanguage` because it's a non-standard property (TS only has types for standard properties). So you will need to write something like `navigator.language || (navigation as any).userLanguage` or define your own type.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -315,4 +325,15 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
+            color_depth: screen.colorDepth,
+            screen_height: screen.height,
+            screen_width: screen.width,
+            time_zone_offset: new Date().getTimezoneOffset().toString(),
+            java_enabled: navigator.javaEnabled(),
```
davidchin: It seems this method is not widely supported (https://developer.mozilla.org/en-US/docs/Web/API/NavigatorPlugins/javaEnabled). Because this is a method, it'd throw an error if it's undefined. So I think you should do a check before calling it - i.e.: `typeof navigator === 'function' ? navigator.javaEnabled() : undefined` (or `false` or any other default value you think it's appropriate).

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -315,4 +325,15 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
     private _updateAdyenComponentState(newState: AdyenCardState) {
         this._adyenComponentState = newState;
     }
+
+    private _getBrowserInfo(): BrowserInfoRequest {
+        return {
+            language: navigator.language || navigator.languages[0],
+            color_depth: screen.colorDepth,
```
davidchin: I don't think there's an equivalent property in IE but I guess it's okay to send back `undefined` if it's unavailable.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
clopezh: Ok I'll write this: `typeof navigator.javaEnabled === 'function' ? navigator.javaEnabled() : false` because if I only check navigator, then `navigator.javaEnabled()` throw me an error that it does not exists in type `never`

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
clopezh: I keep as `screen.colorDepth || 24` 
because : some implementations return 24 for compatibility reasons.
re: https://developer.mozilla.org/en-US/docs/Web/API/Screen/colorDepth

src/payment/is-vaulted-instrument.ts
```diff
@@ -1,5 +1,11 @@
-import { PaymentInstrument, VaultedInstrument } from './payment';
+import { AdyenV2Instrument, FormattedPayload, PaymentInstrument, VaultedInstrument } from './payment';
 
 export default function isVaultedInstrument(instrument: PaymentInstrument): instrument is VaultedInstrument {
+    const formattedPayload = (instrument as FormattedPayload<AdyenV2Instrument>).formattedPayload;
+
+    if (formattedPayload && formattedPayload.bigpay_token) {
+        return Boolean(formattedPayload.bigpay_token.token);
```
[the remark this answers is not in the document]
clopezh: Ok, This change is required because in `payment-request-transformer.ts` is in use `isVaultedInstrument` and in this step the paymentData is all formattedPayload variable, so is not  possible to cast as VaultedInstrument, that's the reason why I added this validation for Adyen that use formattedPayload

clopezh:
@icatalina The changes are already done!, Could you please take a look?? ♻️

src/payment/payment.ts
```diff
@@ -95,6 +96,18 @@ export interface PaypalInstrument {
     };
 }
 
+export interface AdyenV2Instrument {
```
icatalina: This type is actually comprised of 2 different interfaces:

```ts
interface AdyenV2Token {
    browser_info: BrowserInfo;
    credit_card_token?: void;
    bigpay_token: {
        credit_card_number_confirmation?: string;
        verification_value?: string;
        token: string;
    };
}

interface AdyenV2Card {
    browser_info: BrowserInfo;
    credit_card_token: {
        token: string;
    };
    bigpay_token?: void;
}

export type AdyenV2Instrument = AdyenV2Token | AdyenV2Card;
```

clopezh:
@icatalina Thanks to clarify and help with your request! Could you please take a look? ♻️

src/payment/payment-request-transformer.ts
```diff
@@ -35,7 +35,8 @@ export default class PaymentRequestTransformer {
         const orderMeta = checkoutState.order.getOrderMeta();
         const internalCustomer = customer && billingAddress && mapToInternalCustomer(customer, billingAddress);
 
-        const authToken = instrumentMeta && payment.paymentData && isVaultedInstrument(payment.paymentData) ?
+        const authToken = instrumentMeta && payment.paymentData &&
+        (isVaultedInstrument(payment.paymentData) || isFormattedInstrument(payment.paymentData)) ?
```
icatalina: Just some indenting here.

```suggestion
            (isVaultedInstrument(payment.paymentData) || isFormattedInstrument(payment.paymentData)) ?
```

src/payment/is-vaulted-instrument.ts
```diff
@@ -11,3 +11,14 @@ export function isHostedVaultedInstrument(instrument: PaymentInstrument): instru
         !instrument.hasOwnProperty('ccCvv')
     );
 }
+
+export function isFormattedInstrument(instrument: PaymentInstrument): instrument is FormattedPayload<FormattedVaultedInstrument> {
```
icatalina:
```suggestion
export function isFormattedVaultedInstrument(instrument: PaymentInstrument): instrument is FormattedPayload<FormattedVaultedInstrument> {
```

PR #729, opened 2019-10-16
PAYMENTS-4616 Allow vaulting and paying with BraintreePaypal accounts

icatalina:
## What?
- Use an object to configure `BraintreePaypal`
- Support for paying with `BraintreePaypal` vaulted accounts
- Support for vaulting `BraintreePaypal` accounts
- Bump `@bigcommerce/bigpay-client` to support `formattedPayload`

## Why?
- Paypal Vaulting Support

## Testing / Proof
- Unit / Functional

### DIFF
--- a/src/payment/payment.ts
+++ b/src/payment/payment.ts
@@ -72,6 +72,7 @@ export interface HostedInstrument {
 }
 
 export interface PaypalInstrument {
+    vault_payment_instrument: boolean | null;
     device_info: string | null;
     paypal_account: {
         token: string;
--- a/src/payment/payment.ts
+++ b/src/payment/payment.ts
@@ -4,7 +4,7 @@ export default interface Payment {
     paymentData?: PaymentInstrument & PaymentInstrumentMeta;
 }
 
-export type PaymentInstrument = CreditCardInstrument | NonceInstrument | VaultedInstrument | CryptogramInstrument | HostedInstrument | ThreeDSVaultedInstrument;
+export type PaymentInstrument = CreditCardInstrument | NonceInstrument | VaultedInstrument | CryptogramInstrument | HostedInstrument | ThreeDSVaultedInstrument | FormattedPayload<PaypalInstrument>;
 
 export interface PaymentInstrumentMeta {
     deviceSessionId?: string;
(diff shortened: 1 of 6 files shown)


src/payment/payment.ts
```diff
@@ -69,3 +70,16 @@ export interface ThreeDSecureToken {
 export interface HostedInstrument {
     shouldSaveInstrument?: boolean;
 }
+
+export interface PaypalInstrument {
+    vault_payment_instrument: boolean | null;
```
davidchin: How come it's `boolean | null` instead of just `?`

src/payment/strategies/braintree/braintree-payment-processor.ts
```diff
@@ -38,21 +46,19 @@ export default class BraintreePaymentProcessor {
...
 
                 return paypal.tokenize({
-                    amount,
-                    currency,
                     enableShippingAddress: true,
-                    flow: 'checkout',
-                    locale: storeLanguage,
-                    offerCredit,
+                    flow: shouldSaveInstrument ? 'vault' : 'checkout',
+                    offerCredit: false,
```
davidchin: Do you need `offerCredit: false` when `config` already provides that value?

src/payment/strategies/braintree/braintree-paypal-payment-strategy.ts
```diff
@@ -103,16 +104,53 @@ export default class BraintreePaypalPaymentStrategy implements PaymentStrategy {
...
 
         if (nonce) {
-            return Promise.resolve({ ...payment, paymentData: { nonce, method } });
+            return Promise.resolve({ ...payment, paymentData: this._formattedPayload(nonce) });
         }
 
-        const tokenizedCard = this._braintreePaymentProcessor
-            .paypal(grandTotal, storeLanguage, currency.code, this._credit);
+        if (isVaultedInstrument(paymentData)) {
+            if (!this._paymentMethod.config.isVaultingEnabled) {
```
davidchin: `config` could be destructed along with `nonce` 💅

src/payment/payment.ts
icatalina: I'm pretty sure big-pay requires values to be either defined or null. I'm not 100% sure, that's at least the pattern big-pay client follows.

We can in the future try to work out which values must actually be `null` and which ones can just be `?: <type>`.

src/payment/strategies/braintree/braintree-payment-processor.ts
icatalina: sorted

PR #730, opened 2019-10-17
PAYMENTS-4759 Add support to get and filter instruments

icatalina:
## What?
- Add support for account instruments
- Add support to filter instruments based on payment method type

## Why?
- We want to support account-type instruments required for paypal vaulting
- We would like to future proof the SDK so new instruments added to the API don't break old copies of the Checkout SDK.
- **From now onwards:** In order to instrument types that are not of type `card` you need to provide the `paymentMethod` to `getAllInstruments`

## Testing / Proof
- Unit / Functional

### DIFF
--- a/src/payment/instrument/instrument-response-body.ts
+++ b/src/payment/instrument/instrument-response-body.ts
@@ -5,16 +5,27 @@ export interface InstrumentError {
     message: string;
 }
 
-export interface InternalInstrument {
+export type InternalInstrument = CardInternalInstrument | AccountInternalInstrument;
+
+export interface BaseInternalInstrument {
     bigpay_token: string;
     default_instrument: boolean;
     provider: string;
-    iin: string;
-    last_4: string;
+    trusted_shipping_address: boolean;
+}
+
+export interface CardInternalInstrument extends BaseInternalInstrument {
+    brand: string;
     expiry_month: string;
     expiry_year: string;
-    brand: string;
-    trusted_shipping_address: boolean;
+    iin: string;
+    last_4: string;
+    method_type: 'card';
+}
+
+export interface AccountInternalInstrument extends BaseInternalInstrument {
+    external_id: string;
+    method_type: 'paypal';
 }
 
 export interface InstrumentsResponseBody {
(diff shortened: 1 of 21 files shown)


src/payment/instrument/instrument-response-body.ts
```diff
@@ -5,16 +5,27 @@ export interface InstrumentError {
     message: string;
 }
 
-export interface InternalInstrument {
+export  type InternalInstrument = CardInternalInstrument | AccountInternalInstrument;
```
davidchin: Extra space in front of `type` 💅

src/payment/instrument/instrument-selector.ts
```diff
@@ -19,7 +22,25 @@ export type InstrumentSelectorFactory = (state: InstrumentState) => InstrumentSe
 export function createInstrumentSelectorFactory(): InstrumentSelectorFactory {
     const getInstruments = createSelector(
         (state: InstrumentState) => state.data,
-        instruments => () => instruments
+        instruments => (paymentMethod?: PaymentMethod) => {
+            if (!instruments) {
+                return undefined;
```
davidchin: Just `return` is enough 💅

src/payment/instrument/supported-payment-instruments.ts
```diff
@@ -0,0 +1,42 @@
+import Instrument from './instrument';
+
+interface SupportedInstruments {
+    [key: string]: Partial<Instrument>;
```
davidchin: Maybe `Pick<Instrument, 'provider' | 'method'>` is more accurate so `provider` and `method` are not marked as optional.

src/payment/instrument/instrument-selector.ts
```diff
@@ -19,7 +22,25 @@ export type InstrumentSelectorFactory = (state: InstrumentState) => InstrumentSe
...
+            }
+
+            if (!paymentMethod) {
+                const supportedCardInstruments = filter(supportedInstruments, { method: 'card' });
+
+                return flatMap(supportedCardInstruments, cardInstrument => filter(instruments, cardInstrument));
```
davidchin: Is the order of `instruments` important? i.e.: Right now, it seems the order of `Instrument` is based on `supportedInstruments` and then `instruments`. If it's not important, then what you have here is good :)

icatalina: We only show one instrument type at a given time, so as long as we retain the order for every type of instrument it should be ok.

PR #734, opened 2019-10-25
CHECKOUT-4509 Optional param to exclude shippings options

capsula4:
## What?
Add param to optionally exclude shipping options

## Why?
At the moment we always include shipping options when shipping address is updated. Sometimes we want to update fields that we know won't affect shipping, so it's savy not to fetch shipping options in this scenario.

## Testing / Proof
unit

Coverage:
Screen Shot 2019-10-25 at 6 17 00 pm

### DIFF
--- a/src/common/http-request/join-or-merge-includes.ts
+++ b/src/common/http-request/join-or-merge-includes.ts
@@ -1,0 +1,17 @@
+import joinIncludes from './join-includes';
+import mergeIncludes from './merge-includes';
+
+/**
+ * Joins or merges a base list of includes with a set of additional includes.
+ */
+export default function joinOrMergeIncludes<T extends string>(
+    baseIncludes: T[],
+    includeDictionaryOrList: { [key in T]?: boolean } | T[] = []
+): string {
+    return Array.isArray(includeDictionaryOrList) ?
+        joinIncludes([
+            ...baseIncludes,
+            ...includeDictionaryOrList,
+        ]) :
+        mergeIncludes(baseIncludes, includeDictionaryOrList);
+}
--- a/src/shipping/consignment-action-creator.ts
+++ b/src/shipping/consignment-action-creator.ts
@@ -4,7 +4,7 @@ import { Observable, Observer } from 'rxjs';
 
 import { AddressRequestBody } from '../address';
 import { Cart } from '../cart';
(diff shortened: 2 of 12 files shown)


src/shipping/consignment-action-creator.ts
```diff
@@ -150,7 +151,7 @@ export default class ConsignmentActionCreator {
 
     updateAddress(
         address: AddressRequestBody,
-        options?: RequestOptions
+        options?: RequestOptions | UpdateShippingAddressRequestOptions
```
davidchin: I think this could just be `UpdateShippingAddressRequestOptions` as it is a subset of `RequestOptions`

src/shipping/shipping-request-options.ts
```diff
@@ -2,6 +2,13 @@ import { RequestOptions } from '../common/http-request';
 
 import { AmazonPayShippingInitializeOptions } from './strategies/amazon';
 
+/*
+ * A set of options for configuring requests related to the default shipping method.
+ */
+export interface UpdateShippingAddressRequestOptions extends RequestOptions {
+    includeShippingOptions?: boolean;
```
davidchin: I think the current way of passing optional includes is just an `include` flag, like `CheckoutParams`. And we pass the option like this.

```ts
loadCheckout('123', { params: { include: CheckoutIncludes.PhysicalItemsCategoryNames } })
```

Do you want to introduce a new convention?

capsula4:
@davidchin  let me know if it's good now

src/common/http-request/join-or-merge-includes.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import joinIncludes from './join-includes';
+import mergeIncludes from './merge-includes';
+
+/**
+ * Joins or merges a base list of includes with a set of additional includes.
+ */
+export default function joinOrMergeIncludes(
+    baseIncludes: string[],
+    includeDictionaryOrList: CheckoutIncludeParam | CheckoutIncludes[] = []
```
davidchin: I think `includeDictionaryOrList` needs to be a generic type instead of `CheckoutIncludeParam | CheckoutIncludes[]`, as this function belongs to the `common` so ideally it should not be tied to a specific set of params.

PR #737, opened 2019-10-30
INT-1434 Support for Klarna V2

rlopz:
[INT-1434](https://jira.bigcommerce.com/browse/INT-1434)
## What?
KlarnaV2 strategy has been created and keep Klarna V1 working as well

## Why?
Klarna wants to display each  category as a payment method and load a specific widget for them on the checkout page

## Testing / Proof
[VIDEO DEMO](https://drive.google.com/file/d/1tOaap_EP3JY-WRSOd0ckgGk2c3Ldh3cW/view?usp=sharing)

## Siblings PRs
[INT-1434 CHECKOUT-JS](https://github.com/bigcommerce/checkout-js/pull/165)
[INT-2301 NG-PAYMENTS](https://github.com/bigcommerce-labs/ng-payments/pull/959)

### DIFF
--- a/src/payment/strategies/klarnav2/klarnav2-payment-initialize-options.ts
+++ b/src/payment/strategies/klarnav2/klarnav2-payment-initialize-options.ts
@@ -1,0 +1,23 @@
+import { KlarnaLoadResponse } from './klarna-payments';
+
+/**
+ * A set of options that are required to initialize the KlarnaV2 payment method.
+ *
+ * When KlarnaV2 is initialized, a list of payment options will be displayed for the customer to choose from.
+ * Each one with its own widget.
+ */
+export default interface KlarnaV2PaymentInitializeOptions {
+    /**
+     * The ID of a container which the payment widget should insert into.
+     */
+    container: string;
+
+    /**
+     * A callback that gets called when the widget is loaded and ready to be
+     * interacted with.
+     *
+     * @param response - The result of the initialization. It indicates whether
+     * or not the widget is loaded successfully.
+     */
+    onLoad?(response: KlarnaLoadResponse): void;
+}
--- a/src/payment/strategies/klarnav2/klarnav2-payment-strategy.ts
+++ b/src/payment/strategies/klarnav2/klarnav2-payment-strategy.ts
@@ -1,0 +1,170 @@
+import { includes } from 'lodash';
+
+import { Address } from '../../../address';
+import { BillingAddress } from '../../../billing';
+import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
+import { InvalidArgumentError, MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType } from '../../../common/error/errors';
+import { OrderActionCreator, OrderRequestBody } from '../../../order';
+import { OrderFinalizationNotRequiredError } from '../../../order/errors';
+import { RemoteCheckoutActionCreator } from '../../../remote-checkout';
+import { PaymentMethodCancelledError, PaymentMethodInvalidError } from '../../errors';
+import PaymentMethodActionCreator from '../../payment-method-action-creator';
+import { PaymentInitializeOptions, PaymentRequestOptions } from '../../payment-request-options';
+import PaymentStrategy from '../payment-strategy';
+
+import KlarnaPayments, { KlarnaAddress, KlarnaAuthorizationResponse, KlarnaLoadResponse, KlarnaUpdateSessionParams } from './klarna-payments';
+import KlarnaV2ScriptLoader from './klarnav2-script-loader';
+
+export default class KlarnaV2PaymentStrategy implements PaymentStrategy {
+    private _klarnaPayments?: KlarnaPayments;
+    private _supportedEUCountries = ['AT', 'DE', 'DK', 'FI', 'GB', 'NL', 'NO', 'SE', 'CH'];
+
+    constructor(
+        private _store: CheckoutStore,
+        private _orderActionCreator: OrderActionCreator,
+        private _paymentMethodActionCreator: PaymentMethodActionCreator,
+        private _remoteCheckoutActionCreator: RemoteCheckoutActionCreator,
+        private _klarnav2ScriptLoader: KlarnaV2ScriptLoader
+    ) {}
+
+    initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
+        return this._klarnav2ScriptLoader.load()
+            .then(klarnaPayments => { this._klarnaPayments = klarnaPayments; })
+            .then(() => this._loadPaymentsWidget(options))
+            .then(() => this._store.getState());
+    }
+
+    deinitialize(): Promise<InternalCheckoutSelectors> {
+        return Promise.resolve(this._store.getState());
+    }
+
+    execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        if (!payload.payment) {
+            throw new InvalidArgumentError('Unable to proceed because "payload.payment" argument is not provided.');
+        }
+
+        const { payment: { paymentData, ...paymentPayload } } = payload;
+        const { gatewayId } = paymentPayload;
+
+        if (!gatewayId) {
+            throw new InvalidArgumentError('Unable to proceed because "payload.payment.gatewayId" argument is not provided.');
+        }
+
+        return this._authorize(paymentPayload.methodId, gatewayId)
+            .then(({ authorization_token: authorizationToken }) => this._store.dispatch(
+                this._remoteCheckoutActionCreator.initializePayment(gatewayId, { authorizationToken })
+            ))
+            .then(() => this._store.dispatch(
+                this._orderActionCreator.submitOrder({
+                    ...payload,
+                    payment: paymentPayload,
+                    // Note: API currently doesn't support using Store Credit with Klarna.
+                    // To prevent deducting customer's store credit, set it as false.
+                    useStoreCredit: false,
+                }, options)
+            ));
+    }
+
+    finalize(): Promise<InternalCheckoutSelectors> {
+        return Promise.reject(new OrderFinalizationNotRequiredError());
+    }
+
+    private _loadPaymentsWidget(options: PaymentInitializeOptions): Promise<KlarnaLoadResponse> {
+        if (!options.klarnav2) {
+            throw new InvalidArgumentError('Unable to load widget because "options.klarnav2" argument is not provided.');
+        }
+
+        const { methodId, klarnav2: { container, onLoad } } = options;
+        const state = this._store.getState();
+
+        return new Promise<KlarnaLoadResponse>(resolve => {
+            const paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(methodId);
+
+            if (!this._klarnaPayments || !paymentMethod.clientToken) {
+                throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+            }
+
+            this._klarnaPayments.init({ client_token: paymentMethod.clientToken });
+            this._klarnaPayments.load({ container, payment_method_category: paymentMethod.id }, response => {
+                if (onLoad) {
+                    onLoad(response);
+                }
+                resolve(response);
+            });
+        });
+    }
+
+    private _getUpdateSessionData(billingAddress: BillingAddress, shippingAddress?: Address): KlarnaUpdateSessionParams {
+        if (!includes(this._supportedEUCountries, billingAddress.countryCode)) {
+            return {};
+        }
+
+        const data: KlarnaUpdateSessionParams = {
+            billing_address: this._mapToKlarnaAddress(billingAddress, billingAddress.email),
+        };
+
+        if (shippingAddress) {
+            data.shipping_address = this._mapToKlarnaAddress(shippingAddress, billingAddress.email);
+        }
+
+        return data;
+    }
+
+    private _mapToKlarnaAddress(address: Address, email?: string): KlarnaAddress {
+        const klarnaAddress: KlarnaAddress = {
+            street_address: address.address1,
+            city: address.city,
+            country: address.countryCode,
+            given_name: address.firstName,
+            family_name: address.lastName,
+            postal_code: address.postalCode,
+            region: address.stateOrProvince,
+            email,
+        };
+
+        if (address.address2) {
+            klarnaAddress.street_address2 = address.address2;
+        }
+
+        if (address.phone) {
+            klarnaAddress.phone = address.phone;
+        }
+
+        return klarnaAddress;
+    }
+
+    private async _updateOrder(gatewayId: string) {
+        await this._paymentMethodActionCreator.loadPaymentMethod(gatewayId).toPromise();
+    }
+
+    private _authorize(methodId: string, gatewayId: string): Promise<KlarnaAuthorizationResponse> {
+        return new Promise<KlarnaAuthorizationResponse>((resolve, reject) => {
+            const billingAddress = this._store.getState().billingAddress.getBillingAddress();
+            const shippingAddress = this._store.getState().shippingAddress.getShippingAddress();
+
+            if (!billingAddress) {
+                throw new MissingDataError(MissingDataErrorType.MissingBillingAddress);
+            }
+
+            const updateSessionData = this._getUpdateSessionData(billingAddress, shippingAddress);
+
+            this._updateOrder(gatewayId);
+
+            if (!this._klarnaPayments) {
+                throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+            }
+
+            this._klarnaPayments.authorize({ payment_method_category: methodId }, updateSessionData, res => {
+                if (res.approved) {
+                    return resolve(res);
+                }
+
+                if (res.show_form) {
+                    return reject(new PaymentMethodCancelledError());
+                }
+
+                reject(new PaymentMethodInvalidError());
+            });
+        });
+    }
+}
--- a/src/payment/create-payment-strategy-registry.ts
+++ b/src/payment/create-payment-strategy-registry.ts
@@ -31,6 +31,7 @@ import { CreditCardPaymentStrategy } from './strategies/credit-card';
 import { CyberSourcePaymentStrategy } from './strategies/cybersource/index';
 import { createGooglePayPaymentProcessor, GooglePayBraintreeInitializer, GooglePayPaymentStrategy, GooglePayStripeInitializer } from './strategies/googlepay';
 import { KlarnaPaymentStrategy, KlarnaScriptLoader } from './strategies/klarna';
+import { KlarnaV2PaymentStrategy, KlarnaV2ScriptLoader } from './strategies/klarnav2';
 import { LegacyPaymentStrategy } from './strategies/legacy';
 import { MasterpassPaymentStrategy, MasterpassScriptLoader } from './strategies/masterpass';
 import { NoPaymentDataRequiredPaymentStrategy } from './strategies/no-payment';
@@ -144,6 +145,16 @@ export default function createPaymentStrategyRegistry(
             paymentMethodActionCreator,
             remoteCheckoutActionCreator,
             new KlarnaScriptLoader(scriptLoader)
+        )
+    );
+
+    registry.register(PaymentStrategyType.KLARNAV2, () =>
+        new KlarnaV2PaymentStrategy(
+            store,
+            orderActionCreator,
+            paymentMethodActionCreator,
+            remoteCheckoutActionCreator,
+            new KlarnaV2ScriptLoader(scriptLoader)
         )
     );
 
--- a/src/payment/payment-request-options.ts
+++ b/src/payment/payment-request-options.ts
@@ -7,6 +7,7 @@ import { ChasePayInitializeOptions } from './strategies/chasepay';
 import { CreditCardPaymentInitializeOptions } from './strategies/credit-card';
 import { GooglePayPaymentInitializeOptions } from './strategies/googlepay';
 import { KlarnaPaymentInitializeOptions } from './strategies/klarna';
+import { KlarnaV2PaymentInitializeOptions } from './strategies/klarnav2';
 import { MasterpassPaymentInitializeOptions } from './strategies/masterpass';
 import { PaypalExpressPaymentInitializeOptions } from './strategies/paypal';
 import { SquarePaymentInitializeOptions } from './strategies/square';
@@ -73,6 +74,12 @@ export interface PaymentInitializeOptions extends PaymentRequestOptions {
      * They can be omitted unless you need to support Klarna.
      */
     klarna?: KlarnaPaymentInitializeOptions;
+
+    /**
+     * The options that are required to initialize the KlarnaV2 payment method.
+     * They can be omitted unless you need to support KlarnaV2.
+     */
+    klarnav2?: KlarnaV2PaymentInitializeOptions;
 
     /**
      * The options that are required to initialize the Masterpass payment method.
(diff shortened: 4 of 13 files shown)


src/payment/strategies/klarnav2/klarnav2-payment-strategy.ts
```diff
@@ -0,0 +1,186 @@
...
+                if (onLoad) {
+                    onLoad(response);
+                }
+                resolve(response);
+            });
+        });
+    }
+
+    private _getUpdateSessionData(billingAddress: BillingAddress, shippingAddress?: Address): KlarnaUpdateSessionParams {
+        if (!includes(this._supportedEUCountries, billingAddress.countryCode)) {
```
vmparra: You should send this directly from BCApp

```JavaScript
        const { methodId } = options;

        this._paymentMethod = this._store.getState().paymentMethods.getPaymentMethod(methodId);

        if (!this._paymentMethod || !this._paymentMethod.initializationData) {
        }

        const _supportedCountries = { !this._paymentMethod.initializationData };

```

src/payment/payment-strategy-action-creator.ts
```diff
@@ -104,12 +105,18 @@ export default class PaymentStrategyActionCreator {
...
 
-            return concat(
-                of(createAction(PaymentStrategyActionType.InitializeRequested, undefined, { methodId })),
-                this._strategyRegistry.getByMethod(method)
+            if (strategyVersion) {
+                strategy = this._strategyRegistry.getByMethod(method, strategyVersion)
                     .initialize({ ...options, methodId, gatewayId })
-                    .then(() => createAction(PaymentStrategyActionType.InitializeSucceeded, undefined, { methodId }))
-            );
+                    .then(() => createAction(PaymentStrategyActionType.InitializeSucceeded, undefined, {methodId}));
```
vmparra: space between curly braces and argument

src/payment/strategies/klarnav2/klarnav2-payment-initialize-options.ts
```diff
@@ -0,0 +1,28 @@
...
+ * A set of options that are required to initialize the Klarna payment method.
+ *
+ * When Klarna is initialized, a widget will be inserted into the DOM. The
+ * widget has a list of payment options for the customer to choose from.
+ */
+export default interface KlarnaV2PaymentInitializeOptions {
+    /**
+     * The ID of a container which the payment widget should insert into.
+     */
+    container: string;
```
vmparra: Is should be called `containerId`

src/payment/strategies/klarnav2/klarnav2-payment-initialize-options.ts
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * The payment_method_category specifies which of Klarna’s customer offerings
+     * (Pay now, Pay later or Slice it) that is being shown.
+     */
+    payment_method_category: string;
```
vmparra: If names are fixed could you create an `enum` for such purpose?
 
```javascript
export enum PaymentMethodCategory {
    PayNow = 'Pay Now',
    PayLater = 'Pay Later',
    SliceIt = 'Slice It'
}

File: src/payment/strategies/klarnav2/klarnav2-payment-strategy.ts [hunk 5]
```diff
@@ -0,0 +1,186 @@
...
+    private _loadWidget(options: PaymentInitializeOptions): Promise<KlarnaLoadResponse> {
+        if (!options.klarnav2) {
+            throw new InvalidArgumentError('Unable to load widget because "options.klarna" argument is not provided.');
+        }
+
+        const {methodId, klarnav2: {container, payment_method_category, onLoad}} = options;
+        const state = this._store.getState();
+
+        return new Promise<KlarnaLoadResponse>(resolve => {
+            const paymentMethod = state.paymentMethods.getPaymentMethod(methodId);
```
vmparra: Empty line after declaration

src/payment/strategies/klarnav2/klarnav2-payment-strategy.ts
```diff
@@ -0,0 +1,186 @@
...
+            if (!paymentMethod) {
+                throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+            }
+
+            if (!this._klarnaPayments || !paymentMethod.clientToken) {
+                throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+            }
+            this._klarnaPayments.init({client_token: paymentMethod.clientToken});
```
vmparra: Empty line after code block

src/payment/strategies/klarnav2/klarnav2-payment-initialize-options.ts
rlopz: I based on others `initialize-options` files and many of them are using `container` instead of `containerId`, actually there are two that are using `containerId` (`Adyen` and `Stripe`) so I think the standard is `container`

src/payment/strategies/klarnav2/klarnav2-payment-initialize-options.ts
rlopz: I think there is no need to do that change

src/payment/strategies/klarnav2/klarnav2-payment-strategy.ts
rlopz: As far as I know checkout team does not want us to use `initializationData` field unless it is really necessary due this field is an `Any` type

rlopz:
@vmparra ♻️ I have addressed your comments, could you please take another look?

rlopz:
@bigcommerce/checkout This is the first approach to face `KlarnaV2` in which we are using the same module for both strategies (`Klarna` and `KlarnaV2`) and adding a new way to choose the strategy since right now we are only able to choose by `gateway` or `paymentId`. Anyway we know that the first approach does not follow the normal flow of bigcommerce so we are also thinking in another approach which is about creating a new module called `module.klarnav2` that inherit from `module.klarna` so we can choose the strategy by the gatewayId and we would not modify some generic files on SDK. Please let us know which approach do you prefer us to follow.

docs/interfaces/klarnapaymentinitializeoptions.md
```diff
@@ -1,14 +1,14 @@
-[@bigcommerce/checkout-sdk](../README.md) > [KlarnaPaymentInitializeOptions](../interfaces/klarnapaymentinitializeoptions.md)
+[@bigcommerce/checkout-sdk](../README.md) > [Klarnav2PaymentInitializeOptions](../interfaces/klarnapaymentinitializeoptions.md)
 
-# KlarnaPaymentInitializeOptions
+# Klarnav2PaymentInitializeOptions
```
capsula4: this is auto-generated so no need to update it

src/payment/payment-method.ts
```diff
@@ -12,4 +12,5 @@ export default interface PaymentMethod {
     nonce?: string;
     initializationData?: any;
     returnUrl?: string;
+    strategyVersion?: string;
```
capsula4: Also might be worth avoiding having a new `strategyVersion` param. You could just simple have a strategy called `klarnav2`. That should simplify the PR quite a bit.

rlopz: Actually I have created a new strategy called `klarnav2` but this is the first approach where does not exists a module for `klarnav2` so I can not choose the strategy with the `gatewayId` nor with `providerId` so I had to implement another way to choose the strategy that is why I add this new field. Also you can see this [comment](https://github.com/bigcommerce/checkout-sdk-js/pull/737#issuecomment-552703673). Further more right now I'm working on another approach that is about using same strategy and module for both versions of `Klarna`.

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -153,32 +212,59 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
         return klarnaAddress;
     }
 
-    private _authorize(): Promise<any> {
+    private async _updateOrder() {
+        await this._paymentMethodActionCreator.loadPaymentMethod('klarna').toPromise();
+    }
+
+    private _authorize(category: string): Promise<any> {
```
clopezh: This private method needs to return a specific type not `any` type.

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -153,32 +212,59 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
...
         return new Promise((resolve, reject) => {
-            const billingAddress = this._store.getState().billingAddress.getBillingAddress();
-            const shippingAddress = this._store.getState().shippingAddress.getShippingAddress();
 
-            if (!billingAddress) {
-                throw new MissingDataError(MissingDataErrorType.MissingBillingAddress);
-            }
+            return new Promise<InternalCheckoutSelectors>(() => {
```
clopezh: Please fix the identation 🍹

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -153,32 +212,59 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
...
+                    const billingAddress = this._store.getState().billingAddress.getBillingAddress();
+                    const shippingAddress = this._store.getState().shippingAddress.getShippingAddress();
 
-            if (!this._klarnaCredit) {
-                throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
-            }
+                    if (!billingAddress) {
```
clopezh: Please fix the identation 🍹

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -153,32 +212,59 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
...
+                        throw new MissingDataError(MissingDataErrorType.MissingBillingAddress);
+                    }
 
-            const updateSessionData = this._getUpdateSessionData(billingAddress, shippingAddress);
+                    const updateSessionData = this._getUpdateSessionData(billingAddress, shippingAddress);
```
clopezh: Please fix the identation 🍹

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -153,32 +212,59 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
...
 
-            this._klarnaCredit.authorize(updateSessionData, res => {
-                if (res.approved) {
-                    return resolve(res);
-                }
+                    if (category !== 'klarna') {
```
clopezh: Please fix the identation 🍹

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -153,32 +212,59 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
...
+                        this._updateOrder();
 
-                if (res.show_form) {
-                    return reject(new PaymentMethodCancelledError());
-                }
+                        if (!this._klarnaPayments) {
```
clopezh: Please fix the identation 🍹

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -153,32 +212,59 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
...
+                            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+                        }
 
-                reject(new PaymentMethodInvalidError());
-            });
+                        this._klarnaPayments.authorize({ payment_method_category: category }, updateSessionData, res => {
```
clopezh: Please fix the identation 🍹

src/payment/strategies/klarna/klarna-payments.ts
```diff
@@ -0,0 +1,57 @@
+export default interface KlarnaPayments {
+    authorize(options: KlarnaAuthorizeOptions, data: KlarnaUpdateSessionParams, callback: (res: KlarnaAuthorizationResponse) => void): void;
```
clopezh: You can create only one data type instead of have two.

i.e.
`authorize(KlarnaOptions): KlarnaResponses`

src/payment/strategies/klarna/klarna-script-loader.spec.ts
```diff
@@ -11,7 +11,7 @@ describe('KlarnaScriptLoader', () => {
     });
 
     it('loads widget script', () => {
-        klarnaScriptLoader.load();
+        klarnaScriptLoader.loadCredit();
 
         expect(scriptLoader.loadScript).toHaveBeenCalledWith(
             '//credit.klarnacdn.net/lib/v1/api.js'
```
clopezh: You expect this string and in klarna script loader is removed.

src/payment/strategies/klarna/klarna-script-loader.ts
```diff
@@ -1,17 +1,25 @@
 import { ScriptLoader } from '@bigcommerce/script-loader';
 
 import KlarnaCredit from './klarna-credit';
+import KlarnaPayments from './klarna-payments';
 import KlarnaWindow from './klarna-window';
+import KlarnaV2Window from './klarnav2-window';
 
-const SDK_URL = '//credit.klarnacdn.net/lib/v1/api.js';
```
clopezh: Here you removed sdk_url and you used in test

rlopz: actually I did not removed it I just change the constant name that's why I'm still doing unit test of it

src/payment/strategies/klarna/klarna-script-loader.spec.ts
rlopz: Actually is not removed I just change the name of the constant

rlopz:
@clopezh Thanks for the review, identation and unit test comments did not apply and the rest of them have been addressed. Could you please take another look ♻️

src/payment/strategies/klarna/klarna-payment-strategy.spec.ts
```diff
@@ -68,6 +73,15 @@ describe('KlarnaPaymentStrategy', () => {
             load: jest.fn((_, callback) => callback({ show_form: true })),
         };
 
+        klarnaPayments = {
+            authorize: jest.fn((params, data, callback) => {
+                params && data ? callback({ approved: true, authorization_token: 'bar' }) :
```
capsula4: do you need this condition? I see no difference in what's being executed

src/payment/strategies/klarna/klarna-payment-strategy.spec.ts
```diff
@@ -179,31 +282,19 @@ describe('KlarnaPaymentStrategy', () => {
 
             await strategy.initialize({ methodId: paymentMethod.id, klarna: { container: '#container' } });
 
-            strategy.execute(payload);
+            await strategy.execute(payload);
```
capsula4: if not required is better not to add await, these test seems to only be performing assertions on what happens when `strategy.execute(payload)` is executed, not necessarily making assertions of what happens when it finishes.

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -65,9 +73,9 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
 
         const { payment: { paymentData, ...paymentPayload } } = payload;
 
-        return this._authorize()
+        return this._authorize(paymentPayload.methodId)
             .then(({ authorization_token: authorizationToken }) => this._store.dispatch(
-                this._remoteCheckoutActionCreator.initializePayment(paymentPayload.methodId, { authorizationToken })
+                this._remoteCheckoutActionCreator.initializePayment('klarna', { authorizationToken })
```
capsula4: isnt `paymentPayload.methodId === 'klarna' ` ? do we need to pass it hard-coded?

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -114,6 +144,35 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
             }));
     }
 
+    private _loadPaymentsWidget(options: PaymentInitializeOptions): Promise<KlarnaLoadResponse> {
+        if (!options.klarnav2) {
+            throw new InvalidArgumentError('Unable to load widget because "options.klarna" argument is not provided.');
```
capsula4: "options.klarnav2"

src/payment/payment-method.ts
capsula4: wouldnt it be possible to call in the SDK the methodID `klarnav2`?

```js
// PaymentStrategyType.KLARNAV2 = 'klarnav2'
 registry.register(PaymentStrategyType.KLARNAV2, () =>
        new KlarnaPaymentV2Strategy(
            store,
            orderActionCreator,
            paymentMethodActionCreator,
            remoteCheckoutActionCreator,
            new KlarnaScriptLoader(scriptLoader)
        )
    );
```

This way you use a different class and avoid having if conditions in the main class. You could always make `KlarnaPaymentV2Strategy extends KlarnaPaymentStrategy` and internally just reuse the same logic.

capsula4:
It feels a bit odd to me that KlarnaV2 is implemented in KlarnaPaymentStrategy. There are a few conditions in the code that could be removed by having something like 
`KlarnaV2PaymentStrategy extends KlarnaPaymentStrategy`

Also, I think you could be able to register `KlarnaV2PaymentStrategy` against the 3 new methods IDs?

```js
const klarnaV2PaymentStrategy = new KlarnaV2PaymentStrategy(
            store,
            orderActionCreator,
            paymentMethodActionCreator,
            remoteCheckoutActionCreator,
            new KlarnaScriptLoader(scriptLoader)
        );

registry.register(PaymentStrategyType.KLARNAV2_PAYNOW, klarnaV2PaymentStrategy);
registry.register(PaymentStrategyType.KLARNAV2_PAYLATER, klarnaV2PaymentStrategy);
registry.register(PaymentStrategyType.KLARNAV2_PAYOVERTIME, klarnaV2PaymentStrategy);
```

What do you think?

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -29,7 +31,13 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
     ) {}
 
     initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
-        return this._klarnaScriptLoader.load()
+        const { klarnav2 } = options;
+
+        if (klarnav2) {
+            return this._initializeV2(options);
```
davidchin: You should create a new strategy for V2 instead of adding a bunch of `if` conditions to make V1 work with V2 as well. There will be some duplications but it's more important to keep the separation than to keep the code DRY. If you want to re-use part of the code, you can extract some common functionalities into separate classes or functions and compose them with the V1 and V2 strategies.

The API should also return `gateway` as `klarnav2` to avoid name collision with V1.

src/payment/strategies/klarna/klarna-payment-strategy.ts
davidchin: They are hard-coding this because Klarna V2 is a multi-option gateway, meaning method ID will be something like `pay_later`.

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -114,6 +144,35 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
...
+    private _loadPaymentsWidget(options: PaymentInitializeOptions): Promise<KlarnaLoadResponse> {
+        if (!options.klarnav2) {
+            throw new InvalidArgumentError('Unable to load widget because "options.klarna" argument is not provided.');
+        }
+
+        const { methodId, klarnav2: { container, payment_method_category, onLoad } } = options;
+        const state = this._store.getState();
+
+        return new Promise<KlarnaLoadResponse>(resolve => {
+            const paymentMethod = state.paymentMethods.getPaymentMethod(methodId);
```
davidchin: You can use the new method I've added recently - `getPaymentMethodOrThrow` - so you don't have to do the check you're doing below.

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -84,6 +92,28 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
         return Promise.reject(new OrderFinalizationNotRequiredError());
     }
 
+    private _initializeV2(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
+        return this._klarnaScriptLoader.loadPayments()
+            .then(klarnaPayments => { this._klarnaPayments = klarnaPayments; })
+            .then(() => {
+                this._unsubscribe = this._store.subscribe(
+                    state => {
+                        if (state.paymentStrategies.isInitialized(options.methodId)) {
```
davidchin: I don't think you'll need this check at all if you just move the entire `subscribe` call just above `.then(() => this._store.getState())` (after the initial `_loadPaymentsWidget` call). That'll effectively say only react to changes to the `outstandingBalance` field after the initial loading of the widget.

src/payment/strategies/klarna/klarnav2-payment-initialize-options.ts
```diff
@@ -0,0 +1,28 @@
...
+export default interface KlarnaV2PaymentInitializeOptions {
+    /**
+     * The ID of a container which the payment widget should insert into.
+     */
+    container: string;
+    /**
+     * The payment_method_category specifies which of Klarna’s customer offerings
+     * (Pay now, Pay later or Slice it) that is being shown.
+     */
+    payment_method_category: string;
```
davidchin: As far as I know, this is just `methodId`, which you already have access to in the `initialize` method. Therefore you should be using that instead of passing a new parameter that essentially carries the same information.

davidchin:
@capsula4 I think for these multi-options providers, the methods they provide are dynamic. It's only possible to register a strategy against a method if we know the method in advance. So what we usually do is to just register a strategy against the provider name (a.k.a. `gateway`).

I agree that v2 implementation shouldn't be embedded with v1. But I don't recommend using inheritance because v2 is not a subset of v1 (it's not a "is a" relationship).

rlopz:
Hello @davidchin, thanks for the review. I agree with you about having two strategies for klarna and klarnav2 is the best option for the frontend, but creating a new strategy for klarnav2 implies duplicate the backend implementation, and in the tech design we proposed to the @bigcommerce/payments team to update the existing module to support the new feature and the team approved it.
If we decide to change the approach to split the current integration into 2 separate modules, this will require to duplicate components in BC-App and bigpay which we will need to double check with payments team again.
Also recently we have fixed some bugs related to klarna, so having two implementations will required us to fix new coming bugs in both versions

capsula4:
> If we decide to change the approach to split the current integration into 2 separate modules, this will require to duplicate components in BC-App and bigpay which we will need to double check with payments team again.
Can't still use the same module? Just maybe need to map to different gateways (`klarna`, and `klarnav2`) to use the same module?

> Also recently we have fixed some bugs related to klarna, so having two implementations will required us to fix new coming bugs in both versions
You should still be able to extract common functionality into a function and reuse it in both places?

rlopz:
Hi @davidchin thanks for the review, I think I already addressed all your comments if I have missed something please let me know ♻️

src/payment/strategies/klarnav2/klarnav2-payment-strategy.ts
```diff
@@ -0,0 +1,185 @@
...
+        private _paymentMethodActionCreator: PaymentMethodActionCreator,
+        private _remoteCheckoutActionCreator: RemoteCheckoutActionCreator,
+        private _klarnav2ScriptLoader: KlarnaV2ScriptLoader
+    ) {}
+
+    initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
+        return this._klarnav2ScriptLoader.load()
+            .then(klarnaPayments => { this._klarnaPayments = klarnaPayments; })
+            .then(() => {
+                this._unsubscribe = this._store.subscribe(
```
davidchin: This whole `subscribe` block seems to be doing nothing so you can probably just remove it. As a result, you can also remove the `_unsubscribe` call in `deinitialize`.

PR #739, opened 2019-10-30
feat(checkout): CHECKOUT-4465 Make state/province optional for certain countries based on requireState flag

ankurbigcomm:
## What?
The backend would now return list of countries along with a flag for each country that specifies if the state or province is required for that country. The functionality is currently guarded by an experiment, value for which is also part of the country response. 

Based on the experiment toggle and the value of the requireState flag the state/province field will be made required/optional in both the billing and shipping address.

## Why?
Certain countries don't need the state/province as a mandatory field to uniquely identify billing or shipping address. For these countries shopper should be allowed to save his/her billing and shipping address without selecting the state or province. 

## Testing / Proof
Unit test cases updated accordingly and pass. Values verified in DB and behaviour toggles with change in DB values.

DB State: 
Screen Shot 2019-10-31 at 10 38 21 am

Required: 
Screen Shot 2019-10-31 at 10 29 18 am

Optional 
Screen Shot 2019-10-31 at 10 27 41 am

### DIFF
--- a/src/form/form-selector.ts
+++ b/src/form/form-selector.ts
@@ -73,12 +73,12 @@ export function createFormSelectorFactory(): FormSelectorFactory {
     }
 
     function processProvince(field: FormField, country?: Country): FormField {
-        const { subdivisions = [] } = country || {};
+        const { subdivisions = [], requiresState } = country || {};
 
         if (!subdivisions.length) {
             return {
                 ...field,
-                required: false,
+                required: requiresState == null ? false : requiresState,
             };
         }
 
@@ -91,7 +91,7 @@ export function createFormSelectorFactory(): FormSelectorFactory {
             ...field,
             name: 'stateOrProvinceCode',
             options: { items },
-            required: true,
+            required: requiresState == null ? true : requiresState,
             type: 'array',
             fieldType: 'dropdown',
             itemtype: 'string',
--- a/src/geography/country.ts
+++ b/src/geography/country.ts
@@ -3,6 +3,7 @@ export default interface Country {
     name: string;
     hasPostalCodes: boolean;
     subdivisions: Region[];
+    requiresState: boolean;
 }
 
 export interface Region {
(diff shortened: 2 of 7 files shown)


src/shipping/shipping-country-reducer.spec.ts
```diff
@@ -35,7 +35,8 @@ describe('shippingCountryReducer()', () => {
         const action: LoadShippingCountriesAction = {
             type: ShippingCountryActionType.LoadShippingCountriesSucceeded,
             payload: [
-                { code: 'JP', name: 'Japan', hasPostalCodes: false, subdivisions: [] },
+                { code: 'JP', name: 'Japan', hasPostalCodes: false, subdivisions: [], isRequiresStateExperimentOn: false,
```
capsula4: I'm thinking that the SDK should probably unaware of the experiment. Probably the experiment should simply be applied at an API level. If it's false, then `requiresState` should always be true whenever there are subdivisions.

ankurbigcomm: The changes are now complete. Removed all dependency from isRequriesStateExperimentOn

src/form/form-selector.ts
```diff
@@ -73,12 +73,12 @@ export function createFormSelectorFactory(): FormSelectorFactory {
     }
 
     function processProvince(field: FormField, country?: Country): FormField {
-        const { subdivisions = [] } = country || {};
+        const { subdivisions = [], requiresState = false } = country || {};
```
capsula4: might be safer to not default to `false`, you might want to check if it's undefined and default to the existing behaviour if this property does not exist.

ankurbigcomm: removed default and added check to use default if requiresState is null or undefined.

src/geography/country.ts
```diff
@@ -3,6 +3,7 @@ export default interface Country {
     name: string;
     hasPostalCodes: boolean;
     subdivisions: Region[];
+    requiresState: boolean;
```
capsula4: maybe mark this as optional `requiresState?`

PR #740, opened 2019-11-07
feat(payment): INT-2062 Support iDEAL payment provider through Adyen V2 gateway

vmparra:
[INT-2062](https://jira.bigcommerce.com/browse/INT-2062)

## What?
Support iDEAL APM through AdyenV2 gateway

## Why?
As a merchant I want to be able to support online back transfers using iDEAL as APM if store country is Netherlands and selected currency for the order is Euro.

## Testing / Proof
Unit Testing

## Dependencies
[BigPay](https://github.com/bigcommerce/bigpay/pull/2169)

Ideal 1 0 2020-01-13 14_37_34

### DIFF
--- a/src/payment/strategies/adyenv2/adyenv2-initialize-options.ts
+++ b/src/payment/strategies/adyenv2/adyenv2-initialize-options.ts
@@ -1,12 +1,12 @@
 import { Omit } from '../../../common/types';
 
-import { AdyenCreditCardComponentOptions, AdyenThreeDS2Options } from './adyenv2';
+import { AdyenAdditionalActionOptions, AdyenCreditCardComponentOptions, AdyenIdealComponentOptions, AdyenThreeDS2Options } from './adyenv2';
 
 /**
  * A set of options that are required to initialize the AdyenV2 payment method.
  *
  * Once AdyenV2 payment is initialized, credit card form fields, provided by the
- * payment provider as iFrames, will be inserted into the current page. These
+ * payment provider as IFrames, will be inserted into the current page. These
  * options provide a location and styling for each of the form fields.
  */
 export default interface AdyenV2PaymentInitializeOptions {
@@ -16,7 +16,8 @@ export default interface AdyenV2PaymentInitializeOptions {
     containerId: string;
 
     /**
-     * The location to insert the Adyen 3DS V2 component.
+     * @deprecated The location to insert the Adyen 3DS V2 component.
+     * Use additionalActionOptions instead as this property will be removed in the future
      */
     threeDS2ContainerId: string;
 
@@ -26,12 +27,18 @@ export default interface AdyenV2PaymentInitializeOptions {
     cardVerificationContainerId?: string;
 
     /**
-     * ThreeDS2Options
+     * @deprecated
+     * Use additionalActionOptions instead as this property will be removed in the future
      */
     threeDS2Options: AdyenThreeDS2Options;
 
     /**
+     * A set of options that are required to initialize additional payment actions.
+     */
+    additionalActionOptions: AdyenAdditionalActionOptions;
+
+    /**
      * Optional. Overwriting the default options
      */
-    options?: Omit<AdyenCreditCardComponentOptions, 'onChange'>;
+    options?: Omit<AdyenCreditCardComponentOptions, 'onChange'> | AdyenIdealComponentOptions;
 }
--- a/src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
+++ b/src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
@@ -13,7 +13,7 @@ import PaymentActionCreator from '../../payment-action-creator';
 import { PaymentInitializeOptions, PaymentRequestOptions } from '../../payment-request-options';
 import PaymentStrategy from '../payment-strategy';
 
-import { AdyenCardState, AdyenCheckout, AdyenComponent, AdyenConfiguration, AdyenError, ResultCode, ThreeDS2ComponentType, ThreeDS2OnComplete, ThreeDS2Result } from './adyenv2';
+import { AdyenAction, AdyenActionType, AdyenAdditionalAction, AdyenCardState, AdyenCheckout, AdyenComponent, AdyenComponentType, AdyenConfiguration, AdyenError, ResultCode, ThreeDS2OnComplete, ThreeDS2Result } from './adyenv2';
 import AdyenV2PaymentInitializeOptions from './adyenv2-initialize-options';
 import AdyenV2ScriptLoader from './adyenv2-script-loader';
 
@@ -60,7 +60,7 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
                 this._adyenCheckout = adyenCheckout;
 
                 const adyenPaymentComponent = this._adyenCheckout.create(
-                    paymentMethod.id,
+                    paymentMethod.method,
                     {
                         ...adyenv2.options,
                         onChange: (state: AdyenCardState) => {
@@ -74,7 +74,7 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
                 this._adyenPaymentComponent = adyenPaymentComponent;
 
                 if (adyenv2.cardVerificationContainerId) {
-                    const adyenCardVerificationComponent = this._adyenCheckout.create('securedfields', {
+                    const adyenCardVerificationComponent = this._adyenCheckout.create(AdyenComponentType.SecuredFields, {
                         onChange: (state: AdyenCardState) => {
                             this._updateAdyenComponentState(state);
                         },
@@ -143,7 +143,7 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
                 return this._store.dispatch(this._paymentActionCreator.submitPayment(paymentPayload));
             })
             .catch(error => {
-                if (!(error instanceof RequestError) || !some(error.body.errors, { code: 'three_d_secure_required' })) {
+                if (!(error instanceof RequestError) || (!some(error.body.errors, { code: 'three_d_secure_required' }) && !some(error.body.errors, { code: 'additional_action_required' }))) {
                     return Promise.reject(error);
                 }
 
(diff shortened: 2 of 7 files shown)


src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -270,6 +276,25 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
+    private _handleFromAction(action: AdyenV2Action, paymentMethodId: string): Promise<any> {
+        return new Promise(() => {
+            if (!this._adyenCheckout) {
+                throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+            }
+
+            if (paymentMethodId === AdyenV2PaymentMethodType.IDEAL) {
+                const { onLoad } = this._getAdyenV2PaymentInitializeOptions().threeDS2Options;
+                const adyenComponent = this._adyenCheckout
+                    .createFromAction(action);
```
davidchin: Unnecessary line break 🍹

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -270,6 +276,25 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
+                const container = this._getAdyenV2PaymentInitializeOptions().threeDS2ContainerId;
```
davidchin: Are you trying to support a 3DS flow? It doesn't seem to be - you're trying to support online bank transfer instead. But here, you're using the container ID and options for 3DS flow? I don't think that's appropriate, you either use the default options and `onload` callback or define something new for this flow.

vmparra: Both 3DS1 or online back transfers are full redirects asking for additional info and a 3DS2 challenge shopper. As you can see, all have in common one thing, they all are waiting for additional info to be captured.

My suggestion is to simply rename the container to something more generic like `needs-action-container-id` be able to mount every component where additional info is required.

vmparra: i.e. `scheme-adyen-component-field-additional-action`, or `ideal-adyen-component-field-additional-action`

vmparra: In fact, I'll rather suggest to switch `payment method`and `payment provider` order to have something like `adyen-scheme-additional-action-component-field`

vmparra:
@davidchin Please take another 👀 ♻️

src/payment/strategies/adyenv2/adyenv2-initialize-options.ts
```diff
@@ -16,9 +16,9 @@ export default interface AdyenV2PaymentInitializeOptions {
     containerId: string;
 
     /**
-     * The location to insert the Adyen 3DS V2 component.
+     * The location to insert the additional action component.
      */
-    threeDS2ContainerId: string;
+    additionalActionContainerId: string;
```
davidchin: I'm sorry but you can't rename fields once it's merged to master. You can only make additive changes that are backward compatible. One way to do this is to read from both properties, but mark the first one as deprecated (so both 3DS and Bank Transfer will use the same container). Another way is introduce a second field called `bankTransferContainerId`, which is only used when bank transfer is required. The first approach might be an easy solution as the client doesn't need to do anything and it can be compatible with other action types such as QR code etc..., but if in the future, 3DS requires a different treatment than bank transfer, you won't be able to do it. The second approach gives you more control over what the client needs to provide in order to support 3DS and bank transfer. But in the future, if we need to support more actions, we'll need to add more fields and ask the clients to update. Judging on the Adyen documentation, it seems the first approach might align better with what they offer https://docs.adyen.com/checkout/components-web#step-4-additional-front-end

src/payment/strategies/adyenv2/adyenv2-initialize-options.ts
```diff
@@ -16,10 +16,15 @@ export default interface AdyenV2PaymentInitializeOptions {
     containerId: string;
 
     /**
-     * The location to insert the Adyen 3DS V2 component.
+     * @deprecated The location to insert the Adyen 3DS V2 component.
```
davidchin: When you deprecate this field, it means you'll need to use `additionalActionContainerId` in the places where you're using `threeDS2ContainerId` right now. i.e.: 

```ts
const options = this._getAdyenV2PaymentInitializeOptions();

challengeComponent.mount(`#${options.additionalActionContainerId || options.threeDS2ContainerId}`);
```

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -270,6 +274,24 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
     }
 
+    private _handleFromAction(action: AdyenV2Action, paymentMethodId: string): Promise<any> {
+        return new Promise(() => {
+            if (!this._adyenCheckout) {
+                throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+            }
+
+            if (paymentMethodId === AdyenV2PaymentMethodType.IDEAL) {
+                const { onLoad } = this._getAdyenV2PaymentInitializeOptions().threeDS2Options;
```
davidchin: Since you've decided to generalise 3DS and other additional payment actions, you'll need to deprecate this `threeDS2Options` field as well and introduce a new field, something like `additionalActionOptions`. Similarly, you'll need to read from both in places where `threeDS2Options` field is used at the moment.

vmparra:
@davidchin Please take another 👀 ♻️

src/payment/strategies/adyenv2/adyenv2.ts
```diff
@@ -158,16 +162,11 @@ export interface InputDetail {
...
-     */
-    widgetSize?: string;
-
+export interface AdditionalActionCallbacks {
     /**
-     * A callback that gets called when adyen component is mounted
+     * A callback that gets called when adyen component is loaded
      */
-    onLoad(cancel: () => void): void;
+    onLoad(cancel: () => void, shouldShow: boolean): void;
```
davidchin: Based on this `shouldShow` flag, it seems to me that the requirements for 3DS and Bank Transfer are quite different, but you're trying to meet them with a single solution. IMO, they should be tackled separately.

3DS requires us to present a challenge to the shopper, which they must complete in order to proceed. So it requires two things on our end - a visible container to insert the challenge widget into, and a way for us to dismiss or unmount the challenge widget if the shopper chooses not to continue.

On the other hand, Bank Transfer doesn't require us to present any UI component to the shopper when the initial payment fails. Instead, it just needs to redirect the shopper elsewhere. Therefore it doesn't really need a visible container, and it doesn't have anything to unmount.

Having `shouldShow` flag is an indicator that the client needs to treat the two scenarios differently. I'm guessing you need this flag because in UCO we're presenting the 3DS challenge inside a modal, which gets triggered inside the `onLoad` callback. As you're using the same `onLoad` callback for Bank Transfer, the modal for 3DS will also be triggered for Bank Transfer flow unless it is manually suppressed. It is _a_ solution - but it is very specific to how we (UCO) present the 3DS challenge to shoppers. Imagine there's a custom checkout that decides to present the 3DS challenge inline inside of a modal; this `shouldShow` flag won't have any use.

So it seems to me that the most appropriate solution is to not to introduce a new `additionalActionOptions` and not deprecate `threeDS2Options`. Instead, you should treat 3DS and Bank Transfer separately. Correct me if I'm wrong, but it seems for Bank Transfer, the "component" that gets returned from `createFromAction` is not a visible one. Instead it is a hidden component whose responsibility is to redirect shoppers. If that's the case, then there's actually no need for the client to provide a separate container ID and a callback so you can use it to unmount it. You can probably just create new hidden container behind the scene and just mount it there.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -143,7 +143,7 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
                 return this._store.dispatch(this._paymentActionCreator.submitPayment(paymentPayload));
             })
             .catch(error => {
-                if (!(error instanceof RequestError) || !some(error.body.errors, { code: 'three_d_secure_required' })) {
+                if (!(error instanceof RequestError) || (!some(error.body.errors, { code: 'three_d_secure_required' }) && !some(error.body.errors, { code: 'BankTransferRequired' }))) {
```
davidchin: I think error codes from our payment service should be in snake case. I understand Adyen's `createFromAction` method expects `BankTransferRequired` as the action type. So what you should do here instead is to map the error code coming from BigPay to the format Adyen expects.

vmparra:
@davidchin Comments have been addressed ♻️

vmparra:
@davidchin Tour comments have been addressed, could you please take a 👁 ?

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -262,22 +266,46 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
 
-            const threeDS2Container = this._getAdyenV2PaymentInitializeOptions().threeDS2ContainerId;
+            const { threeDS2ContainerId, additionalActionOptions } = this._getAdyenV2PaymentInitializeOptions();
 
-            onLoad(() => {
-                challengeComponent.unmount();
-                reject();
-            });
+            if (onBeforeLoad) {
+                onBeforeLoad(() => {
```
davidchin: Can you have both `onBeforeLoad` and `onLoad`? But only pass the cancel callback to `onLoad`. i.e.:

```ts
if (onBeforeLoad) {
    onBeforeLoad(AdyenComponentType.IDEAL);
}

challengeComponent.mount(`#${additionalActionOptions.containerId || threeDS2ContainerId}`);

if (onLoad) {
    onLoad(() => {
        challengeComponent.unmount();
        reject();
    }, AdyenComponentType.IDEAL);
}
```

vmparra:
[BigPay](https://github.com/bigcommerce/bigpay/pull/2169) dependency is fixed, could you please approve both? And if it is possible [UCO](https://github.com/bigcommerce/checkout-js/pull/172) share as well? ♻️

PR #741, opened 2019-11-08
INT-1902 Update supported payment instrument

YosafatSanchez:
## What? [INT-1902](https://jira.bigcommerce.com/browse/INT-1902)
As a requirement from David Aland, we need to add a new field in BigPay DB, so to support this change is needed all these changes

## Why?
To support the new approach of vaulted instrument for Adyen V2

## Testing / Proof
No test were affected

### DIFF
--- a/src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
+++ b/src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
@@ -105,7 +105,7 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
                     return Promise.reject(error);
                 }
 
-                if (error.body.three_ds_result.result_code === ResultCode.IdentifyShopper) {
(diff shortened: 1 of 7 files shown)


src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -137,7 +137,7 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
                         });
                 }
 
-                if (error.body.three_ds_result.result_code === ResultCode.ChallengeShopper) {
+                if (error.body.three_ds_result.code === ResultCode.ChallengeShopper) {
```
icatalina: This breaks the contract for previous versions of the SDK. It should not be an issue because adyen v2 is not live yet, but we need to be mindful when doing these kind of changes.

PR #742, opened 2019-11-11
feat(checkout): INT-1780 Enable card vaulting for barclaycard

FranciscoOnt:
## What?
Create a payment strategy for barclaycard that supports vaulting, based on the regular `offsite-payment-strategy`.

## Why?
To make vaulting instruments and paying with vaulted instruments possible with this offsite provider, as the regular `offsite-payment-strategy` does not support vaulting.

## Testing / Proof
[Demo Video](https://drive.google.com/drive/folders/12sLa822lvuD8TmSe0L9J3Ki4ga1yCu9Y?usp=sharing)
Screen Shot 2019-12-02 at 11 10 43 AM


src/payment/strategies/barclaycard/barclaycard-payment-strategy.ts
```diff
@@ -0,0 +1,54 @@
...
+            throw new PaymentArgumentInvalidError(['payment']);
+        }
+
+        await this._store.dispatch(this._orderActionCreator.submitOrder(payload, options));
+
+        return await this._store.dispatch(this._paymentActionCreator.initializeOffsitePayment(
+            payment.methodId,
+            payment.gatewayId,
+            {
+                ...paymentData,
```
icatalina: Just out of curiosity, why are we destructuring this?

FranciscoOnt: @icatalina  Im working on a very similar PR (#746 ), and it's more likely for that PR to be merged first, so I was thinking into making them as similar as possible to have fewer conflicts to solve when the other PR gets merged.

FranciscoOnt: Im working on a PR that modifies the same files (#746) that is more likely to be merged  before this one, so  I'm trying to keep files similar to have the fewer conflicts as possible

FranciscoOnt:
@icatalina, could you please re-review? ♻️

src/payment/payment-action-creator.ts
```diff
@@ -37,10 +37,11 @@ export default class PaymentActionCreator {
 
     initializeOffsitePayment(
         methodId: string,
-        gatewayId?: string
+        gatewayId?: string,
+        paymentData: PaymentInstrument = {}
```
davidchin: I believe this should just be `instrumentId` and `shouldSaveInstrument` as `PaymentInstrument` contains other fields such as card number and other sensitive data which are not applicable to an offsite payment.

FranciscoOnt:
@davidchin , I made the changes you suggested in the latest commit , could you please re-review? ♻️

davidchin:
Hi @FranciscoOnt, thanks for one of the changes I've asked. You might have missed the part where I asked you to remove `BarclaycardPaymentStrategy` and update `OffsitePaymentStrategy` instead. Could you please take a look again? Thanks.

FranciscoOnt:
I've removed the `BarclaycardPaymentStrategy` as requested @davidchin ♻️

src/payment/payment.ts
```diff
@@ -80,6 +80,11 @@ export interface PaypalInstrument {
     };
 }
 
+export interface BarclaycardInstrument {
```
davidchin: Not sure why this payload needs to tie with a specific payment method. It seems they are just `HostedInstrument` and `VaultedInstrument` in a different format. I believe moving forward we want to use `FormattedPayload<T>`. So perhaps you could just create two new interfaces and call them `FormattedHostedInstrument` and `FormattedVaultedInstrument`?

src/payment/payment-action-creator.ts
```diff
@@ -37,10 +37,18 @@ export default class PaymentActionCreator {
...
-        gatewayId?: string
+        gatewayId?: string,
+        instrumentId?: string,
+        shouldSaveInstrument?: boolean
     ): ThunkAction<InitializeOffsitePaymentAction, InternalCheckoutSelectors> {
         return store => {
-            const payload = this._paymentRequestTransformer.transform({ gatewayId, methodId }, store.getState());
+            const paymentData = {
+                formattedPayload: {
+                    bigpay_token: instrumentId ? instrumentId : null,
```
davidchin: I think you only ever want to send one or the other - not both, because it doesn't make sense for you to indicate that you want to vault an instrument when you're already paying with a vaulted instrument. The constraint could either be applied on the interface or implementation level. i.e.:

```ts
let paymentData: FormattedPayload<FormattedHostedInstrument | FormattedVaultedInstrument> | undefined;

if (instrumentId) {
    paymentData = { bigpay_token: instrumentId };
} else if (shouldSaveInstrument) {
    paymentData = { vault_payment_instrument: shouldSaveInstrument };
}
```

PR #743, opened 2019-11-12
feat(payment): INT-1902 Update payment method id and support gateway in instrument selector

clopezh:
## What?
According to this request: https://github.com/bigcommerce/bigpay/pull/1910#discussion_r345040613, this change is needed to make Adyen V2 works

## Why?
If this change is not merge, Adyen V2 will not work as expected

## Testing / Proof
Unit / Functional / Manual

## Depends on
[BigPay](https://github.com/bigcommerce/bigpay/pull/1910)


src/payment/instrument/supported-payment-instruments.ts
```diff
@@ -5,49 +5,49 @@ interface SupportedInstruments {
 }
 
 const supportedInstruments: SupportedInstruments = {
-    scheme_adyenv2: {
+    'adyenv2.scheme': {
```
capsula4: what's the reason for this change? having dots in the key might make it harder to construct the object.

clopezh: I did this suggestion following a @icatalina DM (from slack) and he tell me that it's ok to keep as this. If I compose like: 
```
adyenv2: {
   scheme: { }
}
```

Then I will need to modify the type of `SupportedInstruments` interface, and it will have two types: either this structure or the current one.

However @icatalina is aware of this, he has all the context of this implementation.

PR #746, opened 2019-11-15
feat(checkout): INT-1778 Create a method to get browser info

FranciscoOnt:
## What?
Create a method to get browser info and add it to the offsite payment flow.

## Why?
Barclaycard requires browser info for 3DS2, but it can be reused/extended for other providers

## Testing / Proof
Pending

### DIFF
--- a/src/common/browser-info/browser-info.ts
+++ b/src/common/browser-info/browser-info.ts
@@ -4,5 +4,6 @@ export default interface BrowserInfo {
     language: string;
     screen_height: number;
     screen_width: number;
+    time_zone?: string;
     time_zone_offset: string;
 }
--- a/src/payment/payment-action-creator.ts
+++ b/src/payment/payment-action-creator.ts
@@ -1,12 +1,14 @@
 import { createAction, ThunkAction } from '@bigcommerce/data-store';
+import { merge } from 'lodash';
 import { concat, from, of } from 'rxjs';
 import { catchError, switchMap } from 'rxjs/operators';
 
 import { InternalCheckoutSelectors } from '../checkout';
+import { BrowserInfo } from '../common/browser-info';
 import { throwErrorAction } from '../common/error';
 import { OrderActionCreator } from '../order';
 
-import Payment, { FormattedHostedInstrument, FormattedPayload, FormattedVaultedInstrument } from './payment';
+import Payment, { FormattedBrowserInfo, FormattedHostedInstrument, FormattedPayload, FormattedVaultedInstrument } from './payment';
 import { InitializeOffsitePaymentAction, PaymentActionType, SubmitPaymentAction } from './payment-actions';
 import PaymentRequestSender from './payment-request-sender';
 import PaymentRequestTransformer from './payment-request-transformer';
@@ -39,15 +41,20 @@ export default class PaymentActionCreator {
         methodId: string,
         gatewayId?: string,
         instrumentId?: string,
-        shouldSaveInstrument?: boolean
+        shouldSaveInstrument?: boolean,
+        browserInfo?: BrowserInfo
     ): ThunkAction<InitializeOffsitePaymentAction, InternalCheckoutSelectors> {
         return store => {
-            let paymentData: FormattedPayload<FormattedHostedInstrument | FormattedVaultedInstrument> | undefined;
+            let paymentData: FormattedPayload<FormattedBrowserInfo | FormattedHostedInstrument | FormattedVaultedInstrument> | undefined;
 
             if (instrumentId) {
                 paymentData = { formattedPayload: { bigpay_token: instrumentId } };
             } else if (shouldSaveInstrument) {
                 paymentData = { formattedPayload: { vault_payment_instrument: shouldSaveInstrument } };
+            }
+
+            if (browserInfo) {
+                paymentData = merge(paymentData, { formattedPayload: { browser_info: browserInfo }});
             }
 
             const payload = this._paymentRequestTransformer.transform({ gatewayId, methodId, paymentData }, store.getState());
--- a/src/payment/payment.ts
+++ b/src/payment/payment.ts
@@ -10,7 +10,7 @@ export default interface Payment {
 export type PaymentInstrument = (
     CreditCardInstrument |
     CryptogramInstrument |
-    FormattedPayload<AdyenV2Instrument | PaypalInstrument | FormattedHostedInstrument | FormattedVaultedInstrument> |
+    FormattedPayload<AdyenV2Instrument | FormattedBrowserInfo | PaypalInstrument | FormattedHostedInstrument | FormattedVaultedInstrument> |
     HostedCreditCardInstrument |
     HostedInstrument |
     NonceInstrument |
@@ -109,6 +109,10 @@ interface AdyenV2Card {
         token: string;
     };
     bigpay_token?: void;
+}
+
+export interface FormattedBrowserInfo {
+    browser_info: BrowserInfo | null;
 }
 
 export interface FormattedHostedInstrument {
--- a/src/payment/strategies/offsite/offsite-payment-strategy.ts
+++ b/src/payment/strategies/offsite/offsite-payment-strategy.ts
@@ -1,4 +1,5 @@
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
+import { getBrowserInfo } from '../../../common/browser-info';
 import { OrderActionCreator, OrderPaymentRequestBody, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
 import { PaymentArgumentInvalidError } from '../../errors';
@@ -32,8 +33,12 @@ export default class OffsitePaymentStrategy implements PaymentStrategy {
                     payment.methodId,
                     payment.gatewayId,
                     instrumentId,
-                    shouldSaveInstrument))
-            );
+                    shouldSaveInstrument,
+                    {
+                        ...getBrowserInfo(),
+                        time_zone: new Date().toLocaleTimeString('en-us', {timeZoneName: 'short'}).split(' ')[2],
+                    }
+                )));
     }
 
     finalize(options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
--- a/src/payment/strategies/offsite/offsite-payment-strategy.spec.ts
+++ b/src/payment/strategies/offsite/offsite-payment-strategy.spec.ts
@@ -102,7 +102,21 @@ describe('OffsitePaymentStrategy', () => {
     it('initializes offsite payment flow', async () => {
         await strategy.execute(payload, options);
 
-        expect(paymentActionCreator.initializeOffsitePayment).toHaveBeenCalledWith(options.methodId, options.gatewayId, null, null);
+        expect(paymentActionCreator.initializeOffsitePayment).toHaveBeenCalledWith(
+            options.methodId,
+            options.gatewayId,
+            null,
+            null,
+            expect.objectContaining({
+                color_depth: expect.any(Number),
+                java_enabled: expect.any(Boolean),
+                language: expect.any(String),
+                screen_height: expect.any(Number),
+                screen_width: expect.any(Number),
+                time_zone: expect.any(String),
+                time_zone_offset: expect.any(String),
+            })
+        );
         expect(store.dispatch).toHaveBeenCalledWith(initializeOffsitePaymentAction);
     });
 


src/payment/strategies/barclaycard/barclaycard.ts
```diff
@@ -0,0 +1,13 @@
+export interface BrowserInfoRequest {
+    device_os: DeviceOS;
```
capsula4: I think the intention of the SDK is to keep all the fields using `camelCase`, however they should be converted to `snake_case`  in https://github.com/bigcommerce/bigpay-client-js/blob/7f163b5aa6c2ec4b18a0b12890cdde3a569c2118/src/payment/offsite-payment-mappers/payload-mapper.js#L67

src/payment/strategies/barclaycard/barclaycard-payment-strategy.ts
```diff
@@ -0,0 +1,83 @@
...
+            time_zone: new Date().toLocaleTimeString('en-us', {timeZoneName: 'short'}).split(' ')[2],
+        };
+    }
+
+    private _getDeviceOs(): DeviceOS {
+        let deviceOS: DeviceOS = DeviceOS.Other;
+        if (navigator.userAgent.indexOf('Win') !== -1) {
+            deviceOS = DeviceOS.Windows;
+        } else if (navigator.userAgent.indexOf('like Mac') !== -1) {
+            deviceOS = DeviceOS.IOS;
```
capsula4: should we differentiate between iOS and MacOS ?

src/payment/payment.ts
```diff
@@ -1,10 +1,12 @@
+import { BrowserInfoRequest as BarclaycardBrowserInfoRequest } from './strategies/barclaycard';
+
 export default interface Payment {
     methodId: string;
     gatewayId?: string;
     paymentData?: PaymentInstrument & PaymentInstrumentMeta;
 }
 
-export type PaymentInstrument = CreditCardInstrument | NonceInstrument | VaultedInstrument | CryptogramInstrument | HostedInstrument | ThreeDSVaultedInstrument | FormattedPayload<PaypalInstrument>;
+export type PaymentInstrument = CreditCardInstrument | NonceInstrument | VaultedInstrument | CryptogramInstrument | HostedInstrument | ThreeDSVaultedInstrument | FormattedPayload<PaypalInstrument | BarclaycardBrowserInfoRequest>;
```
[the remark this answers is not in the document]
FranciscoOnt: If I create the type with the field, I would need extra logic in the mapper specific for the  `browser_info` to maintain the convention of sending every field individually  and additional login in bigpay to grab every field from the object the mapper, I would prefer send every field inside the formattedPayload  than inside a nested field.

src/payment/strategies/barclaycard/barclaycard-payment-strategy.ts
FranciscoOnt: According to Barclaycard's docs, these are the possible values
Screen Shot 2019-11-19 at 2 06 59 PM

FranciscoOnt:
@capsula4  could you please re-review? ♻️

src/payment/strategies/barclaycard/barclaycard-payment-strategy.ts
```diff
@@ -0,0 +1,82 @@
+import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
+import { OrderActionCreator, /* OrderPaymentRequestBody,*/  OrderRequestBody } from '../../../order';
```
capsula4: maybe remove `OrderPaymentRequestBody` if not in used

src/payment/strategies/barclaycard/barclaycard-payment-strategy.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+        expect(orderActionCreator.finalizeOrder).toHaveBeenCalledWith(getOrder().orderId, options);
+        expect(store.dispatch).toHaveBeenCalledWith(finalizeOrderAction);
+    });
+
+    it('does not finalize order if order is not created', async () => {
+        const state = store.getState();
+
+        jest.spyOn(state.order, 'getOrder').mockReturnValue(getIncompleteOrder());
+
+        try {
```
capsula4: For exceptions, it's better to test by using `expect(() => ...).toThrow()` rather than try catch, but just a suggestion
```js
expect(() => strategy.finalize()).toThrow(OrderFinalizationNotRequiredError)
```

src/payment/payment.ts
```diff
@@ -80,6 +82,10 @@ export interface PaypalInstrument {
     };
 }
 
+export interface BayclardcardInstrument {
+    browserInfo: BrowserInfoRequest;
```
capsula4: just realised that the idea behind `formattedPayload` is that the payload is already formatted for bigpay https://github.com/bigcommerce/bigpay-client-js/pull/82, to avoid mapping the content in bigpay. So what you had before was actually aligned with what we started 🤦‍♂

FranciscoOnt:
I made the changes you suggested @capsula4 , could you please re-review? ♻️

src/payment/strategies/barclaycard/barclaycard-payment-strategy.ts
```diff
@@ -0,0 +1,82 @@
...
+
+        await this._store.dispatch(this._orderActionCreator.submitOrder(payload, options));
+
+        return await this._store.dispatch(this._paymentActionCreator.initializeOffsitePayment(
+            payment.methodId,
+            payment.gatewayId,
+            {
+                ...paymentData,
+                formattedPayload: {
+                    ...this._getBrowserInfo(),
```
capsula4: shouldn't this be `browser_info: this._getBrowserInfo()` ? I would suggest not to merge anything that you haven't tested it actually works. Might need to wait for bigpay PR to be merged first

FranciscoOnt: By doing this the fields in are sent one by one to bigpay instead of being sent as an object, also it's easier for bigpay to grab every field  individually, and that is how it's actually implemented.

There should not be a problem for waiting for bigpay, as even if the fields are sent, they would just be ignored if there is no logic in bigpay to handle them, as bigpay is not strongly-typed as the sdk.

capsula4: yeah I was saying that because I thought bigpay is expecting a `browser_info` object https://github.com/bigcommerce/bigpay/pull/2054/files#diff-3d02eaae8a0c8141fff8fd2a0344b915R32

TomA-R: Considering BigPay now _does_ have this info under the `browser_info` property, are we sure this is the correct format? Is it tested in the current state?

FranciscoOnt: The `browser_info` property on bigpay is used to send the info to the provider not to receive it so it should continue working as expected.

icatalina: If we end up keeping it like this, there's no need for the `...`:

```ts
formattedPayload: this._getBrowserInfo(),
```

icatalina:
I just realised, aren't these 2 PRs almost the same? https://github.com/bigcommerce/checkout-sdk-js/pull/726

I can see we are using a `browser_info` key within `formattedPayload` in the other PR. Could we please agree on a single format?

https://github.com/bigcommerce/checkout-sdk-js/pull/726/files#diff-a839c90c95b5c3aaa5ff88e50647438eR93-R100

```ts
formattedPayload: {
    bigpay_token: {
        credit_card_number_confirmation: paymentData.ccNumber,
        token: paymentData.instrumentId,
        verification_value: paymentData.ccCvv,
    },
    browser_info: this._getBrowserInfo(),
},
```

FranciscoOnt:
@icatalina , I decided to add the `browser_info` key to make it consistent with #726 ♻️

icatalina:
Since most of the data is the same between both PRs, could we try to standarize in checkout and figure out and compose what's needed by the payment provider in bigpay?

image

e.g.: Instead of sending resolution, just send `screen_heigh` and `screen_width` and make bigpay compose them? figure out device_os based on user agent, the same way we do in the UI, and I guess `time_zone` and `time_zone_offset` can also be combined... 

For use agent we might want to just read it from the headers as opposed to send it twice, one in the header and one in the body of the request... 🤔

src/payment/payment.ts
```diff
@@ -80,6 +82,10 @@ export interface PaypalInstrument {
     };
 }
 
+export interface BarclaycardInstrument {
+    browser_info: BarclaycardBrowserInfoRequest;
```
davidchin: Is there any reason why it needs to be tied with a specific payment method? It seems to me that it's just some generic information about the user's device. And some of the information the server can already derive from the user agent of a request so the client probably doesn't need to send them separately.

What I want to see is that the client sends the minimal amount of information that's not tied to a particular payment method. The server then receives the information and transforms it to the payload schema / format that the payment method expects. 

The problem with the current approach is that what happens if another payment method requires us to send browser information as well. But that one requires us to send it in a different format. But we can't because `browser_info` is tied with Barclaycard.

src/payment/strategies/barclaycard/barclaycard-payment-strategy.ts
```diff
@@ -0,0 +1,82 @@
...
+
+        await this._store.dispatch(this._orderActionCreator.submitOrder(payload, options));
+
+        return await this._store.dispatch(this._paymentActionCreator.initializeOffsitePayment(
+            payment.methodId,
+            payment.gatewayId,
+            {
+                ...paymentData,
+                formattedPayload: {
+                    browser_info: this._getBrowserInfo(),
```
davidchin: If we can define a standard contract for sending additional device information, we wouldn't need a specific strategy implementation for Barclaycard.

FranciscoOnt:
@davidchin @icatalina , I've changed the approach for this, but still not sure how to figure out `device_os` properly, could you please take a look? ♻️

src/common/browser-info/get-browser-info.ts
```diff
@@ -0,0 +1,27 @@
...
+        screen_height: screen.height,
+        screen_width: screen.width,
+        time_zone: new Date().toLocaleTimeString('en-us', {timeZoneName: 'short'}).split(' ')[2],
+        time_zone_offset: new Date().getTimezoneOffset().toString(),
+    };
+}
+
+function _getDeviceOs(): DeviceOS {
+    let deviceOS: DeviceOS = DeviceOS.Other;
+    if (navigator.userAgent.indexOf('Win') !== -1) {
```
icatalina: Since all we are doing here is parsing the `UserAgent`, should we just move all this logic to bigpay and figure out the device_os from there?

We can read the User Agent directly from the header in bigpay and work out what the value for device_os should be.

src/payment/payment.ts
```diff
@@ -80,6 +82,10 @@ export interface PaypalInstrument {
     };
 }
 
+export interface FormattedBrowserInfo {
```
icatalina: Is the idea that we can send just `FormattedBrowserInfo`? I would expect this is in addition to the rest of the payment information.

src/payment/strategies/offsite/offsite-payment-strategy.ts
```diff
@@ -32,7 +33,8 @@ export default class OffsitePaymentStrategy implements PaymentStrategy {
                     payment.methodId,
                     payment.gatewayId,
                     instrumentId,
-                    shouldSaveInstrument))
+                    shouldSaveInstrument,
+                    getBrowserInfo()))
```
icatalina: That should go in a new line:

```suggestion
                    getBrowserInfo()
                )));
```

src/payment/payment.ts
FranciscoOnt: Yes, as very few providers require this info to be sent, I believe it would be better to be able to send it only when necessary. Also if the we send the browser info outside the formattedPayload it would require the mappers to be changed as well to allow the browser info fields to be sent.

FranciscoOnt:
@icatalina thanks for the feedback, I've made the changes as suggested, could you please re-review? ♻️

src/common/browser-info/browser-info.ts
```diff
@@ -4,5 +4,6 @@ export default interface BrowserInfo {
     language: string;
     screen_height: number;
     screen_width: number;
+    time_zone?: string;
```
icatalina: I know `time_zone` is only required for a particular case, but since it is pretty cheap to send it, I would argue we should send it all the time.

In fact, assuming we can deduce `time_zone_offset` from `time_zone` we could just replace one with the other and when we need `time_zone_offset` just use `time_zone` in the backend to generate it.

src/payment/strategies/offsite/offsite-payment-strategy.ts
```diff
@@ -32,8 +33,12 @@ export default class OffsitePaymentStrategy implements PaymentStrategy {
                     payment.methodId,
                     payment.gatewayId,
                     instrumentId,
-                    shouldSaveInstrument))
-            );
+                    shouldSaveInstrument,
+                    {
+                        ...getBrowserInfo(),
+                        time_zone: new Date().toLocaleTimeString('en-us', {timeZoneName: 'short'}).split(' ')[2],
```
icatalina: Please, add `time_zone` to `getBrowserInfo`. There's no need to only generate this for this particular case.

Also, it seems like `timeZoneName` might not be supported in `ie`. Can you please check:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString

image

src/payment/payment.ts
```diff
@@ -111,6 +111,10 @@ interface AdyenV2Card {
     bigpay_token?: void;
 }
 
+export interface FormattedBrowserInfo {
+    browser_info: BrowserInfo | null;
```
icatalina: `browser_info` should not be `null`. If we have a type specifically to hold `browser_info` it should always contain it.

src/payment/strategies/offsite/offsite-payment-strategy.ts
```diff
@@ -32,8 +33,12 @@ export default class OffsitePaymentStrategy implements PaymentStrategy {
                     payment.methodId,
                     payment.gatewayId,
                     instrumentId,
-                    shouldSaveInstrument))
-            );
+                    shouldSaveInstrument,
+                    {
+                        ...getBrowserInfo(),
```
icatalina: There's no need for the execute method to know about `getBrowserInfo()`. It is always good to try and hide as much as possible the implementation details. Just send it for all `offsitePayments`.

src/payment/payment-action-creator.ts
```diff
@@ -39,17 +41,22 @@ export default class PaymentActionCreator {
         methodId: string,
         gatewayId?: string,
         instrumentId?: string,
-        shouldSaveInstrument?: boolean
+        shouldSaveInstrument?: boolean,
+        browserInfo?: BrowserInfo
     ): ThunkAction<InitializeOffsitePaymentAction, InternalCheckoutSelectors> {
         return store => {
-            let paymentData: FormattedPayload<FormattedHostedInstrument | FormattedVaultedInstrument> | undefined;
+            let paymentData: FormattedPayload<FormattedBrowserInfo | FormattedHostedInstrument | FormattedVaultedInstrument> | undefined;
```
icatalina: If I read this correctly, we are saying that you could never send `FormattedHostedInstrument` or `FormattedVaultedInstrument` together with `FormattedBrowserInfo` but your code seems to be merging them together.

cbandam:
This PR is not needed since the requirement was addressed from BigPay with this other PR https://github.com/bigcommerce/bigpay/pull/2054

PR #752, opened 2019-11-25
CHECKOUT-4203: Add factory for creating hosted payment forms

davidchin:
## What?
Add `HostedFormFactory` for creating hosted payment forms. It returns an instance of `HostedForm`, which you can use to collect and submit payment details within a set of cross-origin iframes.

## Why?
This is a part of the work required for the hosted payment form project. Please note that the code required for the iframes will come in the next PR.

## Testing / Proof
CircleCI

### DIFF
--- a/src/hosted-form/hosted-form-order-data-transformer.ts
+++ b/src/hosted-form/hosted-form-order-data-transformer.ts
@@ -1,0 +1,41 @@
+import { omit } from 'lodash';
+
+import { ReadableCheckoutStore } from '../checkout';
+import { MissingDataError, MissingDataErrorType } from '../common/error/errors';
+import { OrderPaymentRequestBody } from '../order';
+import { HostedCreditCardInstrument } from '../payment';
+
+import HostedFormOrderData from './hosted-form-order-data';
+
+export default class HostedFormOrderDataTransformer {
+    constructor(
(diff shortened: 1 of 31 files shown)


src/hosted-form/hosted-form-order-data-transformer.ts
```diff
@@ -0,0 +1,41 @@
...
+import { HostedCreditCardInstrument } from '../payment';
+
+import HostedFormOrderData from './hosted-form-order-data';
+
+export default class HostedFormOrderDataTransformer {
+    constructor(
+        private _store: ReadableCheckoutStore
+    ) {}
+
+    transform(payload: OrderPaymentRequestBody): HostedFormOrderData {
```
capsula4: I wonder if this is just a transformer, since it also takes the store as an argument, it's more like `HostedFormOrderDataBuilder`? 🍹

davidchin: I had a chat with Luis regarding this and agreed that naming it as `builder` might be confusing as the implementation doesn't really follow the builder pattern. In effect, the class is responsible for transforming one type of payload data object into another. So it seems `transformer` might the most appropriate. I can revisit this if I can find a better name for this in the future. For now, we're happy to proceed with what I have.

PR #753, opened 2019-11-25
CHECKOUT-4204: Format credit card field values

davidchin:
## What?
* Format credit card number and expiry date values.
* Originally the code lives in UCO. I've copied it across to this repo with some slight modifications.

## Why?
We want to format the values for better user experience. Also, we need to preserve the existing behaviour.

## Testing / Proof
CircleCI

### DIFF
--- a/src/hosted-form/iframe-content/card-number-formatter.ts
+++ b/src/hosted-form/iframe-content/card-number-formatter.ts
@@ -1,0 +1,37 @@
+import { number } from 'card-validator';
+import creditCardType from 'credit-card-type';
+import { max } from 'lodash';
+
+const NUMBER_SEPARATOR = ' ';
+
+export default class CardNumberFormatter {
+    format(value: string): string {
+        const { card } = number(value);
+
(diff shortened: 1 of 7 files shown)


src/hosted-form/iframe-content/card-number-formatter.ts
```diff
@@ -0,0 +1,39 @@
+import { number } from 'card-validator';
+import creditCardType from 'credit-card-type';
+import { max } from 'lodash';
+
+export default class CardNumberFormatter {
+    constructor(
+        private _separator: string = ' '
```
RobinOng: Just wondering why we need this in the constructor?

davidchin: I guess I don't really need it. The separator could just be a hard coded constant.

davidchin: I've updated the code and removed the parameter from the constructor.

PR #755, opened 2019-11-26
CHECKOUT-4203: Add text input element responsible for accepting user input inside iframe

davidchin:
## What?
Add `HostedInput` class, which is responsible for accepting user inputs inside an iframe.

## Why?
It is intended to be loaded and initialised inside a hosted field iframe. It also acts as a base class for specialised fields, such as card number and expiry field.

## Testing / Proof
CircleCI

### DIFF
--- a/src/hosted-form/iframe-content/hosted-input-validator.ts
+++ b/src/hosted-form/iframe-content/hosted-input-validator.ts
@@ -1,0 +1,77 @@
+import { cvv, expirationDate, number } from 'card-validator';
+import { pick } from 'lodash';
+import { object, string, StringSchema, ValidationError } from 'yup';
+
+import HostedInputValidateErrorData from './hosted-input-validate-error-data';
+import HostedInputValues from './hosted-input-values';
+
+interface HostedInputValidateOptions {
+    isCardCodeRequired?: boolean;
+}
+
+export default class HostedInputValidator {
+    async validate(values: HostedInputValues, options?: HostedInputValidateOptions): Promise<HostedInputValidateErrorData[]> {
+        const schemas: { [key in keyof HostedInputValues]: StringSchema } = {
+            cardExpiry: this._getCardExpirySchema(),
+            cardName: this._getCardNameSchema(),
+            cardNumber: this._getCardNumberSchema(),
+        };
+
+        if (options && options.isCardCodeRequired) {
+            schemas.cardCode = this._getCardCodeSchema();
+        }
+
+        try {
+            await object(pick(schemas, Object.keys(values)))
+                .validate(values, { abortEarly: false });
+
+            return [];
+        } catch (error) {
+            if (error.name !== 'ValidationError') {
+                throw error;
+            }
+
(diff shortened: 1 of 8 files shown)


src/hosted-form/iframe-content/hosted-input-validator.ts
```diff
@@ -0,0 +1,77 @@
...
+import HostedInputValidateErrorData from './hosted-input-validate-error-data';
+import HostedInputValues from './hosted-input-values';
+
+interface HostedInputValidateOptions {
+    isCardCodeRequired?: boolean;
+}
+
+export default class HostedInputValidator {
+    async validate(values: HostedInputValues, options?: HostedInputValidateOptions): Promise<HostedInputValidateErrorData[]> {
+        const config: { [key in keyof HostedInputValues]: StringSchema } = {
```
capsula4: maybe name it something like `inputSchemas` or `fieldSchemas`

src/hosted-form/iframe-content/hosted-input-validator.ts
```diff
@@ -0,0 +1,77 @@
...
+            cardNumber: this._getCardNumberSchema(),
+        };
+
+        if (options && options.isCardCodeRequired) {
+            config.cardCode = this._getCardCodeSchema();
+        }
+
+        try {
+            await object(pick(config, Object.keys(values)))
+                .validate(values, { abortEarly: false });
```
capsula4: could you use `validateSync` for this?, to avoid using await.

src/hosted-form/iframe-content/hosted-input.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    setValue(value: string): void {
+        this._processChange(value);
+    }
+
+    isTouched(): boolean {
+        return this._isTouched;
+    }
+
+    attach(): void {
```
capsula4: should the `containerId` be an argument of attach() instead of the constructor?

src/hosted-form/iframe-content/hosted-input-validator.ts
davidchin: I've renamed the variable to `schemas` for brevity.

src/hosted-form/iframe-content/hosted-input-validator.ts
davidchin: Prefer to keep it as `validate` if that's alright with you. Using `async/await` essentially makes the code read as if I'm using `validateSync`.

src/hosted-form/iframe-content/hosted-input.ts
davidchin: I guess it could. But I don't really need it as an argument for my use case. Right not it's in the constructor and can't be changed for the lifetime of the instance. Potentially we could move it to `attach` if there's a need for it in the future.

capsula4: yeah the main reason was so that the code is more self documented `attach(containerId: string): void`

PR #762, opened 2019-12-03
fix(spam-protection): CHECKOUT-4560 Fix spam protection not working for braintree in non-chrome browsers

RobinOng:
## What?
Spam protection is preventing braintree paypal to opening the modal.

**Note:** marked `CheckoutService#intializeSpamProtection` as deprecated since using this method will break Braintree Paypal

## Why?
Safari and FireFox have the limit where modal has to be opened with 1 second after user interaction, whereas Chrome has the limit of 5 seconds.

To fix it, we changed to flow slightly where we will validate the checkout immediately after `validateSpamProtection` is called, instead of when creating an order. This is prevent the scenario where the token expired and shopper have to complete the challenge multiple times.

## Testing / Proof
Added unit tests and will be testing with `checkout-js`.

Screen Shot 2019-12-04 at 5 41 06 pm
Screen Shot 2019-12-04 at 5 40 36 pm


src/spam-protection/spam-protection-request-sender.ts
```diff
@@ -0,0 +1,26 @@
...
+    ].join(','),
+};
+
+export default class SpamProtectionRequestSender {
+    constructor(
+        private _requestSender: RequestSender
+    ) {}
+
+    validate(checkoutId: string, token: string, { timeout }: RequestOptions = {}): Promise<Response<Checkout>> {
+        const url = `/api/storefront/checkouts/${checkoutId}/spam-protection`;
```
davidchin: Does the endpoint need to return a response body? Is it because the checkout object changes its state (i.e.: has an additional flag) after validation? Or could it just return a status code? If you do want to send back the checkout object, perhaps you don't need to ask for the additional includes?

src/spam-protection/spam-protection-action-creator.ts
```diff
@@ -0,0 +1,88 @@
...
+                            return throwError(createErrorAction(SpamProtectionActionType.SubmitFailed, error));
+                        }
+
+                        const checkout = state.checkout.getCheckout();
+
+                        if (!checkout) {
+                            throw new MissingDataError(MissingDataErrorType.MissingCheckout);
+                        }
+
+                        this._requestSender.validate(checkout.id, token);
```
davidchin: This looks like an asynchronous request. Should we wait for it to complete?

src/spam-protection/spam-protection-action-creator.ts
```diff
@@ -0,0 +1,88 @@
...
+                return empty();
+            }
+
+            return concat(
+                of(createAction(SpamProtectionActionType.ExecuteRequested, undefined)),
+                this._googleRecaptcha.execute()
+                    .pipe(take(1))
+                    .pipe(switchMap(({ error, token }) => {
+                        if (error || !token) {
+                            return throwError(createErrorAction(SpamProtectionActionType.SubmitFailed, error));
```
davidchin: You might want to just throw the error here. And catch it with a `catchError` operator at the end of the observable chain, and re-throw it as an error action. The reason is that there's another potential point of failure (`requestSender.validate`) that you might want to catch and throw the same error action for it.

src/spam-protection/spam-protection-action-creator.ts
```diff
@@ -0,0 +1,88 @@
...
+    ) {}
+
+    initialize(options?: SpamProtectionOptions): ThunkAction<SpamProtectionAction, InternalCheckoutSelectors> {
+        return store => Observable.create((observer: Observer<SpamProtectionAction>) => {
+            const spamProtectionElementId = options ? options.containerId : 'spamProtectionContainer';
+
+            if (!options && !document.getElementById(spamProtectionElementId)) {
+                const spamProtectionElement = document.createElement('div');
+                spamProtectionElement.setAttribute('id', spamProtectionElementId);
+                document.body.append(spamProtectionElement);
```
davidchin: I believe you want to use `appendChild` instead. AFAIK `append` is not supported in IE.

src/spam-protection/spam-protection-action-creator.ts
```diff
@@ -0,0 +1,88 @@
...
+            }
+
+            const state = store.getState();
+            const config = state.config.getConfig();
+
+            if (!config) {
+                throw new MissingDataError(MissingDataErrorType.MissingCheckoutConfig);
```
davidchin: Perhaps this observable action can be written using pipe and operators instead? You can keep it in this form but you'll need to find a way to catch this error and transform it into an error action.

src/spam-protection/spam-protection-action-creator.ts
RobinOng: Thanks!

src/spam-protection/spam-protection-request-sender.ts
RobinOng: Yup, the value for `shouldRunSpamProtectionCheck` will change in the checkout object after verification. I removed the includes.

src/spam-protection/spam-protection-action-creator.ts
RobinOng: Oops, I missed this.

src/spam-protection/spam-protection-action-creator.ts
```diff
@@ -0,0 +1,84 @@
...
+                    .pipe(take(1))
+                    .pipe(switchMap(({ error, token }) => {
+                        if (error || !token) {
+                            throw new SpamProtectionFailedError();
+                        }
+
+                        return this._requestSender.validate(checkout.id, token)
+                            .then(({ body }) => createAction(SpamProtectionActionType.ExecuteSucceeded, body));
+                    }))
+                    .pipe(
```
davidchin: Prefer for this `pipe` to append to the end of `concat` instead of this inner observable.

RobinOng: Thanks DC! Fixed.

PR #763, opened 2019-12-06
CHECKOUT-4203: Submit card details via hosted fields for stored instrument verification

davidchin:
## What?
Make all fields of a hosted form optional so it's possible to use it for stored instrument verification.

## Why?
Right now, we assume every hosted form must have a card number, name and expiry field. For stored instrument verification (required when shipping to an untrusted shipping address), we only need to show the card number or card code field to the shopper.

## Testing / Proof
CircleCI


davidchin:
Hrmmm I just realised stored instrument verification requires a different validation schema.

src/hosted-form/hosted-form.ts
```diff
@@ -46,25 +45,23 @@ export default class HostedForm {
...
-        return await this._getNumberField().submit(
+        return await this._getFirstField().submit(
             this._fields.map(field => field.getType()),
             this._payloadTransformer.transform(payload)
         );
     }
 
     async validate(): Promise<void> {
-        return await this._getNumberField().validate();
+        return await this._getFirstField().validate();
```
capsula4: why do we only validate the first field?

src/hosted-form/iframe-content/hosted-input-validator.ts
```diff
@@ -1,39 +1,60 @@
...
-            cardExpiry: this._getCardExpirySchema(),
-            cardName: this._getCardNameSchema(),
-            cardNumber: this._getCardNumberSchema(),
-        };
+    constructor(
+        private _cardInstrument?: CardInstrument
+    ) {}
 
+    async validate(values: HostedInputValues, fields?: HostedFieldType[]): Promise<HostedInputValidateResults> {
+        const requiredFields = fields || Object.keys(values);
```
capsula4: shouldn't you always be passing the fields? What's the scenario where you get the fields based on the values?

src/hosted-form/hosted-form.ts
davidchin: `validate` actually validates the entire form. The reason is because some fields are related - i.e.: CVV depends on the card number. I could probably rename it to `validateForm`.

src/hosted-form/iframe-content/hosted-input-validator.ts
davidchin: I guess I could make `fields` a required parameter, or infer the field types based on the `values`.

src/hosted-form/hosted-form.ts
capsula4: yeah I think validate is meant to validate the whole form, I was mostly asking how by validating the first field, we also validate other fields, by just looking at `this._getFirstField().validate();` it's hard to guess how it validates the whole form

src/hosted-form/iframe-content/hosted-input-validator.ts
capsula4: yeah what I was thinking is that if you rely on values only, you could potentially miss to pass a value and validate the schema based on the values itself. I feel it would be easier to either stick to always relying on values, or making `fields` mandatory

PR #764, opened 2019-12-09
CHECKOUT-4571 Add StepTracker service

capsula4:
## What?
Adds step tracker service. This has been extracted from https://github.com/bigcommerce/checkout-js/tree/master/src/app/analytics with the difference that the backfill is off by default, and a custom step order can be passed.

## Why?
So custom checkouts built using Checkout SDK can benefit from enhance e-commerce.

## Testing / Proof
unit

@bigcommerce/checkout

### DIFF
--- a/src/analytics/analytics-step-tracker.ts
+++ b/src/analytics/analytics-step-tracker.ts
@@ -1,0 +1,459 @@
+import { keys } from 'lodash';
+
+import { LineItemMap } from '../cart';
+import { Checkout, CheckoutService } from '../checkout';
+import { InvalidArgumentError } from '../common/error/errors';
+import { ShopperCurrency, StoreProfile } from '../config';
+import { Coupon } from '../coupon';
+import { Order } from '../order';
+import { ShippingOption } from '../shipping';
+
+import { AnalyticsTracker } from './analytics-tracker-window';
+import StepTracker from './step-tracker';
+
+export interface StepTrackerConfig {
+    checkoutSteps?: AnalyticStepType[];
+}
+
+export type AnalyticStepType = 'customer' | 'shipping' | 'billing' | 'payment';
+
+const ORDER_ITEMS_STORAGE_KEY = 'ORDER_ITEMS';
+
+export enum AnalyticStepId {
+    CUSTOMER = 1,
+    SHIPPING,
+    BILLING,
+    PAYMENT,
+}
+
+const ANALYTIC_STEPS: { [key: string]: AnalyticStepId } = {
+    customer: AnalyticStepId.CUSTOMER,
+    shipping: AnalyticStepId.SHIPPING,
+    billing: AnalyticStepId.BILLING,
+    payment: AnalyticStepId.PAYMENT,
+};
+
+export default class AnalyticsStepTracker implements StepTracker {
+    private _checkoutStarted: boolean = false;
+    private _completedSteps: { [key: string]: boolean } = {};
+    private _viewedSteps: { [key in AnalyticStepId]?: boolean; } = {};
+    private _analyticStepOrder: AnalyticStepType[] = [
+        'customer',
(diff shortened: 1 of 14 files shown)


src/analytics/create-step-tracker.ts
```diff
@@ -0,0 +1,55 @@
...
+import StepTracker from './step-tracker';
+
+/**
+ * Creates an instance of `StepTracker`.
+ *
+ * @remarks
+ * ```js
+ * const checkoutService = createCheckoutService();
+ * await checkoutService.loadCheckout();
+ * const stepTracker = createStepTracker(checkoutService, { shouldBackfillPreviousSteps: true });
```
capsula4: potentially the config options could only have `checkoutSteps: [ ... ]`, if not passed it will assume the default, if passed empty won't backfill.

src/analytics/analytics-step-tracker.ts
```diff
@@ -0,0 +1,450 @@
...
+    ];
+
+    constructor(
+        private checkoutService: CheckoutService,
+        private storage: StorageFallback,
+        private analytics: AnalyticsTracker,
+        { checkoutSteps }: StepTrackerConfig = {}
+    ) {
+        if (checkoutSteps !== undefined) {
+            this.analyticStepOrder = checkoutSteps;
```
davidchin: We might need a tighter check - to ensure that the consumer can only pass in the allowed steps. This is because right now it only has a compiler check, which is not sufficient for apps that are not using TS.

src/analytics/analytics-step-tracker.ts
```diff
@@ -0,0 +1,450 @@
...
+            cart: {
+                lineItems,
+                discountAmount,
+                id,
+            },
+        } = checkout;
+
+        const extraItemsData = this.saveExtraItemsData(id, lineItems);
+
+        this.analytics.track('Checkout Started', this.getTrackingPayload({
```
davidchin: Is it fine to send "Checkout Started" event again after a page refresh?

src/analytics/create-step-tracker.ts
```diff
@@ -0,0 +1,55 @@
...
+        throw new MissingDataError(MissingDataErrorType.MissingCheckoutConfig);
+    }
+
+    const { isAnalyticsEnabled } = config.checkoutSettings;
+
+    if (isAnalyticsEnabled && isAnalyticsTrackerWindow(window)) {
+        return new AnalyticsStepTracker(
+            checkoutService,
+            localStorageFallback,
+            window.analytics,
```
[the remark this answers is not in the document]
capsula4: That sounds good 👍

src/analytics/analytics-step-tracker.ts
capsula4: that's current behaviour and I don't think merchants are complaining. I'm concerned that if I make a change here it could break the existing implementation. For example, if checkout is reloaded (after modifying the cart) the saved information could go out of sync.

src/analytics/create-step-tracker.ts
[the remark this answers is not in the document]
capsula4: We automatically load segment when they enable analytics in CP. However it doesn't hurt to add a warning.

src/analytics/analytics-step-tracker.ts
davidchin: I see, that's a good point. I agree, it's better to keep it this way.

src/analytics/create-step-tracker.ts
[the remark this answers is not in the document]
davidchin: Yes I'm aware of that. I guess I might be overly cautious. What you have right now is fine. In terms of version incompatibility, it won't be an issue once I have the CDN version ready, which is very soon.

PR #766, opened 2019-12-10
refactor(provider): Use AdyenV2 custom card components for TSV and vaulting

vmparra:
## What? [INT-2181](https://jira.bigcommerce.com/browse/INT-2181)
Added support fort encrypted information for vaulted instruments and TSV on AdyenV2

## Why?
According to PCI compliance on Adyen, we need to also encrypt cvv and credit card number for TSV vaulted instruments

## Testing / Proof
image

Ping @bigcommerce/payments @bigcommerce/intersys-integrations

### DIFF
--- a/src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
+++ b/src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
@@ -194,13 +194,33 @@ describe('AdyenV2PaymentStrategy', () => {
         const identifyShopperError = getIdentifyShopperError();
         const challengeShopperError = getChallengeShopperError();
         const redirectShopperError = getRedirectShopperError();
-        let adyenComponent: AdyenComponent;
+        let adyenPaymentComponent: AdyenComponent;
+        let adyenCardVerificationComponent: AdyenComponent;
         let componentOptions: ThreeDS2DeviceFingerprintComponentOptions | ThreeDS2ChallengeComponentOptions;
         let options: PaymentInitializeOptions;
         let threeDS2Component: AdyenComponent;
 
         beforeEach(() => {
+            let handleOnChange: (state: AdyenCardState) => {};
             options = getAdyenInitializeOptions();
+
+            adyenPaymentComponent = {
+                mount: jest.fn(() => {
+                    handleOnChange(getValidCardState());
+
+                    return;
+                }),
+                unmount: jest.fn(),
+            };
+
+            adyenCardVerificationComponent = {
+                mount: jest.fn(() => {
+                    handleOnChange(getValidCardState());
+
+                    return;
+                }),
+                unmount: jest.fn(),
+            };
 
             threeDS2Component = {
                 mount: jest.fn(() => {
@@ -214,39 +234,8 @@ describe('AdyenV2PaymentStrategy', () => {
                 }),
                 unmount: jest.fn(),
             };
-            adyenComponent = {
-                mount: jest.fn(),
-                unmount: jest.fn(),
-            };
 
             jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod').mockReturnValue(getAdyenV2());
-        });
-
-        afterEach(() => {
-            jest.resetAllMocks();
-        });
-
-        it('submits order without payment data', async () => {
-            const payload = getOrderRequestBody();
-            const options = { methodId: 'adyenv2' };
-
-            await strategy.execute(payload, options);
-
-            expect(orderActionCreator.submitOrder).toHaveBeenCalledWith(omit(payload, 'payment'), options);
-            expect(store.dispatch).toHaveBeenCalledWith(submitOrderAction);
-        });
-
-        it('submits payment with adyen state', async () => {
-            let handleOnChange: (state: AdyenCardState) => {};
-
-            adyenComponent = {
-                mount: jest.fn(() => {
-                    handleOnChange(getValidCardState());
-
-                    return;
-                }),
-                unmount: jest.fn(),
-            };
 
             jest.spyOn(adyenV2ScriptLoader, 'load').mockReturnValue(Promise.resolve(adyenCheckout));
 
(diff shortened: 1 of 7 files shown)


src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -25,7 +25,7 @@ import { getErrorPaymentResponseBody, getVaultedInstrument } from '../../payment
 
 import { AdyenV2PaymentStrategy, AdyenV2ScriptLoader, ThreeDS2ChallengeComponentOptions, ThreeDS2ComponentType, ThreeDS2DeviceFingerprintComponentOptions } from '.';
 import { AdyenCardState, AdyenCheckout, AdyenComponent } from './adyenv2';
-import { getAdyenCheckout, getAdyenInitializeOptions, getChallengeShopperError, getIdentifyShopperError, getIdentifyShopperErrorResponse, getInvalidCardState, getRedirectShopperError, getValidCardState, getValidChallengeResponse } from './adyenv2.mock';
+import { getAdyenCheckout,  getAdyenInitializeOptions, getChallengeShopperError, getIdentifyShopperError, getIdentifyShopperErrorResponse, getInvalidCardState, getRedirectShopperError, getValidCardState, getValidChallengeResponse } from './adyenv2.mock';
```
clopezh: This space is not necessary 🍹

clopezh: This space is not necessary 🍹

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -71,6 +72,16 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
 
                 this._adyenComponent = adyenComponent;
 
+                const adyenCustomCardComponent = this._adyenCheckout.create('securedfields', {
+                    onChange: (state: AdyenCardState) => {
+                        this._updatePaymentMethodState(state);
+                    },
+                });
+
+                adyenCustomCardComponent.mount(`#${paymentMethod.id}-tsv`);
```
davidchin: The container ID needs to be passed in instead of being hardcoded here. The user of the SDK needs to be able to specify where the component can be inserted.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -181,12 +213,12 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
         return this._adyenv2;
     }
 
-    private _getStateContainer(): string {
-        if (!this._stateContainer) {
-            return '{}';
+    private _getPaymentMethodState(): AdyenCardState {
+        if (!this._paymentMethodState) {
+            throw new InvalidArgumentError('"stateContainer" argument was not provided during initialization.');
```
davidchin: This is not the reason why `_paymentMethodState` is undefined. There's no way for the user of the SDK to fix this problem because `stateContainer` is not an argument that they can provide during initialisation. It is just a private property of the class that you've defined.

I think I've said this before. Defining a getter like this is an anti-pattern because it prevents you from throwing an appropriate error. The error you throw depends on the context of the call. And sometimes, it's fine to not throw an error because it's in the expected state. You should avoid doing this in the future.

In this particular case, it seems you might not be able to proceed with submitting the payment if the `execute` method is called before the shopper has interacted with the payment component and selected a card (before the `onChange` is called). So at the beginning of the `execute` method, you can put a guard statement to stop the code execution unless `_paymentMethodState` is defined.

```ts
const paymentMethodState = this._paymentMethodState;

if (!paymentMethodState) {
}
```

src/payment/is-vaulted-instrument.ts
```diff
@@ -1,5 +1,11 @@
-import { PaymentInstrument, VaultedInstrument } from './payment';
+import { AdyenV2Instrument, FormattedPayload, PaymentInstrument, VaultedInstrument } from './payment';
 
 export default function isVaultedInstrument(instrument: PaymentInstrument): instrument is VaultedInstrument {
+    const formattedPayload = (instrument as FormattedPayload<AdyenV2Instrument>).formattedPayload;
```
davidchin: I don't think this is correct, because this type checker is to determine whether `instrument` is the type of `VaultedInstrument`. It looks like you want to check if `instrument` is the type of `AdyenV2Instrument`. As you can tell, `AdyenV2Instrument` and `VaultedInstrument` are two separate types. Even when you go with the suggestion I've made regarding the ditching the provider-specific interface in favour of a common interface, it will still be different to `VaultedInstrument`. What you should do instead to define a new type checking function.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -441,20 +402,10 @@ describe('AdyenV2PaymentStrategy', () => {
                 ThreeDS2ComponentType.ThreeDS2Challenge,
                 componentOptions
             );
-            expect(adyenCheckout.create).toHaveBeenCalledTimes(3);
+            expect(adyenCheckout.create).toHaveBeenCalledTimes(4);
```
davidchin: Not sure why you're asserting that a method has been called 4 times. Don't you think it's a bit random? I think you're testing a state change but have not reset the spy before the state change. If you do that, you can probably write a more precise assertion and make the test a lot easier to read.

vmparra: In fact we are calling (now) `create` method 4 times.
1.- Payment component (i.e. scheme, ideal, etc.)
2.- treeDS2Component (fingerprint)
3.- treeDS2Component (challenge)
4.- customFields component (this is the one been added)

vmparra:
@davidchin ♻️ Changes have been addressed, please take another look

PR #767, opened 2019-12-11
[WIP] fix(integrations): INT-2059 Lowercase validation for methodId

ric-sanchez:
## What? [INT-2059](https://jira.bigcommerce.com/browse/INT-2059)

Adding lowercase validation for methodId on Payment Method retrieval

## Why?

When checking out using Adyen as Gateway some of the Payment Methods comes in mixed Upper and Lower Case, as you can see in the example below:

image

So, the `getPaymentMethod` method transforms the selected `methodId` and all the payment method listed on the `paymentMethods` array to lowercase in order to make a proper comparison and retrieve the selected Payment Method data.

## Testing / Proof

int-2059

Ping @bigcommerce/checkout @bigcommerce/payments

### DIFF
--- a/src/checkout-buttons/strategies/paypal/paypal-button-strategy.spec.ts
+++ b/src/checkout-buttons/strategies/paypal/paypal-button-strategy.spec.ts
@@ -146,6 +146,7 @@ describe('PaypalButtonStrategy', () => {
 
             await strategy.initialize(options);
         } catch (error) {
+            console.log(error);
             expect(error).toBeInstanceOf(MissingDataError);
         }
     });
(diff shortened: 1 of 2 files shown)


icatalina:
Hi! I might be missing some context here. How come we need this? is there a mismatch between 2 apis? can't we fix the returned values?

ric-sanchez:
> Hi! I might be missing some context here. How come we need this? is there a mismatch between 2 apis? can't we fix the returned values?

I think the solution of transform to lowercase the order selected method id and the ids on the Adyend accepted Methods on the lodash `find` callback function works for us because we don't alter any of the values outside of the context of the `getPaymentMethod` function, and we are returning the info needed by the `finalize` function for the proper selected method.

ric-sanchez:
Closing PR: Changes will be addressed on BigPay

PR #773, opened 2020-01-08
CHECKOUT-4187: Build distribution files for static server

davidchin:
## What?
Build distribution files for a static server.

## Why?
We want to make the SDK available on CDN. In order to that, we'll need to set up a server and deploy the built files there. A separate private repository and build process have been set up to handle that part. Here, we just have to push the built files to the server repository so that it can handle the rest.

## Testing / Proof
CircleCI

### DIFF
--- a/webpack-server.config.js
+++ b/webpack-server.config.js
@@ -1,0 +1,79 @@
+const path = require('path');
+const semver = require('semver');
+const { DefinePlugin } = require('webpack');
+const WebpackAssetsManifest = require('webpack-assets-manifest');
+
+const { transformManifest } = require('./scripts/webpack');
+const { babelLoaderRule, baseConfig, libraryEntries, libraryName, srcPath } = require('./webpack-common.config');
+
+const version = require('./package.json').version;
+const versionDir = `v${semver.major(version)}`;
+const outputPath = path.join(__dirname, 'dist-server', versionDir);
+
+function getServerConfig() {
+    return {
+        ...baseConfig,
+        name: 'umd',
+        entry: libraryEntries,
+        output: {
+            filename: '[name]-[contenthash:8].js',
+            library: libraryName,
+            libraryTarget: 'umd',
+            path: outputPath,
+        },
+        module: {
+            rules: [
+                babelLoaderRule,
+                ...baseConfig.module.rules,
+            ],
+        },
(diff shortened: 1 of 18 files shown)


webpack.config.js
```diff
@@ -1,89 +1,55 @@
...
+                ...baseConfig.module.rules,
             ],
-        }),
-    }),
+        },
+    };
+}
 
-    Object.assign({}, assetConfig, {
-        name: 'cjs',
```
icatalina: We need to keep the names for the task as it is being use to run them in isolation: 

https://github.com/bigcommerce/checkout-sdk-js/blob/master/package.json#L27-L28

webpack-server.config.js
```diff
@@ -0,0 +1,95 @@
...
+            path: outputPath,
+        },
+        module: {
+            rules: [
+                babelLoaderRule,
+                ...baseConfig.module.rules,
+            ],
+        },
+        plugins: [
+            new AsyncHookPlugin({
```
icatalina: Since all the `AsyncHookPlugin` is doing is making these 2 task run sequentially it might be better to just use named configurations and actually run one after the other.

```sh
"build:cdn": "webpack webpack-server.config.js --config-name cdn",
"build:loader": "webpack webpack-server.config.js --config-name loader",
"build:container": "npm run build:cdn && npm run build:loader",
```

or

```sh
"build:cdn": "webpack webpack-server.config.js --config-name cdn && webpack webpack-server.config.js --config-name loader",
```

webpack-server.config.js
```diff
@@ -0,0 +1,79 @@
...
+                MANIFEST_JSON: JSON.stringify(require(path.join(outputPath, 'manifest.json'))),
+            }),
+        ],
+    };
+}
+
+// This configuration is for building distribution files for the static server
+// instead of the NPM package.
+function getConfigs(options, argv) {
+    if (argv.configName === 'umd-loader') {
```
davidchin: I know out of the box Webpack filters the config based on `configName`. But due to the fact that `umd-loader` config requires `manifest.json` to be present, it'd fail unless I don't try to create the config at all. So I think this solution makes sense - by default build the UMD files unless `config-name umd-loader` is passed.

PR #776, opened 2020-01-09
feat(payments): PAYPAL-140 Render paypal buttons with paypal sdk

helga1507:
## What?
Render Smart Payment Buttons on the cart page

## Why?
Change old paypal scripts to new implementation

## Testing / Proof
Screen Shot 2020-01-09 at 12 27 11 PM


BC-ACherednichenko:
Hello @davidchin, could you check this PR, maybe you have any thoughts or suggestions? We still have some work to do on it and we are not going to merge it right now, but it is very close to the final version, but we would like to check if you have suggestions for improvements, so we can be ready to merge in a few weeks


cc @helga1507

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-strategy.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+            clientId: 'abc',
+        };
+
+        options = {
+            containerId: 'paypalcommerce-container1',
+            methodId: CheckoutButtonMethodType.PAYPALCOMMERCE,
+            paypalCommerce: paypalOptions,
+        };
+
+        orderID = 'ORDER_ID';
```
BC-ACherednichenko: @helga1507 can we add TODO here to make sure we do not miss it

BC-ACherednichenko: please disregard my note above, we can leave it for testing purposes

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-strategy.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    constructor(
+        private _store: CheckoutStore,
+        private _checkoutActionCreator: CheckoutActionCreator,
+        private _paypalScriptLoader: PaypalCommerceScriptLoader,
+        private _formPoster: FormPoster,
+        private _requestSender: RequestSender
+    ) {}
+
+    initialize(options: CheckoutButtonInitializeOptions): Promise<void> {
```
davidchin: I'm happy for you to use `async/await` if you find it more readable. But if you're happy with this, I'm also good with it. 🍹

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-strategy.ts
```diff
@@ -0,0 +1,82 @@
...
+        const state = this._store.getState();
+        const paymentMethod = state.paymentMethods.getPaymentMethod(options.methodId);
```
davidchin: You can use `getPaymentMethodOrThrow` so you don't have to implement a null check for `paymentMethod`.

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-options.ts
```diff
@@ -0,0 +1,6 @@
+export interface PaypalCommerceButtonInitializeOptions {
+    /**
+     * The Client ID of the Paypal App
+     */
+    clientId: string;
```
davidchin: Do you require the caller to pass the client ID in? I thought it'd be a part of the `PaymentMethod` object. If so, you shouldn't need to pass it from the outside because the button strategy itself should have access to the `PaymentMethod` object and read the required field.

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-strategy.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    private _setupPayment(): Promise<string> {
+
+        return this._store.dispatch(this._checkoutActionCreator.loadDefaultCheckout())
+            .then(state => {
+                const url = '/api/storefront/payment/paypalcommerce';
+                const headers = {
+                    'X-API-INTERNAL': INTERNAL_USE_ONLY,
+                    'Content-Type': 'application/x-www-form-urlencoded',
```
davidchin: I'm curious why the content-type is not `application/json`? This looks like a regular storefront API so the content type should just be JSON.

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-strategy.ts
```diff
@@ -0,0 +1,82 @@
...
+                };
+                const cart = state.cart.getCart();
+                const currentCartId = cart ? cart.id : '';
+
+                return this._requestSender.post(url, {
+                    headers,
+                    body: {cartId: currentCartId},
```
davidchin: Missing spaces around the '{}'

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-strategy.ts
```diff
@@ -0,0 +1,82 @@
...
+                });
+            })
+            .then(res => res.body.orderId)
+            .catch(error => {
+                throw error;
```
davidchin: Unless you want to catch the error and do something with it (i.e.: transform it and handle it), you don't need to call `catch` and re-throw the error.

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-strategy.ts
```diff
@@ -0,0 +1,82 @@
...
+            });
+    }
+
+    private _tokenizePayment(paymentId: string, data: ApproveDataOptions) {
+        if (!data.orderID) {
```
davidchin: I assume this order ID is a PayPal reference (not our order ID because at this point order is not created yet)?

src/payment/strategies/paypalCommerce/paypalCommerce-script-loader.ts
```diff
@@ -0,0 +1,27 @@
+import { ScriptLoader } from '@bigcommerce/script-loader';
+
+import { PaymentMethodClientUnavailableError } from '../../errors';
+
+import { PaypalCommerceHostWindow, PaypalCommerceSDK } from './paypalCommerce-sdk';
+
+export default class PaypalCommerceScriptLoader {
```
davidchin: Can we have kebab case for file and folder names? 💅

src/payment/strategies/paypalCommerce/paypalCommerce-script-loader.ts
```diff
@@ -0,0 +1,27 @@
...
+    private _window: PaypalCommerceHostWindow;
+
+    constructor(
+        private _scriptLoader: ScriptLoader
+    ) {
+        this._window = window;
+    }
+
+    async loadPaypalCommerce(clientId: string = '', currency: string = 'USD'): Promise<PaypalCommerceSDK> {
```
davidchin: I think it might be better to not set a default value for both parameters. Things would not work properly without a client ID. And it's not correct to assume that the default currency is USD. If there are no default values to the method, the compiler can help us by warning us when we forget to pass one of these required arguments in.

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-strategy.ts
helga1507: Thank you!
Yes, it's more

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-strategy.ts
helga1507: Thank you!

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-strategy.ts
helga1507: Thank you

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-strategy.ts
helga1507: Okey, thank you

src/payment/strategies/paypalCommerce/paypalCommerce-script-loader.ts
helga1507: Sure!

src/payment/strategies/paypalCommerce/paypalCommerce-script-loader.ts
helga1507: Thank you!
I agree with you. It will be better

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-options.ts
helga1507: Thank you!
Actually, I'm not sure. I did it like 'PayPal Express' and I think it needs only for paypal commerce buttons. And it has sent during initialization buttons
But I think it's not part of PaymentMethod

davidchin: Right I see, indeed it's not getting returned. I wonder if we can add that field to our API so we don't have to pass it as a template variable. I know for Amazon we're passing `clientId` under `initializationData`. But it's not ideal because `initializationData` is a schema-less field. Perhaps we can make it a top-level field instead? cc @BC-ACherednichenko

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-strategy.ts
helga1507: Yes, but this api works for paypal express too. And request with application/json doesn't work

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-strategy.ts
helga1507: Yes, it's paypal reference

helga1507: Maybe could be better to use "MissingDataErrorType.MissingPayment" ?

src/checkout-buttons/strategies/paypalCommerce/paypalCommerce-button-options.ts
BC-ACherednichenko: @davidchin, yep, it looks much better to pass `client-id` through `PaymentMethod`, we will check it out

src/payment/strategies/paypal-commerce/paypal-commerce-script-loader.ts
```diff
@@ -0,0 +1,32 @@
...
+    ) {
+        this._window = window;
+    }
+
+    async loadPaypalCommerce(clientId: string, currency: string = 'USD'): Promise<PaypalCommerceSDK> {
+        if (!clientId) {
+            throw new InvalidArgumentError();
+        }
+
+        const scriptSrc = `https://www.paypal.com/sdk/js?currency=${currency}&client-id=${clientId}`;
```
BC-ACherednichenko: @helga1507 can we also pass locale https://developer.paypal.com/docs/checkout/reference/customize-sdk/#locale, to the script params, 
I think it would be better to have an object with parameters, instead of arguments (because there could be a lot arguments)

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -0,0 +1,75 @@
...
+    initialize(options: CheckoutButtonInitializeOptions): Promise<void> {
+        const state = this._store.getState();
+        const paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(options.methodId);
+        const paypalOptions = options.paypalCommerce;
+
+        if (!paypalOptions || !paypalOptions.clientId) {
+            throw new InvalidArgumentError();
+        }
+
+        return this._paypalScriptLoader.loadPaypalCommerce(paypalOptions.clientId)
```
BC-ACherednichenko: @helga1507 should not we pass currency from the quote?

src/payment/strategies/paypal-commerce/paypal-commerce-script-loader.ts
BC-ACherednichenko: @helga1507 , I am creating separate tickets for additional params that need to be passed, it would be out of the scope of this PR
I think we should add an object that would have all properties for script initialization for future

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private _setupPayment(): Promise<string> {
+
+        return this._store.dispatch(this._checkoutActionCreator.loadDefaultCheckout())
+            .then(state => {
+                const url = '/api/storefront/payment/paypalcommerce';
+                const headers = {
+                    'X-API-INTERNAL': INTERNAL_USE_ONLY,
+                    'Content-Type': 'application/x-www-form-urlencoded',
```
davidchin: I'm still not sure why we need to post the content type as `application/x-www-form-urlencoded`. The body should be passed in as a `JsonBody` as an argument for the controller action. @BC-ACherednichenko can you take a look at how other post/put actions are written?

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
helga1507: Should. I added it 
Thank you!

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
I removed the doc file and canceled the changes in package-lock.json

BC-ACherednichenko: @davidchin, this part of the functionality is consistent with PayPal express `api/storefront/payment/paypalexpress`, we decided to use this route `api/storefront/payment/{storefront}` instead of creating a new one, because and it does the same thing as PayPalCommerce does, once PayPal express eliminated we can rebuild this controller to accept JSON. 
 This route, in general, accepts data in form-urlencoded format, but in comparison with PayPal Commerce implementation, it sends a request differently 

 https://github.com/bigcommerce/checkout-sdk-js/blob/8044df22b714020444a523e96ee96b7c89b010b8/src/checkout-buttons/strategies/paypal/paypal-button-strategy.ts#L89

```js
        return this._store.dispatch(this._checkoutActionCreator.loadDefaultCheckout())
            .then(state => {
                const cart = state.cart.getCart();
                const cartId = cart ? cart.id : '';

                return actions.request.post(`${this._host}/api/storefront/payment/paypalexpress`, { merchantId, cartId }, {
                    headers: {
                        'X-API-INTERNAL': INTERNAL_USE_ONLY,
                    },
                });
            })
```

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
```diff
@@ -0,0 +1,25 @@
...
+    };
+}
+
+export interface PaypalCommerceHostWindow extends Window {
+    paypal?: PaypalCommerceSDK;
+}
+
+export interface PaypalCommerceScriptOptions {
+    'client-id': string;
+    currency: string;
```
BC-ACherednichenko: `currency `

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -27,35 +28,45 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
...
+                const config = state.config.getStoreConfig();
+                const cart = state.cart.getCart();
+
+                if (!config) {
+                    throw new MissingDataError(MissingDataErrorType.MissingCheckoutConfig);
+                }
+
+                const paramsScript = {
+                    'client-id': paypalOptions.clientId,
+                    currency: config.currency.code,
```
davidchin: Do you want to pass the store's default currency code or the currency code of the cart?

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -27,35 +28,45 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
...
-        return this._store.dispatch(this._checkoutActionCreator.loadDefaultCheckout())
-            .then(state => {
-                const url = '/api/storefront/payment/paypalcommerce';
-                const headers = {
-                    'X-API-INTERNAL': INTERNAL_USE_ONLY,
-                    'Content-Type': 'application/x-www-form-urlencoded',
-                };
-                const cart = state.cart.getCart();
-                const currentCartId = cart ? cart.id : '';
+    private _setupPayment(cart?: Cart): Promise<string> {
```
davidchin: Can `cart` ever be optional? If it's not defined, it means the data hasn't be loaded for some reason and we should throw an error to stop the code execution (similar to what you're doing with the missing config check).

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
```diff
@@ -18,3 +18,8 @@ export interface PaypalCommerceSDK {
 export interface PaypalCommerceHostWindow extends Window {
     paypal?: PaypalCommerceSDK;
 }
+
+export interface PaypalCommerceScriptOptions {
+    'client-id': string;
+    'currency': string;
```
davidchin: Probably don't need the quotation if the key doesn't contain `-`.

src/payment/strategies/paypal-commerce/paypal-commerce-script-loader.ts
```diff
@@ -14,12 +14,15 @@ export default class PaypalCommerceScriptLoader {
...
     }
 
-    async loadPaypalCommerce(clientId: string, currency: string = 'USD'): Promise<PaypalCommerceSDK> {
-        if (!clientId) {
+    async loadPaypalCommerce(options: PaypalCommerceScriptOptions): Promise<PaypalCommerceSDK> {
+        if (!options || !options['client-id']) {
             throw new InvalidArgumentError();
         }
+        const params = (Object.keys(options) as Array<keyof PaypalCommerceScriptOptions>)
+            .map(key => `${key}=${options[key]}`)
```
davidchin: I'm thinking instead of making an exception and use a hyphenated key `client-id`, what if you make the key camel case and map it into a hyphenated key. That way we can keep the casing of the keys consistent.

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
BC-ACherednichenko: @davidchin , we are planning to pass transactional currency (the one that later will be pushed to PayPal API)

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
helga1507: You're right. I added check to cart
Thank you

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
davidchin: I'm a little bit confused. 🤔If PayPal express doesn't require a different content-type and it uses the same route / controller, why does PayPal Commerce require something different?

src/payment/strategies/paypal-commerce/paypal-commerce-script-loader.ts
helga1507: Okay. Thank you!

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
BC-ACherednichenko: per our conversation
in PayPal commerce we send a request with `RequestSender` - our own library,   in PayPal express it being sent thru actions object that passed from PayPal side on SDK side, it has different params
https://github.com/bigcommerce/checkout-sdk-js/blob/71980ef5ec64693851a1b8856ca559df091c12ea/src/payment/strategies/paypal/paypal-sdk.ts#L56

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
BC-ACherednichenko: @davidchin , agree, I have double-checked on it - for the case when transactional currency differs from display one, in my case display CAD -> transactional USD, in cart.currency we have the correct one - transactional USD
image
@helga1507  - can you please apply these changes

helga1507: @BC-ACherednichenko Sure

PR #777, opened 2020-01-09
CHECKOUT-4205: Reload checkout page if for some reason form expires or invalidates after initial page load

davidchin:
## What?
Reload the checkout page if, for some reason, the form expires or invalidates after the initial page load.

## Why?
This error scenario can potentially happen. This is a measure to prevent it from happening. By reloading the page, the shopper can generate a new payment form and recover from the error.

I've also added in a check to make sure the shopper won't get into a redirect loop - just in case if there's anything unexpected that prevents the recovery attempt from working.

## Testing / Proof
CircleCI

### DIFF
--- a/src/hosted-form/hosted-field.spec.ts
+++ b/src/hosted-form/hosted-field.spec.ts
@@ -1,7 +1,8 @@
 import { IframeEventListener, IframeEventPoster } from '../common/iframe';
+import { BrowserStorage } from '../common/storage';
 
 import { InvalidHostedFormConfigError, InvalidHostedFormError } from './errors';
-import HostedField from './hosted-field';
+import HostedField, { LAST_RETRY_KEY } from './hosted-field';
 import { HostedFieldEvent, HostedFieldEventType } from './hosted-field-events';
 import HostedFieldType from './hosted-field-type';
 import { getHostedFormOrderData } from './hosted-form-order-data.mock';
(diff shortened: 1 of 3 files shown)


src/hosted-form/hosted-field.spec.ts
```diff
@@ -12,11 +13,15 @@ describe('HostedField', () => {
...
     let eventListener: Pick<IframeEventListener<HostedInputEventMap>, 'listen'>;
+    let location: Pick<Location, 'replace'>;
+    let storage: Pick<BrowserStorage, 'getItem' | 'setItem'>;
 
     beforeEach(() => {
         container = document.createElement('div');
         eventPoster = { post: jest.fn(), setTarget: jest.fn() };
         eventListener = { listen: jest.fn() };
+        location = { replace: jest.fn() };
+        storage = new BrowserStorage('EmbeddedCheckout');
```
capsula4: `new BrowserStorage('HostedFields');` 🍹

davidchin: Oops, my bad

src/hosted-form/hosted-form-factory.ts
```diff
@@ -11,6 +12,8 @@ import HostedForm from './hosted-form';
 import HostedFormOptions, { HostedCardFieldOptionsMap, HostedStoredCardFieldOptionsMap } from './hosted-form-options';
 import HostedFormOrderDataTransformer from './hosted-form-order-data-transformer';
 
+const STORAGE_PREFIX = 'BigCommerce.HostedField';
```
capsula4: `const STORAGE_NAMESPACE` ? 🍹 or maybe even avoid creating the const?

PR #779, opened 2020-01-13
fix(payment): PAYMENTS-4997 After deleting a PP account from checkout,…

Tharaae:
… all accounts with same email are deleted but not removed from available vaulted accounts list unless page is refreshed

## What?
After an account instrument is deleted from checkout manage modal, the available vaulted accounts when returning from instruments manage modal are updated to fetch the new vaulted accounts list from backend, rather than filtering it on frontend.

## Why?
Deleting a vaulted account actually deletes all vaulted accounts with the same email address on DB. This should be reflected properly on UI. 

Before this PR, If a PP account is vaulted twice as two instruments for the same customer on the same store, when one of those instruments is deleted, the other instrument for the same PP account is also deleted on DB. However, after returning from Manage modal to checkout page, the user finds the other instrument of the same account in the list of available vaulted accounts while it's actually deleted. The correct list is not fetched unless the page in refreshed.

## Testing / Proof
Unit Tests
Screen records

**Before**
Screen Recording 2020-01-14 at 10 46 19 am

**After**
Screen Recording 2020-01-14 at 10 13 24 am

### DIFF
--- a/src/checkout/checkout-service.spec.ts
+++ b/src/checkout/checkout-service.spec.ts
@@ -996,17 +996,21 @@ describe('CheckoutService', () => {
     describe('#deleteInstrument()', () => {
         it('deletes an instrument', async () => {
             const instrumentId = '456';
-            const action = of(createAction('DELETE_INSTRUMENT'));
-
+            const deleteAction = of(createAction('DELETE_INSTRUMENT'));
             jest.spyOn(instrumentActionCreator, 'deleteInstrument')
-                .mockReturnValue(action);
+                .mockReturnValue(deleteAction);
 
             jest.spyOn(store, 'dispatch');
+
+            const loadAction = of(createAction('LOAD_INSTRUMENTS'));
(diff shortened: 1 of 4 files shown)


icatalina:
I feel a bit weird about moving this logic to the UI. Shouldn't we just retrieve the list again from the BE and update the UI based on that?

Tharaae:
> I feel a bit weird about moving this logic to the UI. Shouldn't we just retrieve the list again from the BE and update the UI based on that?

The logic is already on the UI; it's just not complete. The deleted account is filtered out from the list in the reducer (on the UI). What is done in this PR is just updating the filter to return the correct list.

Tharaae:
@icatalina Changed the implementation to re-fetch from backend rather than filtering out on UI.
A `TODO` thing is to get BigPay "delete vaulted instrument" function to return the updated list after deletion, so we need only one backend call instead of two.

PR #784, opened 2020-01-21
feat(payment): INT-2286 Use credit_card as payment method instead of card

FranciscoOnt:
## What? [INT-2286](https://jira.bigcommerce.com/browse/INT-2286)
Rename supported payment instrument to `credit_card`.

## Why?
To match with the new naming convention, so the stored credit cards are able to be  displayed and used.

## SIbling PRs
https://github.com/bigcommerce/bigpay/pull/2181
https://github.com/bigcommerce/bigcommerce/pull/33239

## Testing / Proof
Screen Shot 2020-01-21 at 10 53 56 AM

### DIFF
--- a/src/payment/instrument/supported-payment-instruments.ts
+++ b/src/payment/instrument/supported-payment-instruments.ts
@@ -49,7 +49,7 @@ const supportedInstruments: SupportedInstruments = {
         provider: 'paymetric',
         method: 'credit_card',
     },
-    'barclaycard.card': {
+    'barclaycard.credit_card': {
         provider: 'barclaycard',
         method: 'credit_card',
     },


src/payment/instrument/supported-payment-instruments.ts
```diff
@@ -49,7 +49,7 @@ const supportedInstruments: SupportedInstruments = {
         provider: 'paymetric',
         method: 'credit_card',
     },
-    'barclaycard.card': {
+    'barclaycard.credit_card': {
```
leeBigCommerce: Why does 'barclaycard.credit_card' include the method type when all others are only the provider?

Also, will there be any use cases of  'barclaycard.card' which will subsequently stop working after this change?

FranciscoOnt: Barclaycard is a provider with multiple APMs (Alternate Payment Methods), given this scenario, the backend always maps the provider as the `gateway` for every method, then when the `gateway` field is present the sdk concatenates `gateway +  method` to get the key that is used to search in this file.

file `src/payment/instrument/instrument-selector.ts`:
``` ts
    const paymentMethodKey = paymentMethod.gateway ? `${paymentMethod.gateway}.${paymentMethod.id}` : paymentMethod.id;
    const currentMethod = supportedInstruments[paymentMethodKey];
```

The backend changes needed to rename `card` to `credit_card` in the backend have already been merged, so the cases where 'barclaycard.card' is used will no longer be happening.

PR #788, opened 2020-02-02
CHECKOUT-4655: Fix SagePay by making it compatible with hosted payment forms

davidchin:
## What?
* Fix SagePay by making it compatible with hosted payment forms.
* This is done by:
  * Making `SagePayPaymentStrategy` inherit from `CreditCardPaymentStrategy`. It should inherit because SagePay is a type of credit card payment method. It has additional code because it has a special 3DS flow.
  * And add support for 13 digit PAN by configuring `card-validator`. We need to support it because SagePay's test credit card is in that format.

## Why?
So that shoppers can pay with SagePay using a hosted payment form.

## Testing / Proof
Unit


src/common/types/card-validator.d.ts
```diff
@@ -0,0 +1,37 @@
+import 'card-validator';
+
+// Merge @types/card-validator with missing methods
```
davidchin: We need this because `@types/card-validator` is out of date and doesn't have the type definitions that match with the implementation.

src/hosted-form/iframe-content/card-number-formatter.ts
```diff
@@ -1,5 +1,4 @@
-import { number } from 'card-validator';
-import creditCardType from 'credit-card-type';
```
davidchin: Don't need to pull in a separate package because 'card-validator' already has `creditCardType` (it is the same package).

src/common/types/card-validator.d.ts
capsula4: maybe add a `todo: remove this once @types/card-validator` is up to date with implementation?

PR #790, opened 2020-02-06
CHECKOUT-4655: Make Cardinal 3DS work with hosted payment form

davidchin:
## What?
Make Cardinal 3DS flow for Cybersource and PayPal Pro work with hosted payment form.

## Why?
Currently, we're including CC details in `Cardinal.continue` call - which is a call to prompt the shopper to complete the authentication required by the issuer of their card. But because we're using the Hybrid integration approach, it seems we don't have to do that because we are already passing the order and payment details to Cardinal when we make a lookup request to check the enrolment of the card. However, we do need to pass the BIN number to Songbird (Cardinal's client library) so that it can collect the device data required for the card issuer.

Based on the above, I had to change the code so that credit card details can be sent within the hosted payment iframes, while allowing Songbird work as intended in the host page (which is collecting device data and prompting authentication).

## Testing / Proof
Unit / Functional

### DIFF
--- a/src/hosted-form/iframe-content/hosted-input-aggregator.spec.ts
+++ b/src/hosted-form/iframe-content/hosted-input-aggregator.spec.ts
@@ -54,6 +54,17 @@ describe('HostedInputAggregator', () => {
             .toEqual(frames.map(frame => frame.hostedInput));
     });
 
+    it('does not throw error if there are other iframes in parent window that belong to different origin than itself', () => {
+        frames.push({
+            get hostedInput() {
+                throw new DOMException();
+            },
+        } as unknown as HostedInputWindow);
+
+        expect(aggregator.getInputs())
+            .toEqual(frames.slice(0, -1).map(frame => frame.hostedInput));
+    });
+
     it('gathers all adjacent hosted inputs that satisfy filter', () => {
         expect(aggregator.getInputs(field => includes([HostedFieldType.CardCode, HostedFieldType.CardExpiry], field.getType())))
             .toEqual([frames[0].hostedInput, frames[1].hostedInput]);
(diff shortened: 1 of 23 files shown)


davidchin:
cardinal-cybersource

src/payment/strategies/cardinal/cardinal-three-d-secure-flow.ts
```diff
@@ -1,150 +1,100 @@
...
 
-    start(payment: Payment): Promise<InternalCheckoutSelectors> {
-        if (!payment.paymentData) {
-            throw new MissingDataError(MissingDataErrorType.MissingPayment);
-        }
-
-        if (!isCreditCardLike(payment.paymentData) && !isVaultedInstrument(payment.paymentData)) {
-            throw new InvalidArgumentError();
+    async start(
+        execute: PaymentStrategy['execute'],
```
capsula4: 👏

src/payment/strategies/cybersource/cybersource-payment-strategy.ts
```diff
@@ -1,63 +1,53 @@
...
-
-        if (!payment) {
-            throw new MissingDataError(MissingDataErrorType.MissingPayment);
+    async execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        const { payment: { methodId = '' } = {} } = payload;
+        const { paymentMethods: { getPaymentMethodOrThrow } } = this._store.getState();
+
+        if (getPaymentMethodOrThrow(methodId).config.is3dsEnabled) {
+            return this._threeDSecureFlow.start(
+                super.execute.bind(this),
```
capsula4: 👌

src/hosted-form/iframe-content/hosted-input-aggregator.spec.ts
```diff
@@ -54,6 +54,17 @@ describe('HostedInputAggregator', () => {
             .toEqual(frames.map(frame => frame.hostedInput));
     });
 
+    it('ignores other irrelevant frames', () => {
```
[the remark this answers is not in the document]
davidchin: Good point. It happens when you're trying to access another frame that does belong to the same origin as the caller.

PR #792, opened 2020-02-10
CHECKOUT-4669: Create version-specific loader file and reference it in hosted payment form

davidchin:
## What?
* Create a version-specific loader file and reference it in hosted payment form.
* Bonus: Added a command for releasing alpha versions

## Why?
So that the version used in UCO can be aligned with the version used in the payment form. 

## Testing / Proof
CircleCI

### DIFF
--- a/scripts/circleci/update-server.sh
+++ b/scripts/circleci/update-server.sh
@@ -12,8 +12,7 @@ mkdir -p /tmp/repo-server/public
 
 # Replace the existing dist files with new ones
 RELEASE_VERSION_DIR=$(echo $RELEASE_VERSION | cut -d . -f 1)
-rm -rf /tmp/repo-server/public/$RELEASE_VERSION_DIR
-cp -r dist-server/$RELEASE_VERSION_DIR /tmp/repo-server/public
+cp -rf dist-server/$RELEASE_VERSION_DIR /tmp/repo-server/public
 
 # Commit the new dist files and push to upstream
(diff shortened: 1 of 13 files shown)


webpack-common.config.js
```diff
@@ -12,31 +15,39 @@ const libraryEntries = {
...
-            {
-                test: /\.[tj]s$/,
-                include: srcPath,
-                loader: 'ts-loader',
-            },
+            ],
+        },
+        plugins: [
+            new DefinePlugin({
+                'process.env.NODE_ENV': 'process.env.NODE_ENV',
```
icatalina: You need to keep this in the cjs

scripts/circleci/update-server.sh
```diff
@@ -12,8 +12,8 @@ mkdir -p /tmp/repo-server/public
 
 # Replace the existing dist files with new ones
 RELEASE_VERSION_DIR=$(echo $RELEASE_VERSION | cut -d . -f 1)
-rm -rf /tmp/repo-server/public/$RELEASE_VERSION_DIR
-cp -r dist-server/$RELEASE_VERSION_DIR /tmp/repo-server/public
+cp -rf dist-server/$RELEASE_VERSION_DIR /tmp/repo-server/public
+cp -rf dist-server/$RELEASE_VERSION /tmp/repo-server/public
```
icatalina: Are you meant to keep both?

PR #794, opened 2020-02-10
CHECKOUT-4640 Add support for marketing emails consent

capsula4:
## What?
Add support for marketing emails consent for guest shoppers

## Why?
GDPR compliance

## Testing / Proof
unit

@bigcommerce/checkout

### DIFF
--- a/src/billing/billing-address-action-creator.ts
+++ b/src/billing/billing-address-action-creator.ts
@@ -1,27 +1,30 @@
 import { createAction, createErrorAction, ThunkAction } from '@bigcommerce/data-store';
 import { Response } from '@bigcommerce/request-sender';
-import { Observable, Observer } from 'rxjs';
+import { concat, defer, empty, merge, of, Observable, Observer } from 'rxjs';
+import { catchError } from 'rxjs/operators';
 
 import { Checkout, InternalCheckoutSelectors } from '../checkout';
+import { throwErrorAction } from '../common/error';
 import { MissingDataError, MissingDataErrorType } from '../common/error/errors';
 import { RequestOptions } from '../common/http-request';
-import { GuestCredentials } from '../customer';
+import { CustomerAction, CustomerActionType, CustomerRequestSender, GuestCredentials } from '../customer';
 
 import { BillingAddressRequestSender } from '.';
 import { BillingAddressUpdateRequestBody } from './billing-address';
 import { BillingAddressActionType, ContinueAsGuestAction, UpdateBillingAddressAction } from './billing-address-actions';
-import { UnableToContinueAsGuestError } from './errors';
+import { UnableToContinueAsGuestError, UpdateCustomerError } from './errors';
 
 export default class BillingAddressActionCreator {
     constructor(
-        private _requestSender: BillingAddressRequestSender
+        private _requestSender: BillingAddressRequestSender,
+        private _customerRequestSender: CustomerRequestSender
     ) {}
 
     continueAsGuest(
         credentials: GuestCredentials,
         options?: RequestOptions
-    ): ThunkAction<ContinueAsGuestAction, InternalCheckoutSelectors> {
-        return store => Observable.create((observer: Observer<ContinueAsGuestAction>) => {
+    ): ThunkAction<ContinueAsGuestAction | CustomerAction, InternalCheckoutSelectors> {
+        return store => {
             const state = store.getState();
             const checkout = state.checkout.getCheckout();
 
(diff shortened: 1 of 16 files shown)


src/billing/billing-address-action-creator.ts
```diff
@@ -37,7 +37,7 @@ export default class BillingAddressActionCreator {
 
             const billingAddress = state.billingAddress.getBillingAddress();
 
-            let billingAddressRequestBody;
+            let billingAddressRequestBody: Partial<BillingAddressUpdateRequestBody> & GuestCredentials;
```
davidchin: This does look a little a bit odd. Are we planning to update the `marketingConsent` flag using the billing address endpoint?

capsula4: I'm waiting to see what happens in the API first. I agree it feels a bit odd. I would say what is odd is that we have modelled the guest shopper email in the cart as the billing address email.

It makes sense to send the consent to receive emails in the same payload the email is being sent. If the email in billing address changes (and no consent is being sent), the consent should be revoked. You can see the discussion here https://github.com/bigcommerce/bigcommerce/pull/33506

davidchin: Ok cool, I'll hold off approving this PR until the discussion is finalised. But besides that, the changes look good.

capsula4:
@davidchin I've updated the implementation accordingly to what we have agreed with Valentin, let me know what you think

src/billing/billing-address-action-creator.ts
```diff
@@ -104,6 +114,33 @@ export default class BillingAddressActionCreator {
...
     }
 
+    private _updateCustomerConsent(
+        {
+            email,
+            marketingEmailConsent,
+        }: GuestCredentials,
+        options?: RequestOptions
+    ): Observable<CustomerAction> {
+        if (!marketingEmailConsent) {
```
capsula4: there's no use case to require opting out. The opt out is most commonly offered via an unsubscribe link in the first email received. Same as we do we newsletter subscriptions in storefront.

src/billing/billing-address-action-creator.ts
```diff
@@ -50,17 +53,24 @@ export default class BillingAddressActionCreator {
...
+                    of(createAction(BillingAddressActionType.ContinueAsGuestRequested)),
+                    defer(async () => {
+                        const { body } = await this._createOrUpdateBillingAddress(
+                            checkout.id,
+                            billingAddressRequestBody,
+                            options
+                        );
+
+                        return createAction(BillingAddressActionType.ContinueAsGuestSucceeded, body);
+                    }).pipe(
```
davidchin: I'd probably move this `pipe` to the end of `concat` because your intention is to catch any error that could happen within this `concat` block. Having said that, there's only one observable that could potentially fail - that is the `defer`. So it really makes no difference at the moment. But if that's the case, you can just simplify this by using a regular `try/catch` block within the `defer`. 🍹

src/billing/billing-address-action-creator.ts
```diff
@@ -104,6 +114,35 @@ export default class BillingAddressActionCreator {
...
+        return concat(
+            of(createAction(CustomerActionType.UpdateCustomerRequested)),
+            defer(async () => {
+                const { body } = await this._customerRequestSender.updateCustomer({
+                    email,
+                    acceptsMarketing: marketingEmailConsent,
+                }, options);
+
+                return createAction(CustomerActionType.UpdateCustomerSucceeded, body);
+            }).pipe(
```
davidchin: Same comment as above

src/billing/billing-address-action-creator.ts
```diff
@@ -104,6 +114,35 @@ export default class BillingAddressActionCreator {
...
     }
 
+    private _updateCustomerConsent(
+        {
+            email,
+            marketingEmailConsent,
+        }: GuestCredentials,
+        options?: RequestOptions
+    ): Observable<CustomerAction> {
+        if (marketingEmailConsent === undefined) {
```
davidchin: To be safe, maybe check for both `undefined` and `null`.

src/customer/customer-request-sender.ts
```diff
@@ -10,6 +11,13 @@ export default class CustomerRequestSender {
         private _requestSender: RequestSender
     ) {}
 
+    updateCustomer(customer: CurrentCustomer, { timeout }: RequestOptions = {}): Promise<Response<CurrentCustomer>> {
+        const url = `/api/storefront/customer`;
```
davidchin: Just use regular `''` instead of back ticks if there's no string interpolation. 🍹 💅

PR #797, opened 2020-02-21
refactor(payment): INT-2350 Use createFromAction for every payment method

vmparra:
## What?
...

## Why?
...

## Testing / Proof
image

### DIFF
--- a/src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
+++ b/src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
@@ -227,7 +227,26 @@ describe('AdyenV2PaymentStrategy', () => {
                 await strategy.initialize(options);
                 await strategy.execute(getOrderRequestBodyWithVaultedInstrument());
 
-                expect(paymentActionCreator.submitPayment).toHaveBeenCalledWith(expect.objectContaining({methodId: 'scheme', paymentData: {formattedPayload: {bigpay_token : {credit_card_number_confirmation: 'ENCRYPTED_CARD_NUMBER', token: '123', verification_value: 'ENCRYPTED_CVV'}, browser_info: {color_depth: 24, java_enabled: false,  language: 'en-US', screen_height: 0, screen_width: 0, time_zone_offset: expect.anything()}}}}));
+                expect(paymentActionCreator.submitPayment).toHaveBeenCalledWith(expect.objectContaining({
+                    methodId: 'scheme',
+                    paymentData: {
+                        formattedPayload: {
+                            bigpay_token : {
+                                credit_card_number_confirmation: 'ENCRYPTED_CARD_NUMBER',
+                                token: '123',
+                                verification_value: 'ENCRYPTED_CVV',
+                            },
+                            browser_info: {
+                                color_depth: 24,
+                                java_enabled: false,
+                                language: 'en-US',
+                                screen_height: 0,
+                                screen_width: 0,
+                                time_zone_offset: expect.anything(),
+                            },
+                        },
+                    },
+                }));
                 expect(adyenCheckout.create).toHaveBeenCalledTimes(2);
             });
 
--- a/src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
+++ b/src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
@@ -1,34 +1,29 @@
 import { createClient as createPaymentClient } from '@bigcommerce/bigpay-client';
 import { createAction, createErrorAction } from '@bigcommerce/data-store';
-import { createFormPoster, FormPoster } from '@bigcommerce/form-poster';
 import { createRequestSender } from '@bigcommerce/request-sender';
 import { createScriptLoader, createStylesheetLoader } from '@bigcommerce/script-loader';
-import { omit } from 'lodash';
 import { of, Observable } from 'rxjs';
 
 import { createCheckoutStore, CheckoutRequestSender, CheckoutStore, CheckoutValidator } from '../../../checkout';
 import { getCheckoutStoreState } from '../../../checkout/checkouts.mock';
 import { InvalidArgumentError, MissingDataError, NotInitializedError, RequestError } from '../../../common/error/errors';
-import { getResponse } from '../../../common/http-request/responses.mock';
 import { FinalizeOrderAction, OrderActionCreator, OrderActionType, OrderRequestSender, SubmitOrderAction } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
 import { PaymentInitializeOptions } from '../../../payment';
-import { PaymentArgumentInvalidError } from '../../errors';
+import { PaymentArgumentInvalidError, PaymentMethodCancelledError } from '../../errors';
 import PaymentActionCreator from '../../payment-action-creator';
 import { PaymentActionType, SubmitPaymentAction } from '../../payment-actions';
 import { getAdyenV2 } from '../../payment-methods.mock';
 import PaymentRequestSender from '../../payment-request-sender';
 import PaymentRequestTransformer from '../../payment-request-transformer';
-import { getErrorPaymentResponseBody, getVaultedInstrument } from '../../payments.mock';
 
-import { AdyenComponentType, AdyenPaymentMethodType, AdyenV2PaymentStrategy, AdyenV2ScriptLoader, ThreeDS2ChallengeComponentOptions, ThreeDS2DeviceFingerprintComponentOptions } from '.';
-import { AdyenCardState, AdyenCheckout, AdyenComponent } from './adyenv2';
-import { getAdyenCheckout, getAdyenInitializeOptions, getBankTransferError, getChallengeShopperError, getIdentifyShopperError, getIdentifyShopperErrorResponse, getInvalidCardState, getOrderRequestBody, getRedirectShopperError, getValidCardState, getValidChallengeResponse } from './adyenv2.mock';
+import { AdyenAdditionalActionState, AdyenComponentState, AdyenError, AdyenPaymentMethodType, AdyenV2PaymentStrategy, AdyenV2ScriptLoader, ResultCode } from '.';
+import { AdyenCheckout, AdyenComponent } from './adyenv2';
+import { getAdditionalActionError, getAdyenCheckout, getAdyenError, getComponentState, getInitializeOptions, getInitializeOptionsWithNoCallbacks, getInitializeOptionsWithUndefinedWidgetSize, getOrderRequestBody, getOrderRequestBodyWithoutPayment, getOrderRequestBodyWithVaultedInstrument, getUnknownError } from './adyenv2.mock';
 
 describe('AdyenV2PaymentStrategy', () => {
     let finalizeOrderAction: Observable<FinalizeOrderAction>;
     let adyenV2ScriptLoader: AdyenV2ScriptLoader;
-    let formPoster: FormPoster;
     let orderActionCreator: OrderActionCreator;
     let paymentActionCreator: PaymentActionCreator;
     let store: CheckoutStore;
@@ -55,7 +50,6 @@ describe('AdyenV2PaymentStrategy', () => {
 
         adyenV2ScriptLoader = new AdyenV2ScriptLoader(scriptLoader, stylesheetLoader);
 
-        formPoster = createFormPoster();
         store = createCheckoutStore(getCheckoutStoreState());
 
         finalizeOrderAction = of(createAction(OrderActionType.FinalizeOrderRequested));
@@ -63,9 +57,6 @@ describe('AdyenV2PaymentStrategy', () => {
         submitPaymentAction = of(createAction(PaymentActionType.SubmitPaymentRequested));
 
         jest.spyOn(store, 'dispatch');
-
-        jest.spyOn(formPoster, 'postForm')
-            .mockImplementation((_url, _data, callback = () => {}) => callback());
 
         jest.spyOn(orderActionCreator, 'finalizeOrder')
             .mockReturnValue(finalizeOrderAction);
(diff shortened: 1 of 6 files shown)


ValeriaRivas:
💚 Tested
QE Details:

Attempt to | Result to Verify | Result
-- | -- | --
pay with a 3DS1 card (AdyenV2) | Second step authentication is triggered and successful checkout | Pass
Pay with a 3DS2 card (Adyen V2) | Second step authentication is triggered and successful checkout | Pass
pay with a 3DS1 card (Elavon)| Second step authentication is triggered and successful checkout | Pass
pay with a 3DS1 card (Sagepay) | Second step authentication is triggered and successful checkout | Pass

src/payment/strategies/adyenv2/adyenv2-initialize-options.ts
```diff
@@ -30,12 +30,12 @@ export default interface AdyenV2PaymentInitializeOptions {
...
      * Use additionalActionOptions instead as this property will be removed in the future
      */
-    threeDS2Options: AdyenThreeDS2Options;
+    threeDS2Options: ThreeDS2Options;
 
     /**
      * A set of options that are required to initialize additional payment actions.
      */
-    additionalActionOptions: AdyenAdditionalActionOptions;
+    additionalActionOptions: AdditionalActionOptions;
```
davidchin: Can you please keep the `Adyen` prefix? This is because all the interfaces eventually get exported under the same namespace. So it's better to keep the prefix to avoid possible name collision?

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -22,14 +21,13 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
     private _adyenv2?: AdyenV2PaymentInitializeOptions;
     private _adyenPaymentComponent?: AdyenComponent;
     private _adyenCardVerificationComponent?: AdyenComponent;
-    private _adyenComponentState?: AdyenCardState;
+    private _componentState?: ComponentState;
```
davidchin: You can remove the `adyen` prefix for private variables, properties, methods etc... But for exported interface names, as I said earlier, you should keep the prefix.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -143,66 +141,37 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
-                                            paymentData: {
-                                                ...payment.paymentData,
-                                                shouldSaveInstrument,
-                                            },
-                                        }))
-                                    );
-                            }
-
+                return this._handleFromAction(error.body.provider_data)
+                    .then((payment: Payment) =>
```
davidchin: You can drop the `Payment` type declaration because it's already implied by the return value of `_handleFromAction`

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -143,66 +141,37 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
+                        this._store.dispatch(this._paymentActionCreator.submitPayment({
+                            ...payment,
+                            paymentData: {
+                                ...payment.paymentData,
+                                shouldSaveInstrument,
+                            },
+                        }))
+                    )
+                    .catch(error => {
```
davidchin: You might want to extract this catch block into a private method, it seems to be repeated one after another.

```ts
.catch(error => {
    if (!(error instanceof RequestError) || !some(error.body.errors, { code: 'additional_action_required' })) {
        return Promise.reject(error);
    }

    return this._handleFromAction(error.body.provider_data)
        .then((payment: Payment) =>
            this._store.dispatch(this._paymentActionCreator.submitPayment({
                ...payment,
                paymentData: {
                    ...payment.paymentData,
                    shouldSaveInstrument,
                },
            }))
        );
})
```

A second question is, do we really need to have two `catch` blocks? As in, is it possible for a provider to ask for two additional actions?

vmparra: At least there's one case where two additional actions can be fired and it's when [Challenge](https://docs.adyen.com/checkout/3d-secure/native-3ds2/web-component) flow can be preceded by a frictionless flow.

vmparra:
@davidchin Changes have been addressed, please take another look ♻️

src/payment/strategies/adyenv2/adyenv2-initialize-options.ts
```diff
@@ -30,7 +30,7 @@ export default interface AdyenV2PaymentInitializeOptions {
      * @deprecated
      * Use additionalActionOptions instead as this property will be removed in the future
      */
-    threeDS2Options: AdyenThreeDS2Options;
+    threeDS2Options: ThreeDS2Options;
```
capsula4: as David's suggested, can you keep the prefix `Adyen` for this? Specially because it's deprecated and some clients could be consuming this type, we might be better off keeping this as it was, until we can remove all deprecated stuff.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -242,131 +193,79 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
+            AdyenPaymentMethodType.Sofort,
+        ];
 
-            if (onLoad) {
-                onLoad(() => {
-                    fingerprintComponent.unmount();
-                });
-            }
-        });
+        return !excludedPaymentMethods.find(x => paymentMethodName === x);
```
capsula4: prefer `name` over `x`

```js
return !excludedPaymentMethods.find(name => paymentMethodName === name);
```

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -242,131 +193,79 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
-            if (onBeforeLoad) {
-                onBeforeLoad();
-            }
+                }))
+            );
+    }
 
-            fingerprintComponent.mount(`#${additionalActionOptions.containerId || threeDS2ContainerId}`);
+    private _shouldMountPaymentComponent(paymentMethodName: string): boolean {
+        const excludedPaymentMethods = [
```
capsula4: What's the relationship between `_shouldMountPaymentComponent` and `excludedPaymentMethods` ? should the list be called something like `unmountablePaymentMethods` ?

vmparra: Purpose for that list is to _not_ mount components of those payment method types.

vmparra:
@capsula4  Changes have been addressed, please take another look ♻️

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -142,83 +153,23 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
-                return new Promise(() => {
-                    this._formPoster.postForm(error.body.three_ds_result.acs_url, {
-                        PaReq: error.body.three_ds_result.payer_auth_request,
-                        TermUrl: error.body.three_ds_result.callback_url,
-                        MD: error.body.three_ds_result.merchant_data,
-                    });
-                });
-            });
+            .catch(error => this._processPayment(error, shouldSaveInstrument))
+            .catch(error => this._processPayment(error, shouldSaveInstrument));
```
capsula4: is not very obvious that if something fails (catching the error) then you proceed to process the payment. Also is not obvious why you are calling it twice. Might be worth trying to make the code a bit more self explanatory

```js
.catch(error => {
  if (error.type == 'additional_action_required') {
    this._processAdditionalActionRequired(error, shouldSaveInstrument));
  }
  throw error;
})`
```

If the idea behind is to call itself as long as there is a `additional_action_required`, maybe considering something recursive instead of hard coding it twice?

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -100,14 +111,14 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
 
-                if (!adyenComponentState) {
+                if (!componentState) {
                     throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
                 }
-                if (paymentData && isVaultedInstrument(paymentData)) {
 
-                    const { encryptedCardNumber, encryptedSecurityCode } = adyenComponentState.data.paymentMethod;
+                if (paymentData && isVaultedInstrument(paymentData)) {
+                    const { encryptedCardNumber, encryptedSecurityCode } = (componentState as CardState).data.paymentMethod;
```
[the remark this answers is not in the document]
vmparra: Yes, we do, this it is required now that we handle multiple component states (CardState, WechatState, etc.).

vmparra:
@capsula4 Changes have been addressed, please take another look ♻️

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -242,131 +162,111 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
-            fingerprintComponent.mount(`#${additionalActionOptions.containerId || threeDS2ContainerId}`);
+        const payment = await this._handleAction(error.body.provider_data);
 
-            if (onLoad) {
-                onLoad(() => {
-                    fingerprintComponent.unmount();
-                });
-            }
-        });
+        return await this._store.dispatch(this._paymentActionCreator.submitPayment({
```
capsula4: usually when you use await you don't use `.catch()` but instead `try { } catch() { }`
```ja
try {
  return await this._store.dispatch(this._paymentActionCreator.submitPayment({
} catch (error) {
  return this._processAdditionalAction(error, shouldSaveInstrument
}
```

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -242,318 +116,337 @@ describe('AdyenV2PaymentStrategy', () => {
...
-                .toThrow(PaymentArgumentInvalidError);
-        });
+                await strategy.initialize(options);
+
+                expect(adyenCheckout.create).toHaveBeenCalledTimes(1);
+            });
 
-        it('throws error if order is missing', async () => {
-            const state = store.getState();
+            it('Initializes AliPay payment method', async () => {
```
capsula4: It's better to have the test title linked to what you are asserting, e.g.:

```js
it('does not call adyenCheckout.create when initializing AliPay')
```

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -242,318 +116,337 @@ describe('AdyenV2PaymentStrategy', () => {
...
-                    });
+            it('throws an error when payment is not present', async () => {
+                await strategy.initialize(options);
 
-                    return;
-                }),
-                unmount: jest.fn(),
-            };
+                try {
+                    await strategy.execute(getOrderRequestBodyWithoutPayment());
```
capsula4: it's better use `expect(() => ...).toThrow(PaymentArgumentInvalidError)`

The reason is, if accidentally a bug is introduced and strategy.execute stops throwing an exception, the assertion in  the "catch" block will never be executed, and the unit test won't catch the introduced bug.

```js
expect(async( ) => await strategy.execute(getOrderRequestBodyWithoutPayment()))
  .toThrowError(PaymentArgumentInvalidError));
```

another less common way is to manually call fail, but I would discourage this as it's not the pattern that we use.
```js
try {
  // code that will throw an exception
  fail();
} catch() {
 // expect..
}
```

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -242,318 +116,337 @@ describe('AdyenV2PaymentStrategy', () => {
...
+                } catch (error) {
+                    expect(error).toBeInstanceOf(PaymentArgumentInvalidError);
+                }
 
-            jest.spyOn(adyenV2ScriptLoader, 'load').mockReturnValue(Promise.resolve(adyenCheckout));
+                expect(adyenCheckout.create).toHaveBeenCalledTimes(2);
```
capsula4: is not obvious as why you are expecting this to be called twice, mainly because the title of the test doesn't mention anything around it.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -242,318 +116,337 @@ describe('AdyenV2PaymentStrategy', () => {
...
-        it('returns 3DS2 IdentifyShopper flow and responses ChallengeShopper flow', async () => {
-            const threeDS2ChallengeComponent: AdyenComponent = {
-                mount: jest.fn(() => {
-                    componentOptions.onComplete({
-                        data: {
-                            threeDS2Token: 'challengeToken',
-                        },
-                    });
+                try {
+                    await strategy.execute(getOrderRequestBody());
```
capsula4: as before, prefer `expect(() => ...).toThrowError()`

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -242,318 +116,337 @@ describe('AdyenV2PaymentStrategy', () => {
...
+                }
 
-            jest.spyOn(adyenV2ScriptLoader, 'load').mockReturnValue(Promise.resolve(adyenCheckout));
+                expect(adyenCheckout.create).toHaveBeenCalledTimes(2);
+            });
 
-            jest.spyOn(paymentActionCreator, 'submitPayment')
-                .mockReturnValueOnce(of(createErrorAction(PaymentActionType.SubmitPaymentFailed, identifyShopperError)))
-                .mockReturnValueOnce(submitPaymentAction);
+            it('returns UNKNOWN_ERROR', async () => {
```
capsula4:
```js
it('returns UNKNOWN_ERROR when submitPayment fails')
```

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -242,318 +116,337 @@ describe('AdyenV2PaymentStrategy', () => {
...
-                unmount: jest.fn(),
-            };
+                expect(paymentActionCreator.submitPayment).toHaveBeenCalledTimes(1);
+                expect(adyenCheckout.create).toHaveBeenCalledTimes(2);
+            });
 
-            jest.spyOn(adyenCheckout, 'create')
-                .mockImplementationOnce(jest.fn((_type, options) => {
-                    componentOptions = options;
+            it('tries to pay with invalid component state', async () => {
```
capsula4:
```
it('does not submit payment when trying to pay with invalid component state')
```

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -242,318 +116,337 @@ describe('AdyenV2PaymentStrategy', () => {
...
 
-                    return threeDS2ChallengeComponent;
-                }));
+                await strategy.initialize(options);
 
-            jest.spyOn(adyenV2ScriptLoader, 'load').mockReturnValue(Promise.resolve(adyenCheckout));
+                try {
+                    await strategy.execute(getOrderRequestBody());
+                } catch (error) {
+                    expect(error).toBeInstanceOf(NotInitializedError);
```
capsula4: I don't think this belongs here. I don't think this is being executed.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -242,318 +116,337 @@ describe('AdyenV2PaymentStrategy', () => {
...
 
-        it('returns 3DS2 IdentifyShopper flow and Adyen response onError', async () => {
-            additionalActionComponent = {
-                mount: jest.fn(() => {
-                    componentOptions.onError({
-                        errorCode: 'errorCode',
-                        message: 'errorMessage',
-                    });
+            it('throws an error when payment is not present', async () => {
+                await strategy.initialize(options);
```
capsula4: it seems that you are calling `await strategy.initialize(options);` in every single test. Might it be worth moving this to `beforeEach` block of `#execute`?

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -242,318 +116,337 @@ describe('AdyenV2PaymentStrategy', () => {
...
-            await strategy.execute(getOrderRequestBody());
-
-            expect(paymentActionCreator.submitPayment).toHaveBeenCalledTimes(3);
-            expect(paymentActionCreator.submitPayment).toHaveBeenLastCalledWith({
-                methodId: 'scheme',
-                paymentData: {
-                    nonce: JSON.stringify({
-                        threeDS2Token: 'challengeToken',
-                        paymentData: 'paymentData',
+            it('pays with vaulted instrument', async () => {
```
capsula4:
```js
it('calls submitPayment when paying with vaulted instrument', async () => {
```

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -242,318 +116,337 @@ describe('AdyenV2PaymentStrategy', () => {
...
+                jest.spyOn(paymentActionCreator, 'submitPayment')
+                    .mockReturnValueOnce(submitPaymentAction);
+
+                await strategy.initialize(options);
+                await strategy.execute(getOrderRequestBodyWithVaultedInstrument());
+
+                expect(paymentActionCreator.submitPayment).toHaveBeenCalledTimes(1);
```
capsula4: not required, but it's recommended to add some expectations around the arguments
e.g.
```hs
expect(paymentActionCreator.submitPayment).toHaveBeenCalledWith(expect.objectMatching({
  paymentData: ....,
}));
```

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
vmparra: it does, it test

```javascript
                if (!componentState) {
                }

File: src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts [hunk 21]
```diff
@@ -242,318 +116,337 @@ describe('AdyenV2PaymentStrategy', () => {
...
+                expect(adyenCheckout.create).toHaveBeenCalledTimes(2);
+            });
+
+            it('additional action component fires back onError', async () => {
```
capsula4: it('calls adyenCheckout.createFromAction when payment fails with identifyShopperError', async () => {

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -242,318 +116,337 @@ describe('AdyenV2PaymentStrategy', () => {
...
-                },
-            };
-            const executeOptions = { methodId: 'adyenv2' };
+                expect(adyenCheckout.create).toHaveBeenCalledTimes(2);
+                expect(adyenCheckout.createFromAction).toHaveBeenCalledTimes(1);
+            });
 
-            await strategy.initialize(options);
-            await strategy.execute(payload, executeOptions);
+            it('Initiates ChallengeShopper but user cancels', async () => {
```
capsula4:
```js
it('unmounts additional component when payment fails with challengeShopperError')
```

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
vmparra: _Some_ negative tests are using different  initialization `options`  i.e.

- ` options = getInitializeOptionsWithNoCallbacks();` 

- `options = getInitializeOptionsWithUndefinedWidgetSize();`

vmparra:
@capsula4 Changes have been addressed, please take another look ♻️

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -242,318 +116,322 @@ describe('AdyenV2PaymentStrategy', () => {
...
+                jest.spyOn(paymentActionCreator, 'submitPayment')
+                    .mockReturnValueOnce(submitPaymentAction);
 
-                    return threeDS2ChallengeComponent;
-                }));
+                await strategy.initialize(options);
+                await strategy.execute(getOrderRequestBodyWithVaultedInstrument());
 
-            jest.spyOn(adyenV2ScriptLoader, 'load').mockReturnValue(Promise.resolve(adyenCheckout));
+                expect(paymentActionCreator.submitPayment).toHaveBeenCalledWith(expect.objectContaining({methodId: 'scheme', paymentData: {formattedPayload: {bigpay_token : {credit_card_number_confirmation: 'ENCRYPTED_CARD_NUMBER', token: '123', verification_value: 'ENCRYPTED_CVV'}, browser_info: {color_depth: 24, java_enabled: false,  language: 'en-US', screen_height: 0, screen_width: 0, time_zone_offset: expect.anything()}}}}));
```
capsula4: would be good if you can format this so it's easier to be read

PR #799, opened 2020-02-21
feat(payments): PAYPAL-200 Create order request to bcapp 

helga1507:
## What?
Create request for creation order in bcapp
Create request to bigpay for creation payment 

## Why?
Payment order with PayPal Commerce

## Testing / Proof
Screen Shot 2020-02-20 at 6 58 00 PM

### DIFF
--- a/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
+++ b/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
@@ -32,11 +32,11 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
                 const config = state.config.getStoreConfig();
                 const cart = state.cart.getCart();
 
-                if (!config || !config.currency) {
+                if (!config) {
                     throw new MissingDataError(MissingDataErrorType.MissingCheckoutConfig);
                 }
 
-                if (!cart || !cart.id) {
+                if (!cart) {
                     throw new MissingDataError(MissingDataErrorType.MissingCart);
                 }
 
--- a/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
+++ b/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
@@ -32,11 +32,11 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
                 const config = state.config.getStoreConfig();
                 const cart = state.cart.getCart();
 
-                if (!config) {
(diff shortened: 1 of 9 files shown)


src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -32,11 +32,11 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
                 const config = state.config.getStoreConfig();
                 const cart = state.cart.getCart();
 
-                if (!config) {
+                if (!config || !config.currency) {
```
davidchin: I think `currency` is not an optional property so it's okay for you to not do a null check here.

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -32,11 +32,11 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
...
                     throw new MissingDataError(MissingDataErrorType.MissingCheckoutConfig);
                 }
 
-                if (!cart) {
+                if (!cart || !cart.id) {
```
davidchin: Same here, `cart.id` is not defined as an optional so you can also omit the null check here.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -0,0 +1,55 @@
...
+            formattedPayload: {
+                vault_payment_instrument: null,
+                device_info: null,
+                paypal_account: {
+                    order_id: paymentMethod.initializationData.orderId,
+                },
+            },
+        };
+
+        return this._store.dispatch(this._orderActionCreator.submitOrder(order, options))
```
davidchin: Prefer `async/await` for new code. 🍹 

PS: I know in some part of our code we use `async/await`, and in others, we use the regular `then` promise chain. The reason is simply because the former is newer code, whereas the latter is older code.

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
helga1507: Okay. Thank you!

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
helga1507: Thank You!

Issue #701, opened 2019-09-11
updateShippingAddress() does not have Custom Items in lineItems payload


LiHsuanChen:
When a custom item is added to cart, it doesn't show up in the lineItems payload when updateShippingAddress() is called. This is causing error when I only have custom items in cart and want to call updateShippingAddress();

capsula4:
Hi Terry, thanks for reporting this bug.

We have created a ticket in our internal board so a backend engineer can take a look at this. We will update the ticket when we have more information.

Greets,
Luis

kyle-bolt:
Is there an update on this request?

capsula4:
Sorry for the delay in tackling this. We thought there was no support to do this from the Storefront API. However this issue only required a fix at the Checkout SDK level. Next time you are welcome to submit a PR yourself!

PR #802, opened 2020-02-26
CHECKOUT-4416 Include custom items IDs when calling updateShippingAddress

capsula4:
## What?
Include custom items IDs when in payload when calling `updateShippingAddress`

## Why?
In case a cart only has custom items, `updateShippingAddress` would fail since there were no items ids in the payload.

Reported in https://github.com/bigcommerce/checkout-sdk-js/issues/701

## Testing / Proof
unit

@bigcommerce/checkout

### DIFF
--- a/src/shipping/consignment-action-creator.spec.ts
+++ b/src/shipping/consignment-action-creator.spec.ts
@@ -5,6 +5,7 @@ import { from, of } from 'rxjs';
 import { catchError, toArray } from 'rxjs/operators';
 
 import { Address } from '../address';
+import { getCart } from '../cart/carts.mock';
 import { createCheckoutStore, Checkout, CheckoutRequestSender, CheckoutStore } from '../checkout';
 import { getCheckout, getCheckoutState, getCheckoutStoreState } from '../checkout/checkouts.mock';
 import { InvalidArgumentError, MissingDataError } from '../common/error/errors';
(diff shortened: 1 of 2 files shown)


src/shipping/consignment-action-creator.spec.ts
```diff
@@ -887,6 +888,19 @@ describe('consignmentActionCreator', () => {
         });
 
         it('sends request to update shipping address in first consignment', async () => {
+            jest.spyOn(store.getState().cart, 'getCart').mockReturnValue({
```
icatalina: It might be better to keep these 2 cases separate. One for just checking the payload get sent as is, one for _adding custom products to the payload_? What do you think?

capsula4: `updateShippingAddress` **always** sends the payload (address), **plus** all the shippable items in the cart. This used to be only physical, now it has been extended to include custom items.

src/shipping/consignment-action-creator.ts
```diff
@@ -301,8 +301,14 @@ export default class ConsignmentActionCreator {
                 .map(item => ({
                     itemId: item.id,
                     quantity: item.quantity,
-                })
-            ),
+                }))
+                .concat(
+                    (cart.lineItems && cart.lineItems.customItems || [])
+                        .map(item => ({
```
davidchin: Instead of duplicating the `map` function, would it be better if you just concat the `physicalItems` and `customItems` before `map`? i.e.:

```ts
const { physicalItems, customItems = [] } = cart.lineItems;

[...physicalItems, ...customItems].map(item => ({
    itemId: item.id,
    quantity: item.quantity,
}));
```

capsula4: Thanks, that worked! I tried using `physicalItems.concat()` but it was complaining that the types were different.

capsula4: @davidchin  let me know if it's good to go

PR #808, opened 2020-03-05
feat(payment): PAYPAL-227 Pass theme settings to PP script

helga1507:
## What?
Get theme settings and pass it to PayPal buttons initialization script
Edit currency - now form cart (not from config)

## Why?
For changing styles smart-buttons

## Testing / Proof
Screen Shot 2020-03-05 at 12 00 06 PM

### DIFF
--- a/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.spec.ts
+++ b/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.spec.ts
@@ -12,7 +12,7 @@ import { INTERNAL_USE_ONLY } from '../../../common/http-request';
 import { ConfigActionCreator, ConfigRequestSender } from '../../../config';
 import { PaymentMethod, PaymentMethodActionType } from '../../../payment';
 import { getPaypalCommerce } from '../../../payment/payment-methods.mock';
-import { ButtonsOptions, PaypalCommerceHostWindow, PaypalCommerceScriptLoader, PaypalCommerceScriptOptions, PaypalCommerceSDK } from '../../../payment/strategies/paypal-commerce';
+import { ButtonsOptions, PaypalCommerceHostWindow, PaypalCommerceScriptLoader, PaypalCommerceScriptOptions, PaypalCommerceSDK, StyleButtonColor, StyleButtonLabel } from '../../../payment/strategies/paypal-commerce';
 import { getPaypalCommerceMock } from '../../../payment/strategies/paypal-commerce/paypal-commerce.mock';
 import { CheckoutButtonInitializeOptions } from '../../checkout-button-options';
 import CheckoutButtonMethodType from '../checkout-button-method-type';
@@ -47,7 +47,13 @@ describe('PaypalCommerceButtonStrategy', () => {
         formPoster = createFormPoster();
         paypalScriptLoader = new PaypalCommerceScriptLoader(getScriptLoader());
 
-        paypalOptions = {};
+        paypalOptions = {
+            style: {
+                color: StyleButtonColor.white,
+                label: StyleButtonLabel.buynow,
+                height: 45,
+            },
+        };
 
         options = {
             containerId: 'paypalcommerce-container1',
@@ -126,6 +132,7 @@ describe('PaypalCommerceButtonStrategy', () => {
         expect(paypal.Buttons).toHaveBeenCalledWith({
             createOrder: expect.any(Function),
             onApprove: expect.any(Function),
+            style: paypalOptions.style,
         });
     });
 
@@ -135,7 +142,7 @@ describe('PaypalCommerceButtonStrategy', () => {
         expect(render).toHaveBeenCalled();
     });
 
-    it('check clientId, currency, commit, intent, disableFunding in paypalCommerce script', async () => {
+    it('calls loadPaypalCommerce with expected arguments', async () => {
         jest.spyOn(paypalScriptLoader, 'loadPaypalCommerce')
             .mockImplementation((options: PaypalCommerceScriptOptions) => {
                 (window as PaypalCommerceHostWindow).paypal = paypal;
@@ -156,7 +163,7 @@ describe('PaypalCommerceButtonStrategy', () => {
         await strategy.initialize(options);
     });
 
-    it('check disableFunding in paypalCommerce script, if isPayPalCreditAvailable = true', async () => {
+    it('calls loadPaypalCommerce without disableFunding when isPayPalCreditAvailable = true', async () => {
         paymentMethod.initializationData.isPayPalCreditAvailable = true;
 
         await store.dispatch(of(createAction(PaymentMethodActionType.LoadPaymentMethodsSucceeded, [paymentMethod])));
@@ -250,6 +257,55 @@ describe('PaypalCommerceButtonStrategy', () => {
         }
     });
 
+    describe('validate style for PaypalCommerce checkout button', () => {
+        it('invalid all data', async () => {
+            const style: any = {
+                layout: 'aaa',
+                color: 'aaa',
+                shape: 'aaa',
+                height: 5,
+                label: 'aaa',
+                tagline: true,
+            };
+
+            await strategy.initialize({ ...options, paypalCommerce: { style } });
+
+            expect(paypal.Buttons).toHaveBeenCalledWith({
+                createOrder: expect.any(Function),
+                onApprove: expect.any(Function),
+                style: { height: 25 },
+            });
+        });
+
+        it('invalid height and valid other', async () => {
+            const style: any = {
+                height: 100,
+                tagline: true,
+                layout: 'horizontal',
+            };
+
+            await strategy.initialize({ ...options, paypalCommerce: { style } });
+
+            expect(paypal.Buttons).toHaveBeenCalledWith({
+                createOrder: expect.any(Function),
+                onApprove: expect.any(Function),
+                style: { tagline: true, layout: 'horizontal', height: 55 },
+            });
+        });
+
+        it('invalid height - not number', async () => {
+            const style: any = { height: '' };
+
+            await strategy.initialize({ ...options, paypalCommerce: { style } });
+
+            expect(paypal.Buttons).toHaveBeenCalledWith({
+                createOrder: expect.any(Function),
+                onApprove: expect.any(Function),
+                style: {},
+            });
+        });
+    });
+
     describe('throws error during initialize', () => {
         it('without clientId', async () => {
             paymentMethod.initializationData.clientId = null;
@@ -290,28 +346,6 @@ describe('PaypalCommerceButtonStrategy', () => {
                 await strategy.initialize(options);
             } catch (error) {
                 expect(error).toBeInstanceOf(MissingDataError);
-            }
-        });
-
-        it('throw error without config', async () => {
-            const state = getCheckoutStoreState();
-            const config = state.config;
-            delete config.data;
-
-            store = createCheckoutStore({ ...state, config });
-
-            strategy = new PaypalCommerceButtonStrategy(
-                store,
-                checkoutActionCreator,
-                paypalScriptLoader,
-                formPoster,
-                requestSender
-            );
-
-            try {
-                await strategy.initialize(options);
-            } catch (error) {
-                expect(error).toEqual(new MissingDataError(MissingDataErrorType.MissingCheckoutConfig));
             }
         });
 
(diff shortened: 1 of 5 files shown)


src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -22,43 +22,84 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
...
+                };
+
+                if (paypalOptions && paypalOptions.style) {
+                    buttonParams.style = this.validateStyleParams(paypalOptions.style);
+                }
+
+                const paramsScript: PaypalCommerceScriptOptions = {
                     clientId: paymentMethod.initializationData.clientId,
-                    currency: config.currency.code,
+                    currency: cart.currency.code,
```
helga1507: @davidchin I changed currency for more intention-revealing. 
Thank you for your help!

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.spec.ts
```diff
@@ -126,6 +132,50 @@ describe('PaypalCommerceButtonStrategy', () => {
         expect(paypal.Buttons).toHaveBeenCalledWith({
             createOrder: expect.any(Function),
             onApprove: expect.any(Function),
+            style: paypalOptions.style,
+        });
+    });
+
+    it('validate style for PaypalCommerce checkout button', async () => {
+        const style = {
```
capsula4: can you do `const style as PaypalButtonStyleOptions` to avoid the casting on line 150?

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.spec.ts
```diff
@@ -126,6 +132,50 @@ describe('PaypalCommerceButtonStrategy', () => {
         expect(paypal.Buttons).toHaveBeenCalledWith({
             createOrder: expect.any(Function),
             onApprove: expect.any(Function),
+            style: paypalOptions.style,
+        });
+    });
+
+    it('validate style for PaypalCommerce checkout button', async () => {
```
capsula4: it's better to be more descriptive for the unit tests, for example
`it calls paypal API with the provided styles`

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.spec.ts
```diff
@@ -126,6 +132,50 @@ describe('PaypalCommerceButtonStrategy', () => {
...
+        await strategy.initialize({ ...options, paypalCommerce: paypalOptions });
+
+        expect(paypal.Buttons).toHaveBeenCalledWith({
+            createOrder: expect.any(Function),
+            onApprove: expect.any(Function),
+            style: { height: 25 },
+        });
+    });
+
+    it('validate style (height, tagline) for PaypalCommerce checkout button', async () => {
```
capsula4: `it calls paypal API with height maximum of 55`

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.spec.ts
```diff
@@ -135,6 +185,46 @@ describe('PaypalCommerceButtonStrategy', () => {
         expect(render).toHaveBeenCalled();
     });
 
+    it('check clientId, currency, commit, intent, disableFunding in paypalCommerce script', async () => {
```
capsula4:
```js
it('calls loadPaypalCommerce with expected arguments')
```

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.spec.ts
```diff
@@ -135,6 +185,46 @@ describe('PaypalCommerceButtonStrategy', () => {
...
+
+                expect(options).toMatchObject(obj);
+
+                return Promise.resolve(paypal);
+            });
+
+        await strategy.initialize(options);
+    });
+
+    it('check disableFunding in paypalCommerce script, if isPayPalCreditAvailable = true', async () => {
```
capsula4:
```js
it('calls loadPaypalCommerce without disableFunding when isPayPalCreditAvailable = true')
```

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -22,43 +22,84 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
...
-                    });
+                    .then(paypal => paypal.Buttons(buttonParams).render(`#${options.containerId}`));
             });
     }
 
     deinitialize(): Promise<void> {
         return Promise.resolve();
     }
 
+    validateStyleParams(style: PaypalButtonStyleOptions): PaypalButtonStyleOptions {
```
capsula4: Probably this should be a private method, 
```
private _validateStyleParams(style: PaypalButtonStyleOptions): PaypalButtonStyleOptions {
```

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -22,43 +22,84 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
...
+
+                const paramsScript: PaypalCommerceScriptOptions = {
                     clientId: paymentMethod.initializationData.clientId,
-                    currency: config.currency.code,
+                    currency: cart.currency.code,
+                    commit: false,
+                    intent: paymentMethod.initializationData.intent,
                 };
 
+                if (!paymentMethod.initializationData.isPayPalCreditAvailable) {
```
capsula4: prefer destructuring for better readability
```js
const { isPayPalCreditAvailable } = paymentMethod.initializationData;

if (!isPayPalCreditAvailable) {
```

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
```diff
@@ -3,8 +3,43 @@ export interface ApproveDataOptions {
     orderID: string;
 }
 
+export enum StyleButtonLabel {
```
capsula4: I think this would be more easily consumed externally if exposed as strings rather than enums. Not all of our clients use typescript, so passing integers might not have a good readability for a JS consumer.

e.g:
```js
{
  layout: 1,
  color: 2,
  shape: 3
}
```

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -22,43 +22,84 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
...
+        if (color && !StyleButtonColor[color]) {
+            delete updatedStyle.color;
+        }
+
+        if (shape && !StyleButtonShape[shape]) {
+            delete updatedStyle.shape;
+        }
+
+        if (typeof height === 'number') {
+            updatedStyle.height = height < 25
```
capsula4: you could do something like  
```js
updatedStyle.height = Math.min(Math.max(height, 55), 25);
```

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.spec.ts
helga1507: Thank you!

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.spec.ts
helga1507: Thank you!

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
helga1507: Sure, I agree. Thank you!

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.spec.ts
helga1507: Thank you!

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.spec.ts
helga1507: Thank you!

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.spec.ts
helga1507: But I test here (and in the previous function) validation params in style. So in 'options' can be different style from clients who don't use typescript. Paypal's script and smart buttons don't work with wrong style, although these are just styles. So I wanted validate data and replace wrong params.
I'll create describe with validate params in different cases

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.spec.ts
helga1507: I can't use in this case `PaypalButtonStyleOptions` because I would like to pass invalid data for check function of validation. Maybe in this case will be better `style: any` ?

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
helga1507: Enum needs only for `_validateStyleParams()`. Because I don't want to duplicate code like 
`layout?: vertical | horizontal` and in _validateStyleParams() `if layout !== 'vertical' ...`

Actually, `layout: 'paypal'` - valid, because
`layout?: keyof typeof StyleButtonLayout` = `layout?: vertical | horizontal`

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
helga1507: That's great idea. Thank you!
But tslint doesn't understand value
Can I write `type height = 25...55; updatedStyle.height = Math.min(Math.max(height, 25), 55) as height;` ?

I'm not sure, that's valid

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
capsula4: The enum is not really just needed internally for `validateStyleParams`
This is also exposed externally. The styles can be provided as part of the argument to initialise the payment method, it will get eventually exported as part of the documentation.
https://github.com/bigcommerce/checkout-sdk-js/blob/master/docs/interfaces/paymentinitializeoptions.md

I think you can still use the enum, but define strings instead of numbers. An example: https://github.com/bigcommerce/checkout-sdk-js/blob/06c19c2b35d76aa084ee317758588f43a84ba056/src/payment/strategies/square/square-form.ts#L78

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
capsula4: you are right it's a bit annoying that TSlint doesn't recognize it, you could keep it as you had it then to avoid overriding the type.

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
helga1507: Yes, I know about documentation. But we still need validate this data, because script of payPal doesn't work with wrong style. PayPal doesn't validate it (so weird) and with one mistake in style param you don't see smart buttons

I tried to do like in `DigitalWalletType`. But in this case I can't validate without duplicate code. Now I check data like this `StyleButtonLabel[style.label]`, because `style.label` is key. If I define strings instead of numbers key doesn't match value

Or do you mean create enum with the same key and value? Like this 
`export enum StyleButtonLabel {
    paypal = 'paypal',
    checkout = 'checkout',
    buynow = 'buynow',
    pay = 'pay',
    installment = 'installment',
}`

capsula4: yeah, you can create the enum using strings instead of numbers so it's easier to be consumed externally.

helga1507: Ok, I'll do it
Thank you!

helga1507:
@capsula4 Sure. Thanks a lot for your help!

PR #818, opened 2020-03-22
feat(payment): INT-2437 Add support for GooglePay on Adyen

vmparra:
## What?
I want to be able to offer my world shoppers GooglePay payment method through Adyen

## Why?
So that I can give them alternative methods to pay so that they are more likely to purchase.

## Testing / Proof

- Manual

- Unit

1. Revert this PR

## Sibling PRs
[UCO](https://github.com/bigcommerce/checkout-js/pull/253)

### DIFF
--- a/src/checkout-buttons/strategies/googlepay/googlepay-adyenv2-button-strategy.spec.ts
+++ b/src/checkout-buttons/strategies/googlepay/googlepay-adyenv2-button-strategy.spec.ts
@@ -10,7 +10,7 @@ import { getConfigState } from '../../../config/configs.mock';
 import { getCustomerState } from '../../../customer/customers.mock';
 import { PaymentMethod } from '../../../payment';
 import { getPaymentMethodsState } from '../../../payment/payment-methods.mock';
-import { createGooglePayPaymentProcessor, GooglePayAuthorizeNetInitializer, GooglePayPaymentProcessor } from '../../../payment/strategies/googlepay';
+import { createGooglePayPaymentProcessor, GooglePayAdyenV2Initializer, GooglePayPaymentProcessor } from '../../../payment/strategies/googlepay';
 import { getGooglePaymentDataMock } from '../../../payment/strategies/googlepay/googlepay.mock';
 import { CheckoutButtonInitializeOptions } from '../../checkout-button-options';
 import CheckoutButtonMethodType from '../checkout-button-method-type';
@@ -50,7 +50,7 @@ describe('GooglePayCheckoutButtonStrategy', () => {
 
         paymentProcessor = createGooglePayPaymentProcessor(
             store,
-            new GooglePayAuthorizeNetInitializer()
+            new GooglePayAdyenV2Initializer()
         );
 
         formPoster = createFormPoster();
@@ -88,7 +88,7 @@ describe('GooglePayCheckoutButtonStrategy', () => {
 
     describe('#initialize()', () => {
         it('Creates the button', async () => {
-            checkoutButtonOptions = getCheckoutButtonOptions(CheckoutButtonMethodType.GOOGLEPAY_AUTHORIZENET);
+            checkoutButtonOptions = getCheckoutButtonOptions(CheckoutButtonMethodType.GOOGLEPAY_ADYENV2);
 
             await strategy.initialize(checkoutButtonOptions);
 
@@ -96,7 +96,7 @@ describe('GooglePayCheckoutButtonStrategy', () => {
         });
 
         it('Validates if strategy has been initialized', async () => {
-            checkoutButtonOptions = getCheckoutButtonOptions(CheckoutButtonMethodType.GOOGLEPAY_AUTHORIZENET);
+            checkoutButtonOptions = getCheckoutButtonOptions(CheckoutButtonMethodType.GOOGLEPAY_ADYENV2);
 
             await strategy.initialize(checkoutButtonOptions);
 
@@ -104,7 +104,7 @@ describe('GooglePayCheckoutButtonStrategy', () => {
         });
 
         it('fails to initialize the strategy if no container id is supplied', async () => {
-            checkoutButtonOptions = getCheckoutButtonOptions(CheckoutButtonMethodType.GOOGLEPAY_AUTHORIZENET, Mode.UndefinedContainer);
+            checkoutButtonOptions = getCheckoutButtonOptions(CheckoutButtonMethodType.GOOGLEPAY_ADYENV2, Mode.UndefinedContainer);
 
             try {
                 await strategy.initialize(checkoutButtonOptions);
@@ -114,7 +114,7 @@ describe('GooglePayCheckoutButtonStrategy', () => {
         });
 
         it('fails to initialize the strategy if no valid container id is supplied', async () => {
-            checkoutButtonOptions = getCheckoutButtonOptions(CheckoutButtonMethodType.GOOGLEPAY_AUTHORIZENET, Mode.InvalidContainer);
+            checkoutButtonOptions = getCheckoutButtonOptions(CheckoutButtonMethodType.GOOGLEPAY_ADYENV2, Mode.InvalidContainer);
 
             await expect(strategy.initialize(checkoutButtonOptions))
                 .rejects.toThrow(InvalidArgumentError);
@@ -125,7 +125,7 @@ describe('GooglePayCheckoutButtonStrategy', () => {
         let containerId: string;
 
         beforeAll(() => {
-            checkoutButtonOptions = getCheckoutButtonOptions(CheckoutButtonMethodType.GOOGLEPAY_AUTHORIZENET, Mode.GooglePayAuthorizeNet);
+            checkoutButtonOptions = getCheckoutButtonOptions(CheckoutButtonMethodType.GOOGLEPAY_ADYENV2, Mode.GooglePayAdyenV2);
             containerId = checkoutButtonOptions.containerId;
         });
 
@@ -156,7 +156,7 @@ describe('GooglePayCheckoutButtonStrategy', () => {
 
     describe('#handleWalletButtonClick', () => {
         it('handles wallet button event', async () => {
-            checkoutButtonOptions = getCheckoutButtonOptions(CheckoutButtonMethodType.GOOGLEPAY_AUTHORIZENET, Mode.GooglePayAuthorizeNet);
+            checkoutButtonOptions = getCheckoutButtonOptions(CheckoutButtonMethodType.GOOGLEPAY_ADYENV2, Mode.GooglePayAdyenV2);
 
             jest.spyOn(paymentProcessor, 'displayWallet').mockReturnValue(Promise.resolve(getGooglePaymentDataMock()));
             jest.spyOn(paymentProcessor, 'handleSuccess').mockReturnValue(Promise.resolve());
@@ -166,7 +166,7 @@ describe('GooglePayCheckoutButtonStrategy', () => {
             await strategy.initialize(checkoutButtonOptions);
 
             walletButton.click();
-            expect(paymentProcessor.initialize).toHaveBeenCalledWith(CheckoutButtonMethodType.GOOGLEPAY_AUTHORIZENET);
+            expect(paymentProcessor.initialize).toHaveBeenCalledWith(CheckoutButtonMethodType.GOOGLEPAY_ADYENV2);
         });
     });
 });
--- a/src/checkout-buttons/strategies/googlepay/googlepay-braintree-button-strategy.spec.ts
+++ b/src/checkout-buttons/strategies/googlepay/googlepay-braintree-button-strategy.spec.ts
@@ -12,9 +12,10 @@ import { getCustomerState } from '../../../customer/customers.mock';
 import { PaymentMethod } from '../../../payment';
 import { getPaymentMethodsState } from '../../../payment/payment-methods.mock';
 import { BraintreeScriptLoader, BraintreeSDKCreator } from '../../../payment/strategies/braintree';
-import { createGooglePayPaymentProcessor, GooglePaymentData, GooglePayBraintreeInitializer, GooglePayPaymentProcessor } from '../../../payment/strategies/googlepay';
+import { createGooglePayPaymentProcessor, GooglePayBraintreeInitializer, GooglePayPaymentProcessor } from '../../../payment/strategies/googlepay';
 import { getGooglePaymentDataMock } from '../../../payment/strategies/googlepay/googlepay.mock';
 import { CheckoutButtonInitializeOptions } from '../../checkout-button-options';
+import CheckoutButtonMethodType from '../checkout-button-method-type';
 
 import GooglePayButtonStrategy from './googlepay-button-strategy';
 import { getCheckoutButtonOptions, getPaymentMethod, Mode } from './googlepay-button.mock';
@@ -67,29 +68,23 @@ describe('GooglePayCheckoutButtonStrategy', () => {
             paymentProcessor
         );
 
-        jest.spyOn(store, 'dispatch')
-            .mockReturnValue(Promise.resolve(store.getState()));
-
-        jest.spyOn(paymentProcessor, 'initialize')
-            .mockReturnValue(Promise.resolve());
-
-        jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod')
-            .mockReturnValue(paymentMethod);
-
         jest.spyOn(formPoster, 'postForm')
             .mockReturnValue(Promise.resolve());
+        jest.spyOn(store, 'dispatch')
+            .mockReturnValue(Promise.resolve(store.getState()));
+        jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod')
+            .mockReturnValue(paymentMethod);
+        jest.spyOn(paymentProcessor, 'initialize')
+            .mockReturnValue(Promise.resolve());
+        jest.spyOn(paymentProcessor, 'deinitialize');
 
-        container = document.createElement('div');
-        container.setAttribute('id', 'googlePayCheckoutButton');
         walletButton = document.createElement('a');
         walletButton.setAttribute('id', 'mockButton');
-
         jest.spyOn(paymentProcessor, 'createButton')
             .mockReturnValue(walletButton);
 
-        jest.spyOn(paymentProcessor, 'deinitialize');
-
-        container.appendChild(walletButton);
+        container = document.createElement('div');
+        container.setAttribute('id', 'googlePayCheckoutButton');
         document.body.appendChild(container);
     });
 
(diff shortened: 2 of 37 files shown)


src/checkout-buttons/strategies/googlepay/googlepay-adyenv2-button-strategy.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+            if (checkoutButtonOptions.containerId) {
+                const button = document.getElementById(checkoutButtonOptions.containerId);
+
+                if (button) {
+                    expect(button.firstChild).toBe(null);
+                }
+            }
+
+            container = document.createElement('div');
+            document.body.appendChild(container);
```
clopezh, line 175: Can these code be part of afterEach() call? 🍹

src/checkout-buttons/strategies/googlepay/googlepay-adyenv2-button-strategy.spec.ts
```diff
@@ -0,0 +1,216 @@
...
```
clopezh, line 190: re: last comment

src/customer/strategies/googlepay/googlepay-adyenv2-customer-strategy.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+                const button = document.getElementById(customerInitializeOptions.googlepayadyenv2.container);
+
+                if (button) {
+                    expect(button.firstChild).toBe(null);
+                }
+            }
+
+            // Prevent "After Each" failure
+            container = document.createElement('div');
+            document.body.appendChild(container);
```
clopezh, line 178: Can these code be part of afterEach() call? 🍹

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -214,59 +217,72 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
-                        },
-                    },
-                });
-        }
+            resolve(paymentComponent);
+        });
     }
 
-    private async _processAdditionalAction(error: any, shouldSaveInstrument?: boolean): Promise<any> {
+    private async _processAdditionalAction(error: unknown, shouldSaveInstrument?: boolean): Promise<InternalCheckoutSelectors> {
```
clopezh, line 279: I like `unknown` instead of any 👍

src/payment/strategies/adyenv2/adyenv2-script-loader.ts
```diff
@@ -2,29 +2,32 @@ import { ScriptLoader, StylesheetLoader } from '@bigcommerce/script-loader';
...
+import { AdyenClient, AdyenConfiguration, AdyenHostWindow } from './adyenv2';
 
 export default class AdyenV2ScriptLoader {
     constructor(
         private _scriptLoader: ScriptLoader,
         private _stylesheetLoader: StylesheetLoader,
         private _window: AdyenHostWindow = window
-    ) {}
+    ) {
+    }
```
clopezh, line 13:
```suggestion
    ) { }
```

ValeriaRivas:
💚 Tested
QE Details:
Auth and capture, auth only, capture, void, refund, partially refund tested. All working correctly.

vmparra:
@clopezh I've just addressed your comments, please take another 👁️

src/checkout-buttons/strategies/googlepay/googlepay-adyenv2-button-strategy.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+    afterEach(() => {
+        document.body.removeChild(container);
+    });
+
+    it('creates an instance of GooglePayButtonStrategy', () => {
+        expect(strategy).toBeInstanceOf(GooglePayButtonStrategy);
+    });
+
+    describe('#initialize()', () => {
+        describe('Payment method exist', () => {
```
capsula4, line 93: all tests seems to be inside this describe block "Payment method exist", maybe you could remove it since it only adds extra indentation?

src/checkout-buttons/strategies/googlepay/googlepay-adyenv2-button-strategy.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+
+            jest.spyOn(paymentProcessor, 'displayWallet').mockReturnValue(Promise.resolve(getGooglePaymentDataMock()));
+            jest.spyOn(paymentProcessor, 'handleSuccess').mockReturnValue(Promise.resolve());
+            jest.spyOn(paymentProcessor, 'updateShippingAddress').mockReturnValue(Promise.resolve());
+
+            await strategy.initialize(checkoutButtonOptions);
+
+            walletButton.click();
+
+            expect(paymentProcessor.initialize).toHaveBeenCalled();
```
capsula4, line 169: wouldn't this be called even if `walletButton.click()`  line was removed? what are you exactly expecting to happen after `walletButton.click();` is called?

```js
// ... rest of the test
await strategy.initialize(checkoutButtonOptions);
expect(paymentProcessor.initialize).toHaveBeenCalled()

walletButton.click();
// expect ??
```

src/checkout-buttons/strategies/googlepay/googlepay-button-strategy.ts
```diff
@@ -71,13 +71,17 @@ export default class GooglePayButtonStrategy implements CheckoutButtonStrategy {
 
         return this._googlePayPaymentProcessor.displayWallet()
             .then(paymentData => this._googlePayPaymentProcessor.handleSuccess(paymentData)
-            .then(() => {
-                if (paymentData.shippingAddress) {
-                    this._googlePayPaymentProcessor.updateShippingAddress(paymentData.shippingAddress);
-                }
-            }))
+                .then(() => {
```
capsula4, line 74: Might be easier to keep the indentation as it was, but close the brackets on the line before, e.g:
```js
.then(paymentData => this._googlePayPaymentProcessor.handleSuccess(paymentData))
.then(() => {
  if (paymentData.shippingAddress) {
  // ...
```

or if you want to make it even better, switch to async/await

```js
const paymentData = await this._googlePayPaymentProcessor.displayWallet();
await this._googlePayPaymentProcessor.handleSuccess(paymentData);

if (paymentData.shippingAddress) {
  this._googlePayPaymentProcessor.updateShippingAddress(paymentData.shippingAddress);
}

await this._onPaymentSelectComplete();
```

src/checkout-buttons/strategies/googlepay/googlepay-stripe-button-strategy.spec.ts
```diff
@@ -107,108 +99,72 @@ describe('GooglePayCheckoutButtonStrategy', () => {
...
+            jest.spyOn(paymentProcessor, 'displayWallet').mockReturnValue(Promise.resolve(getGooglePaymentDataMock()));
             jest.spyOn(paymentProcessor, 'handleSuccess').mockReturnValue(Promise.resolve());
             jest.spyOn(paymentProcessor, 'updateShippingAddress').mockReturnValue(Promise.resolve());
 
-            await strategy.initialize(googlePayOptions).then(() => {
-                walletButton.click();
-            });
+            await strategy.initialize(checkoutButtonOptions);
+
+            walletButton.click();
```
capsula4, line 167: same as in the other test, what's the difference when walletButton is provided? is it that `paymentProcessor.initialize` is only called after the wallet is clicked? If so, might be better to be explicit about it:

```js
await strategy.initialize(checkoutButtonOptions);
expect(paymentProcessor.initialize).not.toHaveBeenCalled();
walletButton.click();
expect(paymentProcessor.initialize).toHaveBeenCalledWith(/** ...arg */);
```

src/customer/strategies/googlepay/googlepay-adyenv2-customer-strategy.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+        container = document.createElement('div');
+        container.setAttribute('id', 'googlePayCheckoutButton');
+        document.body.appendChild(container);
+    });
+
+    afterEach(() => {
+        document.body.removeChild(container);
+    });
+
+    it('creates an instance of GooglePayCustomerStrategy', () => {
```
capsula4, line 87: same as the other file, this test might not be needed

src/customer/strategies/googlepay/googlepay-adyenv2-customer-strategy.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+            await strategy.deinitialize();
+
+            const button = document.getElementById(containerId);
+
+            expect(button).toHaveProperty('firstChild', null);
+        });
+    });
+
+    describe('#signIn()', () => {
+        it('throws error if trying to sign in programmatically', async () => {
```
capsula4, line 152: `it('always throws error')` ?

src/customer/strategies/googlepay/googlepay-adyenv2-customer-strategy.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+            jest.spyOn(store.getState().payment, 'getPaymentId')
+                .mockReturnValue(paymentId);
+
+            const options = {
+                methodId: 'googlepayadyenv2',
+            };
+
+            await strategy.signOut(options);
+
+            expect(store.getState).toHaveBeenCalledTimes(3);
```
capsula4, line 202: maybe also add `expect(await strategy.signOut(options)).toEqual(store.getState())` ?

src/customer/strategies/googlepay/googlepay-adyenv2-customer-strategy.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+                },
+            };
+
+            jest.spyOn(paymentProcessor, 'displayWallet').mockReturnValue(Promise.resolve(getGooglePaymentDataMock()));
+            jest.spyOn(paymentProcessor, 'handleSuccess').mockReturnValue(Promise.resolve());
+            jest.spyOn(paymentProcessor, 'updateShippingAddress').mockReturnValue(Promise.resolve());
+
+            await strategy.initialize(customerInitializeOptions);
+
+            walletButton.click();
```
capsula4, line 221: same comment as other files

src/checkout-buttons/strategies/googlepay/googlepay-braintree-button-strategy.spec.ts
```diff
@@ -113,108 +105,72 @@ describe('GooglePayCheckoutButtonStrategy', () => {
...
+            jest.spyOn(paymentProcessor, 'displayWallet').mockReturnValue(Promise.resolve(getGooglePaymentDataMock()));
             jest.spyOn(paymentProcessor, 'handleSuccess').mockReturnValue(Promise.resolve());
             jest.spyOn(paymentProcessor, 'updateShippingAddress').mockReturnValue(Promise.resolve());
 
-            await strategy.initialize(googlePayOptions).then(() => {
-                walletButton.click();
-            });
+            await strategy.initialize(checkoutButtonOptions);
+
+            walletButton.click();
```
capsula4, line 173: same

src/customer/strategies/googlepay/googlepay-customer-strategy.ts
```diff
@@ -95,33 +99,31 @@ export default class GooglePayCustomerStrategy implements CustomerStrategy {
...
         event.preventDefault();
 
         return this._googlePayPaymentProcessor.displayWallet()
             .then(paymentData => this._googlePayPaymentProcessor.handleSuccess(paymentData)
-            .then(() => {
-                if (paymentData.shippingAddress) {
-                    this._googlePayPaymentProcessor.updateShippingAddress(paymentData.shippingAddress);
-                }
-            }))
+                .then(() => {
```
capsula4, line 108: same as I commented in other file, maybe close the brackets earlier to avoid extra indentation or switch to async/await. Thanks!

src/payment/create-payment-strategy-registry.spec.ts
```diff
@@ -52,6 +52,11 @@ describe('CreatePaymentStrategyRegistry', () => {
         expect(paymentStrategy).toBeInstanceOf(AdyenV2PaymentStrategy);
     });
 
+    it('can instantiate goolgepayadyenv2', () => {
```
capsula4, line 55: typo, `google`

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -214,59 +213,72 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
+        return new Promise(resolve => {
+            if (adyenv2.cardVerificationContainerId) {
+                cardVerificationComponent = adyenClient.create(AdyenComponentType.SecuredFields, {
+                    ...adyenv2.options,
+                    onChange: componentState => this._updateComponentState(componentState),
+                    onError: componentState => this._updateComponentState(componentState),
+                });
 
-                paymentComponent.mount(`#${adyenv2.containerId}`);
+                cardVerificationComponent.mount(`#${adyenv2.cardVerificationContainerId}`);
```
capsula4, line 229: what if this fails because the provided containerID doesnt exist? shouldn't you need to reject the returned promise?

src/payment/strategies/adyenv2/adyenv2-script-loader.ts
```diff
@@ -2,29 +2,31 @@ import { ScriptLoader, StylesheetLoader } from '@bigcommerce/script-loader';
...
-            this._scriptLoader.loadScript(`https://checkoutshopper-${configuration.environment}.adyen.com/checkoutshopper/sdk/3.6.0/adyen.js`),
-        ])
-        .then(() => {
-            if (!this._window.AdyenCheckout) {
-                throw new PaymentMethodClientUnavailableError();
-            }
+    load(configuration: AdyenConfiguration): Promise<AdyenClient> {
+        const promises = [];
+
+        promises.push(this._stylesheetLoader.loadStylesheet(`https://checkoutshopper-${configuration.environment}.adyen.com/checkoutshopper/sdk/3.6.0/adyen.css`));
```
capsula4, line 17: using `push()` it's better suited for when you have conditional promises that are not always meant to be executed, e.g.
```js
if (X) {
  promises.push(/** .... */)
}
```

In this case, might be better to keep what it was there before
```js
await Promise.all([
  // ...,
  // ...
]);

if (!this._window.AdyenCheckout) {
}

return new this._window.AdyenCheckout(configuration);
```

src/payment/strategies/googlepay/googlepay-adyenv2-initializer.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+                getAdyenV2PaymentMethodMock(),
+                false
+            );
+
+            expect(initialize).toEqual(getAdyenV2PaymentDataRequest());
+        });
+    });
+
+    describe('#teardown', () => {
+        it('teardown the initializer', () => expect(googlePayInitializer.teardown()).resolves.toBeUndefined());
```
capsula4, line 28: `it returns a promise that always resolves?`

also, even though the test itself is one line, might be better to keep the title and the test in different lines.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -183,4 +184,20 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
                 });
         }, { methodId }), { queueId: 'widgetInteraction' });
     }
+
+    private async _paymentInstrumentSelected(paymentData: GooglePaymentData) {
+        if (!this._methodId) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+        }
+
+        const methodId = this._methodId;
```
capsula4, line 193: Minor but you might not need this variable

vmparra:
♻️ @capsula4 Changes have been addressed, please take another 👀

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -175,12 +173,18 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
         return this._store.dispatch(this._paymentStrategyActionCreator.widgetInteraction(() => {
             return this._googlePayPaymentProcessor.displayWallet()
                 .then(paymentData => this._paymentInstrumentSelected(paymentData))
-                .then(() => onPaymentSelect())
+                .then(() => {
+                    if (onPaymentSelect) {
```
davidchin, line 177: `onPaymentSelect` has a default value `() => {}` (a no-op function) so you shouldn't need to do a null check.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -175,12 +173,18 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+                        onPaymentSelect();
+                    }
+                })
                 .catch(error => {
                     if (error.statusCode !== 'CANCELED') {
-                        onError(error);
+                        if (onError) {
```
davidchin, line 183: Same here.

src/payment/strategies/googlepay/googlepay-adyenv2-initializer.ts
```diff
@@ -0,0 +1,108 @@
...
+                    parameters: {
+                        gateway: 'adyen',
+                        gatewayMerchantId,
+                    },
+                },
+            }],
+            transactionInfo: {
+                currencyCode,
+                totalPriceStatus: 'FINAL',
+                totalPrice: round(outstandingBalance, 2).toFixed(2),
```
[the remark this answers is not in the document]
vmparra: Google Pay requires `TotalPrice`  to match `^[0-9]+(\.[0-9][0-9])?$` **RegEx** 

image

Another reason is that there are currencies that have 0 decimal places like Japanese yen.

src/checkout-buttons/strategies/googlepay/googlepay-adyenv2-button-strategy.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+
+            await strategy.initialize(checkoutButtonOptions);
+
+            expect(paymentProcessor.initialize).toHaveBeenCalledTimes(1);
+        });
+
+        it('fails to initialize the strategy if no container id is supplied', () => {
+            checkoutButtonOptions = getCheckoutButtonOptions(CheckoutButtonMethodType.GOOGLEPAY_ADYENV2, Mode.UndefinedContainer);
+
+            expect(() => strategy.initialize(checkoutButtonOptions)).toThrow(InvalidArgumentError);
```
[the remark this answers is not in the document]
davidchin: Ok, I'll take this back. I think you'll need to write `async/await` with `try/catch` block instead, because putting `async/await` inside the `expect` function won't actually halt the Jest test runner.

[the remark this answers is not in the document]
capsula4: I think you can do it without try/catch, using the following syntax:
```js
await expect(strategy.initialize(checkoutButtonOptions))
  .rejects.toThrow(InvalidArgumentError));
```

src/payment/strategies/googlepay/googlepay-adyenv2-initializer.ts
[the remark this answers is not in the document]
davidchin: Yes, as I've noted in my previous comment, `.toFixed(2)` would always give you back two decimal places irrespective of the currency. And you shouldn't do it for currencies that don't have cents/decimals. The RegExp also says decimals are optional but here you're hardcoding to two decimal places with `toFixed`. Anyway, the doc doesn't say it'd be a problem even if you send back `100.00` for Yen so I guess this is okay. But what you really need to do is just `${round(outstandingBalance, 2)}` to cap the decimal places to two if there are any.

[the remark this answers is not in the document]
vmparra: We experienced this problem back then when we released Google Pay on Stripe & Google Pay on Braintree.

We found that if we but a 1 US dollar item there were times where rounding itself didn't do the job.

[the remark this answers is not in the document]
davidchin: Ok cool, yes, I had a look at the RegExp again. It accepts either 0 decimal or 2 decimals (not 1) so `round` is probably not enough. Cool, let's keep the `toFixed` as it seems GooglePay doesn't really care about Yen having decimals.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -31,33 +32,38 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
 
-        this._scriptLoader.load(configuration)
-            .then(adyenCheckout => {
-                this._adyenCheckout = adyenCheckout;
-                this._mountComponent(paymentMethod.method);
-            });
+        try {
+            this._paymentComponent = await this._mountPaymentComponent(paymentMethod);
+        } catch (error) {
+            throw error;
```
davidchin, line 56: If you're not handling the error here, you don't need a `try/catch`. You can just let the error bubble up.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -31,33 +32,38 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
...
+        }
+
+        if (paymentMethod.method === AdyenPaymentMethodType.CreditCard ||
+            paymentMethod.method === AdyenPaymentMethodType.Bancontact) {
+            try {
+                this._cardVerificationComponent = await this._mountCardVerificationComponent();
+            } catch (error) {
+                throw error;
```
davidchin, line 64: Same here

vmparra:
♻️ @davidchin Changes have been addressed, please take another 👀

PR #824, opened 2020-03-26
feat(payments): INT-2452 Add billing and shipping for klarna

AlanMunozN:
## What? [INT-2452](https://jira.bigcommerce.com/browse/INT-2452)
I added billing and shipping when the country is AU/NZ

## Why?
Because we want to send billing and shipping when the country is AU/NZ

## Testing / Proof

[Video](https://drive.google.com/file/d/1n91rq9tI50MpXqcuu9aA77__MJ9BKB8P/view?usp=sharing)

Screen Shot 2020-03-30 at 2 36 55 PM

### DIFF
--- a/src/payment/strategies/klarna/klarna-payment-strategy.ts
+++ b/src/payment/strategies/klarna/klarna-payment-strategy.ts
@@ -10,6 +10,7 @@ import { RemoteCheckoutActionCreator } from '../../../remote-checkout';
 import { PaymentMethodCancelledError, PaymentMethodInvalidError } from '../../errors';
 import PaymentMethodActionCreator from '../../payment-method-action-creator';
 import { PaymentInitializeOptions, PaymentRequestOptions } from '../../payment-request-options';
+import { supportedCountries, supportedCountriesRequiringStates } from '../klarnav2';
 import PaymentStrategy from '../payment-strategy';
 
 import KlarnaCredit, { KlarnaAddress, KlarnaLoadResponse, KlarnaUpdateSessionParams } from './klarna-credit';
@@ -18,7 +19,6 @@ import KlarnaScriptLoader from './klarna-script-loader';
 export default class KlarnaPaymentStrategy implements PaymentStrategy {
     private _klarnaCredit?: KlarnaCredit;
     private _unsubscribe?: (() => void);
-    private _supportedEUCountries = ['AT', 'DE', 'DK', 'FI', 'GB', 'NL', 'NO', 'SE', 'CH'];
 
     constructor(
         private _store: CheckoutStore,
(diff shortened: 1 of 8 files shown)


pgonzalezr:
Looks good from QE side

Billing/shipping addresses are properly handled by Klarna

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -115,7 +116,7 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
     }
 
     private _getUpdateSessionData(billingAddress: BillingAddress, shippingAddress?: Address): KlarnaUpdateSessionParams {
-        if (!includes(this._supportedEUCountries, billingAddress.countryCode)) {
+        if (!includes(this._supportedEUCountries, billingAddress.countryCode) && !includes(this._supportedOCCountries, billingAddress.countryCode)) {
```
capsula4, line 119: maybe you can merge both arrays for this check?

```js
!includes([ ...this._supportedEUCountries, _supportedOCCountries]), billingAddress.countryCode)
```

src/payment/strategies/klarna/klarna-payment-strategy.ts
```diff
@@ -130,6 +131,10 @@ export default class KlarnaPaymentStrategy implements PaymentStrategy {
         return data;
     }
 
+    private _needsStateCode(countryCode: string) {
+        return includes(this._supportedOCCountries, countryCode);
```
capsula4, line 135: do you need state code for NZ? I believe NZ doesn't have state codes.

Also, it might make more sense to have two different arrays to better express the intent of them
```js
private _countriesRequiringStates = ['AU', 'NZ'];
private _supportedCountries = ['AT', 'DE', 'DK', 'FI', 'GB', 'NL', 'NO', 'SE', 'CH', ...this._countriesRequiringStates];
```

src/payment/strategies/klarnav2/klarnav2-payment-strategy.ts
```diff
@@ -18,6 +18,7 @@ import KlarnaV2ScriptLoader from './klarnav2-script-loader';
 export default class KlarnaV2PaymentStrategy implements PaymentStrategy {
     private _klarnaPayments?: KlarnaPayments;
     private _supportedEUCountries = ['AT', 'DE', 'DK', 'FI', 'GB', 'NL', 'NO', 'SE', 'CH'];
+    private _supportedOCCountries = ['AU', 'NZ'];
```
capsula4, line 21: if this is always the same as klarna v1, could you extract these values in a different file so they are in sync?

AlanMunozN:
@capsula4  I made changes ♻️ ,It's ready for another look

src/payment/strategies/klarna/klarna-payment-strategy.ts
AlanMunozN: yes, It's required by Klarna.

capsula4: As discussed offline, I don't think NZ will return any stateOrProvinceCode, as it's not a dropdown in our UI.

Screen Shot 2020-04-02 at 9 49 46 am

AlanMunozN: That's right I'm going to apply this change, thanks for the explanation

PR #825, opened 2020-03-27
feat(payment): PAYPAL-288 Disable Credit/Debit card SPB

helga1507:
## What?
Disable Credit/Debit card SPB

## Why?
This button will not be available to merchants as part of PPCP BC integration

## Testing / Proof
Screen Shot 2020-03-27 at 7 24 27 PM

### DIFF
--- a/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
+++ b/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
@@ -1,11 +1,9 @@
 import { FormPoster } from '@bigcommerce/form-poster';
-import { RequestSender } from '@bigcommerce/request-sender';
 
 import { Cart } from '../../../cart';
 import { CheckoutActionCreator, CheckoutStore } from '../../../checkout';
 import { InvalidArgumentError, MissingDataError, MissingDataErrorType } from '../../../common/error/errors';
-import { ContentType, INTERNAL_USE_ONLY } from '../../../common/http-request';
-import { ApproveDataOptions, ButtonsOptions, PaypalButtonStyleOptions, PaypalCommerceScriptLoader, PaypalCommerceScriptOptions, StyleButtonColor, StyleButtonLabel, StyleButtonLayout, StyleButtonShape  } from '../../../payment/strategies/paypal-commerce';
+import { ApproveDataOptions, ButtonsOptions, DisableFundingType, PaypalButtonStyleOptions, PaypalCommerceInitializationData, PaypalCommerceRequestSender, PaypalCommerceScriptLoader, PaypalCommerceScriptOptions, StyleButtonColor, StyleButtonLabel, StyleButtonLayout, StyleButtonShape  } from '../../../payment/strategies/paypal-commerce';
 import { CheckoutButtonInitializeOptions } from '../../checkout-button-options';
 import CheckoutButtonStrategy from '../checkout-button-strategy';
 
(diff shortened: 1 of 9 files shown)


src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -68,6 +59,32 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
...
+    private _setupPayment(cart: Cart): Promise<string> {
+        const cartId = cart.id;
+        const url = '/api/storefront/payment/paypalcommerce';
+        const body = { cartId };
+        const headers = {
+            'X-API-INTERNAL': INTERNAL_USE_ONLY,
+            'Content-Type': ContentType.Json,
+        };
+
+        return this._requestSender.post(url, { headers, body })
```
capsula4: Instead of using the request sender directly, you could follow the same approach as the rest of the strategies.

You can create a `PaypalCommerceRequestSender` that has a `setupPayment(carId)` public method.

Also, instead of using `.then` for promises, we prefer the newer `async/await` syntax

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
```diff
@@ -54,10 +54,18 @@ export interface PaypalCommerceHostWindow extends Window {
...
+
+const ScriptDisableFunding = [ 'credit', 'card' ];
+
 export interface PaypalCommerceScriptOptions {
     clientId: string;
     currency?: string;
     commit?: boolean;
     intent?: 'capture' | 'authorize';
-    disableFunding?: 'credit';
+    disableFunding?: Array<typeof ScriptDisableFunding[number]>;
```
davidchin: A simpler way to declare this property is `Array<'credit' | 'card'>`.

helga1507: Thank you!
But in this case I see error "Type 'string' is not assignable to type 'credit' | 'card''." I can rewrite some code for this, but it's not so elegant :slightly_smiling_face:
So I think it would be better use `Array<typeof ScriptDisableFunding[number]>`

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
I created PaypalCommerceRequestSender. It needs for my next task too
And changed all `. then`

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
davidchin: From what I can tell, you're not seeing the `'string' is not assignable` error because `Array<typeof ScriptDisableFunding[number]>` essentially means `string[]`. This is because with this line 

```ts
const ScriptDisableFunding = [ 'credit', 'card' ]
```

you're declaring a variable `ScriptDisableFunding` with `string[]` as its type (it is implicitly declared by TS). Therefore `Array<typeof ScriptDisableFunding[number]>` translates to `Array<string[][number]>`, which is simply `string[]`.

I think what you're trying to do here is to declare this property to only accept an array containing either `'credit'` or `'card'`, not any string. If you want to achieve that, `Array<'credit' | 'card'>` is the correct declaration. Therefore you should look at fixing the line that gives you the error instead. For example, in this file `paypal-commerce-button-strategy.ts`, you have

```ts
const disableFunding = [ 'card' ];
```

This is similar to the issue I described above where the type for that variable is implicitly declared as `string[]`. So when you assign it to a property that has a stricter type, you'll get a type error. Instead, what you need to do is to either provide a type for that variable explicitly, i.e.: 

```ts
const disableFunding: Array<'credit' | 'card'>
```
or inline the declaration with the return value: i.e.:

```ts
return {
    clientId,
    commit: false,
    currency: cart.currency.code,
    disableFunding: isPayPalCreditAvailable ? ['card', 'credit'] : ['card'],
    intent,
};
```

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -16,58 +14,66 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
...
+
+        return paypal.Buttons(buttonParams).render(`#${options.containerId}`);
     }
 
     deinitialize(): Promise<void> {
         return Promise.resolve();
     }
 
+    private async _setupPayment(cartId: string): Promise<string> {
+        const res = await this._paypalCommerceRequestSender.setupPayment(cartId);
```
capsula4: minor, but you could do
```js
const { orderId } = await this._paypalCommerceRequestSender.setupPayment(cartId);

return orderId;
```

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -16,58 +14,66 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
...
+
+        return res.orderId;
+    }
+
+    private _tokenizePayment(paymentId: string, data: ApproveDataOptions) {
```
capsula4: you could this instead (sorry for not picking it up before)
```js
private _tokenizePayment(paymentId: string, { orderId }: ApproveDataOptions) {
  if (!orderID) {
    // ..
  }
```

src/payment/strategies/paypal-commerce/paypal-commerce-script-loader.spec.ts
```diff
@@ -39,7 +39,7 @@ describe('PaypalCommerceScriptLoader', () => {
                 .mockImplementation((url: string) => {
                     (window as PaypalCommerceHostWindow).paypal = paypal;
 
-                    ['client-id=aaa', 'currency=EUR', 'disable-funding=credit', 'intent=capture' ].forEach(str => {
+                    ['client-id=aaa', 'currency=EUR', 'disable-funding=credit,card', 'intent=capture' ].forEach(str => {
```
capsula4: one extra whitespace at the very end of the array 🍹

helga1507: Thank you!

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
helga1507: You are right! Thank you!

I created type disableFundingType for avoidance duplicate code. Because here there are can be more options (you can see it here https://developer.paypal.com/docs/checkout/reference/customize-sdk/#disable-funding)

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -16,58 +14,66 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
...
+        state = await this._store.dispatch(this._checkoutActionCreator.loadDefaultCheckout());
+        const cart = state.cart.getCart();
+        const paypalOptions = options.paypalCommerce;
+
+        if (!cart) {
+            throw new MissingDataError(MissingDataErrorType.MissingCart);
+        }
+
+        const buttonParams: ButtonsOptions = {
+            createOrder: async () => this._setupPayment(cart.id),
```
davidchin, line 40: There's no `await` for this block so you can remove the `async` here.

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
```diff
@@ -54,10 +59,18 @@ export interface PaypalCommerceHostWindow extends Window {
     paypal?: PaypalCommerceSDK;
 }
 
+export interface PaypalCommerceInitializationData {
+    clientId: string;
+    intent?: 'capture' | 'authorize';
+    isPayPalCreditAvailable?: boolean;
+}
+
+export type disableFundingType = Array<'credit' | 'card'>;
```
davidchin, line 68: Can you capitalise this as you're defining a new type?

helga1507: Sure. Thank you!

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
helga1507: Sure. Thank you!

PR #827, opened 2020-03-30
CHECKOUT-4791 Use subscriptions endpoint for marketing consent

capsula4:
## What?
Use https://github.com/bigcommerce/bigcommerce/pull/34306 subscriptions endpoint for setting email marketing consent.

## Why?
We changed our architectural approach tu subscriptions, to be exposed as a different entity from the customer entity.

## Testing / Proof
unit

@bigcommerce/checkout


src/billing/billing-address-action-creator.ts
```diff
@@ -120,27 +121,16 @@ export default class BillingAddressActionCreator {
...
-        ).pipe(
-            catchError(error => throwErrorAction(
-                CustomerActionType.UpdateCustomerFailed,
-                new UpdateCustomerError(error)
-            ))
-        );
+        return this._subscriptionActionCreator.updateSubscriptions({
+            email,
+            acceptsMarketingNewsletter: marketingEmailConsent,
+            acceptsAbandonedCartEmails: marketingEmailConsent,
```
davidchin: How come `acceptsMarketingNewsletter` and `acceptsAbandonedCartEmails` both reflect the same `marketingEmailConsent` parameter? Our UI might have a single checkbox, but I guess there's a reason for our API to distinguish between the two therefore we should reflect that on the SDK level.

src/billing/billing-address-reducer.ts
```diff
@@ -58,13 +58,13 @@ function errorsReducer(
     case BillingAddressActionType.UpdateBillingAddressFailed:
         return objectSet(errors, 'updateError', action.payload);
 
-    case CustomerActionType.UpdateCustomerRequested:
-    case CustomerActionType.UpdateCustomerSucceeded:
+    case SubscriptionsActionType.UpdateSubscriptionsRequested:
```
davidchin: Since update subscriptions is a separate action, I think it'd be better to store the status and error separately than storing them against `continueAsGuestError` and `isContinuingAsGuest` respectively, and have a separate selector each to reflect these values. For example, it'd be odd for the caller to get `isContinuingAsGuest` as `true` when `updateSubscriptions` is called; instead, the caller would expect to call `isUpdatingSubscriptions` to get such value.

src/billing/billing-address-action-creator.ts
capsula4: Mostly thought about setting both values because we already released this API (`marketingEmailConsent`) and this way, avoid breaking changes. The name is generic enough to encompass both newsletter and abandoned cart emails.

However, the truth is, that this was not having any effect when passed as it requires a CP setting to be turned on (hidden behind an experiment). Happy to "break" the API.

src/billing/billing-address-reducer.ts
capsula4: Also related to your other comment, I think the intent here was to provide an easy way to continue as guest and subscribe that email to newsletter, in a parallel, by using a single SDK action.

Since I was merging these two actions (update billing address & update subscription), I was running into a problem that `isContinuingAsGuest` could get stuck in `true` when updating subscription failed, that's why I'm doing everything here (https://github.com/bigcommerce/checkout-sdk-js/pull/805/files)

Maybe I could still keep a new the state separately, but I think I might need to keep these listeners here?

src/billing/billing-address-action-creator.ts
davidchin: I see. Before we have one flag, now we have two. I think what you can do is to introduce two new flags `acceptsMarketingNewsletter` and `acceptsAbandonedCartEmails` and deprecate the old flag `marketingEmailConsent`. For backward compatibility, if `marketingEmailConsent` is passed as `true`, both `acceptsMarketingNewsletter` and `acceptsAbandonedCartEmails` are passed as `true`. The caller can otherwise pass the values separately using the new flags.

I think it's worth having the new method `updateSubscriptions` because the SDK serves as a JS client for the checkout storefront API. So if there's an endpoint for it, we should have a corresponding JS method for it unless there's a good reason not to.

src/billing/billing-address-reducer.ts
davidchin: Right I see what you mean, both `continueAsGuest` and `updateSubscriptions` actions update the customer's subscription consent. I think what you should do is to, as you said, keep what you have here, but also create a new reducer and selector to specifically store and retrieve the value for subscription updates. With them, the caller can specifically retrieve the status and error state for the new `updateSubscriptions` method.

src/customer/guest-credentials.ts
```diff
@@ -1,5 +1,6 @@
-export default interface GuestCredentials {
+import { Subscriptions } from '../subscription';
+
+export type GuestCredentials = Partial<Subscriptions> & {
     id?: string;
     email: string;
-    marketingEmailConsent?: boolean;
```
capsula4, line 4: since this was an experimental feature, I have decided to remove it to reduce the code that we have to maintain.

src/subscription/subscriptions-reducer.spec.ts
```diff
@@ -0,0 +1,46 @@
+import { createAction } from '@bigcommerce/data-store';
+
+import { RequestError } from '../common/error/errors';
+import { getErrorResponse } from '../common/http-request/responses.mock';
+import { SubscriptionsActionType } from '../subscription';
+
+import subscriptionsReducer from './subscriptions-reducer';
+import SubscriptionsState from './subscriptions-state';
+
+describe('billingAddressReducer', () => {
```
davidchin, line 10: This is probably a copy-and-paste mistake. Can you please rename this to `subscriptionsReducer` instead?

src/checkout/create-internal-checkout-selectors.ts
```diff
@@ -96,6 +99,7 @@ export function createInternalCheckoutSelectorsFactory(): InternalCheckoutSelect
             shippingAddress,
             shippingCountries,
             shippingStrategies,
+            subscriptions,
```
davidchin, line 102: Can you also export the new selectors you've created in `CheckoutStoreStatusSelector` and `CheckoutStoreErrorSelector` so that they can be called externally?

PR #829, opened 2020-04-01
CHECKOUT-4789: Apply polyfills to external dependencies for targeted environments

davidchin:
## What?
Apply polyfills to external dependencies for targeted environments

## Why?
Before this change, `babel-env` doesn't apply to `node_modules` dependencies (it only applies to our own code). As a result, the UMD bundle doesn't work on its own in IE11 if some of those dependencies use features that need to be polyfilled.

Right now this problem only affects payment buttons in PDP and cart page because they are provided by the UMD bundle. I believe it has been a problem since October last year when we bumped the version from 1.21.0 to 1.41.0 (however, I'm surprised that we haven't had any support ticket regarding it 🤔). 

On the other hand, UCO uses the CJS bundle and has a different way of applying the `babel-env` so it doesn't get affected by this problem (it applies the entire `core-js` library for its targeted environment instead of applying based on usage).

## Testing / Proof
|Before|After|
|-|-|
|Screen Shot 2020-04-02 at 10 00 42 am|Screen Shot 2020-04-02 at 9 58 58 am|

### DIFF
--- a/webpack-common.config.js
+++ b/webpack-common.config.js
@@ -49,29 +49,48 @@ async function getBaseConfig() {
     };
 };
 
-const babelLoaderRule = {
-    test: /\.[tj]s$/,
-    include: srcPath,
-    loader: 'babel-loader',
-    options: {
-        presets: [
-            [
-                '@babel/preset-env',
-                {
-                    corejs: 3,
-                    targets: [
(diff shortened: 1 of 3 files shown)


TomA-R:
🙇

PR #832, opened 2020-04-09
feat(payment): INT-2410 Add stripe account as configuration

clopezh:
## What?
Add stripe account when Stripe V3 script is loaded

## Why?
Stripe is requiring to add Stripe Account to identify each merchant

## Testing / Proof
image
image

### DIFF
--- a/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
+++ b/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
@@ -30,7 +30,7 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
         private _stripeScriptLoader: StripeV3ScriptLoader
     ) {}
 
-    initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
+    async initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
         const stripeOptions = options.stripev3;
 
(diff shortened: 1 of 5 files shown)


src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -43,7 +43,7 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
             throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
         }
 
-        return this._stripeScriptLoader.load(paymentMethod.initializationData.stripePublishableKey)
+        return this._stripeScriptLoader.load(paymentMethod.initializationData.stripePublishableKey, paymentMethod.initializationData.stripeConnectedAccount)
```
capsula4, line 46: this line seem to have got a bit too long (prefer 120 char max limit per line)

```js
return this._stripeScriptLoader.load(
  paymentMethod.initializationData.stripePublishableKey, 
  paymentMethod.initializationData.stripeConnectedAccount
)
  .then(stripeJs => {
```

Also not required, but it might be good if you can change this line to use async/await
```js
const stripeJs = await this._stripeScriptLoader.load(
  paymentMethod.initializationData.stripePublishableKey, 
  paymentMethod.initializationData.stripeConnectedAccount
);

this._stripeV3Client = stripeJs;
// ...
```

PR #834, opened 2020-04-10
feat(payment): PAYPAL-265 Markup flow

helga1507:
## What?
For markup flow (flow when we start PayPal commerce flow on the checkout page, not on cart page), we need to implement logic that will send a request to `api/storefront/payment/paypalcommerce/`

after customer has chosen PayPal Commerce method and clicked Place Order.
After FE received a link from 
`api/storefront/payment/paypalcommerce/`, we need to forward customer to url based on the link

After approving in the paypal payment should continue 
 

## Why?
Paying in checkout page

## Testing / Proof
Screen Shot 2020-04-10 at 2 53 29 PM

### DIFF
--- a/src/common/overlay/overlay-style.ts
+++ b/src/common/overlay/overlay-style.ts
@@ -1,0 +1,83 @@
+interface OverlayStyleOptions {
+    background?: string;
+    id?: string;
+    transitionDuration?: number;
+    classLayout?: string;
+    classOverlayText?: string;
+    classClose?: string;
+}
+
+export default function getOverlayStyle(options: OverlayStyleOptions): HTMLElement {
+    const { id, background, transitionDuration, classLayout, classOverlayText, classClose } = options;
+    const styles = document.createElement('style');
+    const addClassLayout = classLayout ? `, .${classLayout}` : '';
+
+    styles.id = `${id}--styles`;
+    styles.type = 'text/css';
+    styles.innerText = `
+        #${id}${addClassLayout} {
+            display: block;
+            height: 100%;
+            width: 100%;
+            left: 0;
+            top: 0;
+            position: fixed;
+            z-index: 2147483647;
+        }
+        #${id} {
+            transition: opacity ${transitionDuration}ms ease-out;
+            background: ${background};
+            opacity: 0;
+        }
+    `;
+
+    if (classOverlayText) {
+        styles.innerText += `
+             #${id} {
+                display: flex;
+                align-items: center;
+                justify-content: center;
+             }
+             #${id} .${classOverlayText} {
+                color: white;
+                max-width: 330px;
+                font-size: 1.2em;
+                text-align: center;
+             }
+        `;
+    }
+
+    if (classClose) {
+        styles.innerText += `
+            #${id} {
+                opacity: 1;
+            }
+            .${classLayout} .${classClose} {
+                position: fixed;
+                right: 16px;
+                top: 16px;
+                width: 16px;
+                height: 16px;
+                opacity: 0.6;
+                cursor: pointer;
+                z-index: 3147483647;
+            }
+            .${classLayout} .${classClose}::after, .${classLayout} .${classClose}::before {
+                position: absolute;
+                left: 8px;
+                content: '';
+                height: 16px;
+                width: 2px;
+                background-color: #fff;
+            }
+            .${classLayout} .${classClose}::after {
+                transform: rotate(-45deg);
+            }
+            .${classLayout} .${classClose}::before {
+                transform: rotate(45deg);
+            }
+        `;
+    }
+
+    return styles;
+}
--- a/src/common/overlay/overlay.ts
+++ b/src/common/overlay/overlay.ts
@@ -1,36 +1,54 @@
+import getOverlayStyle from './overlay-style';
+
 export interface OverlayOptions {
     background?: string;
     id?: string;
     transitionDuration?: number;
+    hasCloseButton?: boolean;
+    innerHtml?: HTMLElement | DocumentFragment;
 }
 
 export interface OverlayShowOptions {
     onClick?(event: MouseEvent): void;
+    onClickClose?(event: MouseEvent): void;
+}
+
+export interface Elements {
+    element: HTMLElement;
+    mainElement: HTMLElement;
+    closeElement?: HTMLElement;
 }
 
 export default class Overlay {
     private _element: HTMLElement;
+    private _mainElement: HTMLElement;
+    private _closeElement?: HTMLElement;
     private _unregisterClick?: () => void;
 
     constructor(options?: OverlayOptions) {
-        this._element = this._createElement(options);
+        const { element, closeElement, mainElement } = this._createElements(options);
+
+        this._element = element;
+        this._closeElement = closeElement;
+        this._mainElement = mainElement;
     }
 
     show(options?: OverlayShowOptions): void {
-        if (this._element.parentElement) {
+
+        if (this._mainElement.parentElement) {
             return;
         }
 
         this._registerClick(options);
 
-        document.body.appendChild(this._element);
+        document.body.appendChild(this._mainElement);
 
         // Fade In
         setTimeout(() => this._element.style.opacity = '1');
     }
 
     remove(): void {
-        if (!this._element.parentElement) {
+        if (!this._mainElement.parentElement) {
             return;
         }
 
@@ -43,30 +61,66 @@ export default class Overlay {
         setTimeout(() => this._element.style.opacity = '0');
     }
 
-    private _createElement(options?: OverlayOptions): HTMLElement {
+    private _createElements(options?: OverlayOptions): Elements {
         const element = document.createElement('div');
         const {
             background = 'rgba(0, 0, 0, 0.8)',
-            id = null,
+            id = 'checkoutOverlay',
             transitionDuration = 400,
+            hasCloseButton,
+            innerHtml,
         } = options || {};
-
-        element.style.background = background;
-        element.style.display = 'block';
-        element.style.height = '100%';
-        element.style.left = '0px';
-        element.style.opacity = '0';
-        element.style.position = 'fixed';
-        element.style.top = '0px';
-        element.style.transition = `opacity ${transitionDuration}ms ease-out`;
-        element.style.width = '100%';
-        element.style.zIndex = '2147483647';
+        let mainElement = element;
+        let closeElement: HTMLElement | undefined;
+        let classLayout: string | undefined;
+        let classClose: string | undefined;
+        let classOverlayText: string | undefined;
 
         if (id) {
             element.id = id;
         }
 
-        return element;
+        if (innerHtml) {
+            const overlayText = document.createElement('div');
+            classOverlayText = `${id}--overlayText`;
+            overlayText.className = classOverlayText;
+            overlayText.appendChild(innerHtml);
+
+            element.appendChild(overlayText);
+        }
+
+        if (hasCloseButton) {
+            classClose = `${id}--close`;
+            closeElement = document.createElement('div');
+            closeElement.className =  classClose;
+
+            classLayout = `${id}--layout`;
+            mainElement = document.createElement('div');
+            mainElement.className = classLayout;
+
+            mainElement.appendChild(element);
+            mainElement.appendChild(closeElement);
+        }
+
+        mainElement.appendChild(getOverlayStyle({ id, background, transitionDuration, classLayout, classOverlayText, classClose }));
+
+        return { element, closeElement, mainElement };
+    }
+
+    private _addEventListener(element?: HTMLElement, onClick?: (event: MouseEvent) => void): void {
+        if (!element || !onClick) {
+           return;
+        }
+
+        element.addEventListener('click', onClick);
+    }
+
+    private _removeEventListener(element?: HTMLElement, onClick?: (event: MouseEvent) => void): void {
+        if (!element || !onClick) {
+            return;
+        }
+
+        element.removeEventListener('click', onClick);
     }
 
     private _registerClick(options?: OverlayShowOptions): void {
@@ -74,32 +128,34 @@ export default class Overlay {
             this._unregisterClick();
         }
 
-        if (options && options.onClick) {
-            const { onClick } = options;
+        if (options) {
+            const { onClick, onClickClose } = options;
 
-            this._element.addEventListener('click', onClick);
+            this._addEventListener(this._element, onClick);
+            this._addEventListener(this._closeElement, onClickClose);
 
             this._unregisterClick = () => {
-                this._element.removeEventListener('click', onClick);
+                this._removeEventListener(this._element, onClick);
+                this._removeEventListener(this._closeElement, onClickClose);
                 this._unregisterClick = undefined;
             };
         }
     }
 
     private _removeAfterTransition(): void {
-        const handeTransition: (event: Event) => void = event => {
+        const handleTransition: (event: Event) => void = event => {
             // NOTE: `event` is not correctly typed in this version of TS
             if ((event as TransitionEvent).propertyName !== 'opacity') {
                 return;
             }
 
-            if (this._element.parentElement) {
-                this._element.parentElement.removeChild(this._element);
+            if (this._mainElement.parentElement) {
+                this._mainElement.remove();
             }
 
-            this._element.removeEventListener('transitionend', handeTransition);
+            this._element.removeEventListener('transitionend', handleTransition);
         };
 
-        this._element.addEventListener('transitionend', handeTransition);
+        this._element.addEventListener('transitionend', handleTransition);
     }
 }
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.spec.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.spec.ts
@@ -1,0 +1,51 @@
+import { MissingDataError } from '../../../common/error/errors';
+
+import { PaypalCommercePaymentProcessor } from './index';
+
+// tslint:disable:no-non-null-assertion
+describe('PaypalCommercePaymentProcessor', () => {
+    let paypalCommercePaymentProcessor: PaypalCommercePaymentProcessor;
+    const focus = jest.fn();
+    const close = jest.fn();
+
+    beforeEach(() => {
+        window.open = jest.fn();
+
+        paypalCommercePaymentProcessor = new PaypalCommercePaymentProcessor();
+    });
+
+    it('call window.open', async () => {
+        try {
+            await paypalCommercePaymentProcessor.paymentPayPal('approveUrl');
+        } catch (error) {
+            expect(window.open).toHaveBeenCalled();
+        }
+    });
+
+    it('throw error when window closed', async () => {
+        await expect(paypalCommercePaymentProcessor.paymentPayPal('approveUrl')).rejects.toThrow(MissingDataError);
+    });
+
+    it('check event listener on message', async () => {
+        window.open = jest.fn(() => ({ focus, close }));
+        const map: {[key: string]: any}  = {};
+
+        window.addEventListener = jest.fn((event, cb) => {
+            if (event === 'message') {
+                setTimeout(() => {
+                    cb({
+                        origin: 'https://www.paypal.com',
+                        data: JSON.stringify({ operation: 'return_to_merchant', updateParent: true }),
+                    });
+                });
+            }
+
+            map[event] = cb;
+        });
+
+        const res = await paypalCommercePaymentProcessor.paymentPayPal('approveUrl');
+
+        expect(res).toEqual(true);
+    });
+});
+// tslint:enable:no-non-null-assertion
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
@@ -1,0 +1,111 @@
+import { MissingDataError, MissingDataErrorType } from '../../../common/error/errors';
+import { Overlay } from '../../../common/overlay';
+
+const modalWidth = 450;
+const modalHeight = 600;
+
+export interface ProcessorOptions {
+    overlay?: {
+        helpText?: string;
+        continueText?: string;
+    };
+}
+
+export default class PaypalCommercePaymentProcessor {
+    private _window = window;
+    private _popup?: WindowProxy | null;
+    private _overlay?: Overlay;
+
+    constructor() {}
+
+    initialize({ overlay }: ProcessorOptions) {
+        this._overlay = new Overlay({ hasCloseButton: true, innerHtml: this._getOverlayElements(overlay) });
+    }
+
+    paymentPayPal(approveUrl: string): Promise<boolean> {
+        return new Promise((resolve, reject) => {
+            const paramsWindow =  this._getParamsWindow();
+
+            const closeWindow = (isResolve: boolean, isRemoveOverlay: boolean = true) => {
+                this._window.removeEventListener('message', messageHandler);
+
+                if (this._popup) {
+                    this._popup.close();
+                    this._popup = undefined;
+                }
+
+                if (isRemoveOverlay && this._overlay) {
+                    this._overlay.remove();
+                }
+
+                isResolve
+                    ? resolve(true)
+                    : reject(new MissingDataError(MissingDataErrorType.MissingPayment));
+            };
+
+            const messageHandler = (event: MessageEvent) => {
+                if (event.origin !== 'https://www.sandbox.paypal.com' && event.origin !== 'https://www.paypal.com') {
+                    return;
+                }
+
+                const data = JSON.parse(event.data);
+
+                if (data.operation === 'return_to_merchant' && data.updateParent) {
+                    this._window.removeEventListener('message', messageHandler);
+                    closeWindow(true);
+                }
+            };
+
+            this._window.addEventListener('message', messageHandler);
+            this._popup = this._window.open(approveUrl, 'PPFrame', paramsWindow);
+
+            const popupTick = setInterval(() => {
+                if (!this._popup || this._popup.closed) {
+                    clearInterval(popupTick);
+
+                    closeWindow(false);
+                }
+            }, 500);
+
+            if (this._overlay) {
+                this._overlay.show({
+                    onClick: () => this._popup ? this._popup.focus() : closeWindow(false),
+                    onClickClose: () => closeWindow(false, false),
+                });
+            }
+        });
+    }
+
+    deinitialize(): void {
+        this._overlay = undefined;
+    }
+
+    private _getOverlayElements(options: ProcessorOptions['overlay'] = {}): DocumentFragment {
+        const fragment = document.createDocumentFragment();
+        const helpText = document.createElement('div');
+        const continueText = document.createElement('strong');
+
+        helpText.className = 'paypal-commerce-overlay_text';
+        helpText.innerText = options.helpText || 'Don\'t see the secure PayPal browser? We\'ll help you re-launch the window to complete your flow. You might need to enable pop-ups in your browser in order to continue.';
+
+        continueText.className = 'paypal-commerce-overlay_link';
+        continueText.innerText = options.continueText || 'Click to continue';
+        continueText.style.marginTop = '15px';
+        continueText.style.display = 'block';
+        continueText.style.color = 'white';
+        continueText.style.textDecoration = 'underline';
+
+        fragment.appendChild(helpText);
+        fragment.appendChild(continueText);
+
+        return fragment;
+    }
+
+    private _getParamsWindow(): string {
+        return `
+            left=${Math.round((window.screen.height - modalWidth) / 2)},
+            top=${Math.round((window.screen.width - modalHeight) / 2)},
+            height=${modalHeight},width=${modalWidth},status=yes,toolbar=no,menubar=no,resizable=yes,scrollbars=no
+        `;
+    }
+}
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
@@ -1,42 +1,46 @@
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
-import { MissingDataError, MissingDataErrorType } from '../../../common/error/errors';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
 import { PaymentArgumentInvalidError } from '../../errors';
 import PaymentActionCreator from '../../payment-action-creator';
-import { PaymentRequestOptions } from '../../payment-request-options';
+import { PaymentInitializeOptions, PaymentRequestOptions } from '../../payment-request-options';
 import PaymentStrategy from '../payment-strategy';
+
+import { PaypalCommercePaymentProcessor, PaypalCommerceRequestSender } from './index';
 
 export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
 
     constructor(
         private _store: CheckoutStore,
         private _orderActionCreator: OrderActionCreator,
-        private _paymentActionCreator: PaymentActionCreator
+        private _paymentActionCreator: PaymentActionCreator,
+        private _paypalCommerceRequestSender: PaypalCommerceRequestSender,
+        private _paypalCommercePaymentProcessor: PaypalCommercePaymentProcessor
     ) {}
 
-    initialize(): Promise<InternalCheckoutSelectors> {
+    initialize({ paypalcommerce }: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
+        this._paypalCommercePaymentProcessor.initialize({ overlay: paypalcommerce && paypalcommerce.overlay });
+
         return Promise.resolve(this._store.getState());
     }
 
     async execute(payload: OrderRequestBody, options: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
         const { payment, ...order } = payload;
-        const paymentMethod = this._store.getState().paymentMethods.getPaymentMethod(options.methodId);
+        const state = this._store.getState();
+        const paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(options.methodId);
 
         if (!payment) {
             throw new PaymentArgumentInvalidError(['payment']);
         }
 
-        if (!paymentMethod || !paymentMethod.initializationData.orderId) {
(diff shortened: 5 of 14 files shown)


src/payment/strategies/paypal-commerce/paypal-commerce-overlay.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    beforeEach(() => {
+        overlay = new OverlayPaypal();
+    });
+
+    afterEach(() => {
+        const element = document.getElementById(idOverlay);
+
+        if (element) {
+            element.parentElement!.removeChild(element);
+        }
```
golcinho, line 16: I think it could a little bit simpler

```
const element = document.getElementById(idOverlay);

if (element) {
    element.remove();
}
```

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,67 @@
...
+                    closeWindow(false);
+                }
+            }, 500);
+
+            this._overlay.show({
+                onClick: () => {
+                    this._popup
+                        ? this._popup.focus()
+                        : closeWindow(false);
+                },
```
golcinho, line 63: 🍺 Could be simple

```
onClick: () => this._popup ? this._popup.focus() : closeWindow(false),
```

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,67 @@
+import { MissingDataError, MissingDataErrorType } from '../../../common/error/errors';
+
+import { OverlayPaypal } from './index';
+
+const widthWindow = 450;
+const heightWindow = 600;
```
golcinho, line 6: 🍺 

- `widthWindow` => `modalWidth` OR `windowWidth`
- `heightWindow` => `modalHeight` OR `windowHeight`

I prefer to use `modal` as a name instead of `window`.
- `window` name related to the global window object so, this name could a little bit confuse;

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,67 @@
...
+                    if (data.operation === 'return_to_merchant' && data.updateParent) {
+                        this._window.removeEventListener('message', handlerMessage);
+                        closeWindow(true);
+                    }
+                }
+            };
+
+            this._window.addEventListener('message', handlerMessage);
+
+            const params = `
```
golcinho, line 43: 🍺 I think you could move this variable outside the class like `widthWindow` variable.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+    const close = jest.fn();
+
+    beforeEach(() => {
+        window.open = jest.fn(() => {
+            return { focus, close };
+        });
+
+        overlay = new OverlayPaypal();
+        paypalCommercePaymentProcessor = new PaypalCommercePaymentProcessor(overlay);
+    });
```
golcinho, line 19: The current implementation of `window.open` need only on the last `it` so, I recommend:

1)
```
beforeEach(() => {
    window.open = jest.fn();

    overlay = new OverlayPaypal();
    paypalCommercePaymentProcessor = new PaypalCommercePaymentProcessor(overlay);
});
```

2)
```
it('check event listener on message', async () => {
    window.open. mockImplementation(() => ({ focus, close }));
....
})
```

src/payment/strategies/paypal-commerce/paypal-commerce-overlay.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+        expect(text!.innerText).toBe('Don\'t see the secure PayPal browser? We\'ll help you re-launch the window to complete your flow. You might need to enable pop-ups in your browser in order to continue.');
+    });
+
+    it('shows link element', () => {
+        overlay.show();
+
+        const link: HTMLElement | null = document.querySelector(`#${idOverlay} .paypal-commerce-overlay_link`);
+
+        expect(link).toBeTruthy();
+        expect(link!.innerText).toBe('Continue');
```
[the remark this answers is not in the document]
helga1507: Yes, but in this case it will be `null`

src/payment/strategies/paypal-commerce/paypal-commerce-overlay.spec.ts
helga1507: Thank you!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
helga1507: Thank you!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
helga1507: I agree. Thank you

src/payment/strategies/paypal-commerce/paypal-commerce-overlay.spec.ts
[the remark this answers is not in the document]
golcinho: okay, thank you!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.spec.ts
helga1507: Thank you!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -1,42 +1,62 @@
...
-    initialize(): Promise<InternalCheckoutSelectors> {
+    async initialize(): Promise<InternalCheckoutSelectors> {
         return Promise.resolve(this._store.getState());
     }
 
     async execute(payload: OrderRequestBody, options: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
         const { payment, ...order } = payload;
-        const paymentMethod = this._store.getState().paymentMethods.getPaymentMethod(options.methodId);
+        const state = this._store.getState();
+        const paymentMethod = state.paymentMethods.getPaymentMethod(options.methodId);
```
davidchin, line 29: Maybe use `getPaymentMethodOrThrow` instead?

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -1,42 +1,62 @@
...
-        if (!paymentMethod || !paymentMethod.initializationData.orderId) {
-            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        if (!paymentMethod) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+        }
+
+        let { orderId } = paymentMethod.initializationData;
+
+        if (!orderId) {
+            const cart = state.cart.getCart();
```
davidchin, line 42: Perhaps you can add a `getCartOrThrow` method to `CartSelector` to simplify this.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -1,42 +1,62 @@
...
+
+            if (!cart) {
+                throw new MissingDataError(MissingDataErrorType.MissingCart);
+            }
+            const { approveUrl, orderId: createdOrderID } = await this._paypalCommerceRequestSender.setupPayment(cart.id);
+
+            await this._paypalCommercePaymentProcessor.paymentPayPal(approveUrl);
+
+            orderId = createdOrderID;
```
davidchin, line 51: Might be cleaner if you have a `getPaypalOrderId` method that can give you back the order ID. In that method you can return early if it is already available, that way you don't need to deal with these temporary variables. 🍹

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -1,42 +1,62 @@
...
 
-        if (!paymentMethod || !paymentMethod.initializationData.orderId) {
-            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        if (!paymentMethod) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+        }
+
+        let { orderId } = paymentMethod.initializationData;
+
+        if (!orderId) {
```
davidchin, line 41: I assume when you call `setupPayment`, the server will go and talk to PayPal, get an order ID from them, and set it in the session. That order ID will then be surfaced through `initializationData`. So what happens if the shopper doesn't complete whatever they have to do in the PayPal modal? It seems when this method is called again, because the order ID is already defined, we won't be showing the modal again to the shopper?

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -1,42 +1,62 @@
...
+            const cart = state.cart.getCart();
+
+            if (!cart) {
+                throw new MissingDataError(MissingDataErrorType.MissingCart);
+            }
+            const { approveUrl, orderId: createdOrderID } = await this._paypalCommerceRequestSender.setupPayment(cart.id);
+
+            await this._paypalCommercePaymentProcessor.paymentPayPal(approveUrl);
```
davidchin, line 49: Perhaps you can inject `PaypalCommerceRequestSender` into `PaypalCommercePaymentProcessor`. With that, you can just provide a cart ID into this method to start the Paypal flow (considering the processor class is a higher level abstraction for Paypal modal).

src/payment/strategies/paypal-commerce/paypal-commerce-overlay.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+        if (element) {
+            element.remove();
+        }
+    });
+
+    it('shows modal element', () => {
+        overlay.show();
+
+        const modal: HTMLElement | null = document.querySelector(`#${idOverlay} .paypal-commerce-overlay_modal`);
```
capsula4, line 22: I dont think you need to specify the type here. It can be inferred from what's returned by querySelector()

src/payment/strategies/paypal-commerce/paypal-commerce-overlay.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+        expect(modal).toBeTruthy();
+    });
+
+    it('shows text element', () => {
+        overlay.show();
+
+        const text: HTMLElement | null = document.querySelector(`#${idOverlay} .paypal-commerce-overlay_text`);
```
capsula4, line 30: same here.

src/payment/strategies/paypal-commerce/paypal-commerce-overlay.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+        expect(text).toBeTruthy();
+        expect(text!.innerText).toBe('Don\'t see the secure PayPal browser? We\'ll help you re-launch the window to complete your flow. You might need to enable pop-ups in your browser in order to continue.');
+    });
+
+    it('shows link element', () => {
+        overlay.show();
+
+        const link: HTMLElement | null = document.querySelector(`#${idOverlay} .paypal-commerce-overlay_link`);
```
capsula4, line 39: same

src/payment/strategies/paypal-commerce/paypal-commerce-overlay.ts
```diff
@@ -0,0 +1,42 @@
...
+    constructor() {
+        super({ id: idOverlay });
+    }
+
+    show(options?: OverlayShowOptions) {
+        super.show(options);
+        const element = document.getElementById(idOverlay);
+
+        if (element) {
+            const modal = document.createElement('div');
```
capsula4, line 16: pretty minor, but in order to avoid the indentation of the whole block, you could do instead

```js
if (!element) {
  return;
}

const text = document.createElement('div');
const link = document.createElement('a');
// ...
```

src/payment/strategies/paypal-commerce/paypal-commerce-overlay.ts
```diff
@@ -0,0 +1,42 @@
...
+            modal.style.position = 'fixed';
+            modal.style.top = '50%';
+            modal.style.left = '50%';
+            modal.style.transform = 'translate(-50%, -50%)';
+            modal.style.color = 'white';
+            modal.style.maxWidth = '330px';
+            modal.style.fontSize = '1.3em';
+
+            text.className = 'paypal-commerce-overlay_text';
+            text.innerText = 'Don\'t see the secure PayPal browser? We\'ll help you re-launch the window to complete your flow. You might need to enable pop-ups in your browser in order to continue.';
```
capsula4, line 30: is there a way to provide a way to customize this text? Might be part of `options`. Its quite likely that it might need to be translated.

src/payment/strategies/paypal-commerce/paypal-commerce-overlay.ts
```diff
@@ -0,0 +1,42 @@
...
+
+            link.className = 'paypal-commerce-overlay_link';
+            link.innerText = 'Continue';
```
capsula4, line 33: same as above ^

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+        try {
+            await paypalCommercePaymentProcessor.paymentPayPal('approveUrl');
+        } catch (error) {
+            expect(window.open).toHaveBeenCalled();
+        }
+    });
+
+    it('throw error when window closed', async () => {
+        try {
+            await paypalCommercePaymentProcessor.paymentPayPal('approveUrl');
```
capsula4, line 29: for testing assertions, please prefer 
```js
await expect(paypalCommercePaymentProcessor.paymentPayPal('approveUrl')).rejects.toThrow(MissingDataError)
```
or
```js
expect(async () => await paypalCommercePaymentProcessor.paymentPayPal('approveUrl'))
  .toThrow(MissingDataError);
```

using try/catch should be avoided

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export default class PaypalCommercePaymentProcessor {
+    private _window = window;
+    private _popup?: Window | null;
+
+    constructor(
+        private _overlay: OverlayPaypal
+    ) {}
+
+    paymentPayPal(approveUrl: string): Promise<any> {
```
capsula4, line 16: Promise<void> ?

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,62 @@
...
+        return new Promise((resolve, reject) => {
+            const closeWindow = (isResolve: boolean) => {
+                this._overlay.remove();
+
+                if (this._popup) {
+                    this._popup.close();
+                }
+
+                isResolve
+                    ? resolve(true)
```
capsula4, line 26: `resolve()`
I don't think you care about the boolean? If it resolves, it means it succeeded.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,62 @@
...
+    private _window = window;
+    private _popup?: Window | null;
+
+    constructor(
+        private _overlay: OverlayPaypal
+    ) {}
+
+    paymentPayPal(approveUrl: string): Promise<any> {
+        return new Promise((resolve, reject) => {
+            const closeWindow = (isResolve: boolean) => {
```
capsula4, line 18: instead of defining functions inside a function, could you instead define them as private functions?

`_closeWindow`
`_ handlerMessage`

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,62 @@
...
+
+                    if (data.operation === 'return_to_merchant' && data.updateParent) {
+                        this._window.removeEventListener('message', handlerMessage);
+                        closeWindow(true);
+                    }
+                }
+            };
+
+            this._window.addEventListener('message', handlerMessage);
+            const paramsModalWindow = `
```
capsula4, line 42: maybe you could define this in a more readable way?
modalWindowParams = [
  [ 'left',  Math.round((window.screen.height - modalWidth) / 2) ],
  [ 'top', Math.round((window.screen.width - modalHeight) / 2) ],
  // ...
].map(value => value.join('=')).join(',');

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,62 @@
...
+                left=${Math.round((window.screen.height - modalWidth) / 2)},
+                top=${Math.round((window.screen.width - modalHeight) / 2)},
+                height=${modalHeight},width=${modalWidth},status=yes,toolbar=no,menubar=no,resizable=yes,scrollbars=no
+            `;
+            this._popup = this._window.open(approveUrl, 'PPFrame', paramsModalWindow);
+
+            const popupTick = setInterval(() => {
```
capsula4, line 49: what's the purpose of this interval?

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
capsula4, line 18: maybe the param should be something like `_closeWindow(shouldRedirect: boolean)`
This way, it's easier to know what to pass to this method / what is expected to happen after its called

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,62 @@
+import { MissingDataError, MissingDataErrorType } from '../../../common/error/errors';
+
+import { OverlayPaypal } from './index';
+
+const modalWidth = 450;
```
capsula4, line 5: is there a way to customize this? Have you tested how this works in mobile?

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
helga1507: Hm, thank you for your suggest!
I tried to do it, but thought it's not so elegantly

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
helga1507: We should check window. It's one way to catch closing window

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
helga1507: But it's not about redirect. This function will stop working and window will close anyway
Why shouldRedirect?

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
helga1507: In mobile window open in new tab. And it's okay for this case

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
helga1507: Hm, I'm not sure. For me it's quite readable. And in your case array was created, after this was iterated over and joined. 
It's so costly for just one string

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,62 @@
...
+                }
+            };
+
+            this._window.addEventListener('message', handlerMessage);
+            const paramsModalWindow = `
+                left=${Math.round((window.screen.height - modalWidth) / 2)},
+                top=${Math.round((window.screen.width - modalHeight) / 2)},
+                height=${modalHeight},width=${modalWidth},status=yes,toolbar=no,menubar=no,resizable=yes,scrollbars=no
+            `;
+            this._popup = this._window.open(approveUrl, 'PPFrame', paramsModalWindow);
```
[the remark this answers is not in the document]
helga1507: Yes, it's okay for this case

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,62 @@
...
+        private _overlay: OverlayPaypal
+    ) {}
+
+    paymentPayPal(approveUrl: string): Promise<any> {
+        return new Promise((resolve, reject) => {
+            const closeWindow = (isResolve: boolean) => {
+                this._overlay.remove();
+
+                if (this._popup) {
+                    this._popup.close();
```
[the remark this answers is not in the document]
helga1507: Yes. Thank you!

src/payment/strategies/paypal-commerce/paypal-commerce-overlay.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const idOverlay = 'paypal-commerce-overlay';
+
+export default class OverlayPaypal extends Overlay {
+
+    constructor() {
+        super({ id: idOverlay });
+    }
+
+    show(options?: OverlayShowOptions) {
```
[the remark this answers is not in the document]
helga1507: Okay.
I wasn't sure that it okay. But I prefer to add that functionality to the existing Overlay
 Thank you!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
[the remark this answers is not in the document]
helga1507: Sure, thank you!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,62 @@
...
+                }
+
+                isResolve
+                    ? resolve(true)
+                    : reject(new MissingDataError(MissingDataErrorType.MissingPayment));
+            };
+
+            const handlerMessage = (event: MessageEvent) => {
+                if (event.origin.includes('sandbox.paypal.com')) {
```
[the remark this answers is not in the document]
helga1507: You are right. Thank you!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,62 @@
+import { MissingDataError, MissingDataErrorType } from '../../../common/error/errors';
+
+import { OverlayPaypal } from './index';
+
+const modalWidth = 450;
+const modalHeight = 600;
+
+export default class PaypalCommercePaymentProcessor {
+    private _window = window;
+    private _popup?: Window | null;
```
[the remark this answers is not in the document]
helga1507: Okay, thank you!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
helga1507: Yes, thank you!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
helga1507: Great! I'll add it. Thank you

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
helga1507: In the modal user approves payment in PayPal. If  it doesn't complete, paymentPayPal throw error and user sees error in the standard modal window
`setupPayment` should be called immediately before call paymentPayPal for actual data

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
helga1507: I rewrite `PaypalCommercePaymentProcessor`  and now I'm not sure that it's good idea.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.spec.ts
helga1507: Thank you!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
helga1507: I implemented this code, but it doesn't work well. In my implementation after `reject` user sees standard modal window with error and in your suggestion modal window doesn't showed. So I prefer to leave my code

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,91 @@
...
+                    this._overlay.remove();
+                }
+
+                isResolve
+                    ? resolve(true)
+                    : reject(new MissingDataError(MissingDataErrorType.MissingPayment));
+            };
+
+            const messageHandler = (event: MessageEvent) => {
+                if (event.origin === 'https://www.sandbox.paypal.com' || event.origin === 'https://www.paypal.com') {
```
davidchin, line 39: I think here you can also benefit from using a guard statement (to reduce the level of nesting). i.e.:

```ts
if (event.origin !== 'https://www.sandbox.paypal.com' && event.origin !== 'https://www.paypal.com') {
    return;
}
```

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,91 @@
...
+            });
+        });
+    }
+
+    private _getOverlayElements(): HTMLElement[] {
+        const text = document.createElement('div');
+        const link = document.createElement('a');
+
+        text.className = 'paypal-commerce-overlay_text';
+        text.innerText = 'Don\'t see the secure PayPal browser? We\'ll help you re-launch the window to complete your flow. You might need to enable pop-ups in your browser in order to continue.';
```
davidchin, line 72: We might want to pass in the text from the client application instead of hard-coding it inside the SDK (through `initialize` options). This is so that we can support non-English text.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,91 @@
...
+
+        link.className = 'paypal-commerce-overlay_link';
+        link.innerText = 'Click to continue';
+        link.style.marginTop = '15px';
+        link.style.display = 'block';
+        link.style.color = 'white';
+        link.style.textDecoration = 'underline';
+
+        return [ text, link ];
```
davidchin, line 81: It might be a good idea to return `DocumentFragment` (you can create one using `document.createDocumentFragment()`). With that, you can call `appendChild` directly on the fragment and you don't need to do a `forEach` to append one by one.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -0,0 +1,91 @@
...
+            this._overlay.show({
+                onClick: () => this._popup ? this._popup.focus() : closeWindow(false),
+                onClickClose: () => closeWindow(false, false),
+            });
+        });
+    }
+
+    private _getOverlayElements(): HTMLElement[] {
+        const text = document.createElement('div');
+        const link = document.createElement('a');
```
davidchin, line 69: Does the link element do anything? It doesn't seem to have a `href` or a click listener.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -1,42 +1,44 @@
...
         private _orderActionCreator: OrderActionCreator,
-        private _paymentActionCreator: PaymentActionCreator
+        private _paymentActionCreator: PaymentActionCreator,
+        private _paypalCommerceRequestSender: PaypalCommerceRequestSender,
+        private _paypalCommercePaymentProcessor: PaypalCommercePaymentProcessor
     ) {}
 
-    initialize(): Promise<InternalCheckoutSelectors> {
+    async initialize(): Promise<InternalCheckoutSelectors> {
         return Promise.resolve(this._store.getState());
```
davidchin, line 22: You can just write `return` here. It's automatically wrapped in a promise inside an `async` method.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -53,4 +55,14 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
     deinitialize(): Promise<InternalCheckoutSelectors> {
         return Promise.resolve(this._store.getState());
     }
+
+    private async _getOrderId(): Promise<string> {
+        const state = this._store.getState();
+        const cart = state.cart.getCartOrThrow();
+        const { approveUrl, orderId } = await this._paypalCommerceRequestSender.setupPayment(cart.id);
```
davidchin, line 62: I think I asked this question before but I'm not sure if I understand this correctly. I assume after this `setupPayment` call, an order ID is set in the session. Imagine the shopper cancels the PayPal modal (triggered by the next line `paymentPayPal`), and they refresh the page in order to try again. This time, because `initializationData.orderId` is defined, `getOrderId` won't be called again and therefore won't trigger the `paymentPayPal` modal?

src/common/overlay/overlay.ts
```diff
@@ -6,31 +8,52 @@ export interface OverlayOptions {
...
 export default class Overlay {
     private _element: HTMLElement;
+    private _styles: HTMLElement;
+    private _closeElement?: HTMLElement;
+    private _layoutElement?: HTMLElement;
     private _unregisterClick?: () => void;
 
-    constructor(options?: OverlayOptions) {
-        this._element = this._createElement(options);
+    constructor(options?: OverlayOptions, isCloseElement?: boolean, innerHtml?: HTMLElement | HTMLElement[]) {
```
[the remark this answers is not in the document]
helga1507: Okay. Thank you!

src/common/overlay/overlay-style.ts
```diff
@@ -0,0 +1,82 @@
...
+    `;
+
+    if (classModal) {
+        styles.innerText += `
+             #${id} {
+                display: flex;
+                align-items: center;
+                justify-content: center;
+             }
+             .${classModal} {
```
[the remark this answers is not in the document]
helga1507: I agree. Thank you!

src/common/overlay/overlay-style.ts
```diff
@@ -0,0 +1,82 @@
...
+             }
+        `;
+    }
+
+    if (classClose) {
+        styles.innerText += `
+            #${id} {
+                opacity: 1;
+            }
+            .${classClose} {
```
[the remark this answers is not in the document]
helga1507: Yes, you are right
But `classClose` locate in `classLayout`, so I wrote `.${classLayout} .${classClose}`. And in other case I added `#${id}`

src/common/overlay/overlay.ts
```diff
@@ -6,31 +8,52 @@ export interface OverlayOptions {
...
+        document.body.appendChild(element);
+        document.body.appendChild(this._styles);
 
         // Fade In
         setTimeout(() => this._element.style.opacity = '1');
     }
 
     remove(): void {
-        if (!this._element.parentElement) {
+        const element = this._layoutElement || this._element;
```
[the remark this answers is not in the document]
helga1507: I understand your doubt. But we need both components. In this case we have `element` (it's darkened div with click event) and `layoutElement` (it needs only for wrapper `element` and `closeElement`), that  are needed for different purpose. If it would be one element we add event listener to `layoutElement ` and click on the `closeElement` won't work

In `remove()` we should remove `element` (if overlay without closeElement) or `layoutElement` (if overlay with closeElement), so I decided to create `mainElement`, that will be `element` or `layout` and remove `styleElement` (I added it to `mainElement`)

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
helga1507: Order ID isn't set in the session. When shopper refresh the page `orderId` is undefined and this functional repeated

helga1507:
Thank you @davidchin for your review!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
BC-ACherednichenko: @davidchin it is a good question, currently, payment ID passed to the Payment checkout in `onApprove` action https://github.com/bigcommerce/checkout-sdk-js/blob/2cc3cb83ce88a2aa8da0231c9425b124b892e29c/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts#L68, only after customer approved payment session on PayPal side. 
However, we are going to revisit this workflow soon enough in the scope of ticket https://jira.bigcommerce.com/browse/PAYPAL-307, and it will work in the way you described (payment ID will be saved into the session during `setupPayment` call)

BC-ACherednichenko:
@davidchin thanks for your review, could you revisit it again, please

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
davidchin: Right I see. Yes, so the thing that I'm confused with is what happens when `POST /api/storefront/payment/paypalcommerce` (`setupPayment`) is called. From I can I see, the server gets an order ID from PayPal, and then sets it in the session before returning to the client. If that's the case, then we'll hit the problem I described above. But perhaps I'm looking at the wrong part of the code. 🤷‍♂️Anyway, I'll message you this afternoon and we can clear this up :)

BC-ACherednichenko: @davidchin I have tried the case you mentioned above, and you are absolutely right, good catch! 
I have removed part of the code that saves ID into the session, this part of early refactoring that I mentioned before can cause issues, because of it was added before Checkout flow and after Cart flow, it passed without any issues. The flow on the FE for this feature works well and should not cause any issues with a session, however, once we decide to continue refactoring of the way we treat order id it will require changes on FE.

PR #835, opened 2020-04-16
refactor(payment): INT-2464 rename the method id to amazonpay

rodna-bc:
## What? [INT-2464](https://jira.bigcommerce.com/browse/INT-2464)
The payment provider changed the name of the new version of amazon and they do not want any url with the old name.

THIS IS A MANDATORY CHANGE, THE PROVIDER ONLY WANT THE NAME OF `amazonpay` OR `amznpay` SO WE CONSIDER IS MORE CLEAR `amazonpay`

**WE DO NOT CHANGE ANY FUNCTIONALITY THAT WAS ALREADY APPROVED AND MERGED, WE ONLY CHANGED THE METHOD ID NAME**

## Why?
The implementation of the new version of amazon is under development yet, so we are on time to change the method id (this method id is used in all bigcommerce urls) to use the new name.

## Sibling PRs
[BCAPP](https://github.com/bigcommerce/bigcommerce/pull/34595)
[ng-payments](https://github.com/bigcommerce-labs/ng-payments/pull/1004)
[BIGPAY](https://github.com/bigcommerce/bigpay/pull/2423)

## Testing / Proof
QA is testing all the functionality. Unit test continue passing


mmartinezintersys:
💚 Tested

QA Details:

- Regression [test suite](https://docs.google.com/spreadsheets/d/1BcTi0gbxkOJCPa83KaeGryU8qu924CdudqSrtGxj-E0/edit?usp=sharing) to verify the current functionality developed

src/checkout-buttons/strategies/checkout-button-method-type.ts
```diff
@@ -1,5 +1,5 @@
 enum CheckoutButtonMethodType {
-    AMAZON_MAXO = 'amazonmaxo',
+    AMAZON_MAXO = 'amazonpay',
```
capsula4, line 2: should we rename the ENUM too?

src/checkout-buttons/checkout-button-options.ts
```diff
@@ -21,7 +21,7 @@ export interface CheckoutButtonInitializeOptions extends CheckoutButtonOptions {
      * The options that are required to facilitate AmazonMaxo. They can be
      * omitted unless you need to support AmazonMaxo.
      */
-    amazonmaxo?: AmazonMaxoButtonInitializeOptions;
+    amazonpay?: AmazonMaxoButtonInitializeOptions;
```
icatalina, line 24: The name of this object `AmazonMaxoButtonInitializeOptions`, together with the comment above will get exposed in the sdk documentation. It is probably a good idea to change these.

capsula4: Oops, yeah missed this bit. Maybe search and replace `Maxo` with `Pay` in the entire codebase to be safe.

rodna-bc:
@icatalina Hi Ignacio! I do not changed the class names and file names because `amazonpay` was used in all the classes and files for amazon v1.  for that reason I decided to keep in amazonMaxo all the classes and file names! (In the other repositories I changed all (classes and files ) to `amazonpay` because for v1 was used `amazon`

src/checkout-buttons/strategies/checkout-button-method-type.ts
rodna-bc: done!

src/checkout-buttons/checkout-button-options.ts
rodna-bc: We have many class names and file names with AmazonMaxo because `AmazonPay` was used for amazon v1! if I change everything to `amazonpay` it will be even more confusing because v1 is also called `amazonPay` for v1 except the methodId which is just  `amazon`. The other option is rename all to amazonPayv2 but the method id must be `amazonpay` @icatalina @capsula4

icatalina: Discussed offline. We are renaming `AmazonMaxoButtonInitializeOptions` to `AmazonPayButtonInitializeOptions`. A follow up PR will rename all the internal objects from `AmazonMaxo...` to `AmazonPayV2...`.

rodna-bc:
@icatalina @capsula4 as discussed online I renamed all to amazon pay v2 😅 could you please check again? ♻️

PR #836, opened 2020-04-16
fix(checkout): CHECKOUT-4774 Handle no hosted fields to be rendered scenario

ankurbigcomm:
## What?
Handle scenario where hosted fields are not available to be rendered.

## Why?
If a merchant has configured vaulted cards with vaulted cvv code since there is no field expected to be filled out, the hosted form should not be rendered.

## Testing Proof
Refer PR https://github.com/bigcommerce/checkout-js/pull/261

### DIFF
--- a/src/payment/strategies/credit-card/credit-card-payment-strategy.spec.ts
+++ b/src/payment/strategies/credit-card/credit-card-payment-strategy.spec.ts
@@ -196,4 +196,58 @@ describe('CreditCardPaymentStrategy', () => {
                 .toHaveBeenCalledWith(loadOrderAction);
         });
     });
+
+    describe('when hosted form is enabled but hosted fields are not present for rendering', () => {
+        let form: Pick<HostedForm, 'attach' | 'submit' | 'validate'>;
+        let initializeOptions: PaymentInitializeOptions;
+        let loadOrderAction: Observable<LoadOrderSucceededAction>;
+        let state: InternalCheckoutSelectors;
+
+        beforeEach(() => {
+            form = {
+                attach: jest.fn(() => Promise.resolve()),
+                submit: jest.fn(() => Promise.resolve()),
+                validate: jest.fn(() => Promise.resolve()),
+            };
+            initializeOptions = {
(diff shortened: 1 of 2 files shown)


src/payment/strategies/credit-card/credit-card-payment-strategy.ts
```diff
@@ -98,4 +104,8 @@ export default class CreditCardPaymentStrategy implements PaymentStrategy {
 
         return isHostedPaymentFormEnabled;
     }
+
+    private _ishostedFieldAvailable(options?: PaymentInitializeOptions): boolean {
```
davidchin, line 108: Can you capitalise the `h`? 💅

src/payment/strategies/credit-card/credit-card-payment-strategy.ts
```diff
@@ -98,4 +104,8 @@ export default class CreditCardPaymentStrategy implements PaymentStrategy {
...
+        return !(values(options && options.creditCard && options.creditCard.form.fields).every(isUndefined));
```
davidchin, line 109: Should probably consider `null` as well. Perhaps `isNil` is more suitable as it covers both `undefined` and `null`.

src/payment/strategies/credit-card/credit-card-payment-strategy.spec.ts
```diff
@@ -196,4 +196,48 @@ describe('CreditCardPaymentStrategy', () => {
...
+                ));
+
+            jest.spyOn(orderActionCreator, 'loadCurrentOrder')
+                .mockReturnValue(loadOrderAction);
+
+            jest.spyOn(formFactory, 'create')
+                .mockReturnValue(form);
+        });
+
+        it('does not create hosted form', async () => {
```
davidchin, line 236: Can you also add a case to test that we're not submitting using the hosted form in this scenario?

ankurbigcomm: Done

PR #839, opened 2020-04-20
CHECKOUT-4827: Add instructions on how to load library through CDN URL

davidchin:
## What?
Add instructions on how to load the library through a CDN URL.

## Why?
We want to document how developers can use the package by loading it through CDN URL.

## Testing / Proof
None

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ The SDK has a convenient application interface for starting and completing a che
 - [Features](#features)
 - [Getting started](#getting-started)
 - [Installation](#installation)
+    - [Using NPM package](#using-npm-package)
+    - [Using CDN URL](#using-cdn-url)
 - [Requirements](#requirements)
     - [Browser support](#browser-support)
     - [Framework](#framework)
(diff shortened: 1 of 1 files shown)


davidchin:
I'll ask MT to update the Wordpress plugin to pull in the package from CDN rather than NPM using the instruction above.

README.md
```diff
@@ -65,19 +67,47 @@ The Checkout JS SDK is the easiest way to build a bespoke checkout into your sto
...
+```
+https://checkout-sdk.bigcommerce.com/v1/loader.js
+```
+
+The main benefit of using the script URL above is that your application can automatically receive backward compatible updates and bug fixes from us, without having to manually perform an upgrade.
+
+Once the above script is loaded, you can use it to load the module that you need for your application. i.e.:
+
+```js
+const module = await checkoutKitLoader.load('checkout-sdk');
```
capsula4, line 91: I'm not very familiar with this approach. Would typescript still work using the CDN approach? is `checkoutKitLoader` a global object that becomes available when referencing `https://checkout-sdk.bigcommerce.com/v1/loader.js` ?

davidchin: So this is JS only. For TS, we'll need to figure out something in the future so that we can provide types statically and the code dynamically.

Regarding your second point on `checkoutKitLoader`, yes you're correct.

PR #843, opened 2020-04-20
INT-2577 Create a strategy for Bolt

FranciscoOnt:
## What?
Create a payment strategy for bolt that allows a `nonce` to be submitted as payment.

## Why?
Bolt is a provider consumes checkout-sdk in its marketplace app, currently it submits orders and payments via the offline payment strategy and using other provider as its gateway, this strategy will allow the bolt app to submit the orders and payments to the corresponding bolt gateway .

## Testing / Proof
Unit testing.

@bigcommerce/checkout  @bigcommerce/payments  @bigcommerce/apex-integrations

### DIFF
--- a/src/payment/create-payment-strategy-registry.ts
+++ b/src/payment/create-payment-strategy-registry.ts
@@ -25,6 +25,7 @@ import { AffirmPaymentStrategy, AffirmScriptLoader } from './strategies/affirm';
 import { AfterpayPaymentStrategy, AfterpayScriptLoader } from './strategies/afterpay';
 import { AmazonPayPaymentStrategy, AmazonPayScriptLoader } from './strategies/amazon-pay';
 import { BlueSnapV2PaymentStrategy } from './strategies/bluesnapv2';
+import { BoltPaymentStrategy } from './strategies/bolt';
 import { createBraintreePaymentProcessor, createBraintreeVisaCheckoutPaymentProcessor, BraintreeCreditCardPaymentStrategy, BraintreePaypalPaymentStrategy, BraintreeScriptLoader, BraintreeSDKCreator, BraintreeVisaCheckoutPaymentStrategy, VisaCheckoutScriptLoader } from './strategies/braintree';
 import { CardinalClient, CardinalScriptLoader, CardinalThreeDSecureFlow } from './strategies/cardinal';
 import { ChasePayPaymentStrategy, ChasePayScriptLoader } from './strategies/chasepay';
@@ -430,6 +431,14 @@ export default function createPaymentStrategyRegistry(
             new StripeScriptLoader(scriptLoader)
         )
     );
+
+    registry.register(PaymentStrategyType.BOLT, () =>
+        new BoltPaymentStrategy(
+            store,
+            orderActionCreator,
+            paymentActionCreator
+        )
+);
 
     return registry;
 }
--- a/src/payment/payment-strategy-type.ts
+++ b/src/payment/payment-strategy-type.ts
@@ -5,6 +5,7 @@ enum PaymentStrategyType {
     AMAZON = 'amazon',
     AUTHORIZENET_GOOGLE_PAY = 'googlepayauthorizenet',
     BLUESNAPV2 = 'bluesnapv2',
+    BOLT = 'bolt',
     CREDIT_CARD = 'creditcard',
     CYBERSOURCE = 'cybersource',
     KLARNA = 'klarna',
--- a/src/order/order-request-body.ts
+++ b/src/order/order-request-body.ts
@@ -1,4 +1,4 @@
-import { CreditCardInstrument, HostedCreditCardInstrument, HostedInstrument, HostedVaultedInstrument, VaultedInstrument } from '../payment';
+import { CreditCardInstrument, HostedCreditCardInstrument, HostedInstrument, HostedVaultedInstrument, NonceInstrument, VaultedInstrument } from '../payment';
 
 /**
  * An object that contains the information required for submitting an order.
@@ -36,8 +36,8 @@ export interface OrderPaymentRequestBody {
     gatewayId?: string;
 
     /**
-     * An object that contains the details of a credit card or vaulted payment
-     * instrument.
+     * An object that contains the details of a credit card, vaulted payment
+     * instrument or nonce instrument.
      */
-    paymentData?: CreditCardInstrument | HostedInstrument | HostedCreditCardInstrument | HostedVaultedInstrument | VaultedInstrument;
+    paymentData?: CreditCardInstrument | HostedInstrument | HostedCreditCardInstrument | HostedVaultedInstrument | NonceInstrument | VaultedInstrument;
 }
--- a/src/payment/strategies/bolt/bolt-payment-strategy.spec.ts
+++ b/src/payment/strategies/bolt/bolt-payment-strategy.spec.ts
@@ -1,0 +1,117 @@
+import { createClient as createPaymentClient } from '@bigcommerce/bigpay-client';
+import { createAction } from '@bigcommerce/data-store';
+import { createRequestSender } from '@bigcommerce/request-sender';
+import { of, Observable } from 'rxjs';
+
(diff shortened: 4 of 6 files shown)


src/payment/strategies/nonce-pass-through/nonce-pass-through-payment-strategy.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+            .mockReturnValue(submitOrderAction);
+
+        jest.spyOn(paymentActionCreator, 'submitPayment')
+            .mockReturnValue(submitPaymentAction);
+
+        strategy = new NoncePassThroughPaymentStrategy(store, orderActionCreator, paymentActionCreator);
+    });
+
+    it('returns checkout state when initializing the strategy', async () => {
+        return expect(strategy.initialize()).resolves.toEqual(store.getState());
```
rlopz, line 69: I just wondering why are you returning a value in a unit test

src/payment/strategies/nonce-pass-through/nonce-pass-through-payment-strategy.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+    });
+
+    it('returns checkout state when deinitializing the strategy', async () => {
+        return expect(strategy.deinitialize()).resolves.toEqual(store.getState());
```
rlopz, line 73: same as above

src/payment/strategies/nonce-pass-through/nonce-pass-through-payment-strategy.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    deinitialize(): Promise<InternalCheckoutSelectors> {
+        return Promise.resolve(this._store.getState());
+    }
+
+    async execute(payload: OrderRequestBody, _options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        const { payment, ...order } = payload;
+
+        if (!payment) {
+            throw new InvalidArgumentError('Unable to submit payment because "payload.payment" argument is not provided.');
```
mauricio-sg, line 29: I think `PaymentArgumentInvalidError` is a more specific error. 🍹

src/payment/strategies/nonce-pass-through/nonce-pass-through-payment-strategy.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+    });
+
+    it('executes the strategy successfully', async () => {
+        const expectedPayload = {
+            methodId: 'foobar',
+            paymentData: {
+                nonce: 'fooNonce',
+            },
+        };
+        await strategy.execute(payload, options);
```
vmparra, line 87: Please insert a line break after unit test setup

src/payment/strategies/nonce-pass-through/nonce-pass-through-payment-strategy.spec.ts
mauricio-sg: oh yeah, remove the `async` or `await` the assertion

src/payment/payment-strategy-type.ts
```diff
@@ -5,6 +5,7 @@ enum PaymentStrategyType {
     AMAZON = 'amazon',
     AUTHORIZENET_GOOGLE_PAY = 'googlepayauthorizenet',
     BLUESNAPV2 = 'bluesnapv2',
+    BOLT = 'bolt',
```
capsula4, line 8: if it's meant to be generic, should we rename it to something like
```js
NONCE = 'nonce'
```

src/payment/strategies/nonce-pass-through/nonce-pass-through-payment-strategy.ts
```diff
@@ -0,0 +1,54 @@
...
+import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
+import { MissingDataError, MissingDataErrorType } from '../../../common/error/errors';
+import { OrderActionCreator, OrderRequestBody } from '../../../order';
+import { OrderFinalizationNotRequiredError } from '../../../order/errors';
+import { PaymentArgumentInvalidError } from '../../errors';
+import PaymentActionCreator from '../../payment-action-creator';
+import { PaymentRequestOptions } from '../../payment-request-options';
+import PaymentStrategy from '../payment-strategy';
+
+export default class NoncePassThroughPaymentStrategy implements PaymentStrategy {
```
capsula4, line 11: for readability should we simply name it `NoncePaymentStrategy`?

FranciscoOnt:
I renamed the strategy @capsula4 , could you please re-review? ♻️

src/payment/create-payment-strategy-registry.ts
```diff
@@ -193,6 +194,14 @@ export default function createPaymentStrategyRegistry(
         )
     );
 
+    registry.register(PaymentStrategyType.NONCE, () =>
+    new NoncePaymentStrategy(
```
capsula4, line 198: indentation is off

src/payment/strategies/nonce/nonce-payment-strategy.ts
```diff
@@ -0,0 +1,54 @@
...
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        if (!paymentData || !isNonceLike(paymentData)) {
+            throw new MissingDataError(MissingDataErrorType.MissingPayment);
+        }
+
+        await this._store.dispatch(this._orderActionCreator.submitOrder(order, _options));
+
+        return this._store.dispatch(this._paymentActionCreator.submitPayment({
```
capsula4, line 45: would this be sent to bigpay? What will bigpay with `method_id: 'nonce'` ? 🤔

FranciscoOnt: when executing the strategy the method is is retrieved from the payload parameter, so if `method_id: 'bolt'` is sent in the payload  it will be submitted  to the respective gateway.

There is also a `gateway` parameter in the `OrderPaymentRequestBody` type, should I use that one instead? @capsula4

FranciscoOnt: I added the gateway parameter so it can be sent to the corresponding gateway.
But if the SDK is used by a third party application, would it  be posible to load the nonce strategy with a `nonce` methodId, but later use `bolt` as methodId when executing the strategy?

capsula4: I'm not exactly sure how this will be articulated in the background. It seems like we want to make a generic thing but in reality is not possible, as we need to pass `bolt`. I think we might be better off calling this `BoltPaymentStrategy`.

I don't have a thorough understanding of how BigPay works, maybe @davidchin can shed some light here.

cmedinilla: HI @capsula4 this is the tech design and diagram of this solution
https://intranet.bigcommerce.com/display/IDT/Bolt+-+Technical+Design

capsula4: Thanks for sharing this, so far, this doesn't seem to be a generic strategy.

Looking at this diagram, we might need a way to make "Bolt server" generic, that is: what URL should BC send the payment data to (along with the nonce), and then, what's the callback URL that bolt server will talk back to BC.

Screen Shot 2020-05-19 at 10 35 55 am

It looks that's something BigPay would need to be able to handle, however I'm not the best person to talk about BigPay design. My understanding is that @davidchin  has already been part of some of the tech discussions around the Bolt integration, so I will defer this to him.

src/payment/create-payment-strategy-registry.ts
```diff
@@ -431,5 +432,13 @@ export default function createPaymentStrategyRegistry(
         )
     );
 
+    registry.register(PaymentStrategyType.BOLT, () =>
+    new BoltPaymentStrategy(
```
davidchin, line 436: Can you fix the indentation in this block?

PR #844, opened 2020-04-22
CHECKOUT-4842: Trigger event when "enter" key is pressed in one of hosted payment fields

davidchin:
## What?
Trigger `onEnter` event callback when "enter" key is pressed in one of the hosted payment fields. With this callback, the client application can submit an order when it is triggered.

## Why?
With this change, shoppers can use their keyboard to place an order after filling in their credit card details.

## Testing / Proof
CircleCI

### DIFF
--- a/src/hosted-form/iframe-content/hosted-autocomplete-fieldset.ts
+++ b/src/hosted-form/iframe-content/hosted-autocomplete-fieldset.ts
@@ -1,6 +1,5 @@
 import { kebabCase } from 'lodash';
 
-import { InvalidHostedFormConfigError } from '../errors';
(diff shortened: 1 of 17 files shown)


src/hosted-form/iframe-content/hosted-autocomplete-fieldset.ts
```diff
@@ -10,21 +9,15 @@ export default class HostedAutocompleteFieldset {
     private _inputs: HTMLInputElement[];
 
     constructor(
-        private _containerId: string,
+        private _form: HTMLFormElement,
```
capsula4, line 12: 👏

PR #845, opened 2020-04-22
CHECKOUT-4799 Add Sign-in via Email support (passwordless login)

capsula4:
## What?
Add sign-in via email support (passwordless login).

## Why?
To provide a facade to `login.php` API and keep state of sent emails.

## Testing / Proof
unit

@bigcommerce/checkout

### DIFF
--- a/src/checkout/checkout-service.ts
+++ b/src/checkout/checkout-service.ts
@@ -15,6 +15,7 @@ import { OrderActionCreator, OrderRequestBody } from '../order';
 import { PaymentInitializeOptions, PaymentMethodActionCreator, PaymentRequestOptions, PaymentStrategyActionCreator } from '../payment';
 import { InstrumentActionCreator } from '../payment/instrument';
 import { ConsignmentsRequestBody, ConsignmentActionCreator, ConsignmentAssignmentRequestBody, ConsignmentUpdateRequestBody, ShippingCountryActionCreator, ShippingInitializeOptions, ShippingRequestOptions, ShippingStrategyActionCreator } from '../shipping';
+import { SignInEmailActionCreator } from '../signin-email';
 import { SpamProtectionActionCreator, SpamProtectionOptions } from '../spam-protection';
 import { StoreCreditActionCreator } from '../store-credit';
 import { Subscriptions, SubscriptionsActionCreator } from '../subscription';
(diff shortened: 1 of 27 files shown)


src/signin-email/signin-email-action-creator.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+                    { type: SignInEmailActionType.SendSignInEmailRequested },
+                    { type: SignInEmailActionType.SendSignInEmailSucceeded, payload: response.body },
+                ]);
+            });
+
+            it('emits error actions if unable to continue as guest', async () => {
+                jest.spyOn(signInEmailRequestSender, 'sendSignInEmail')
+                    .mockReturnValue(Promise.reject(getErrorResponse()));
+
+                const errorHandler = jest.fn();
```
davidchin, line 48: Might be cleaner if you just write something like this
```ts
const errorHandler = jest.fn(action => of(action));
```

And then you can pass it directly to `catchError(errorHandler)` and you can still assert that `errorHandler` gets called.

src/checkout/checkout-store-selector.spec.ts
```diff
@@ -34,6 +34,10 @@ describe('CheckoutStoreSelector', () => {
         expect(selector.getOrder()).toEqual(internalSelectors.order.getOrder());
     });
 
+    it('returns sing-in email', () => {
```
davidchin, line 37: `sign-in` not `sing-in`

src/checkout/checkout-store-selector.ts
```diff
@@ -49,6 +50,13 @@ export default interface CheckoutStoreSelector {
      */
     getConfig(): StoreConfig | undefined;
 
+    /**
+     * Gets the sign-in email.
+     *
+     * @returns The sign-in email object if sent, otherwise undefined
+     */
+    getSignInEmail(): SignInEmail | undefined;
```
davidchin, line 58: The name is a little awkward but I can't come up with anything better so I guess this is good enough.

src/checkout/checkout-service.ts
```diff
@@ -520,6 +522,20 @@ export default class CheckoutService {
...
 
+    /**
+     * Sends a email that contains a single-use sign-in link. When clicked, this link
+     * signs in the customer without requiring any password.
+     *
+     * @param email - The email to be sent the sign-in link.
+     * @param options - Options for the send email request.
+     * @returns A promise that resolves to the current state.
+     */
+    sendSignInEmail(email: string, options?: RequestOptions): Promise<CheckoutSelectors> {
```
davidchin, line 533: Perhaps we can put `@internal` annotation so these new methods don't get published to our docs until this feature is ready to go live?

src/checkout/checkout-store-selector.ts
capsula4: I know.. _naming is hard_

PR #855, opened 2020-04-30
CHECKOUT-4854: Update the instruction on spam protection checks and add missing bullet points to README

davidchin:
## What?
* Update the instruction on spam protection checks.
* Add missing bullet points to README.

## Why?
Better documentation.

## Testing / Proof
None

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,9 +93,9 @@ const service = module.createCheckoutService();
 ```
 
 Currently, there are three modules available for public use:
-**checkout-sdk**: This is the main module that contains all the public exports of the package.
-**checkout-button**: This sub-module can be used to initialize checkout buttons in the storefront once a cart is created (i.e.: cart page).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -258,12 +258,12 @@ await service.removeCoupon('COUPON');
...
 ```
 
-### Initialize spam protection
+### Execute spam protection check
 
 You can also enable bot protection to prevent bots and other types of automated abuse from creating orders. Note that enabling this feature increases checkout friction, which may affect conversions. As such, we recommend leaving this feature out if your store is not encountering bots.
 
 ```js
-await service.initializeSpamProtection({ containerId: 'spamProtectionContainer' });
+await service.executeSpamCheck();
```
capsula4, line 266: might be worth explaining when this is supposed to be called (https://github.com/bigcommerce/checkout-sdk-js/issues/856)

flyingL123: @capsula4 Does this method automatically handle refreshing the recaptcha? If the shopper is on the payment step for 20 mins, will the security check expire when they try to purchase, or is the sdk handling automatically refreshing the token?

PR #863, opened 2020-05-06
INT-2611 Update the klarna session before the widget is loaded (Klarna V2)

rlopz:
[INT-2611](https://jira.bigcommerce.com/browse/INT-2611)
## What?
Hitting an endpoint to update the klarna session

## Why?
Because we want to avoid mismatch between bigcommerce order and klarna order

## Sibling PRs
[INT-2611 BCAPP](https://github.com/bigcommerce/bigcommerce/pull/34949)

### DIFF
--- a/src/payment/strategies/klarnav2/klarnav2-payment-strategy.ts
+++ b/src/payment/strategies/klarnav2/klarnav2-payment-strategy.ts
@@ -79,9 +79,7 @@ export default class KlarnaV2PaymentStrategy implements PaymentStrategy {
             throw new InvalidArgumentError('Unable to proceed because "payload.payment.gatewayId" argument is not provided.');
         }
 
-        this._updateOrder(gatewayId);
-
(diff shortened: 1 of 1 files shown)


src/payment/strategies/klarnav2/klarnav2-payment-strategy.ts
```diff
@@ -67,7 +67,7 @@ export default class KlarnaV2PaymentStrategy implements PaymentStrategy {
         return Promise.reject(new OrderFinalizationNotRequiredError());
     }
 
-    private _loadPaymentsWidget(options: PaymentInitializeOptions): Promise<KlarnaLoadResponse> {
+    private _loadPaymentsWidget(options: PaymentInitializeOptions): Promise<void | KlarnaLoadResponse> {
```
mauricio-sg, line 70: Why `void`?

rlopz: I already removed it, thanks

PR #865, opened 2020-05-08
CHECKOUT-4879 Expose flash messages

capsula4:
## What?
Expose error FlashMessage via `data.getFlashMessages()`

## Why?
So clients using Sign-in Email Link feature can access the error when the link is invalid.

## Testing / Proof
unit

### DIFF
--- a/src/checkout/checkout-store-selector.ts
+++ b/src/checkout/checkout-store-selector.ts
@@ -6,7 +6,7 @@ import { BillingAddress } from '../billing';
 import { Cart } from '../cart';
 import { createSelector } from '../common/selector';
 import { cloneResult as clone } from '../common/utility';
-import { StoreConfig } from '../config';
+import { FlashMessage, FlashMessageType, StoreConfig } from '../config';
 import { Coupon, GiftCertificate } from '../coupon';
 import { Customer } from '../customer';
 import { FormField } from '../form';
@@ -146,6 +146,17 @@ export default interface CheckoutStoreSelector {
      * undefined if otherwise.
      */
     getSelectedPaymentMethod(): PaymentMethod | undefined;
+
+    /**
+     * Gets the available flash messages.
+     *
+     * Flash messages contain messages set by the server,
+     * e.g: when trying to sign in using an invalid email link.
+     *
+     * @param type - The type of flash messages to be returned. Optional
+     * @returns The flash messages if available, otherwise undefined.
+     */
+    getFlashMessages(type?: FlashMessageType): FlashMessage[] | undefined;
 
     /**
      * Gets the current cart.
@@ -453,6 +464,11 @@ export function createCheckoutStoreSelectorFactory(): CheckoutStoreSelectorFacto
         })
     );
 
+    const getFlashMessages = createSelector(
+        ({ config }: InternalCheckoutSelectors) => config.getFlashMessages,
+        getFlashMessages => clone(getFlashMessages)
+    );
+
     return memoizeOne((
         state: InternalCheckoutSelectors
     ): CheckoutStoreSelector => {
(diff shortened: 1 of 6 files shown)


src/checkout/checkout-store-error-selector.ts
```diff
@@ -295,6 +306,19 @@ export function createCheckoutStoreErrorSelectorFactory(): CheckoutStoreErrorSel
         }
     );
 
+    const getInitialError = createSelector(
```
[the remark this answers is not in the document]
davidchin: I see this as a convenience method for getting flash messages. It's nice to have that because it's a bit of a hassle to traverse the `ContextConfig` object in order to retrieve a message. But considering we do have other types of flash messages and they can get set because of a prior action before a redirect, I'm wondering if it's better if we have a method that can return other types of message as well. This is so that we don't end up with other variants of this method to expose other types of messages in the future.

```ts
getFlashMessages(type?: FlashMessageType): FlashMessage[]
```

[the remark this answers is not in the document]
capsula4: I think flash messages in general are quite random and the name itself "flash message" does not tell much to the consumers. Also, I'm not sure what do we use them for aside for errors (in checkout). That's why I thought about exposing only error flash messages initially, using `getInitialError` selector.

If you think is better to anticipate use cases, I'm happy to have a `getFlashMessages` selector, maybe as part of the `data` selector?
`state.data.getStoreConfig()`
`state.data.getFlashMessages()`

currently is
`state.errors.getInitialError()`

thoughts?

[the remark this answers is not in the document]
davidchin: Yes, I understand where you're coming from. But we can't hide the fact that they are modelled as flash messages as they are already exposed through `config.context.flashMessages`.

Anyway, if we are going to have this convenience method, yes, I agree we should have it under `data` instead.

src/checkout/checkout-store-selector.ts
```diff
@@ -147,6 +147,17 @@ export default interface CheckoutStoreSelector {
...
+    /**
+     * Gets the available flash messages.
+     *
+     * Flash messages contain messages set by the server,
+     * e.g: when trying to sign in using an invalid email link.
+     *
+     * @param type - The type of flash messages to be returned. Optional
+     * @returns The flash messages if available, otherwise undefined.
+     */
+    getFlashMessages(type?: FlashMessageType): FlashMessage[] | undefined;
```
davidchin, line 159: I think if you want to import `FlashMessageType` and use it as a value (not a `type` or interface). You'll need to explicitly export it in `bundles/checkout-sdk.ts`. Otherwise, you can't use it in the client app and you'll need to put in `0`, `1` etc... into this method.

capsula4: I have created a build and this seems to work fine. Managed to use in UCO without issues:

Screen Shot 2020-05-11 at 2 21 05 pm

davidchin: Hrm, interesting. I might be wrong but I think if you actually try to run the code in the browser, you'll get an error. This is because the enum is actually not exported to the `js` file (only in the `d.ts` file).

src/config/config.ts
```diff
@@ -108,10 +108,17 @@ export interface CustomizationConfig {
     languageData: any[];
 }
 
+export type FlashMessageType = 'error' | 'info' | 'warning' | 'success';
+
+export interface FlashMessage {
+    type: FlashMessageType;
+    message: string;
```
capsula4, line 115: modifying API to be strings as per https://github.com/bigcommerce/bigcommerce/pull/34969

PR #869, opened 2020-05-13
feat(payment): PAYPAL-365 choses paypalcredit method on the checkout

helga1507:
## What?
Choses paypalcommercecredit method on the checkout page

## Why?
Merchant should see paypalcommercecredit method on the checkout page after chosen this method in the spb

## Testing / Proof
Screen Shot 2020-05-13 at 10 37 35 AM

### DIFF
--- a/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
+++ b/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
@@ -3,11 +3,12 @@ import { FormPoster } from '@bigcommerce/form-poster';
 import { Cart } from '../../../cart';
 import { CheckoutActionCreator, CheckoutStore } from '../../../checkout';
 import { InvalidArgumentError, MissingDataError, MissingDataErrorType } from '../../../common/error/errors';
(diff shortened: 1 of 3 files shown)


src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -48,24 +51,34 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
...
     deinitialize(): Promise<void> {
+        this.providerId = undefined;
+
         return Promise.resolve();
     }
 
-    private async _setupPayment(methodId: string, cartId: string): Promise<string> {
-        const { orderId } = await this._paypalCommerceRequestSender.setupPayment(methodId, cartId);
+    private _handleClickButtonProvider(providerId: string, { fundingSource }: ClickDataOptions): void {
+        this.providerId = fundingSource === 'credit' ? providerId + 'credit' : providerId;
```
BC-ACherednichenko, line 60: @helga1507 what is the default value for `providerId `? I think for now it would be more obvious to replace `providerId + 'credit'` -> `paypalcommercecredit` as long as this provider Id is the only we one case we want to handle

helga1507: Okay. Thank you!

PR #873, opened 2020-05-15
CHECKOUT-4754: Send an additional header to provide information about the checkout variant that is used to pay for an order

davidchin:
## What?
Send an additional header to provide information about the checkout variant that is used to pay for an order.

## Why?
It's an optional parameter that can help us identify whether the default checkout or a custom checkout is used to complete a checkout process.

## Testing / Proof
Unit

### DIFF
--- a/src/config/config-window.ts
+++ b/src/config/config-window.ts
@@ -1,0 +1,3 @@
+export default interface ConfigWindow extends Window {
+    checkoutVariantIdentificationToken?: string;
+}
--- a/src/order/order-request-sender.ts
+++ b/src/order/order-request-sender.ts
@@ -1,4 +1,5 @@
 import { RequestSender, Response } from '@bigcommerce/request-sender';
+import { isNil, omitBy } from 'lodash';
 
 import { joinIncludes, ContentType, RequestOptions } from '../common/http-request';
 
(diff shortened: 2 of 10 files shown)


src/order/order-request-sender.ts
```diff
@@ -35,10 +42,17 @@ export default class OrderRequestSender {
...
-    submitOrder(body: InternalOrderRequestBody, { timeout }: RequestOptions = {}): Promise<Response<InternalOrderResponseBody>> {
+    submitOrder(body: InternalOrderRequestBody, { headers, timeout }: SubmitOrderRequestOptions = {}): Promise<Response<InternalOrderResponseBody>> {
         const url = '/internalapi/v1/checkout/order';
 
-        return this._requestSender.post(url, { body, timeout });
+        return this._requestSender.post(url, {
+            body,
+            headers: omitBy({
+                'X-Checkout-Variant': headers && headers.checkoutVariant,
+                'X-Checkout-SDK-Version': LIBRARY_VERSION,
```
capsula4, line 52: 👏

src/config/config-window.ts
```diff
@@ -0,0 +1,3 @@
+export default interface ConfigWindow extends Window {
+    checkoutVariantIdentificationToken?: string;
```
capsula4, line 2: out of curiosity, what values can this be?

davidchin: It'd be a JWT token containing the variant ID/name.

PR #874, opened 2020-05-15
feat(payments): INT-2668 Increase code coverage for adyenv2 payment s…

vmparra:
[INT-2668](https://jira.bigcommerce.com/browse/INT-2668)

## What?
Increase code coverage for adyenv2 payment strategy

## Why?
To validate we handle properly exceptions when adyen components fail to mount 

## Testing / Proof
image

### DIFF
--- a/src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
+++ b/src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
@@ -19,7 +19,7 @@ import PaymentRequestTransformer from '../../payment-request-transformer';
 
 import { AdyenAdditionalActionState, AdyenComponentState, AdyenError, AdyenPaymentMethodType, AdyenV2PaymentStrategy, AdyenV2ScriptLoader, ResultCode } from '.';
 import { AdyenComponent } from './adyenv2';
-import { getAdditionalActionError, getAdyenClient, getAdyenError, getComponentState, getInitializeOptions, getInitializeOptionsWithNoCallbacks, getInitializeOptionsWithUndefinedWidgetSize, getOrderRequestBody, getOrderRequestBodyWithoutPayment, getOrderRequestBodyWithVaultedInstrument, getUnknownError } from './adyenv2.mock';
+import { getAdditionalActionError, getAdyenClient, getAdyenError, getComponentState, getFailingComponent, getInitializeOptions, getInitializeOptionsWithNoCallbacks, getInitializeOptionsWithUndefinedWidgetSize, getOrderRequestBody, getOrderRequestBodyWithoutPayment, getOrderRequestBodyWithVaultedInstrument, getUnknownError } from './adyenv2.mock';
 
 describe('AdyenV2PaymentStrategy', () => {
     let finalizeOrderAction: Observable<FinalizeOrderAction>;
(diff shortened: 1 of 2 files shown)


src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
```diff
@@ -150,6 +150,30 @@ describe('AdyenV2PaymentStrategy', () => {
...
+                await expect(strategy.initialize(options))
+                    .rejects.toThrow(NotInitializedError);
+            });
+
+            it('fails mounting card verification component', async () => {
+                cardVerificationComponent = getFailingComponent();
+
+                await expect(strategy.initialize(options))
+                    .rejects.toThrow(NotInitializedError);
+            });
```
clopezh, line 175: This test and this one => https://github.com/bigcommerce/checkout-sdk-js/pull/874/files#diff-874c6669fffb1e8c18b7b0bbf4738a5aR153-R158 seems to be the same, you're using the same beforeEach and the same Component `getFailingComponent` and also you're expecting the same error, so maybe these two tests can be converted into a single one. 🍹

src/payment/strategies/adyenv2/adyenv2.mock.ts
```diff
@@ -5,7 +5,7 @@ import Payment from '../../payment';
 import { PaymentInitializeOptions } from '../../payment-request-options';
 import { getCreditCardInstrument, getErrorPaymentResponseBody, getVaultedInstrument } from '../../payments.mock';
 
-import { AdyenAdditionalActionErrorResponse, AdyenClient, AdyenComponentState, AdyenConfiguration, AdyenError, AdyenPaymentMethodType, ResultCode } from './adyenv2';
+import { AdyenAdditionalActionErrorResponse,  AdyenClient,  AdyenComponent, AdyenComponentState, AdyenConfiguration, AdyenError, AdyenPaymentMethodType, ResultCode } from './adyenv2';
```
clopezh, line 8:
```suggestion
import { AdyenAdditionalActionErrorResponse, AdyenClient, AdyenComponent, AdyenComponentState, AdyenConfiguration, AdyenError, AdyenPaymentMethodType, ResultCode } from './adyenv2';
```

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
vmparra: Actually, they are not testing the same failing component, one is testing cardVerificationComponent and the other one paymentComponent.

vmparra:
♻️ @clopezh I've addressed your comments, please take another 👁️

src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts
capsula4: I still don't see how they are different? the code blocks are identical except for the variable where you assign the result to, which seems unrequired.

```
it('fails mounting card verification component', async () => {
  getFailingComponent();

  await expect(strategy.initialize(options))
    .rejects.toThrow(NotInitializedError);
});

Comment (vmparra) [reply src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts:175]: Hi Luis, 

First test covers catch block when exception it thrown while mounting `paymentComponent`

image

Second test covers catch block when exception it thrown while mounting `cardVerificationComponent`

image

**Comment by vmparra** [comment]
♻️ @capsula4  I've addressed your comments, please take another 👁️

File: src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts (hunk 1 above)
Comment (capsula4) [reply src/payment/strategies/adyenv2/adyenv2-payment-strategy.spec.ts:175]: What is confusing of these test is that is hard to know what's exactly going on by checking the test itself. Although it might seem that the overall test is simpler and has less duplicated code, it's harder to follow what the test is trying to achieve. Code in general is encouraged to be DRY, but for unit tests a DAMP approach is preferred:
https://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests

Ideally, a beforeEach block should be a setup that is the same for each test, and each test shouldn't be affecting the setup.

```js
it('throws exceptions when a faulty adyen payment component is returned', async () => {
  jest.spyOn(adyenCheckout, 'create')
    .mockImplementationOnce(jest.fn((_method, options) => {
      const { onChange } = options;
      handleOnChange = onChange;

      return getFailingComponent();
  }));

  await expect(strategy.initialize(options))
    .rejects.toThrow(NotInitializedError);
});
```

capsula4: Also, is it possible that `adyenCheckout.create()` returns a null component? Is that closer to reality than having an object with a `mount` function that throws?

```js
jest.spyOn(adyenCheckout, 'create')
    .mockImplementationOnce(jest.fn(() => null));
```

PR #875, opened 2020-05-17
feat(payment): PAYMENTS-5425 Implement the UX for Carding remediation solution

Tharaae:
## What?
For credit card payment, after checkout sdk sends payment payload to BigPay, if `human_verification_required` response is returned from BigPay, Google Recaptcha instance is created to verify that the shopper is a human. It's invisible if Google thinks the shopper is human, and shows a challenge if Google thinks the shopper is a bot. 

Once ReCAPTCHA is completed, it'll re-send the payment request to BigPay with almost the same payload, except for the additional token that was just created when completing the ReCAPTCHA.

Sibling PR on bigpay-client-js to be merged first: https://github.com/bigcommerce/bigpay-client-js/pull/99

## Why?
To protect our stores from spammers using long lists of credit cards number to place fraudulent orders.

## Testing / Proof
Screen record
Cybersourse-visible (1)


src/spam-protection/get-google-recaptcha-token.ts
```diff
@@ -0,0 +1,15 @@
+import { ReadableDataStore } from '@bigcommerce/data-store';
+
+import { InternalCheckoutSelectors } from '../checkout';
+
+import CardingProtectionActionCreator from './carding-protection-action-creator';
+
+export default function getGoogleRecapthaToken(store: ReadableDataStore<InternalCheckoutSelectors>, error: any) {
+    const { additional_action_required, status } = error;
+    if (status === 'additional_action_required' && additional_action_required && additional_action_required.type === 'recaptcha_v2_verification') {
```
leeBigCommerce, line 9: Same condition as: https://github.com/bigcommerce/checkout-sdk-js/pull/875/files#diff-6403927dbab7260b0f1154bc8ce53ee6R37 ?

This again makes me think it might be nice to think the error body could be a union type and this conditional could be a type guard. Possibly used at the higher scope:

```
isRecaptchaRequiredError(error)
```

src/hosted-form/hosted-form-order-data-transformer.ts
```diff
@@ -42,4 +43,15 @@ export default class HostedFormOrderDataTransformer {
             paymentMethodMeta,
         };
     }
+
+    private _getAdditionalAction(paymentRecaptchaToken?: string) {
+        return paymentRecaptchaToken ?
```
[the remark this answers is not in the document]
Tharaae: I assume in the future there may be more type of `additional_actions` as a response to other `additional_action_required` errors. So, I think having it through a separate private function makes sense to keep logic of different action types. I will move the check on `paymentRecaptchaToken` out of the function though.

src/payment/payment-request-body.ts
```diff
@@ -8,15 +8,21 @@ import { PaymentInstrument } from './payment';
...
     authToken: string;
-    payment?: PaymentInstrument;
     billingAddress?: InternalAddress;
     cart?: InternalCart;
     customer?: InternalCustomer;
     order?: InternalOrder;
     orderMeta?: {
         deviceFingerprint?: string;
     };
+    payment?: PaymentInstrument;
```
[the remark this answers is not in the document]
Tharaae: Alphabetic order

src/payment/payment-request-transformer.ts
```diff
@@ -45,8 +45,8 @@ export default class PaymentRequestTransformer {
         }
 
         return {
+            additionalAction: this._getAdditionalAction(payment.paymentRecaptchaToken),
```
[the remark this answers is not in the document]
Tharaae: same reply 😄

src/payment/payment-request-transformer.ts
```diff
@@ -131,4 +132,15 @@ export default class PaymentRequestTransformer {
                 ccNumber: this._cardNumberFormatter.unformat(values.cardNumber || ''),
             };
     }
+
+    private _getAdditionalAction(paymentRecaptchaToken?: string) {
```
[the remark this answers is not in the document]
Tharaae: They are two different payloads. Nothing prevents having different logic in the future.

src/payment/payment-strategy-action-creator.ts
```diff
@@ -47,8 +47,17 @@ export default class PaymentStrategyActionCreator {
...
                 }
 
-                return strategy
-                    .execute(payload, { ...options, methodId: payment.methodId, gatewayId: payment.gatewayId })
+                return strategy.execute(payload, { ...options, methodId: payment.methodId, gatewayId: payment.gatewayId })
+                    .catch(error => {
+                        return getGoogleRecapthaToken(store, error)
+                            .then((token: string) => {
+                                if (payload.payment) {
+                                    payload.payment.paymentRecaptchaToken = token;
```
davidchin, line 55: Can we create a new object instead here instead of modifying the argument by reference? This is because the `payload` is passed in from the SDK client and we don't want to modify it intentionally.

src/spam-protection/get-google-recaptcha-token.ts
```diff
@@ -0,0 +1,15 @@
+import { ReadableDataStore } from '@bigcommerce/data-store';
+
+import { InternalCheckoutSelectors } from '../checkout';
+
+import CardingProtectionActionCreator from './carding-protection-action-creator';
+
+export default function getGoogleRecapthaToken(store: ReadableDataStore<InternalCheckoutSelectors>, error: any) {
```
davidchin, line 7: There's a typo here, it should say `RecaptchaToken` rather than `RecapthaToken`

src/payment/payment-action-creator.ts
```diff
@@ -31,7 +31,15 @@ export default class PaymentActionCreator {
                     ))
                 )
         ).pipe(
-            catchError(error => throwErrorAction(PaymentActionType.SubmitPaymentFailed, error))
+            catchError(error => {
```
davidchin, line 34: I'd probably just leave `PaymentActionCreator` as it is and let the new `getGoogleRecaptchaToken` function you've created to parse the error response instead.

src/spam-protection/get-google-recaptcha-token.ts
davidchin: `error` should be typed as `unknown` if you don't know the type at this point. This forces you to write and apply a type guard before using it.

src/spam-protection/get-google-recaptcha-token.ts
```diff
@@ -0,0 +1,15 @@
...
+    const { additional_action_required, status } = error;
```
davidchin, line 8: `error` might not be in the shape of an object that you expect so you should check it before using it. If you get rid of the error response parsing logic in `PaymentActionCreator`, here you should be expecting an `Action` object with `payload` property that contains the API response.

src/spam-protection/carding-protection-action-creator.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    constructor(
+        private _recaptchaSitekey: string
+    ) {
+        this._googleRecaptcha = createSpamProtection(createScriptLoader());
+    }
+
+    initialize(options?: SpamProtectionOptions): ThunkAction<SpamProtectionAction, InternalCheckoutSelectors> {
+        return () => concat(
+            of(createAction(SpamProtectionActionType.InitializeRequested, undefined)),
```
davidchin, line 26: I don't think you'll need to have an action creator because unlike the checkout spam protection, this payment spam protection is not a separate action that can be triggered by the user of the SDK. Instead, it is a part of the submit order/payment action so we don't need to keep track of and surface the execution statuses / errors separately. I'd take out all the `createAction` calls in this class, and perhaps rename this class as `PaymentHumanVerificationHandler` and combine it with `getGoogleRecaptchaToken`. So eventually, you can just inject an instance of the handler into `PaymentStrategyActionCreator` and use it directly. i.e.: (`handler.handle(error)`).

src/spam-protection/carding-protection-action-creator.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    execute(): any {
+        return (store: any) => defer(() => {
+            return concat(
+                of(createAction(SpamProtectionActionType.ExecuteRequested, undefined)),
+                this.initialize()(store),
+                this._googleRecaptcha.execute()
+                    .pipe(take(1))
+                    .pipe(switchMap(async ({ error, token }) => {
+                        if (error instanceof SpamProtectionChallengeNotCompletedError) {
```
davidchin, line 56: I'd probably throw different errors instead. This is because `SpamProtectionFailedError` is used for checkout reCaptcha. The recovery mechanism for both is different. So it's important for the client to be able to distinguish between the two.

src/hosted-form/hosted-form-order-data-transformer.ts
```diff
@@ -42,4 +43,15 @@ export default class HostedFormOrderDataTransformer {
...
+            {
+                type: 'recaptcha_v2_verification',
```
davidchin, line 50: If the handler class / function returns a payload object in this format instead of a string token, you probably won't need to have the same payload formatting logic in two different places.

src/payment/payment-request-transformer.ts
davidchin, line 48: The same `additionalAction` object needs to be added to the return value of `transformWithHostedFormData` otherwise the value won't flow through to the hosted form.

src/payment/payment-strategy-action-creator.ts
```diff
@@ -47,8 +47,17 @@ export default class PaymentStrategyActionCreator {
...
+                                }
+
+                                return strategy.execute(payload, { ...options, methodId: payment.methodId, gatewayId: payment.gatewayId });
```
davidchin, line 58: Do we need to recreate the order if addition actions are required for payment submission? If not, then I think the resubmission shouldn't happen at this level because `execute` usually includes a call to create an order. I remember I made a comment about consolidating the `try/catch` logic by moving the code out of `PaymentActionCreator` so you don't end up with two catch blocks. However, if order resubmission is not required, then the reverse should happen instead - the catch block here should be moved down to the `PaymentActionCreator` (and potentially `HostedForm`) instead.

src/spam-protection/payment-human-verification-handler.ts
```diff
@@ -0,0 +1,59 @@
...
+            }
+
+        cardingProtectionElement = document.createElement('div');
+        cardingProtectionElement.setAttribute('id', cardingProtectionElementId);
+        document.body.appendChild(cardingProtectionElement);
+
+        return await this._googleRecaptcha.load(cardingProtectionElementId, this._recaptchaSitekey);
+    }
+
+    execute(): Observable<any> {
```
davidchin, line 34: Instead of `any`, you can type it as `AdditionalAction`

src/payment/additional-action.ts
```diff
@@ -0,0 +1,12 @@
+export default interface AdditionalAction {
```
davidchin, line 1: `PaymentAdditionalAction` might be a clearer name.

src/payment/additional-action.ts
```diff
@@ -0,0 +1,12 @@
...
+    type: string;
+    data: AdditionalActionData;
+}
+
+export type AdditionalActionData = (
```
davidchin, line 6: This type alias is probably not needed.

src/payment/payment-action-creator.ts
```diff
@@ -21,9 +23,19 @@ export default class PaymentActionCreator {
...
-            ))
+            defer(async () => {
+                try {
+                    return await this._paymentRequestSender.submitPayment(
+                        this._paymentRequestTransformer.transform(payment, store.getState())
+                    );
+                } catch (error) {
+                    const additionalAction: AdditionalAction = await getVerificationAdditionalAction(error);
+
+                    return await this._paymentRequestSender.submitPayment(
```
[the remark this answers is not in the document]
Tharaae: Not using `await` does not work; it causes the error to be fired before actually handling it.

src/payment/payment-action-creator.ts
```diff
@@ -21,9 +23,19 @@ export default class PaymentActionCreator {
     submitPayment(payment: Payment): ThunkAction<SubmitPaymentAction, InternalCheckoutSelectors> {
         return store => concat(
             of(createAction(PaymentActionType.SubmitPaymentRequested)),
-            from(this._paymentRequestSender.submitPayment(
-                this._paymentRequestTransformer.transform(payment, store.getState())
-            ))
+            defer(async () => {
+                try {
+                    return await this._paymentRequestSender.submitPayment(
```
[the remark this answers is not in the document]
Tharaae: Not using `await` does not work; it causes the error to be fired before actually handling it.

src/spam-protection/payment-human-verification-handler.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export default class PaymentHumanVerificationHandler {
+    private _googleRecaptcha: GoogleRecaptcha;
+
+    constructor() {
+        this._googleRecaptcha = createSpamProtection(createScriptLoader());
+    }
+
+    async handle(error: RequestError): Promise<PaymentAdditionalAction> {
+        if (this._isPaymentHumanVerificationRequest(error)) {
```
davidchin, line 19: Might be good to do a guard statement instead, i.e.: 💅

```ts
if (!this._isPaymentHumanVerificationRequest(error)) {
    throw error;
}
```

src/spam-protection/payment-human-verification-handler.ts
```diff
@@ -0,0 +1,67 @@
...
+
+import { createSpamProtection } from './';
+import { CardingProtectionChallengeNotCompletedError, CardingProtectionFailedError, SpamProtectionChallengeNotCompletedError } from './errors';
+import GoogleRecaptcha from './google-recaptcha';
+
+export default class PaymentHumanVerificationHandler {
+    private _googleRecaptcha: GoogleRecaptcha;
+
+    constructor() {
+        this._googleRecaptcha = createSpamProtection(createScriptLoader());
```
davidchin, line 15: Can you create the Google ReCaptcha instance outside of this class and inject it into here instead?

src/hosted-form/hosted-form.ts
```diff
@@ -63,7 +67,14 @@ export default class HostedForm {
         return await this._getFirstField().submitForm(
             this._fields.map(field => field.getType()),
             this._payloadTransformer.transform(payload)
-        );
+        )
+            .catch((error: RequestError) =>
```
davidchin, line 71: Sorry, I meant this `catch` block needs to be converted into `async/await` as well.

src/payment/payment-request-body.ts
```diff
@@ -4,19 +4,21 @@ import { InternalCustomer } from '../customer';
 import { InternalOrder } from '../order';
 import { InternalShippingOption } from '../shipping';
 
+import { PaymentAdditionalAction } from '.';
```
davidchin, line 7: Usually we import using the exact path instead of from index when the file is in the same directory (basically whenever you see `'.'`) i.e.: `import PaymentAdditionalAction  from './payment-additional-action'` 💅

src/payment/payment-action-creator.ts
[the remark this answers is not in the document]
Tharaae: Yes sure, when the `await` in the `try` block is omitted, the action required error is not not handled as needed. The `await` in the `catch` similar return statement can be omitted though.
Screen Shot 2020-05-20 at 10 53 25 am

Tharaae:
Thanks @davidchin 🙇. Unit tests are created.

src/hosted-form/hosted-form.spec.ts
```diff
@@ -50,13 +58,67 @@ describe('HostedForm', () => {
...
             fields,
             eventListener,
             payloadTransformer as HostedFormOrderDataTransformer,
-            callbacks
+            callbacks,
+            paymentHumanVerificationHandler
         );
+
+        errorResponse = {
+            body: {
```
davidchin, line 72: You can just call `getErrorPaymentResponseBody` and merge it with the new keys you want to add.

src/hosted-form/hosted-form.spec.ts
```diff
@@ -110,6 +172,55 @@ describe('HostedForm', () => {
...
+        const errorHandler = jest.spyOn(paymentHumanVerificationHandler, 'handle')
+            .mockReturnValue(Promise.resolve(additionalActionMock));
+
+        try {
+            await form.submit(payload);
+        } catch {
+            errorHandler();
+        }
+
+        expect(paymentHumanVerificationHandler.handle)
```
davidchin, line 220: Can you also verify that `errorHandler` is called?

src/hosted-form/hosted-form.ts
```diff
@@ -63,7 +67,14 @@ export default class HostedForm {
         return await this._getFirstField().submitForm(
```
davidchin, line 67: Make sure `async/await` is consistent (don't use a mixture of `then/catch` and `async/await`)

```ts
try {
    return await this._getFirstField().submitForm(
    );
} catch (error) {
    const additionalAction = await this._paymentHumanVerificationHandler.handle(error);

    return await this._getFirstField().submitForm(
        this._payloadTransformer.transform(payload, additionalAction)
    );
}
```

src/payment/payment-action-creator.spec.ts
```diff
@@ -42,7 +50,59 @@ describe('PaymentActionCreator', () => {
             .mockReturnValue(Promise.resolve(getResponse(getPaymentResponseBody())));
 
         orderActionCreator = new OrderActionCreator(orderRequestSender, {} as CheckoutValidator);
-        paymentActionCreator = new PaymentActionCreator(paymentRequestSender, orderActionCreator, paymentRequestTransformer);
+        paymentActionCreator = new PaymentActionCreator(paymentRequestSender, orderActionCreator, paymentRequestTransformer, paymentHumanVerificationHandler);
+
+        errorResponse = {
```
davidchin, line 55: Same comment as above regarding `getErrorPaymentResponseBody`

src/payment/payment-action-creator.spec.ts
```diff
@@ -103,6 +163,40 @@ describe('PaymentActionCreator', () => {
...
+                .mockReturnValue(Promise.resolve(additionalActionMock));
+
+            await from(paymentActionCreator.submitPayment(getPayment())(store))
+                .pipe(
+                    catchError(errorHandler),
+                    toArray()
+                )
+                .toPromise();
+
+            return expect(paymentHumanVerificationHandler.handle).toHaveBeenCalledWith(errorResponse);
```
davidchin, line 181: Can you also verify `errorHandler` is called?

src/payment/payment-action-creator.ts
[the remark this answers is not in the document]
davidchin: Oh yes, you're right! I had a look again, you do need `async` because it's in `try/catch` block.

src/spam-protection/payment-human-verification-handler.ts
```diff
@@ -0,0 +1,66 @@
...
+import { RequestError } from '../common/error/errors';
+import { PaymentAdditionalAction } from '../payment';
+
+import { CardingProtectionChallengeNotCompletedError, CardingProtectionFailedError, SpamProtectionChallengeNotCompletedError } from './errors';
+import GoogleRecaptcha from './google-recaptcha';
+
+export default class PaymentHumanVerificationHandler {
+    private _googleRecaptcha: GoogleRecaptcha;
+
+    constructor(googleRecaptcha: GoogleRecaptcha) {
```
davidchin, line 12: Usually we just use the shorthand, i.e.:

```ts
constructor(
    private _googleRecaptcha: GoogleRecaptcha
) {}
```

So you don't need to initialise the properties separately in the constructor.

PR #877, opened 2020-05-19
CHECKOUT-4897 Add redirect_to to SignInEmail

capsula4:
## What?
Add redirect_to to sign in email request payload.

## Why?
Because by default it redirects to account page. We want this to redirect to checkout.

## Testing / Proof
Unit

### DIFF
--- a/src/signin-email/signin-email-request-sender.ts
+++ b/src/signin-email/signin-email-request-sender.ts
@@ -1,18 +1,30 @@
 import { RequestSender, Response } from '@bigcommerce/request-sender';
 
 import { ContentType, RequestOptions } from '../common/http-request';
+import { parseUrl } from '../common/url';
 
-import { SignInEmail } from './signin-email';
+import { SignInEmail, SignInEmailRequestBody } from './signin-email';
 
 export default class SignInEmailRequestSender {
     constructor(
         private _requestSender: RequestSender
     ) {}
 
-    sendSignInEmail(email: string, { timeout }: RequestOptions = {}): Promise<Response<SignInEmail>> {
+    sendSignInEmail(
+        {
+            email,
+            redirectUrl,
+        }: SignInEmailRequestBody,
+        {
+            timeout,
+        }: RequestOptions = {}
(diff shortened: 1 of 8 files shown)


src/signin-email/signin-email-request-sender.ts
```diff
@@ -2,17 +2,28 @@ import { RequestSender, Response } from '@bigcommerce/request-sender';
...
 export default class SignInEmailRequestSender {
     constructor(
         private _requestSender: RequestSender
     ) {}
 
-    sendSignInEmail(email: string, { timeout }: RequestOptions = {}): Promise<Response<SignInEmail>> {
+    sendSignInEmail(
+        {
+            email,
+            redirect_url = window.top.location.href,
```
davidchin, line 15: The server already strips out the domain and only takes in the relative path but doesn't hurt for the client to just send the relative path `parseUrl(window.top.location.href).pathname`. (we can't just use `location.pathname` because of IE compatibility).

davidchin: Also, I think the key should be in camel case instead i.e.: `redirectUrl`

capsula4: the response of the endpoint is snake case, probably we should modify to take snake case?

If the server takes the relative path, how would that work for embedded checkouts?

davidchin: 1. Yes, in that case, the casing for the request and response body is different and we should keep them consistent. @valentindellangela can you take a look?
2. MSF should be fine because when we generate the email URL, we use the domain of the current storefront context (the storefront which the shopper's viewing). And the redirect path should be relative to that. It's safer this way because we don't want people to potentially get a link in their email that redirects them to an untrusted site as a part of a phishing attack.

capsula4: Addressed 2 already. Probably 1 should be modified in the API.

valentindellangela: sure happy to update 1 later today or tomorrow 👍

PR #879, opened 2020-05-19
INT-2001 Reload widget when apply store credit checkbox state changes

rlopz:
[INT-2001](https://jira.bigcommerce.com/browse/INT-2001)
## What?
We are reloading the widget each time the apply store credit checkbox state changes

## Why?
We need to update the klarna session with the final amount to pay

## Testing / Proof
[DEMO](https://drive.google.com/file/d/1RZMjQds1ZZJoGIDZpZvKBtqP24ZK43Zw/view?usp=sharing)

### DIFF
--- a/src/payment/strategies/klarnav2/klarnav2-payment-strategy.spec.ts
+++ b/src/payment/strategies/klarnav2/klarnav2-payment-strategy.spec.ts
@@ -99,6 +99,8 @@ describe('KlarnaV2PaymentStrategy', () => {
 
         jest.spyOn(scriptLoader, 'load')
             .mockImplementation(() => Promise.resolve(klarnaPayments));
+
+        jest.spyOn(store, 'subscribe');
     });
 
     describe('#initialize()', () => {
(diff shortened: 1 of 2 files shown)


src/payment/strategies/klarnav2/klarnav2-payment-strategy.ts
```diff
@@ -30,11 +31,31 @@ export default class KlarnaV2PaymentStrategy implements PaymentStrategy {
...
+            .then(() => {
+                let currentBalance: number | undefined;
+
+                this._unsubscribe = this._store.subscribe(
+                    state => {
+                        const prevBalance = currentBalance;
+                        const checkout = state.checkout.getCheckout();
+                        currentBalance = checkout && checkout.outstandingBalance;
+
+                        if (prevBalance && prevBalance !== currentBalance) {
```
capsula4, line 43: instead of doing this, you can pass a second parameter to `subscribe()` that will only trigger the callback when the returned value changes.

```js
this._unsubscribe = this._store.subscribe(
// callback,
({ checkout }) => {
  const checkout = checkout.getCheckout();

  return checkout && checkout.outstandingBalance;
}

File: src/payment/strategies/klarnav2/klarnav2-payment-strategy.spec.ts [hunk 2]
```diff
@@ -118,6 +120,10 @@ describe('KlarnaV2PaymentStrategy', () => {
             expect(scriptLoader.load).toHaveBeenCalledTimes(1);
         });
 
+        it('loads store subscribe once', () => {
+            expect(store.subscribe).toHaveBeenCalledTimes(1);
```
capsula4, line 124: you could use `store. notifyState()` to trigger the subscribe function and make sure the widget is being re-initialized.

rlopz:
@capsula4 thanks for the review, Could you please take another look? ♻️

PR #880, opened 2020-05-20
fix(common): INT-2308 Align button Tell me more at checkout section

christianahvilla:
[INT-2308](https://jira.bigcommerce.com/browse/INT-2308)

## What?
Align Visa Checkout button at checkout section.

## Why?
The text is not placed correctly

## Testing / Proof
Captura de Pantalla 2020-05-20 a la(s) 12 20 52 p m

### DIFF
--- a/src/customer/strategies/braintree/braintree-visacheckout-customer-strategy.ts
+++ b/src/customer/strategies/braintree/braintree-visacheckout-customer-strategy.ts
@@ -142,6 +142,8 @@ export default class BraintreeVisaCheckoutCustomerStrategy implements CustomerSt
             <a class="v-learn v-learn-default" style="text-align: right; display: block; font-size: 10px; color: #003366;" href="#" data-locale="en_US">Tell Me More</a>`;
 
         const visaCheckoutButton = document.createElement('div');
+        visaCheckoutButton.style.display = 'flex';
+        visaCheckoutButton.style.flexDirection = 'column';
         visaCheckoutButton.style.visibility = 'hidden';
         visaCheckoutButton.className = buttonClass;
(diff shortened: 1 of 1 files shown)


src/customer/strategies/braintree/braintree-visacheckout-customer-strategy.ts
```diff
@@ -139,9 +139,12 @@ export default class BraintreeVisaCheckoutCustomerStrategy implements CustomerSt
                 role="button"
                 src="${buttonSource}"
                 />
-            <a class="v-learn v-learn-default" style="text-align: right; display: block; font-size: 10px; color: #003366;" href="#" data-locale="en_US">Tell Me More</a>`;
+            <a class="v-learn v-learn-default" style="text-align: right; margin-top: 4px; font-size: 10px; color: #003366;" href="#" data-locale="en_US">Tell Me More</a>`;
```
capsula4, line 142: since you have a class `v-learn`, would be better to override the styles in `checkout-js`.
Also, the fact that you are setting the text colour in the style as:
- won't work on dark themes
- since it's a style attribute, there won't be a way to change the colour using the CSS

I wonder if you can remove all styling and instead style it using CSS?
Another option would be to allow the attributes to be customizable via options, like `buttonClass` (mainly colour)

christianahvilla: @capsula4 I changed the file to not modify the line 142. Could you review again please?

JoseBobadilla20:
💚 Tested
QA Details: I checked the correct position for `Tell Me More` text.

PR #881, opened 2020-05-20
fix(payment): INT-2674 Fix redirect to home on safari

HadesDX:
[INT-2674](https://jira.bigcommerce.com/browse/INT-2674)

## What?
Change onCheckout event for zip strategy

## Why?
to fix a issue where on checkout would be redirected to home

## Testing / Proof
[INT-2674](https://drive.google.com/open?id=1K5mEXBf3ydJyCLdKBzYja6o7gA2ewlel)


pgonzalezr:
Looks good from QE side, this works fine for both Authorize & capture and Auth-only flows in Safari, Chrome & FF

HadesDX:
@capsula4 I didn't add functionality just rearranged it, and I added a test when missing clientToken

cmedinilla:
♻️ @capsula4

PR #890, opened 2020-05-26
refactor(vaulting): INT-2614 renaming values for iban/accountnumber.

pedroterriquez95:
## What? [INT-2614](https://jira.bigcommerce.com/browse/INT-2614)
Renaming iban to account number.

## Why?
Because we remove iban and issuer from the vaulted_accounts table on BigPay, and now we are using another table, who will be reusable and we need something more generic for this field.

## Testing / Proof
image

## Sibling PR

- [Bigpay 2571](https://github.com/bigcommerce/bigpay/pull/2571)
- [JS 305](https://github.com/bigcommerce/checkout-js/pull/305)


pgonzalezr:
Looks good from QE side

Tested:
- Vaulting instruments
- Reusing vaulted instruments
- Deleted vaulted instruments
- Attempted duplicating vaulted instruments

PR #892, opened 2020-05-27
feat(payment): PAYMENTS-5443 Update docs to include human verification…

Tharaae:
… at payment step

## What?
Update Checkout SDK docs to clarify that human verification may be done during submitting payment.

## Why?
Notify developers of this new feature.

## Testing / Proof
--

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,8 @@ await service.initializePayment({
 
 And then, you can ask the customer to provide payment details required by their chosen payment method. If the method is executed successfully, you will create an order and thereby complete the checkout process.
 
+We may require human verification to be completed before payment can be processed, which will be handled during this step.
+
 ```js
 const payment = {
     methodId: 'braintree',


README.md
```diff
@@ -301,7 +301,7 @@ await service.initializePayment({
 
 #### Submit order
 
-And then, you can ask the customer to provide payment details required by their chosen payment method. If the method is executed successfully, you will create an order and thereby complete the checkout process.
+And then, you can ask the customer to provide payment details required by their chosen payment method. If the method is executed successfully, you will create an order and thereby complete the checkout process. If human verification is required before completing the order (either automatically detected or by having the feature on), this wil be done in this step during submitting the payment.
```
TomA-R, line 304: How about something like
```suggestion
And then, you can ask the customer to provide payment details required by their chosen payment method. If the method is executed successfully, you will create an order and thereby complete the checkout process.

We may require human verification to be completed before payment can be processed, which will be handled during this step.
```

PR #901, opened 2020-06-04
feat(common): ORDERS-3323 Ensure coupon codes are upper case.

PascalZajac:
## What?
Convert coupon codes to upper case before sending to analytics.

## Why?
Coupon codes are case insensitive, so `foo` and `FOO` are handled the same by our system, but are being counted as unique entries by downstream analytics consumers, confusing merchants.

## Testing / Proof
Unit test updated.

@bigcommerce/checkout @bigcommerce/orders


PascalZajac:
I couldn't find any evidence that coupon values are sent server-side.

PR #908, opened 2020-06-15
feat(payments): INT-2748 Adding sezzle strategy

roberbnd:
## What?  [INT-2748](https://jira.bigcommerce.com/browse/INT-2748)
Adding sezzle payment in the registry as external payment 

## Why?
It does a laybuy's refactoring because sezzle and laybuy are similar and both are mapped as external payments, it's required to support sezzle, using the same strategy

## Testing / Proof
sezzle

### DIFF
--- a/src/payment/create-payment-strategy-registry.ts
+++ b/src/payment/create-payment-strategy-registry.ts
@@ -34,10 +34,10 @@ import { ConvergePaymentStrategy } from './strategies/converge';
 import { CreditCardPaymentStrategy } from './strategies/credit-card';
 import { CreditCardRedirectPaymentStrategy } from './strategies/credit-card-redirect';
 import { CyberSourcePaymentStrategy } from './strategies/cybersource/index';
+import { ExternalPaymentStrategy } from './strategies/external';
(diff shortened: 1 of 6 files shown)


src/payment/strategies/sezzle/sezzle-payment-strategy.ts
```diff
@@ -0,0 +1,63 @@
+import { FormPoster } from '@bigcommerce/form-poster';
```
cbandam, line 1: This seems like the same strategy as LayBuy. Wouldn´t it be better to refactor both strategies into a single strategy? @clopezh @josesonora

roberbnd:
@cbandam and @AlanMunozN  are you agree with the changes?

AlanMunozN:
Hi @roberbnd, LGTM :+1: , please make sure Laybuy works as expected and I suggest changing the commit message to  **refactor(payment)**

mmartinezintersys:
💚 TESTED

QA Details:

- Verify Sezzle configuration
- Verify Sezzle shows as payment option
- Sanity check for Laubuy

capsula4:
Although there was a refactor here, there was also an addition of a new feature.
In the future I recommend

roberbnd:
please @capsula4 review the commit message.

PR #910, opened 2020-06-16
docs(payment): INT-2119 Add the @alpha modifier to AmazonPayV2

mauricio-sg:
## What?
- Improved documentation.
- Added the `@alpha` modifier to all the new introduced params.

## Why?
To indicate that Amazon Pay V2 is in an early stage of development.

## Testing / Proof
CircleCI

### DIFF
--- a/src/shipping/strategies/amazon-pay-v2/amazon-pay-v2-shipping-initialize-options.ts
+++ b/src/shipping/strategies/amazon-pay-v2/amazon-pay-v2-shipping-initialize-options.ts
@@ -11,6 +11,7 @@ export default interface AmazonPayV2ShippingInitializeOptions {
      * This editAddressButtonId is used to set an event listener, provide an
      * element ID if you want users to be able to select a different shipping
      * address by clicking on a button. It should be an HTML element.
+     * @alpha
      */
     editAddressButtonId?: string;
 }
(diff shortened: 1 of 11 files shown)


src/shipping/strategies/amazon-pay-v2/amazon-pay-v2-shipping-initialize-options.ts
```diff
@@ -1,12 +1,17 @@
...
  */
 export default interface AmazonPayV2ShippingInitializeOptions {
     /**
-     * This editAddressButtonId is used to set an event listener, provide an element ID if you want
-     * users to be able to select a different shipping address by clicking on a button.
-     * It should be an HTML element.
+     * This editAddressButtonId is used to set an event listener, provide an
+     * element ID if you want users to be able to select a different shipping
+     * address by clicking on a button. It should be an HTML element.
+     * @alpha
```
capsula4, line 14: I don't think you need to specify alpha in every single attribute, since the parent is already flagged as alpha in here https://github.com/bigcommerce/checkout-sdk-js/pull/910/files#diff-f42ac23e5aaa52610da2ad0590d0be18R41.

You can keep as this now, but for next time should be enough by flagging the entry points as alpha.

mauricio-sg: Oops, I didn't know. Documentation says _The release tag applies recursively to members of a container (e.g. class or interface)._ But nothing about members. 🤔 I'd keep it in mind next time.

PR #915, opened 2020-06-24
CHECKOUT-4912 Documentation for continueAsGuest and sendSignInEmail

capsula4:
## What?
Improve documentation for:
 - Continue as guest
 - Passwordless Login

## Why?
Because it was missing/incomplete

## Testing / Proof
n/a

@bigcommerce/checkout

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ The SDK has a convenient application interface for starting and completing a che
     - [Initialize instance](#initialize-instance)
         - [Load checkout](#load-checkout)
     - [Sign in customer](#sign-in-customer)
+        - [Passwordless Sign-in](#passwordless-sign-in)
+    - [Continue as guest](#continue-as-guest)
     - [Set shipping details](#set-shipping-details)
         - [Set shipping address](#set-shipping-address)
         - [Set shipping option](#set-shipping-option)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -177,6 +179,24 @@ console.log(state.data.getCart().email);
 console.log(state.data.getBillingAddress().email);
 ```
 
+#### Passwordless Sign-in
+Customers could sign in using a single-use link sent to their email address. Once they click on the link, they will be redirected back to the website as a signed-in user.
```
RobinOng, line 183: Instead of `redirected back to the website`, maybe `redirected back to the store`?

src/checkout/checkout-service.ts
```diff
@@ -523,13 +523,17 @@ export default class CheckoutService {
...
+     *
+     * ```js
+     * checkoutService.sendSignInEmail({ email: 'foo@bar.com', redirectUrl: 'checkout' });
+     * ```
      *
-     * @internal
      * @param signInEmailRequest - The sign-in email request values.
      * @param options - Options for the send email request.
-     * @returns A promise that resolves to the current state.
+     * @returns A promisee that resolves to the current state.
```
RobinOng, line 536: `promise`?

src/checkout/checkout-service.ts
```diff
@@ -553,9 +557,12 @@ export default class CheckoutService {
...
      * Continues to check out as a guest.
      *
-     * The customer is required to provide their email address in order to
-     * continue. Once they provide their email address, it will be stored as a
-     * part of their billing address.
+     * If your checkout settings allow it, your customers could continue the checkout as guests (without signing in).
+     * If you have enabled the checkout setting "Prompt existing accounts to sign in", this information is
+     * exposed as part of the [Customer](docs/interfaces/customer.md) object.
+     *
+     * Once they provide their email address, it will be stored as a
```
RobinOng, line 564: `it will be stored as` you can drop the `a` at the end

src/checkout/checkout-service.ts
```diff
@@ -553,9 +557,12 @@ export default class CheckoutService {
     /**
      * Continues to check out as a guest.
      *
-     * The customer is required to provide their email address in order to
-     * continue. Once they provide their email address, it will be stored as a
-     * part of their billing address.
+     * If your checkout settings allow it, your customers could continue the checkout as guests (without signing in).
```
RobinOng, line 560: I looked through the document, seems like we are using `Checkout Settings` by capitalize the first letter for each word

PR #917, opened 2020-06-25
feat(payments): PAYPAL-508 Mark up flow additional params

helga1507:
## What?
Mark up flow additional params

## Why?
As we do not use PayPal SDK on the front end on markup flow, we do not pass commit=true in SDK params, but PayPal backend still needs to know which mode are going with, in order to show buyer proper info on PayPal side.

## Testing / Proof
Screen Shot 2020-06-25 at 1 14 26 PM

### DIFF
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
@@ -63,7 +63,8 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
     private async _getOrderId(methodId: string): Promise<string> {
         const state = this._store.getState();
         const cart = state.cart.getCartOrThrow();
-        const { approveUrl, orderId } = await this._paypalCommerceRequestSender.setupPayment(methodId, cart.id);
+        const provider = methodId === 'paypalcommercecredit' ? 'paypalcommercecreditcheckout' : 'paypalcommercecheckout';
+        const { approveUrl, orderId } = await this._paypalCommerceRequestSender.setupPayment(provider, cart.id);
 
         if (approveUrl) {
             await this._paypalCommercePaymentProcessor.paymentPayPal(approveUrl);


src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -63,7 +63,7 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
     private async _getOrderId(methodId: string): Promise<string> {
         const state = this._store.getState();
         const cart = state.cart.getCartOrThrow();
-        const { approveUrl, orderId } = await this._paypalCommerceRequestSender.setupPayment(methodId, cart.id);
+        const { approveUrl, orderId } = await this._paypalCommerceRequestSender.setupPayment(`${methodId}checkout`, cart.id);
```
capsula4, line 66: I'm struggling to understand the explanation in the PR with this change. I don't see any additional params?
Also, the commit is marked as `feat`, it seems to me this is more of a `fix` ?

BC-ACherednichenko: @helga1507 I would recommend adding the switch case or any kind of strategy that calls endpoints explicitly instead of such a dynamic behaviour `${methodId}checkout`

helga1507: @BC-ACherednichenko  Okay. Thank you!

helga1507: @capsula4 It's new endpoint from the backend side and this task isn't bug. Why is it `fix`?

capsula4: I thought it was a fix, because when you were previously passing `paypalcommercecredit` to set up the payment, now you are passing `paypalcommercecreditcheckout`.

Previously you were not appending `checkout` to the methodId. What happens if some SDK client is stuck on an older SDK version where `checkout` is not appended? Would it work for them? Do they need to get this update?

BC-ACherednichenko: @capsula4 I agree with @helga1507 that it is the feature - not a fix, as we are splitting our endpoints to two different groups - the regular one for payment buttons,  and another one for the checkout, it is the part of the feature that PayPal requested.
thanks for the question, good point, this release will be compatible with the previous approach, we saved the old endpoints from the checkout and payment buttons so the old version will be still working (it is backward compatible). We will double-check it before release

capsula4: Thanks for explaining! It makes sense to make it a feature commit.

PR #842, opened 2020-04-20
INT-2428 Display vaulted accounts on payment step for Adyenv2

YosafatSanchez:
## What? [INT-2428](https://jira.bigcommerce.com/browse/INT-2428)
Adding functionality to the vaulted account instruments dropdown

## Why?
To show bank account instruments on Adyen APM's

## Testing / Proof
image

@bigcommerce/checkout @bigcommerce/apex-integrations

PR #918, opened 2020-06-28
fix(payment): PAYMENTS-5575 fix paypal and bank instrument clash

leeBigCommerce:
## What?
To add distinct PayPal and Bank instrument types that both extend from a common Account Instrument type.

## Why?
A previous commit (https://github.com/bigcommerce/checkout-sdk-js/pull/842) created a hybrid PayPal/Bank type that did not properly suit either and in turn stopped vaulted PayPal instruments appearing in `checkout-js`.

~~Paired with: https://github.com/bigcommerce/checkout-js/pull/329~~

I had hoped to change the `bank` mapping to have a `method` value of 'bank', but unfortunately the payment api brings back these entities with their brand name as `method`. Not only this, but the entity makes no mention of bank at all, so we're stuck with keeping the same misnomer here too for the time being so the view can match the two sources.  

## Testing / Proof
Screen-Recording-2020-07-13-at-5


src/payment/instrument/instrument-response-transformer.ts
```diff
@@ -41,11 +42,23 @@ export default class InstrumentResponseTransformer {
 
     private _transformVaultedInstruments(vaultedInstruments: InternalInstrument[] = []): PaymentInstrument[] {
         return vaultedInstruments
-            .map(instrument => this._isAccountInstrument(instrument) ? mapToAccountInstrument(instrument) : mapToCardInstrument(instrument) );
+            .map(instrument => {
+                if (this._isPayPalInstrument(instrument)) {
+                    return mapToPayPalInstrument(instrument);
+                } else if (this._isBankInstrument(instrument)) {
+                    return mapToBankInstrument(instrument);
+                }
```
icatalina, line 50: No need to chain the `if` statements:

```suggestion
                if (this._isPayPalInstrument(instrument)) {
                    return mapToPayPalInstrument(instrument);
                }
                
                if (this._isBankInstrument(instrument)) {
                    return mapToBankInstrument(instrument);
                }
```

leeBigCommerce: Done 👍

Tharaae:
LGTM 👍  .. I know it's still WIP, but when finalising this, please remember to fill in testing proof section in the PR description adding screenshots of the related instruments working on checkout page.

PR #919, opened 2020-06-29
CHECKOUT-4973: Initialise hosted payment field within its iframe

davidchin:
## What?
Initialise hosted payment fields within their own iframe.

## Why?
We're changing the implementation to work with newer browsers and provide a better user experience.

## Testing / Proof
Unit

@bigcommerce/checkout @bigcommerce/payments @abhijethbc

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -449,9 +449,9 @@
       }
     },
     "@babel/helper-validator-identifier": {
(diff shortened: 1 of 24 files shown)


davidchin:
@capsula4 we're not doing a top-level redirect anymore. Hence it's a better experience.

PR #922, opened 2020-07-01
feat(payment): PAYPAL-522 Reopen automatically PayPal popup

helga1507:
## What?
Reopen automatically PayPal popup on the Checkout page when purchase request is failing
The goal is to handle a scenario when there is not enough money on a Buyer's account/credit card. Popup should be opened automatically no more than 3 times. When purchase fails the 3rd time then front-end should show an error message 

## Why?
The goal is to handle a scenario when there is not enough money on a Buyer's account/credit card

## Testing / Proof
ezgif com-video-to-gif (2)

### DIFF
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
@@ -87,9 +87,9 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
         try {
             return await this._store.dispatch(this._paymentActionCreator.submitPayment({ ...payment }));
         } catch (error) {
-            const isNotTransactionRejected = !(error instanceof RequestError) || !some(error.body.errors, {code: 'transaction_rejected'});
+            const isInsufficientFundsError = (error instanceof RequestError) && some(error.body.errors, {code: 'transaction_rejected'});
 
-            if (isNotTransactionRejected || attempts > 2) {
+            if (!isInsufficientFundsError || attempts > 2) {
                 return Promise.reject(error);
             }
 
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
@@ -1,7 +1,11 @@
+import { some } from 'lodash';
+
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
+import { RequestError } from '../../../common/error/errors';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
 import { PaymentArgumentInvalidError } from '../../errors';
+import Payment from '../../payment';
 import PaymentActionCreator from '../../payment-action-creator';
 import { PaymentInitializeOptions, PaymentRequestOptions } from '../../payment-request-options';
 import PaymentStrategy from '../payment-strategy';
@@ -9,6 +13,7 @@ import PaymentStrategy from '../payment-strategy';
 import { PaypalCommercePaymentProcessor, PaypalCommerceRequestSender } from './index';
 
 export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
+    private _approveUrl?: string;
 
     constructor(
         private _store: CheckoutStore,
@@ -33,7 +38,7 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
             throw new PaymentArgumentInvalidError(['payment']);
         }
 
-        const orderId = paymentMethod.initializationData.orderId || await this._getOrderId(options.methodId);
+        const orderId = paymentMethod.initializationData.orderId || await this._getOrderIdAndShowPopup(options.methodId);
 
         const paymentData =  {
             formattedPayload: {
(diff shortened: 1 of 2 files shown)


capsula4:
After reading about the initial problem PAYPAL-481, it seems we want to handle the insufficient funds' error in a way that the shopper is aware of the actual issue.

I would expect the solution to be something like showing a error message explaining the solution properly. If there are insufficient funds, how would an automatic retry fix things?

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -60,15 +65,40 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+                return Promise.reject(error);
+            }
+
+            if (!this._approveUrl) {
+                await this._getOrderIdAndShowPopup(methodId);
+            } else {
+                await this._showPopup();
+            }
+
+            return this._processSubmitPayment(payment, methodId, ++attempts);
```
capsula4, line 101: 3d param could be `attempts + 1`

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -60,15 +65,40 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+        }
+    }
+
+    private async _processSubmitPayment(payment: Payment, methodId: string, attempts = 0): Promise<InternalCheckoutSelectors> {
+        try {
+            return await this._store.dispatch(this._paymentActionCreator.submitPayment({ ...payment }));
+        } catch (error) {
+            const isNotTransactionRejected = !(error instanceof RequestError) || !some(error.body.errors, {code: 'transaction_rejected'});
+
+            if (isNotTransactionRejected || attempts > 2) {
```
capsula4, line 91: Why do we try up to 3 times? What does the user see after the third attempt? Might it be worth to always show this when the payment is rejected because of insufficient funds?

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -60,15 +65,40 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+        if (this._approveUrl) {
+            await this._paypalCommercePaymentProcessor.paymentPayPal(this._approveUrl);
+        }
+    }
+
+    private async _processSubmitPayment(payment: Payment, methodId: string, attempts = 0): Promise<InternalCheckoutSelectors> {
+        try {
+            return await this._store.dispatch(this._paymentActionCreator.submitPayment({ ...payment }));
+        } catch (error) {
+            const isNotTransactionRejected = !(error instanceof RequestError) || !some(error.body.errors, {code: 'transaction_rejected'});
```
capsula4, line 89: instead of checking of `isNotTransactionRejected`, might be worth checking for `isInsufficientFundsError` ?

BC-ACherednichenko:
> After reading about the initial problem PAYPAL-481, it seems we want to handle the insufficient funds' error in a way that the shopper is aware of the actual issue.
> 
> I would expect the solution to be something like showing an error message explaining the solution properly. I cannot see, at least from the attached GIF, how this solution fixes the problem.

@capsula4 thanks for the question, the actual case is when the buyer gets insufficient funds error we show the PayPal popup again and PayPal handles the rest. If we send the capture request and the buyer does not have funds PayPal would keep this information in order session and when buyers see PayPal popup it has a msg about insufficient funds. 
image (13)

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
helga1507: Sure, and could be `++attempts`. It's the same 
Why not `++attempts`?

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
helga1507: Thanks for the question!
I changed the gif and you can see message in the pop-up from PayPal (or in the screenshot of Artem). After 3d time pop-up doesn't open again and  user see error message in the modal window. User can start again after reload page

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
helga1507: Good idea. Thanks!

davidchin:
Hi @BC-ACherednichenko and @helga1507, I understand where you're coming from. However, this deviates from the existing user experience pattern that we have with wallet providers. The standard pattern is that the shopper sees a generic error message telling them the payment is unsuccessful. If they wish to try again and their session is still valid with the provider, they will see a more detailed message from the provider themselves like the one shown in the screenshot above. I'm not saying it is better or worst. I'm just saying that it is different.

Also, your proposed solution assumes that the shopper wants to pay with PayPal again, but with a different card. The shopper might not want to do that - they might want to pay with a different wallet or enter a different card directly into checkout. If the popup automatically reappears 3 times, the shopper will need to close it 3 times in order for them to pay with something other than PayPal. 

Anyway, I think we'll need some product/design input here in order to move forward with this. @bc-charlesho.

bc-charlesho:
Hello engineering team. Have you reviewed your proposal for UX change with BC's product design team? The Payments area is normally looked after by Valter Fatia.

BC-ACherednichenko:
@davidchin @bc-charlesho, thanks for the suggestion, we will check it with Valter

PR #793, opened 2020-02-10
fix(checkout): CHECKOUT-4245 handle custom fields for amazon pay

ankurbigcomm:
## What?
Provide support to save custom fields for shipping address when using amazon pay

## Why?
At the moment the custom form fields are not submitted to the API when using amazon pay account.

@bigcommerce/checkout @bigcommerce/payments

PR #925, opened 2020-07-02
fix(shipping): INT-2832 Handle custom fields for AmazonPayV2

mauricio-sg:
## What? [INT-2832](https://jira.bigcommerce.com/browse/INT-2832)
Provide support to save custom fields for the shipping address when using Amazon Pay V2.

## Why?
Because it's not possible to place the order using AmazonPayV2 with custom address fields defined by merchants. Almost the same as #793.

## Testing / Proof
CircleCI

## Dependencies
[checkout-js/pull/333](https://github.com/bigcommerce/checkout-js/pull/333) depends on this PR.

### DIFF
--- a/src/shipping/strategies/amazon-pay-v2/amazon-pay-v2-shipping-strategy.ts
+++ b/src/shipping/strategies/amazon-pay-v2/amazon-pay-v2-shipping-strategy.ts
@@ -1,6 +1,7 @@
 import { noop } from 'rxjs';
 
 import { ConsignmentActionCreator, ShippingStrategyActionCreator } from '../..';
+import { AddressRequestBody } from '../../../address';
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
 import { InvalidArgumentError, MissingDataError, MissingDataErrorType } from '../../../common/error/errors';
 import { PaymentMethod, PaymentMethodActionCreator } from '../../../payment';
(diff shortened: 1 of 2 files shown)


src/shipping/strategies/amazon-pay-v2/amazon-pay-v2-shipping-strategy.ts
```diff
@@ -19,8 +20,15 @@ export default class AmazonPayV2ShippingStrategy implements ShippingStrategy {
         private _shippingStrategyActionCreator: ShippingStrategyActionCreator
     ) {}
 
-    updateAddress(): Promise<InternalCheckoutSelectors> {
-        return Promise.resolve(this._store.getState());
+    updateAddress(address: AddressRequestBody, options?: ShippingRequestOptions): Promise<InternalCheckoutSelectors> {
+        const updateAddressRequestBody = {
+            ...this._store.getState().shippingAddress.getShippingAddress(),
+            customFields: address.customFields,
+        } as AddressRequestBody;
```
capsula4, line 27: instead of using `as`, maybe you could actually construct the payload properly?

```js
const { country, ... updateAddressRequestBody } = this._store.getState().shippingAddress.getShippingAddress();

this._consignmentActionCreator.updateAddress({
  ...updateAddressRequestBody,
  customFields: address.customFields,
}, options)
```

mauricio-sg:
@capsula4 ♻️ 🙇 Let me know if this is fine, thank you!

PR #926, opened 2020-07-09
fix(checkout): CHECKOUT-5006 added displayDateFormat type

ankurbigcomm:
## What?
Added a new type to storeConfig for displayDateFormat and inputDateFormat.

@bigcommerce/checkout


capsula4:
You might need to update the mocks

PR #934, opened 2020-07-15
feat(payment): INT-2801 Pre-populate ACH Billing Info on AdyenV2

vmparra:
## What? [INT-2801](https://jira.bigcommerce.com/browse/INT-2801)
autopopulate billing with billing from checkout

## Why?
To reduce shopper friction.

## Testing / Proof

- Manual
- Unit

### DIFF
--- a/src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
+++ b/src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
@@ -1,5 +1,6 @@
 import { some } from 'lodash';
 
+import { BillingAddress } from '../../../billing';
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
 import { getBrowserInfo } from '../../../common/browser-info';
 import { InvalidArgumentError, NotInitializedError, NotInitializedErrorType, RequestError } from '../../../common/error/errors';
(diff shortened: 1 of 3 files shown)


ValeriaRivas:
💚 Tested

src/payment/strategies/adyenv2/adyenv2.ts
```diff
@@ -215,6 +215,18 @@ export interface AdyenConfiguration {
     onAdditionalDetails?(state: CardState, component?: AdyenComponent): void;
 }
 
+export interface AdyenPlaceholderData {
+    holderName?: string;
```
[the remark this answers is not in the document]
vmparra: Right, because it populates holder name for every APM including card (scheme), and this is not part of the requirement.

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -210,6 +210,34 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
         });
     }
 
+    private _mapDataFromStoreState() {
```
capsula4, line 213: can you add the return type to this function?

src/payment/strategies/adyenv2/adyenv2-payment-strategy.ts
```diff
@@ -245,10 +273,10 @@ export default class AdyenV2PaymentStrategy implements PaymentStrategy {
                 case AdyenPaymentMethodType.ACH:
                 case AdyenPaymentMethodType.Bancontact:
                     paymentComponent = adyenClient.create(paymentMethod.method, {
-                            ...adyenv2.options,
-                            onChange: componentState => this._updateComponentState(componentState),
-                        }
-                    );
+                        ...adyenv2.options,
+                        onChange: componentState => this._updateComponentState(componentState),
+                        data: this._mapDataFromStoreState(),
```
capsula4, line 278: a `mapper` function usually takes one param and converts into something else.
Should this be `this._getAdyenBillingDetails()` ?

vmparra:
Hi @capsula4, I've just addressed your comments. Can you please take another look?

capsula4:
Change looks good, thanks for addressing. Sorry, I didn't notice mention it the first time, but do you think you could add a test for ensuring we pass this new data?

vmparra:
> Change looks good, thanks for addressing. Sorry, I didn't notice mention it the first time, but do you think you could add a test for ensuring we pass this new data?

Hi @capsula4 I've added the proper test case, please 👀

PR #940, opened 2020-07-24
CHECKOUT-4947: Modify Braintree credit card strategy to use hosted form service when feature is enabled

davidchin:
## What?
Modify `BraintreeCreditCardStrategy` to use `BraintreeHostedForm` (introduced in #939) when the feature is enabled and the required options are passed.

## Why?
So we can accept credit card details using Braintree's hosted payment form.

## Testing / Proof
Unit

### DIFF
--- a/src/payment/strategies/braintree/braintree-credit-card-payment-strategy.spec.ts
+++ b/src/payment/strategies/braintree/braintree-credit-card-payment-strategy.spec.ts
@@ -26,6 +26,7 @@ import PaymentRequestSender from '../../payment-request-sender';
 import PaymentRequestTransformer from '../../payment-request-transformer';
 
 import BraintreeCreditCardPaymentStrategy from './braintree-credit-card-payment-strategy';
+import { BraintreePaymentInitializeOptions } from './braintree-payment-options';
 import BraintreePaymentProcessor from './braintree-payment-processor';
 
 describe('BraintreeCreditCardPaymentStrategy', () => {
@@ -44,10 +45,15 @@ describe('BraintreeCreditCardPaymentStrategy', () => {
     beforeEach(() => {
         braintreePaymentProcessorMock = {} as BraintreePaymentProcessor;
         braintreePaymentProcessorMock.initialize = jest.fn();
+        braintreePaymentProcessorMock.initializeHostedForm = jest.fn(() => Promise.resolve());
         braintreePaymentProcessorMock.tokenizeCard = jest.fn(() => Promise.resolve({ nonce: 'my_tokenized_card' }));
+        braintreePaymentProcessorMock.tokenizeHostedForm = jest.fn(() => Promise.resolve({ nonce: 'my_tokenized_card_with_hosted_form' }));
+        braintreePaymentProcessorMock.tokenizeHostedFormForStoredCardVerification = jest.fn(() => Promise.resolve({ nonce: 'my_tokenized_card_verification_with_hosted_form' }));
         braintreePaymentProcessorMock.verifyCard = jest.fn(() => Promise.resolve({ nonce: 'my_verified_card' }));
+        braintreePaymentProcessorMock.verifyCardWithHostedForm = jest.fn(() => Promise.resolve({ nonce: 'my_verified_card_with_hosted_form' }));
         braintreePaymentProcessorMock.getSessionId = jest.fn(() => Promise.resolve('my_session_id'));
         braintreePaymentProcessorMock.deinitialize = jest.fn();
+        braintreePaymentProcessorMock.deinitializeHostedForm = jest.fn(() => Promise.resolve());
 
         paymentMethodMock = { ...getBraintree(), clientToken: 'myToken' };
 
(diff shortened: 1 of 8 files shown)


src/payment/strategies/braintree/braintree-credit-card-payment-strategy.spec.ts
```diff
@@ -115,6 +121,60 @@ describe('BraintreeCreditCardPaymentStrategy', () => {
...
+                        },
+                    },
+                },
+            });
+
+            expect(braintreePaymentProcessorMock.initializeHostedForm)
+                .not.toHaveBeenCalled();
+        });
+
+        it('does not initialize hosted form if configuration is not passed', async () => {
```
capsula4, line 167: do you need a test for `aymentMethodMock.config.isHostedFormEnabled = false;` but hosted form configuration is passed? This could already be covered, just double checking.

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
```diff
@@ -36,8 +38,14 @@ export default class BraintreeCreditCardPaymentStrategy implements PaymentStrate
 
         try {
             this._braintreePaymentProcessor.initialize(paymentMethod.clientToken, options.braintree);
-            this._is3dsEnabled = paymentMethod.config.is3dsEnabled;
 
+            if (this._isHostedPaymentFormEnabled(options.methodId, options.gatewayId) && options.braintree?.form) {
```
capsula4, line 42: if `options.braintree?.form` is not passed, should we throw a specific `InvalidArgument` exception?

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
```diff
@@ -46,34 +54,46 @@ export default class BraintreeCreditCardPaymentStrategy implements PaymentStrate
...
+
+        const {
+            billingAddress: { getBillingAddressOrThrow },
+            order: { getOrderOrThrow },
+            payment: { isPaymentDataRequired },
+        } = state;
 
-        if (!payment) {
-            throw new PaymentArgumentInvalidError(['payment']);
+        if (!isPaymentDataRequired(orderRequest.useStoreCredit) || !orderRequest.payment) {
```
capsula4, line 68: if `!orderRequest.payment`shouldn't we throw `PaymentArgumentInvalidError` ? and not create the order?

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
```diff
@@ -84,37 +104,93 @@ export default class BraintreeCreditCardPaymentStrategy implements PaymentStrate
         throw error;
     }
 
-    private _isUsingVaulting(paymentData: PaymentInstrument): boolean {
-        if (isCreditCardLike(paymentData)) {
-            return Boolean(paymentData.shouldSaveInstrument);
+    private async _preparePaymentData(payment: OrderPaymentRequestBody, billingAddress: Address, orderAmount: number): Promise<PaymentInstrument & PaymentInstrumentMeta> {
+        const commonPaymentData = { deviceSessionId: this._deviceSessionId };
+
+        if (this._isHostedFormInitialized) {
```
capsula4, line 110: this method seems a bit long, do you think it might be worth splitting into different methods, so that there's only one nesting level?
```js
if (this._isHostedFormInitialized) {
  return this._prepareHostedFormPaymentData(...)
}

// ...

File: src/payment/strategies/braintree/braintree-payment-processor.ts [hunk 5]
```diff
@@ -125,13 +99,63 @@ export default class BraintreePaymentProcessor {
...
+        amount: number,
+        nonce: string
+    ): Promise<BraintreeVerifyPayload> {
+        if (!this._threeDSecureOptions) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
         }
 
+        const { addFrame, removeFrame } = this._threeDSecureOptions;
+        const cancelVerifyCard = () => threeDSecure.cancelVerifyCard()
+            .then(response => {
```
capsula4, line 138: do you think it's worth using async/await?

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.spec.ts
davidchin: `does not initialize hosted form if feature is not enabled` should cover that case.

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
davidchin: We should, but for the next major version. To make this change backward compatible, it has to be optional for now.

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
davidchin: Yes, good suggestion 👍

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
davidchin: Yes, I accidentally removed the guard 😄

davidchin:
@capsula4 thanks for your review. I've done the changes, can you please take a look again when you have some free time? Thanks1

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
capsula4: I don't think you need `|| !orderRequest.payment` in here?

Also, any reason why you removed the destructuring (`const { payment, ...order } = orderRequest;`)? you might not need to use `omit`

davidchin: Good spot. Regarding the destructuring, no particular reason - I think at one point I moved the code to a different place and then I changed my mind but didn't revert cleanly.

PR #942, opened 2020-07-27
feat(checkout): INT-2919 Create an strategy in order to use Bolt in Bigcommerce checkout

FranciscoOnt:
[INT-2919](https://jira.bigcommerce.com/browse/INT-2919)
## What?
Create a strategy to use bolt's client script in the bigcommerce checkout.

## Why?
So the Bolt's client script can be used to process the checkout.

## Testing / Proof
(Pending)

bigcommerce/checkout bigcommerce/payments

### DIFF
--- a/src/payment/create-payment-strategy-registry.ts
+++ b/src/payment/create-payment-strategy-registry.ts
@@ -26,7 +26,7 @@ import { AfterpayPaymentStrategy, AfterpayScriptLoader } from './strategies/afte
 import { AmazonPayPaymentStrategy, AmazonPayScriptLoader } from './strategies/amazon-pay';
 import { createAmazonPayV2PaymentProcessor, AmazonPayV2PaymentStrategy } from './strategies/amazon-pay-v2';
 import { BlueSnapV2PaymentStrategy } from './strategies/bluesnapv2';
-import { BoltPaymentStrategy } from './strategies/bolt';
+import { BoltPaymentStrategy, BoltScriptLoader } from './strategies/bolt';
 import { createBraintreePaymentProcessor, createBraintreeVisaCheckoutPaymentProcessor, BraintreeCreditCardPaymentStrategy, BraintreePaypalPaymentStrategy, BraintreeScriptLoader, BraintreeSDKCreator, BraintreeVisaCheckoutPaymentStrategy, VisaCheckoutScriptLoader } from './strategies/braintree';
 import { CardinalClient, CardinalScriptLoader, CardinalThreeDSecureFlow } from './strategies/cardinal';
 import { ChasePayPaymentStrategy, ChasePayScriptLoader } from './strategies/chasepay';
@@ -508,11 +508,13 @@ export default function createPaymentStrategyRegistry(
     );
 
     registry.register(PaymentStrategyType.BOLT, () =>
-        new BoltPaymentStrategy(
-            store,
-            orderActionCreator,
-            paymentActionCreator
-        )
+    new BoltPaymentStrategy(
+        store,
+        orderActionCreator,
+        paymentActionCreator,
+        paymentMethodActionCreator,
+        new BoltScriptLoader(scriptLoader)
+    )
 );
 
     return registry;
--- a/src/payment/create-payment-strategy-registry.ts
+++ b/src/payment/create-payment-strategy-registry.ts
@@ -26,7 +26,7 @@ import { AfterpayPaymentStrategy, AfterpayScriptLoader } from './strategies/afte
 import { AmazonPayPaymentStrategy, AmazonPayScriptLoader } from './strategies/amazon-pay';
 import { createAmazonPayV2PaymentProcessor, AmazonPayV2PaymentStrategy } from './strategies/amazon-pay-v2';
 import { BlueSnapV2PaymentStrategy } from './strategies/bluesnapv2';
-import { BoltPaymentStrategy, BoltScriptLoader } from './strategies/bolt';
+import { BoltAppPaymentStrategy, BoltScriptLoader } from './strategies/bolt';
 import { createBraintreePaymentProcessor, createBraintreeVisaCheckoutPaymentProcessor, BraintreeCreditCardPaymentStrategy, BraintreePaypalPaymentStrategy, BraintreeScriptLoader, BraintreeSDKCreator, BraintreeVisaCheckoutPaymentStrategy, VisaCheckoutScriptLoader } from './strategies/braintree';
 import { CardinalClient, CardinalScriptLoader, CardinalThreeDSecureFlow } from './strategies/cardinal';
 import { ChasePayPaymentStrategy, ChasePayScriptLoader } from './strategies/chasepay';
@@ -508,13 +508,13 @@ export default function createPaymentStrategyRegistry(
     );
 
     registry.register(PaymentStrategyType.BOLT, () =>
-    new BoltPaymentStrategy(
-        store,
-        orderActionCreator,
-        paymentActionCreator,
-        paymentMethodActionCreator,
-        new BoltScriptLoader(scriptLoader)
-    )
+        new BoltAppPaymentStrategy(
+            store,
+            orderActionCreator,
+            paymentActionCreator,
+            paymentMethodActionCreator,
+            new BoltScriptLoader(scriptLoader)
+        )
 );
 
     return registry;
--- a/src/payment/strategies/bolt/bolt-app-payment-strategy.spec.ts
+++ b/src/payment/strategies/bolt/bolt-app-payment-strategy.spec.ts
@@ -25,163 +25,171 @@ import BoltAppPaymentStrategy from './bolt-app-payment-strategy';
 import BoltScriptLoader from './bolt-script-loader';
 
 describe('BoltAppPaymentStrategy', () => {
-  let boltClient: BoltCheckout;
-  let boltScriptLoader: BoltScriptLoader;
-  let options: PaymentRequestOptions;
-  let orderActionCreator: OrderActionCreator;
-  let requestSender: RequestSender;
-  let store: CheckoutStore;
-  let strategy: BoltAppPaymentStrategy;
-  let submitOrderAction: Observable<SubmitOrderAction>;
-  let submitPaymentAction: Observable<SubmitPaymentAction>;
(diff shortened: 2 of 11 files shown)


src/payment/strategies/bolt/bolt-app-payment-strategy.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+    });
+  });
+
+  describe('#finalize()', () => {
+      it('throws error to inform that order finalization is not required', async () => {
+          try {
+              await strategy.finalize();
+          } catch (error) {
+              expect(error).toBeInstanceOf(OrderFinalizationNotRequiredError);
+          }
```
mauricio-sg, line 188: 🍹 
```suggestion
          await expect(strategy.finalize()).rejects.toThrow(OrderFinalizationNotRequiredError);
```

src/payment/strategies/bolt/bolt.mock.ts
```diff
@@ -0,0 +1,24 @@
...
+export function getBoltScriptMock(success: boolean = false): BoltCheckout {
+    return {
+        configure: jest.fn((_cart: object, _hints: {}, callbacks?: BoltCallbacks) => {
+            return getConfiguredBoltMock(success, callbacks || { success: () => {}, close: () => {}});
+        }),
+        setClientCustomCallbacks: jest.fn(),
+    };
+}
+
+export function getConfiguredBoltMock(success: boolean, callbacks: BoltCallbacks): BoltClient {
```
capsula4, line 12: maybe for clarify rename `success` to `shouldSucceed` ?

src/payment/create-payment-strategy-registry.ts
```diff
@@ -508,10 +508,12 @@ export default function createPaymentStrategyRegistry(
     );
 
     registry.register(PaymentStrategyType.BOLT, () =>
-        new BoltPaymentStrategy(
```
capsula4, line 511: Is the previous `BoltPaymentStrategy` still in use?

src/payment/strategies/bolt/bolt-app-payment-strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+import PaymentMethodActionCreator from '../../payment-method-action-creator';
+import { PaymentInitializeOptions, PaymentRequestOptions } from '../../payment-request-options';
+import PaymentStrategy from '../payment-strategy';
+
+import { BoltCheckout, BoltTransacion } from './bolt';
+import BoltScriptLoader from './bolt-script-loader';
+
+export default class BoltAppPaymentStrategy implements PaymentStrategy {
+    private _boltClient?: BoltCheckout;
+    private _methodId!: string;
```
capsula4, line 17: what's the `!` for?

src/payment/strategies/bolt/bolt-app-payment-strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+        this._methodId = options.methodId;
+
+        const state = this._store.getState();
+        const paymentMethod = state.paymentMethods.getPaymentMethod(this._methodId);
+
+        if (!paymentMethod || !paymentMethod.initializationData.publishableKey) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        const publishableKey = paymentMethod.initializationData.publishableKey;
```
capsula4, line 37: minor, you could do
```js
const { publishableKey } = paymentMethod.initializationData;
```

src/payment/strategies/bolt/bolt-app-payment-strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+                reject(new PaymentMethodCancelledError());
+            };
+
+            const callbacks = {
+                success: onSuccess,
+                close: onClose,
+            };
+
+            if (!this._boltClient) {
+                throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
```
capsula4, line 84: should you reject the promise in here?
```
reject(new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized))`

File: src/payment/strategies/bolt/bolt-app-payment-strategy.ts [hunk 7]
```diff
@@ -0,0 +1,106 @@
...
+            }
+
+            this._boltClient.configure({ orderToken }, {}, callbacks).open();
```
capsula4, line 87: can this fail? if so, might be worth checking for failures and rejecting the promise.

src/payment/strategies/bolt/bolt.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export interface BoltCart {
+    orderToken: string;
+}
+
+export interface BoltCallbacks {
+    check?(): boolean;
+    onCheckoutStart?(): void;
+    onPaymentSubmit?(): void;
+    success(transaction: any, callback: () => void): void;
```
capsula4, line 22: should this be `transaction: BoltTransacion` ?

src/payment/strategies/bolt/bolt.ts
```diff
@@ -0,0 +1,39 @@
...
+    close?(): void;
+}
+
+export interface BoltTransacion {
```
capsula4, line 26: there's a typo, `BoltTransaction`

src/payment/strategies/bolt/bolt-app-payment-strategy.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+      );
+  });
+
+  describe('#initialize', () => {
+    it('successfully initializes the bolt strategy and loads the bolt client', async () => {
+        await expect(strategy.initialize(options)).resolves.toEqual(store.getState());
+        expect(boltScriptLoader.load).toHaveBeenCalledWith('publishableKey', true);
+    });
+
+    it('fails to initialize the bolt strategy if no publishableKey is not provided', async () => {
```
capsula4, line 110: `if publishableKey is not provided`

src/payment/strategies/bolt/bolt-app-payment-strategy.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+        store,
+        orderActionCreator,
+        paymentActionCreator,
+        paymentMethodActionCreator,
+        boltScriptLoader
+      );
+  });
+
+  describe('#initialize', () => {
+    it('successfully initializes the bolt strategy and loads the bolt client', async () => {
```
capsula4, line 105: this file seems to have a mix of indentation styles, we use 4 spaces for indentation.

src/payment/strategies/bolt/bolt-app-payment-strategy.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+      },
+    };
+    options = {
+      methodId: 'bolt',
+    };
+
+    it('successfully executes the bolt strategy and submits payment', async () => {
+        await strategy.initialize(options);
+        await strategy.execute(payload);
+        expect(boltClient.configure).toHaveBeenCalledWith(expect.any(Object), {}, expect.any(Object));
```
capsula4, line 132: instead of `expect.any(Object)` could you be more explicit and use something like

`expect.objectContaining({ .... })`

src/payment/strategies/bolt/bolt-app-payment-strategy.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+      paymentMethodMock.clientToken = undefined;
+      await strategy.initialize(options);
+      await expect(strategy.execute(payload)).rejects.toThrow(MissingDataError);
+      expect(boltClient.configure).not.toHaveBeenCalled();
+      expect(paymentActionCreator.submitPayment).not.toHaveBeenCalled();
+    });
+
+    it('fails to execute the bolt strategy if the client script is not loaded', async () => {
+      jest.spyOn(boltScriptLoader, 'load')
+          .mockResolvedValue(undefined);
```
capsula4, line 154: I think this test is irrelevant, as boltScriptLoader cannot return undefined. Either throws an exception or returns the actual client.

src/payment/create-payment-strategy-registry.ts
FranciscoOnt: It will not be used in the payment step, it will only be used by the bolt's app.
I don't know if that strategy will still be able to be used even if it's not in the registry, is that possible @capsula4 ?

src/payment/strategies/bolt/bolt-app-payment-strategy.ts
FranciscoOnt: It allows the variable not to be initialized and not to be assigned in the constructor.

src/payment/strategies/bolt/bolt-app-payment-strategy.ts
FranciscoOnt: If I validate this outside the promise, it will mask the boltClient as possibly undefined
Screen Shot 2020-07-29 at 12 02 37 PM

src/payment/strategies/bolt/bolt-app-payment-strategy.ts
FranciscoOnt: I believe the errors are handled by the client script, also this method of calling their client was suggested by Bolt themselves.

src/payment/strategies/bolt/bolt-app-payment-strategy.spec.ts
FranciscoOnt: the purpose of this is to be able to test the case where the bolt client is not loaded, which is due to typescript trying to validate the bolt client that was declared as `private _boltClient?: BoltCheckout;` as possibly undefined, to avoid this its possible to declare it the same way as the methodId `private _boltClient!: BoltCheckout;`  but not sure if this is a good idea, as every other provider does not do this and instead validates the existence of the loaded script.

src/payment/create-payment-strategy-registry.ts
capsula4: if it's not in the registry most likely is not in use, since the new strategy doesn't really extend the old one. You can do a search in the repo but most likely is not in use.

PR #947, opened 2020-08-06
fix(payment): INT-3010 Fix Zip store credit implementation

FranciscoOnt:
## What? [INT-3010](https://jira.bigcommerce.com/browse/INT-3010)
Restore the `remoteCheckoutActionCreator` calls for Zip

## Why?
This call stores the `useStoreCredit` on the backend, so when loading the payment method, the store credit can still be applied. 
Apply Store Credit unsets this variable in the backend, but is later required in the `loadPaymentMethod`call.
## Testing / Proof
[Demo Video](https://drive.google.com/file/d/1ZH8JuYecxlImYbYcI8abkxvpvHHS8khD/view?usp=sharing)

### DIFF
--- a/src/payment/strategies/zip/zip-payment-strategy.ts
+++ b/src/payment/strategies/zip/zip-payment-strategy.ts
@@ -5,6 +5,7 @@ import { InvalidArgumentError, MissingDataError, MissingDataErrorType, NotInitia
 import { ContentType, INTERNAL_USE_ONLY } from '../../../common/http-request';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
+import { RemoteCheckoutActionCreator } from '../../../remote-checkout';
 import { StoreCreditActionCreator } from '../../../store-credit';
 import { PaymentMethodCancelledError, PaymentMethodDeclinedError, PaymentMethodInvalidError } from '../../errors';
 import PaymentActionCreator from '../../payment-action-creator';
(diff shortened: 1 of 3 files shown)


pgonzalezr:
💚 Looks good from QE side.

Tested:
- Orders without store credit
- Orders that have a 0 balance after applying store credit
- Orders with a greater amount than available store credit

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -56,14 +58,15 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
...
-
-        const { useStoreCredit } = payload;
+        const { isStoreCreditApplied: useStoreCredit } = this._store.getState().checkout.getCheckoutOrThrow();
 
         if (useStoreCredit !== undefined) {
             await this._store.dispatch(this._storeCreditActionCreator.applyStoreCredit(useStoreCredit));
         }
 
+        await this._store.dispatch(this._orderActionCreator.submitOrder(order, options));
+        await this._store.dispatch(this._remoteCheckoutActionCreator.initializePayment(payment.methodId, { useStoreCredit }));
```
capsula4, line 68: I don't think you need to call remoteCheckoutActionCreator. The store credit was probably not working because the order was being created first. Store credit needs to happen first.

FranciscoOnt: It will not work properly if if the remoteCheckoutActionCreator is not called, when store  credit is applied it unsets a session variable in the backend which is validated in the backend when the payment method is loaded (to get the client token) in line 70, without this variable the token is generated without the store credit applied, the remoteCheckoutActionCreator sets the variable and allows the token to be generated with the store credit applied.

capsula4: Thanks for explaining. Part of the idea behind the new StoreCreditActionCreator was to remove the need of using `remoteCheckoutActionCreator` for store credit purposes, but I guess there is more work required to be done in the backend to fix it.

PR #960, opened 2020-08-18
Fix(Payment): Fixed Chase Pay sign in link

JoseAlvarezHdz:
## What? [INT-2901](https://jira.bigcommerce.com/browse/INT-2901)
Implement a Interval function to try to find correct a element and add the event listener to it

## Why?
Because as other payments has the same ID it was adding the event listener to the previous element

## Testing / Proof

testing

This fix will be made in checkout-js


JoseAlvarezHdz:
This fix will be made in checkout-js

PR #980, opened 2020-09-09
feat(payments): PAYPAL-654 Spb checkout

helga1507:
## What?

- Smart payment buttons were added to the checkout page

- For every method  implemented his own button

- Adding an embedded button instead of the submit button if buyer didn't approve PayPal account on the cart page

- Submitting form automatically after approve PayPal account

## Why?
PPCP pop up is blocked on the BC hosted checkout page and Embedded checkout. Pop ups work for SPB at the shopping cart page. 

## Testing / Proof
image


src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -1,47 +1,88 @@
...
-    initialize({ paypalcommerce }: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
-        this._paypalCommercePaymentProcessor.initialize({ overlay: paypalcommerce && paypalcommerce.overlay });
+    async initialize({ methodId, paypalcommerce }: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
+        const state = await this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId));
+        this._paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(methodId);
+        const { initializationData } = this._paymentMethod;
 
-        return Promise.resolve(this._store.getState());
+        if (initializationData.orderId) {
+            this._orderId = initializationData.orderId;
```
davidchin, line 37: It seems you're not rendering the PayPal button if the shopper has already started the flow elsewhere (i.e.: cart page). How would the caller know when they need to render a normal submit button vs not render one because PayPal will insert one themselves? Is it through `isShowEmbeddedSubmitButton` getter you've added? If that's the case, I don't particularly like the way we reveal that information through `CheckoutState`, because it is a provider-specific piece of information that is surfaced through an interface that's not provider-specific. I think it might be better if you surface it through a callback function that's specific to PayPal Commerce. i.e.:

```ts
initializePayment({
    paypalcommerce: {
        onRender() {
            // When this is called, the caller would know that PayPal button is rendered and
            // therefore can remove the standard submit button. And when the user switches 
            // to a different payment method, the caller can restore the standard submit button.
        }
    }
})
```

src/payment/strategies/paypal-commerce/paypal-commerce-hosted-form.ts
```diff
@@ -14,55 +11,44 @@ enum PaypalCommerceHostedFormType {
...
     constructor(
-        private _paypalCommerceRequestSender: PaypalCommerceRequestSender
+        private _paypalCommercePaymentProcessor: PaypalCommercePaymentProcessor
     ) {}
 
-    async initialize(options: PaypalCommerceFormOptions, cartId: string, paypal: PaypalCommerceSDK) {
-        if (!paypal.HostedFields) {
-            throw new PaymentMethodClientUnavailableError();
-        }
+    async initialize(options: PaypalCommerceFormOptions, cartId: string, paramsScript: DataPaypalCommerceScript) {
```
davidchin, line 20: Is there a reason why changes to PayPal hosted credit card fields are bundled as a part of the PR for PayPal flow? It'd be nice to keep them as separate PRs unless they really have to go together.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-initialize-options.ts
```diff
@@ -11,6 +10,8 @@ export interface PaypalCommercePaymentInitializeOptions {
      * consumption.
      */
     form?: PaypalCommerceFormOptions;
+    style?: PaypalButtonStyleOptions;
+    submitForm?(): void;
```
davidchin, line 14: I'm not sure if it can be an optional parameter consider it is required in order to proceed (unless the shopper has started the flow elsewhere, i.e.: cart page).

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -55,22 +96,36 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+        this._isPaypalButton = false;
 
         return Promise.resolve(this._store.getState());
     }
 
-    private async _getOrderId(methodId: string): Promise<string> {
-        const state = this._store.getState();
-        const cart = state.cart.getCartOrThrow();
-        const provider = methodId === 'paypalcommercecredit' ? 'paypalcommercecreditcheckout' : 'paypalcommercecheckout';
-        const { approveUrl, orderId } = await this._paypalCommerceRequestSender.setupPayment(provider, cart.id);
```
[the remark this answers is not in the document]
BC-ACherednichenko: thanks for the suggestion @davidchin, there is an option to send ajax call in advance, however, initially, we have discussed with PayPal an alternative solution for opening the popup via native PayPal SDK, because `window.open` solution is tied to events that PayPal does not guaratee to be working in future (for popup closing) https://github.com/bigcommerce/checkout-sdk-js/pull/980/files#diff-3d44283948e4a50fdfa4cefc761563bfL59 

```js
 const messageHandler = (event: MessageEvent) => {
    if (event.origin !== 'https://www.sandbox.paypal.com' && event.origin !== 'https://www.paypal.com') {
       return;
    }

   const data = JSON.parse(event.data);

   if (data.operation === 'return_to_merchant' && data.updateParent) {
        this._window.removeEventListener('message', messageHandler);
        closeWindow(true);
   }
};
this._window.addEventListener('message', messageHandler);
```
So we decided to move on with the native PayPal buttons solution, as in that case there is a guarantee so it can work in the future + PayPal maintains popup solution for different browsers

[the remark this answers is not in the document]
helga1507: I'm not sure that it's good point. We decided to implement PayPal smart button to checkout. 
With sbp we shouldn't worry about browser support, postMessages, changes on the PayPal side (with `window.open` we're waiting message with `operation` and `updateParent` and if PayPal change it - it wouldn't work in our side)
And now it work well

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -1,47 +1,88 @@
...
-        const orderId = paymentMethod.initializationData.orderId || await this._getOrderId(options.methodId);
-
         const paymentData =  {
             formattedPayload: {
                 vault_payment_instrument: null,
                 set_as_default_stored_instrument: null,
                 device_info: null,
                 paypal_account: {
-                    order_id: orderId,
+                    order_id: this._orderId,
```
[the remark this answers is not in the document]
helga1507: Yes, you are right. Thank you!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-initialize-options.ts
helga1507: Hm, but we have credit cards and for them it's  optional parameter. And as you noticed, If the shopper approved account on the cart page.
And I added checks on this params [here](https://github.com/bigcommerce/checkout-sdk-js/pull/980/files#diff-3344e5b56907eaf188a8208fc12e5a51R48)

src/payment/strategies/paypal-commerce/paypal-commerce-hosted-form.ts
helga1507: Actually, now feature with hosted fields doesn't work for shoppers (we're testing it yet)

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
helga1507: Thank you for suggestions!

Yes, we should render embedded submit button only in case when shopper didn't approve PayPal account (i.e.: cart page). I think submit button depends on the checkout-sdk and in my case I call `enableEmbeddedSubmitButton ` [here](https://github.com/bigcommerce/checkout-sdk-js/pull/980/files#diff-3344e5b56907eaf188a8208fc12e5a51R62) if we should hide standard button. And call `disableEmbeddedSubmitButton` in the `deinitialize` [here](https://github.com/bigcommerce/checkout-sdk-js/pull/980/files#diff-3344e5b56907eaf188a8208fc12e5a51R105) for remove embedded button.
This actions change `isShowEmbeddedSubmitButton` in state and it use in checkout-js [in this PR](https://github.com/bigcommerce/checkout-js/pull/403). And in my point of view it work like `isInitializing`, `isSubmitting` and others ways of communication between two applications

I'll try your idea with `onRender`. But I have some questions. Do you mean call some action in `onRender` for changing state of `PaymentForm` and render standard or embedded button? Correct?

BC-ACherednichenko: @davidchin what are the possible downsides for having communication via `CheckoutState`? I understand the concern about the leakage of the provider-specific data, however, as Olga mentioned above, we think that `isShowEmbeddedSubmitButton` would be the default way of communication between two apps along with `isInitializing`, `isSubmitting`

davidchin (commented):
Nice work @helga1507

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -1,47 +1,97 @@
...
+        private _paymentMethodActionCreator: PaymentMethodActionCreator,
+        private _paypalCommercePaymentProcessor: PaypalCommercePaymentProcessor,
+        private _credit: boolean = false
     ) {}
 
-    initialize({ paypalcommerce }: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
-        this._paypalCommercePaymentProcessor.initialize({ overlay: paypalcommerce && paypalcommerce.overlay });
+    async initialize({ methodId, paypalcommerce }: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
+        const state = await this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId));
+        this._paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(methodId);
```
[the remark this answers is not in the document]
helga1507: You are right. I forgot to remove it. Thank you!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -1,47 +1,97 @@
...
+            this._orderId = initializationData.orderId;
+
+            return this._store.getState();
+        }
+
+        if (!paypalcommerce) {
+            throw new InvalidArgumentError('Unable to initialize payment because "options.paypalcommerce" argument is not provided.');
+        }
+
+        const { container, hidePaymentButton, submitForm, style, onError = noop } = paypalcommerce;
```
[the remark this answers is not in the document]
helga1507: Oh, great. Thank you!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -1,47 +1,97 @@
...
+        const paramsScript = { options: this._getOptionsScript(initializationData, cart) };
+        const buttonParams: ButtonsOptions = { style, onApprove: data => this._tokenizePayment(data, submitForm) };
+
+        await this._paypalCommercePaymentProcessor.initialize(paramsScript);
+
+        try {
+            this._paypalCommercePaymentProcessor.renderButtons(cart.id, container, buttonParams, {
+                fundingKey: this._credit ? 'CREDIT' : 'PAYPAL',
+                paramsForProvider: {isCheckout: true},
+                cbIfEligible: hidePaymentButton,
```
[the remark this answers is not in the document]
helga1507: Yeah, good idea. Thank you!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -1,47 +1,97 @@
...
+            });
+        } catch (error) {
+            onError();
```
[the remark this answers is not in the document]
helga1507: Okay. Thank you

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -1,111 +1,175 @@
...
+    }
+
+    async renderHostedFields(cartId: string, params: ParamsRenderHostedFields, events?: EventsHostedFields): Promise<void> {
+        if (!this._paypal || !this._paypal.HostedFields) {
+            throw new PaymentMethodClientUnavailableError();
+        }
+
+        const { fields, styles } = params;
+
+        if (this._paypal.HostedFields.isEligible()) {
```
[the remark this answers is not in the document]
helga1507: Sure. Thank you!

helga1507:
@davidchin Thank you for review! I added all changes!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-initialize-options.ts
```diff
@@ -1,18 +1,24 @@
...
+
 export interface PaypalCommercePaymentInitializeOptions {
-    overlay?: {
-        helpText?: string;
-        continueText?: string;
-    };
+    container: string;
+    style?: PaypalButtonStyleOptions;
+    submitForm(): void;
+    hidePaymentButton?(): void;
```
davidchin, line 7: Can we name this parameter as `onRenderButton` as well?

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -55,22 +89,30 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
         return Promise.resolve(this._store.getState());
     }
 
-    private async _getOrderId(methodId: string): Promise<string> {
-        const state = this._store.getState();
-        const cart = state.cart.getCartOrThrow();
-        const provider = methodId === 'paypalcommercecredit' ? 'paypalcommercecreditcheckout' : 'paypalcommercecheckout';
-        const { approveUrl, orderId } = await this._paypalCommerceRequestSender.setupPayment(provider, cart.id);
+    private _isPaypalCommerceOptionsPayments(options: PaypalCommercePaymentInitializeOptions | PaypalCommerceCreditCardPaymentInitializeOptions): options is PaypalCommercePaymentInitializeOptions {
+        return !!(options as PaypalCommercePaymentInitializeOptions).container;
```
davidchin, line 99: `return 'container' in options;` so you don't need to use `as` 🍹

helga1507: It doesn't work without `as`. I see 
`TS2339: Property 'container' does not exist on type 'PaypalCommerceInitializeOptions'.   Property 'container' does not exist on type 'PaypalCommerceCreditCardPaymentInitializeOptions'.`

src/payment/strategies/paypal-commerce/paypal-commerce-payment-initialize-options.ts
helga1507: Sure

davidchin (approved):
Great job, LGTM! 👍

PR #986, opened 2020-09-22
feat(payment): PAYPAL-675 Upgrade to 3DS v2 Braintree

bc-fetisov:
## What?
Update Braintree to 3DS v2
https://developers.braintreepayments.com/guides/3d-secure/migration/javascript/v3

## Why?
Because of right now we are supporting v1 only 

## Testing / Proof
3DS v1
Screenshot 2020-09-22 at 14 48 51

3DS v2
Screenshot 2020-09-22 at 14 57 54

### DIFF
--- a/src/payment/strategies/braintree/braintree-payment-processor.ts
+++ b/src/payment/strategies/braintree/braintree-payment-processor.ts
@@ -150,6 +150,9 @@ export default class BraintreePaymentProcessor {
                 amount,
                 nonce,
                 removeFrame,
+                onLookupComplete: (_data, next) => {
+                    next();
+                },
             })
         );
 
--- a/src/payment/strategies/braintree/braintree.ts
+++ b/src/payment/strategies/braintree/braintree.ts
@@ -92,7 +92,7 @@ export interface BraintreeThreeDSecureOptions {
     showLoader?: boolean;
     addFrame(error: Error | undefined, iframe: HTMLIFrameElement): void;
     removeFrame(): void;
-    onLookupComplete(data: any, next: any): void;
+    onLookupComplete(data: BraintreeThreeDSecureVerificationData, next: () => void): void;
 }
 
 export interface BraintreeDataCollector extends BraintreeModule {
@@ -334,5 +334,17 @@ export interface BraintreeError extends Error {
 export interface BraintreeHostedFormError extends BraintreeError {
     details?: {
         invalidFieldKeys?: string[];
+    };
+}
+
+interface BraintreeThreeDSecureVerificationData {
+    lookup: {
+        threeDSecureVersion: string;
+    };
+    paymentMethod: BraintreeVerifyPayload;
+    requiresUserAuthentication: boolean;
+    threeDSecureInfo: {
+        liabilityShiftPossible: boolean;
+        liabilityShifted: boolean;
     };
 }
--- a/src/payment/strategies/braintree/braintree.ts
+++ b/src/payment/strategies/braintree/braintree.ts
@@ -28,6 +28,10 @@ export interface BraintreeDataCollectorCreatorConfig extends BraintreeModuleCrea
     paypal?: boolean;
 }
 
+export interface BraintreeThreeDSecureCreatorConfig extends BraintreeModuleCreatorConfig {
+    version?: number;
+}
+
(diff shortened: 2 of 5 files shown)


helga1507 (approved):

bc-fetisov:
@davidchin could you review this PR please?

PascalZajac:
Is this no longer WIP?

src/payment/strategies/braintree/braintree-payment-processor.ts
```diff
@@ -150,6 +150,9 @@ export default class BraintreePaymentProcessor {
                 amount,
                 nonce,
                 removeFrame,
+                onLookupComplete: (_data, next) => {
```
davidchin, line 153: It appears that `onLookupComplete` is a deprecated option according to the [doc](https://braintree.github.io/braintree-web/current/ThreeDSecure.html#verifyCard). It suggests us to do something like `threeDSecure.on('lookup-complete')` instead.

I also notice that there are a couple of other [deprecated options](https://developers.braintreepayments.com/guides/3d-secure/migration/javascript/v3#can-merchants-customize-the-presentation-of-the-bank-challenge?) that are only relevant for 3DS1, i.e.: `addFrame` and `removeFrame` because, with 3DS2, the presentation of the challenge is controlled by the issuing bank. Perhaps they don't need to be passed if we're migrating everyone to v2.

Another question I have is backwards compatibility. Is this upgrade a backwards compatible upgrade?

src/payment/strategies/braintree/braintree-payment-processor.ts
```diff
@@ -150,6 +150,9 @@ export default class BraintreePaymentProcessor {
...
+                    next();
```
davidchin, line 154: Do we need to inspect [data](https://braintree.github.io/braintree-web/current/ThreeDSecure.html#~verificationData) parameter before calling `next`? i.e.: Is there a scenario where we shouldn't call `next` and have to cancel the verification flow (i.e.: if the lookup fails or authentication is not required)?

src/payment/strategies/braintree/braintree.ts
```diff
@@ -88,6 +92,7 @@ export interface BraintreeThreeDSecureOptions {
     showLoader?: boolean;
     addFrame(error: Error | undefined, iframe: HTMLIFrameElement): void;
     removeFrame(): void;
+    onLookupComplete(data: any, next: any): void;
```
davidchin, line 95: Prefer to have definitive types rather than `any`.

bc-fetisov: Ok, i've added new interface according to https://braintree.github.io/braintree-web/current/ThreeDSecure.html#~verificationData

src/payment/strategies/braintree/braintree-payment-processor.ts
bc-fetisov: Thanks for pointing to me to their SDK reference. I see that couple of their options in it shown as deprecated. Thay obviously have a documentation inconsistency :(
I've made changes according to https://developers.braintreepayments.com/guides/3d-secure/migration/javascript/v3 guide and tested it with provided test cards. All works as expected:
both v1 and v2 test cases are working, so to be honest, i don't see any reason to change something else radically from this point. 
 
> Is this upgrade a backwards compatible upgrade? 

If I understood the question correctly, yes.
`3DS 2 works on top of your existing 3DS 1 integration. The SDK will fall back to 3DS 1 when 3DS 2 is unavailable for the card being verified.`

src/payment/strategies/braintree/braintree-payment-processor.ts
bc-fetisov: Right now i cant see any real cases. But according to this reference
https://braintree.github.io/braintree-web/current/ThreeDSecure.html#cancelVerifyCard its possible to use `cancelVerifyCard` callback type if need.

davidchin: Ok if you can’t think of a real case right now, then I’m happy with what you have got. We can change it later if we discover there’s a real need.

davidchin (approved):
LGTM, nice work! 👍

PR #990, opened 2020-09-28
feat(payment): INT-3086 Rename containerId to elementOrSelector

vmparra:
## What? [INT-3086](https://jira.bigcommerce.com/browse/INT-3086)
...

## Why?
...

## Testing / Proof
-- Unit

1. Revert this PR

### DIFF
--- a/src/payment/strategies/stripev3/stripev3-initialize-options.ts
+++ b/src/payment/strategies/stripev3/stripev3-initialize-options.ts
@@ -1,4 +1,4 @@
-import { IndividualCardElementOptions, StripeElementOptions } from './stripev3';
+import { ElementOrSelector, IndividualCardElementOptions, StripeElementOptions } from './stripev3';
 
 /**
  * A set of options that are required to initialize the Stripe payment method.
@@ -12,7 +12,7 @@ export default interface StripeV3PaymentInitializeOptions {
     /**
      * The location to insert the credit card number form field.
      */
-    containerId: string;
+    containerId: ElementOrSelector;
 
     options?: StripeElementOptions | IndividualCardElementOptions;
 }
--- a/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
+++ b/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
@@ -278,9 +278,9 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
                         stripeElement = this._stripeCardElements[0];
 
                         try {
-                            cardNumberElement.mount(`#${cardNumberElementOptions.containerId}`);
-                            cardExpiryElement.mount(`#${cardExpiryElementOptions.containerId}`);
-                            cardCvcElement.mount(`#${cardCvcElementOptions.containerId}`);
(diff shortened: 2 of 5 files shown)


mauricio-sg (commented):
I'm afraid I didn't get the point of this change. Why did you rename `containerId` to `elementOrSelector` if you are still mounting and querying by id? 🤔

vmparra:
> I'm afraid I didn't get the point of this change. Why did you rename `containerId` to `elementOrSelector` if you are still mounting and querying by id? 🤔

I just think it makes more sense to name it this way. 😄

mauricio-sg (approved):

src/payment/strategies/stripev3/stripev3-initialize-options.ts
```diff
@@ -12,7 +12,7 @@ export default interface StripeV3PaymentInitializeOptions {
     /**
      * The location to insert the credit card number form field.
      */
-    containerId: string;
+    containerId: ElementOrSelector;
```
capsula4, line 15: is this a breaking change?

it's a bit misleading that a `containerId` property can be an `element` or a `selector`

Even if it's just a selector, the selector would be `#inputId` while the containerId would simply be `inputId`

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -278,9 +278,9 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
                         stripeElement = this._stripeCardElements[0];
 
                         try {
-                            cardNumberElement.mount(`#${cardNumberElementOptions.containerId}`);
-                            cardExpiryElement.mount(`#${cardExpiryElementOptions.containerId}`);
-                            cardCvcElement.mount(`#${cardCvcElementOptions.containerId}`);
+                            cardNumberElement.mount(`#${cardNumberElementOptions.elementOrSelector}`);
```
capsula4, line 281: ````#${cardNumberElementOptions.elementOrSelector}````

this still seems to be an id? we are prepending `#`

vmparra: > `#${cardNumberElementOptions.elementOrSelector}`
> 
> this still seems to be an id? we are prepending `#`

Mount supports CSS Selectors, see https://stripe.com/docs/js/element/mount

src/payment/strategies/stripev3/stripev3-initialize-options.ts
vmparra: > is this a breaking change?
> 
> it's a bit misleading that a `containerId` property can be an `element` or a `selector`
> 
> Even if it's just a selector, the selector would be `#inputId` while the containerId would simply be `inputId`
> 
> maybe we should deprecate `containerId` and create a new one `container: ElementOrSelector ` ? And explain properly in the description that it could either be an element or a selector (not an ID)

I don't believe this is a breaking change as we've always been using `mount` with a #

vmparra:
Isn't necessary at the moment.

PR #992, opened 2020-10-01
fix(payment): PAYPAL-726 Switch payment methods

helga1507:
## What?
Fix bug, when buyer can't switch on the checkout between payment methods
Add one script for different methods (https://github.com/paypal/paypal-js)

## Why?
Script changed each time when buyer changed method
I call `createPaypalCommercePaymentProcessor` separately for each method, because in this case we can `close` PayPal buttons for the method that was closed. In previous case (one `createPaypalCommercePaymentProcessor` for all methods) I had errors, because `deinitialize` was called after `initialize` and close button

## Testing / Proof
https://www.loom.com/share/c54ef8c4859e48c1a201f6b5ef281836


BC-ACherednichenko (approved):
LGTM

davidchin (approved):

PR #995, opened 2020-10-05
fix(payment): PAYPAL-202 fix validation

helga1507:
## What?
Fix validation on credit cards

## Why?
Credit cards had issues in validation

## Testing / Proof
https://www.loom.com/share/38386a47ce44459e865bbf336269caa9


davidchin (approved):
Nice one, it's a similar fix we had to do to Braintree recently as well 😄

PR #996, opened 2020-10-06
feat(payment): PAYPAL-706 Add validate before using spb

helga1507:
## What?
Add validate before using spb
Related task: https://github.com/bigcommerce/checkout-js/pull/427

## Why?
We should validate form before submit

## Testing / Proof
https://www.loom.com/share/94ffcc2a7bc04fb8942ad4732746fc4a

### DIFF
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
@@ -50,10 +50,10 @@ export default class PaypalCommercePaymentProcessor {
         const buttonParams: ButtonsOptions = {
             ...params,
             createOrder: () => this._setupPayment(cartId, paramsForProvider),
-            onClick: data => {
+            onClick: async (data, actions) => {
                 this._fundingSource = data.fundingSource;
 
-                params.onClick?.(data);
+                return params.onClick?.(data, actions);
             },
         };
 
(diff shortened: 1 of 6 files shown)


src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -50,10 +50,10 @@ export default class PaypalCommercePaymentProcessor {
         const buttonParams: ButtonsOptions = {
             ...params,
             createOrder: () => this._setupPayment(cartId, paramsForProvider),
-            onClick: data => {
+            onClick: async (data, actions) => {
                 this._fundingSource = data.fundingSource;
 
-                params.onClick?.(data);
+                return params.onClick?.(data, actions);
```
davidchin, line 56: `onClick` doesn't seem to return a promise (unlike `onValidate`). So there's probably no need for `async` or call `return`.

davidchin (approved):
A small comment, otherwise LGTM 👍

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
helga1507: It depends on purpose. You can use it with promise too
Example from PayPal here https://developer.paypal.com/docs/checkout/integration-features/validation/#asynchronous-validation

PR #998, opened 2020-10-07
feat(checkout): INT-3174 Added title as attribute of flashMessage

roberbnd:
## What? [INT-3174](https://jira.bigcommerce.com/browse/INT-3174)
It adds the property title 

## Why?
The title is a property that we can change

## Testing / Proof
https://drive.google.com/file/d/1ftq1CteeynosKSd1roKwBKE5_JGvtbHi/view?usp=sharing

## Siblings PRs
https://github.com/bigcommerce/checkout-js/pull/429
https://github.com/bigcommerce/bigcommerce/pull/37501

### DIFF
--- a/src/config/config-selector.spec.ts
+++ b/src/config/config-selector.spec.ts
@@ -40,10 +40,10 @@ describe('ConfigSelector', () => {
 
     describe('#getFlashMessages', () => {
         const flashMessages: FlashMessage[] = [
-            { type: 'info', message: 'm_info' },
-            { type: 'error', message: 'm_error' },
-            { type: 'warning', message: 'm_warning' },
-            { type: 'success', message: 'm_success' },
+            { type: 'info', message: 'm_info', title: '' },
+            { type: 'error', message: 'm_error', title: '' },
+            { type: 'warning', message: 'm_warning', title: '' },
+            { type: 'success', message: 'm_success', title: '' },
         ];
 
         beforeEach(() => {
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -114,6 +114,7 @@ export type FlashMessageType = 'error' | 'info' | 'warning' | 'success';
 export interface FlashMessage {
     type: FlashMessageType;
     message: string;
+    title?: string;
 }
 
 export interface ContextConfig {


src/config/config-selector.spec.ts
```diff
@@ -40,10 +40,10 @@ describe('ConfigSelector', () => {
 
     describe('#getFlashMessages', () => {
         const flashMessages: FlashMessage[] = [
-            { type: 'info', message: 'm_info' },
-            { type: 'error', message: 'm_error' },
-            { type: 'warning', message: 'm_warning' },
-            { type: 'success', message: 'm_success' },
+            { type: 'info', title: '', message: 'm_info' },
```
vmparra, line 43: Can we move this field to the end of the object initializer?

cmedinilla (approved):
LGTM!

vmparra (changes requested):
Hi @roberbnd I've just left a comment there, please let me know if you agree.

roberbnd:
@vmparra could you review, please? ♻️

vmparra (approved):
LGTM 👍🏻

mmartinezintersys:
💚 TESTED

QA Details:

 - Verify the error dialog title is changed to **Payment was cancelled** when payment is cancelled 
 - Verify the error dialog message is changed to **Please try another payment method** when payment is cancelled

src/config/config.ts
```diff
@@ -114,6 +114,7 @@ export type FlashMessageType = 'error' | 'info' | 'warning' | 'success';
 export interface FlashMessage {
     type: FlashMessageType;
     message: string;
+    title: string;
```
capsula4, line 117: do all flash messages return a title in the API? could they ever be null?

roberbnd: yes, at least  a string `''`. As you can see it's the default value.
https://github.com/bigcommerce/bigcommerce/pull/37501/files#diff-49465e72c7fe395e75d0f7d26783870aecddd0fc85133dd55cf306d975a6c355R176-R190

roberbnd:
@capsula4  could you review, please? ♻️

capsula4 (approved):

src/config/config.ts
```diff
@@ -114,6 +114,7 @@ export type FlashMessageType = 'error' | 'info' | 'warning' | 'success';
 export interface FlashMessage {
     type: FlashMessageType;
     message: string;
+    title: string | null;
```
capsula4, line 117: we usually document `nullable` as `title?: string` rather than `title: string | null`

PR #1002, opened 2020-10-13
feat(payment): INT-3061 renamed mandate field to mandateUrl on order interface

BC-AxelMalagon:
## What? [INT-3061](https://jira.bigcommerce.com/browse/INT-3061)
Renamed mandate field to mandateUrl on order interface

## Why?
In order to match the persistence and communicate the intended usage.

## Sibling PR
[Checkout-JS #435](https://github.com/bigcommerce/checkout-js/pull/435)

## Testing / Proof

### DIFF
--- a/src/order/order.ts
+++ b/src/order/order.ts
@@ -31,7 +31,7 @@ export default interface Order {
     status: string;
     taxes: Tax[];
     taxTotal: number;
-    mandate?: string;
+    mandateUrl?: string;
 }
 
 export type OrderPayments = Array<GatewayOrderPayment | GiftCertificateOrderPayment>;


vmparra (approved):
LGTM 👍🏻

src/order/order.ts
```diff
@@ -31,7 +31,7 @@ export default interface Order {
     status: string;
     taxes: Tax[];
     taxTotal: number;
-    mandate?: string;
+    mandateUrl?: string;
```
capsula4, line 34: where is this new field being consumed?

capsula4 (approved):

src/order/order.ts
BC-AxelMalagon: This field is coming from bigpay.

PR #1003, opened 2020-10-15
fix(payment): PAYPAL-766 Validate cc before submit order

helga1507:
## What?
Validate credit cards before submit order

## Why?
Validation should be performed right after a buyer clicked place order (without allowing call for POST internalapi/v1/checkout/order)

## Testing / Proof
https://www.loom.com/share/2154e46c3f584fd3aaa7b8e27ba6690e

### DIFF
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
@@ -133,7 +133,7 @@ export default class PaypalCommercePaymentProcessor {
         return this._hostedFields.submit(options);
     }
 
-    getValidationStateHostedFields(): { isValid: boolean; fields: PaypalCommerceHostedFieldsState['fields'] } {
+    getHostedFieldsValidationState(): { isValid: boolean; fields: PaypalCommerceHostedFieldsState['fields'] } {
         if (!this._hostedFields) {
             throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
         }
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
@@ -124,16 +124,6 @@ export default class PaypalCommercePaymentProcessor {
         if (!this._hostedFields) {
             throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
         }
-
-        const { fields } = this._hostedFields.getState();
-
-        const formValid = (Object.keys(fields) as Array<keyof PaypalCommerceHostedFieldsState['fields']>)
-            .every(key => fields[key]?.isValid);
-
-        if (!formValid) {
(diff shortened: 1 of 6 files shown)


src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -143,6 +133,19 @@ export default class PaypalCommercePaymentProcessor {
         return this._hostedFields.submit(options);
     }
 
+    validateHostedForm(): { isValid: boolean; fields: PaypalCommerceHostedFieldsState['fields'] } {
```
davidchin, line 136: Correct me if I'm wrong, but it looks like this method doesn't actually validate the form. It only gets the result of the validation result that's done previously (i.e.: when a user type into the form) or the initial state. Does PayPal offer a method to actually do the validation? If not, it'd be good to rename to the method, i.e.: `getValidationState` so it's clear to other people reading the code that it's a getter rather than an action.

Another thing to check is, what happens when a user clicks on the submit button without making any changes to the form? Are the initial values of `isValid` false?

davidchin, line 136: `isValid` parameter is also an unused parameter.

helga1507: Yes, you are right. I will rename method. Thank you!

PayPal doesn't offer a method to actually do the validation. You can see example on the screenshot.
And  the initial values are `isValid=false` when a user clicks on the submit button without making any changes

Screenshot 2020-10-16 at 10 47 04

helga1507: Hm, I use it [here](https://github.com/bigcommerce/checkout-sdk-js/pull/1003/files#diff-50215dc105323ccd925b612e6d34f5cea9d7908cf618b1fcd386faf238fc701aR68)

davidchin: Ah sorry I misread, it's a return value rather than a parameter. 👍

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -143,6 +133,19 @@ export default class PaypalCommercePaymentProcessor {
         return this._hostedFields.submit(options);
     }
 
+    getValidationStateHostedFields(): { isValid: boolean; fields: PaypalCommerceHostedFieldsState['fields'] } {
```
davidchin, line 136: `getHostedFieldsValidationState` might read better.

davidchin (approved):
A small suggestion, otherwise LGTM 👍

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
helga1507: Yeah, thank you!

PR #1005, opened 2020-10-20
feat(payment): PAYPAL-702 Add alternative payment methods

helga1507:
## What?
Add alternative payment methods

## Why?
New implementation

## Testing / Proof
https://www.loom.com/share/f334356c99064d899bf3a7cddba999cb
https://www.loom.com/share/623788b3b9a64034a02b0f184d970e34


src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
```diff
@@ -189,3 +199,9 @@ export interface PaypalCommerceScriptParams  {
     intent?: 'capture' | 'authorize';
     components?: ComponentsScriptType;
 }
+
+export enum PaymentStrategyTypes {
+    credit = 'credit',
```
davidchin, line 204: Capitalise enum keys, i.e.: `Credit = 'credit'`

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -120,4 +122,29 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+                    return 'GIROPAY';
+                case 'przelewy24':
+                    return 'P24';
+                case 'eps':
+                    return 'EPS';
+                case 'ideal':
+                    return 'IDEAL';
+                case 'mybank':
+                    return 'MYBANK';
+                case 'sofort':
```
davidchin, line 143: I think you're missing a `default` case. What happens if none matches? At the moment, it falls back to 'PAYPAL' but is it what's intended?

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
```diff
@@ -189,3 +199,9 @@ export interface PaypalCommerceScriptParams  {
     intent?: 'capture' | 'authorize';
     components?: ComponentsScriptType;
 }
+
+export enum PaymentStrategyTypes {
```
davidchin, line 203: Might be better to namespace it to avoid confusion, i.e.: `PaypalCommercePaymentStrategyType`.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -120,4 +122,29 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
             intent,
         };
     }
+
+    private _getFindingKey(methodId: string): keyof PaypalCommerceSDKFunding {
```
davidchin, line 126: A suggestion. Instead of injecting an enum type into the class and check against it to determine the `fundingKey`, perhaps it's better to inject an object that can directly return the `fundingKey` value based on `methodId/gatewayId`? i.e.: You can have something like 

```ts
class PaypalCommerceFundingKeyResolver {
    resolve(methodId, gatewayId) {
        if (methodId === 'paypalcommerce') {
            return 'PAYPAL';
        }

        if (methodId === 'paypalcommercecredit') {
            return 'PAYLATER';
        }

        if (methodId === 'bancontact' && gatewayId === 'paypalcommercealternativemethods') {
            return 'BANCONTACT';
        }

        // Etc...

        throw new Error('Unable to resolve funding key');
    }
}
```

And then in `PaypalCommercePaymentStrategy`, you can just ask for the value by calling `PaypalCommerceFundingKeyResolver#resolve`? The benefit is that you don't need to introduce another/secondary "strategy type" to provide branching behaviours. Also it's easier to test as you can test the resolver on its own. Thoughts?

helga1507: Thank you @davidchin !
I added function `paypalCommerceFundingKeyResolver` for this

src/payment/strategies/paypal-commerce/paypal-commerce-funding-key-resolver.ts
```diff
@@ -0,0 +1,32 @@
+export default function paypalCommerceFundingKeyResolver(methodId: string, gatewayId?: string) {
```
davidchin, line 1: Can you make this a class and inject it into `PaypalCommercePaymentStrategy` so you can test it in isolation?

src/payment/strategies/paypal-commerce/paypal-commerce-funding-key-resolver.ts
```diff
@@ -0,0 +1,34 @@
+export default class PaypalCommerceFundingKeyResolver {
+    resolve(methodId: string, gatewayId?: string) {
```
davidchin, line 2: Make the return value explicit? i.e.: `resolve(methodId: string, gatewayId?: string): string`

davidchin (approved):
Nice work, LGTM! 👍

src/payment/strategies/paypal-commerce/paypal-commerce-funding-key-resolver.ts
helga1507: Okay

src/payment/strategies/paypal-commerce/paypal-commerce-funding-key-resolver.ts
helga1507: Oh, sure. Thank you!

PR #1012, opened 2020-11-11
feat(payments): PAYPAL-839 Move method_id in execute

helga1507:
## What?
Move method_id in execute

## Why?
During the implementation of APMs, we implemented it they it reflects the name of the provider for each order PayPal commerce.
However, in order to calculate GMV per each particular APM we have to reflect each APM name in the order in the way like - payment method - PayPal (Sofort).

We need to pass `method_id` in request to bigpay for purchase

## Testing / Proof
Manual, unit


davidchin (approved):
Is this a fix? Also, it'd be good to fill out the "why" section of the PR.

helga1507:
@davidchin No, it does not a fix
I added "why" to this PR and will add it in the next PRs

PR #1015, opened 2020-11-17
feat(payment): PAYPAL-787 TypeError: Cannot read property 'orderId' of null

andriiVitvitskyi1990:
## What?
TypeError: Cannot read property 'orderId' of null

## Why?
According to task PAYPAL-787

## Testing / Proof
tested on dev

### DIFF
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
@@ -24,7 +24,7 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
     async initialize({ gatewayId, methodId, paypalcommerce }: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
         const { paymentMethods: { getPaymentMethodOrThrow }, cart: { getCartOrThrow } } = this._store.getState();
         const { initializationData } = getPaymentMethodOrThrow(methodId, gatewayId);
-        const { orderId, buttonStyle } = initializationData;
+        const { orderId, buttonStyle } = initializationData ?? {};
 
         if (orderId) {
             this._orderId = orderId;


src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -26,7 +26,7 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
         const { initializationData } = getPaymentMethodOrThrow(methodId, gatewayId);
         const { orderId, buttonStyle } = initializationData;
```
BC-ACherednichenko, line 27: Will we have the same issue with `buttonStyle`?

andriiVitvitskyi1990: No, if button style will be undefined, this style wont apply. That's it

BC-ACherednichenko (approved):

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -26,7 +26,7 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
 
-        if (orderId) {
+        if (initializationData && orderId) {
```
davidchin, line 29: Have you tested this? It doesn't seem right because `initializationData` is already destructed in the line above. So if  `initializationData` is `null`, it'll break in the line above rather than this one. I think you should do something like this instead:

```ts
const { orderId, buttonStyle } = initializationData ?? {};
```

andriiVitvitskyi1990: fixed

davidchin (approved):

PR #1016, opened 2020-11-17
feat(payment): PAYPAL-848 Add messages to script

helga1507:
## What?
Add `messages` to `components` in PayPal script always in the PaypalCommerceButtonStrategy

## Why?
It needs for credit banners. PayPal recommends us to use SDK for banners in the way it is implemented on the cart page

Related task https://github.com/bigcommerce/content/pull/615

## Testing / Proof
https://www.loom.com/share/252298feb82047f388deb0f55a9a95d6

### DIFF
--- a/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
+++ b/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
@@ -3,7 +3,7 @@ import { FormPoster } from '@bigcommerce/form-poster';
 import { Cart } from '../../../cart';
 import { CheckoutActionCreator, CheckoutStore } from '../../../checkout';
 import { InvalidArgumentError, MissingDataError, MissingDataErrorType } from '../../../common/error/errors';
-import { ApproveDataOptions, ButtonsOptions, ClickDataOptions, ComponentsScriptType, DisableFundingType, PaypalCommerceInitializationData, PaypalCommercePaymentProcessor, PaypalCommerceScriptParams } from '../../../payment/strategies/paypal-commerce';
+import { ApproveDataOptions, ButtonsOptions, ClickDataOptions, DisableFundingType, PaypalCommerceInitializationData, PaypalCommercePaymentProcessor, PaypalCommerceScriptParams } from '../../../payment/strategies/paypal-commerce';
 import { CheckoutButtonInitializeOptions } from '../../checkout-button-options';
 import CheckoutButtonStrategy from '../checkout-button-strategy';
 
@@ -39,7 +39,7 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
         const messagingContainer = options.paypalCommerce?.messagingContainer;
         const isMessagesAvailable = Boolean(messagingContainer && document.getElementById(messagingContainer));
 
-        await this._paypalCommercePaymentProcessor.initialize(this._getParamsScript(initializationData, cart, isMessagesAvailable));
+        await this._paypalCommercePaymentProcessor.initialize(this._getParamsScript(initializationData, cart));
 
         this._paypalCommercePaymentProcessor.renderButtons(cart.id, `#${options.containerId}`, buttonParams);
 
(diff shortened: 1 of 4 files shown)


BC-ACherednichenko (approved):
LGTM

src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -73,25 +73,20 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
...
-            components.push('messages');
-        }
-
         return {
             'client-id': clientId,
             'merchant-id': merchantId,
             commit: false,
             currency: cart.currency.code,
-            components,
+            components: ['buttons', 'messages'],
```
davidchin, line 89: What happens if `messagingContainer` is not passed in? At the moment, it seems you can actually not pass in `messagingContainer` (which is why you still have `isMessagesAvailable` check).

helga1507: Nothing happens. 
Values for `components` needs for PayPal script. After load script we call `payPal.Buttons.render` for rendering smart payment buttons and call `payPal.Messages.render` for rendering credit banners. 
In this case we add always `messages` to `components` to avoid conflict with script [here](https://github.com/bigcommerce/content/pull/615)

davidchin: Ok cool, I guess what you mean is unless `payPal.Messages.render` is called, nothing happens, even if `components` contains `messages`.

davidchin (approved):

src/payment/strategies/paypal-commerce/paypal-commerce-script-loader.ts
```diff
@@ -17,12 +17,14 @@ export default class PaypalCommerceScriptLoader {
     async loadPaypalCommerce(params: PaypalCommerceScriptParams, isProgressiveOnboardingAvailable?: boolean): Promise<PaypalCommerceSDK> {
         this._validateParams(params, isProgressiveOnboardingAvailable);
 
-        const scriptSrc = 'https://unpkg.com/@paypal/paypal-js@1.0.2/dist/paypal.browser.min.js';
+        if (!this._window.paypalLoadScript) {
```
BC-ACherednichenko, line 20: @helga1507 so in the case when banners loaded the SDK we do not load SDK here?

BC-ACherednichenko: @helga1507 can you take a look?

helga1507: Yes. We don't need do load this script again

BC-ACherednichenko: I see there is no currency in the banner script, I believe if we don't load this script again there will be an issue because the currency is the mandatory param for button rendering (the possible method that will be shown on the frontend depends on the currency of on the checkout)

helga1507: We load `unpkg` only for function `paypalLoadScript`. All parameter for script we add and call here `await this._window.paypalLoadScript(params);` ([line 30](https://github.com/bigcommerce/checkout-sdk-js/blob/5c75f24d892c1234852780151cd8a7c492cab0c0/src/payment/strategies/paypal-commerce/paypal-commerce-script-loader.ts#L30))

BC-ACherednichenko: @helga1507 can you break it down for me?https://github.com/bigcommerce/content/pull/615/files#diff-544247c5f4bda100deda99d1efa9c7626bf30376509651c75a1c46c47e507f23R24 we do not pass currency here, hence,  if the banner loaded before the SPBs, SPBs will consume the initialized SDK without currency that has been loaded before in the banner, is it correct?

BC-ACherednichenko: Per our discussion, such an approach makes sense for the new version of the `unpkg `

helga1507:
@davidchin Yes, you are right
Thank you!

PR #1019, opened 2020-11-19
feat(payment): INT-3237 Prevent Cardinal downgrade to 3DSv1

mauricio-sg:
## What? [INT-3237](https://jira.bigcommerce.com/browse/INT-3237)
Allow reconfiguration of the Cardinal SDK and avoid throwing an error if the consumer cancels the challenge.

## Why?
For cardinal flow v2 we need to be able to run the configuration step each time a new `ReferenceId` is generated (this happens each time the user is challenged). Otherwise, browser data won't be collected for the new `ReferenceId` and auth will downgrade to 3DS v1

## Testing / Proof
Manual, Unit / [Demo Video](https://drive.google.com/file/d/1sRMIcIOp9-cJcSElMic-EjtSktQanL_a/view?usp=sharing)

### DIFF
--- a/src/payment/strategies/cardinal/cardinal-client.ts
+++ b/src/payment/strategies/cardinal/cardinal-client.ts
@@ -3,7 +3,7 @@ import { includes } from 'lodash';
 import { Address } from '../../../address';
 import { BillingAddress } from '../../../billing';
 import { MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType } from '../../../common/error/errors';
-import { PaymentMethodCancelledError, PaymentMethodFailedError } from '../../errors';
+import { PaymentMethodFailedError } from '../../errors';
 import { CreditCardInstrument, ThreeDSecureToken, VaultedInstrument } from '../../payment';
 import { ThreeDsResult } from '../../payment-response-body';
 
@@ -22,14 +22,19 @@ export interface CardinalOrderData {
 }
 
 export default class CardinalClient {
+    private _provider: string = '';
+    private _testMode: boolean = false;
     private _sdk?: Promise<CardinalSDK>;
-    private _isConfigured: boolean = false;
+    private _configurationToken: string = '';
 
     constructor(
         private _scriptLoader: CardinalScriptLoader
     ) {}
 
-    load(provider: string, testMode?: boolean): Promise<void> {
+    load(provider: string, testMode = false): Promise<void> {
+        this._provider = provider;
+        this._testMode = testMode;
+
         if (!this._sdk) {
             this._sdk = this._scriptLoader.load(provider, testMode);
         }
@@ -38,8 +43,12 @@ export default class CardinalClient {
     }
 
     configure(clientToken: string): Promise<void> {
-        if (this._isConfigured) {
-            return Promise.resolve();
+        if (!!this._configurationToken) {
+            if (this._configurationToken === clientToken) {
+                return Promise.resolve();
+            } else {
+                this._sdk = this._scriptLoader.load(`${this._provider}.${Date.now()}`, this._testMode);
+            }
         }
 
         return this._getClientSDK()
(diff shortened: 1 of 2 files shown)


josesonora (approved):

FranciscoOnt (approved):
LGTM

src/payment/strategies/cardinal/cardinal-client.ts
```diff
@@ -37,6 +37,19 @@ export default class CardinalClient {
         return this._sdk.then(() => {});
     }
 
+    isConfigured(): boolean {
+        return this._isConfigured;
+    }
+
+    async reconfigure(provider: string, clientToken: string, testMode?: boolean): Promise<void> {
+        this._sdk = undefined;
```
capsula4, line 45: I'm wondering if it might be a better idea to re-use the `configure` method and instead of storing `isConfigured: bool`, store `configurationToken: string`

When trying to call configure() again with a different token to the one being configured, it can re-configure. This would simplifiy the logic for clients and avoid leaking internal logic in `cardinal-three-d-secure-flow-v2.ts `

src/payment/strategies/cardinal/cardinal-three-d-secure-flow-v2.ts
```diff
@@ -112,4 +112,14 @@ export default class CardinalThreeDSecureFlowV2 {
...
     }
+
+    private async _configure(methodId: string, token: string): Promise<void> {
+        if (this._cardinalClient.isConfigured()) {
+            const { config: { testMode } } = this._store.getState().paymentMethods.getPaymentMethodOrThrow(methodId);
+
+            return this._cardinalClient.reconfigure(methodId, token, testMode);
+        }
+
+        return this._cardinalClient.configure(token);
```
capsula4, line 123: as per the suggestions above, could this simply be `this._cardinalClient.configure(token);` ? I'm assuming `methodId` and `testMode` don't really need to be re-configured? maybe those things can be stored internally in the cardinalClient?

src/payment/strategies/cardinal/cardinal-client.ts
mauricio-sg: Sounds good! 👍

mauricio-sg:
Hey @capsula4, thanks for your feedback. Could you re-review again, please? ♻️ 🙇

src/payment/strategies/cardinal/cardinal-client.ts
```diff
@@ -22,24 +22,33 @@ export interface CardinalOrderData {
...
             this._sdk = this._scriptLoader.load(provider, testMode);
         }
 
         return this._sdk.then(() => {});
     }
 
-    configure(clientToken: string): Promise<void> {
-        if (this._isConfigured) {
-            return Promise.resolve();
+    async configure(clientToken: string): Promise<void> {
```
capsula4, line 45: don't think you need `Promise` as return type if it's async

src/payment/strategies/cardinal/cardinal-client.ts
```diff
@@ -22,24 +22,33 @@ export interface CardinalOrderData {
...
+        if (!!this._configurationToken) {
```
capsula4, line 46: wouldn't `this._configurationToken` always be null? I think you are missing  `this._configurationToken = clientToken;` somewhere.

Maybe something like this?
```js
if (this._configurationToken && this._configurationToken === clientToken) {
  return Promise.resolve();
}

this._configurationToken = clientToken;
this._sdk = this._scriptLoader.load(`${this._provider}.${Date.now()}`, 
```

src/payment/strategies/cardinal/cardinal-client.ts
mauricio-sg: Actually, it's not async. I was trying to do something and forgot to delete it. 🤦 Thank you 👍

src/payment/strategies/cardinal/cardinal-client.ts
mauricio-sg: Yes but, I just wanted to make sure to load it when a reconfiguration took place 😬. Otherwise, we may end up with duplicated/wasted scripts the first time `configure` runs because it's supposed to be already loaded by the `prepare` method of` CardinalThreeDSecureFlow[v2]`.

capsula4 (approved):

PR #1021, opened 2020-11-24
CHECKOUT-4941 Expose customer account fields

capsula4:
## What?
Expose customer account fields

## Why?
So they can be used to render customer creation

## Testing / Proof
API is not returning the right form fields names.
https://launchbay.bigcommerce.net/deployments/193438

Screen Shot 2020-11-24 at 4 55 44 pm

@bigcommerce/checkout

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -1,4 +1,4 @@
-import { FormField } from '../form';
+import { FormFields } from '../form';
 
 export default interface Config {
     context: ContextConfig;
@@ -12,7 +12,19 @@ export interface StoreConfig {
     currency: StoreCurrency;
     displayDateFormat: string;
     inputDateFormat: string;
+
+    /**
+     * @deprecated Please use instead the data selectors
+     * @remarks
+     * ```js
+     * const data = CheckoutService.getState().data;
+     * const shippingAddressFields = data.getShippingAddressFields('US');
+     * const billingAddressFields = data.getBillingAddressFields('US');
+     * const customerAccountFields = data.getCustomerAccountFields();
+     * ```
+     */
     formFields: FormFields;
+
     links: StoreLinks;
     paymentSettings: PaymentSettings;
     shopperConfig: ShopperConfig;
@@ -66,11 +78,6 @@ export interface StoreLinks {
     orderConfirmationLink: string;
 }
 
-export interface FormFields {
-    shippingAddressFields: FormField[];
-    billingAddressFields: FormField[];
-}
-
 export interface StoreCurrency {
     code: string;
     decimalPlaces: string;
@@ -91,6 +98,7 @@ export interface CheckoutSettings {
     isAnalyticsEnabled: boolean;
     isCardVaultingEnabled: boolean;
     isCouponCodeCollapsed: boolean;
+    isSignInEmailEnabled: boolean;
     isPaymentRequestEnabled: boolean;
     isPaymentRequestCanMakePaymentEnabled: boolean;
     isSpamProtectionEnabled: boolean;
--- a/src/form/form-fields-action-creator.ts
+++ b/src/form/form-fields-action-creator.ts
@@ -1,0 +1,30 @@
+import { createAction } from '@bigcommerce/data-store';
+import { concat, defer, of, Observable } from 'rxjs';
+import { catchError } from 'rxjs/operators';
+
+import { cachableAction, ActionOptions } from '../common/data-store';
+import { throwErrorAction } from '../common/error';
+import { RequestOptions } from '../common/http-request';
+
+import { FormFieldsActionType, LoadFormFieldsAction } from './form-fields-actions';
+import FormFieldsRequestSender from './form-fields-request-sender';
+
+export default class FormFieldsActionCreator {
+    constructor(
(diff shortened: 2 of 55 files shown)


src/form/form-fields-action-creator.ts
```diff
@@ -0,0 +1,30 @@
...
+        private _formFieldsRequestSender: FormFieldsRequestSender
+    ) {}
+
+    @cachableAction
+    loadFormFields(options?: RequestOptions & ActionOptions): Observable<LoadFormFieldsAction> {
+        return Observable.create((observer: Observer<LoadFormFieldsAction>) => {
+            observer.next(createAction(FormFieldsActionType.LoadFormFieldsRequested));
+
+            this._formFieldsRequestSender.loadFields(options)
+                .then(response => {
```
davidchin, line 21: Prefer `async/await` for new code

src/form/form-selector.ts
```diff
@@ -2,39 +2,59 @@ import { memoizeOne } from '@bigcommerce/memoize';
...
+            const fields = formFields ? formFields.billingAddress : [];
 
             return fields.map((field: any) => processField(field, countries, selectedCountry));
         }
     );
 
+    const getCustomerAccountFields = createSelector(
+        (state: FormFieldsState) => state.data,
+        formFields => () => {
+            return formFields ? formFields.customerAccount : [];
```
davidchin, line 44: You can omit `return` if there's no `{}`? 🍹

src/config/config.ts
```diff
@@ -12,7 +10,6 @@ export interface StoreConfig {
     currency: StoreCurrency;
     displayDateFormat: string;
     inputDateFormat: string;
-    formFields: FormFields;
```
davidchin, line 15: Removing `formFields` from the interface would be a breaking change though? It's okay to remove the field from the API because the config endpoint is a private endpoint. But from SDK perspective, `Config` interface is not a private interface so we can't just remove it. 

In order to maintains backwards compatibility, perhaps we can use the data from the new form-field endpoint and just map it to the existing `Config` interface?

capsula4: That's a good point. However, I doubt clients would be consuming these fields directly rather than using the selector which takes care of populating states and provinces. Might be a bit late, but might be worth tagging the data returned in the store config as "beta" or "internal only"? thoughts?

src/form/form-fields-action-creator.ts
capsula4: For some reason async/await doesn't seem to work in this scenario, not sure why but this is what I got:
```js
    loadFormFields(options?: RequestOptions & ActionOptions): Observable<LoadFormFieldsAction> {
        return Observable.create(async (observer: Observer<LoadFormFieldsAction>) => {
            observer.next(createAction(FormFieldsActionType.LoadFormFieldsRequested));

            try {
                const response = await this._formFieldsRequestSender.loadFields(options);

                observer.next(createAction(FormFieldsActionType.LoadFormFieldsSucceeded, response.body));
                observer.complete();
            } catch (error) {
                observer.error(createErrorAction(FormFieldsActionType.LoadFormFieldsFailed, error));
            }
        });
    }
```

With that code, checkout-action-creator.spec.ts fails because `LoadFormFieldsSucceeded` action is not present.

davidchin: I can't immediately tell what's wrong with the code at a glance. Perhaps use [SubscriptionsActionCreator](https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/subscription/subscriptions-action-creator.ts) as an example for reference?

src/config/config.ts
davidchin: Maybe, but we can only guess. It's like making an API change hoping nobody's relying on the change 😄. I supposed it's a different story if we can conditionally exclude the change from people who depend on it. Anyway, I think it should be fairly simple change and I can see you've already started working on it.

In regards to adding `beta` and `internal only` to `Config`, I don't think we should? Our Open Checkout depends on it. And it does surface data that are necessary to build a checkout (otherwise we won't be using it). We can, however, add a `deprecated` annotation to the `formFields` field and suggest people to use `getShippingAddressFields`, `getBillingAddressFields` etc... instead.

capsula4: I have added `@deprecated` tag for now, since it cannot really be made "internal only" if it's been part of the public API.

capsula4:
@davidchin  this should be ready for another review, thanks! :bow:

src/config/config.ts
```diff
@@ -12,7 +12,19 @@ export interface StoreConfig {
...
+     * @deprecated Please use instead the data selectors
+     * @remarks
+     * ```js
+     * const data = CheckoutService.getState().data;
+     * const shippingAddressFields = data.getShippingAddressFields('US');
+     * const billingAddressFields = data.getBillingAddressFields('US');
+     * const customerAccountFields = data.getCustomerAccountFields();
+     * ```
+     */
+    formFields?: FormFields;
```
davidchin, line 26: It should still `formFields` (without `?`)

capsula4: fixed

davidchin (approved):
Cool, this LGTM, nice work! 👍

PR #1017, opened 2020-11-18
feat(payments): PAYPAL-847 Add PPCP to README

helga1507:
## What?
Add PPCP to README

## Why?
For the checkout-sdk readme file, we need to mention that Paypal Commerce has specific options for the initialization (as it is mentioned for Amazon)

## Testing / Proof
Screenshot 2020-11-18 at 12 38 30

@bigcommerce/checkout @bigcommerce/payments

PR #1022, opened 2020-11-24
feat(payments): PAYPAL-868 Checkout-sdk PayPal initialization issue

andriiVitvitskyi1990:
## What?
Checkout-sdk PayPal initialization issue

## Why?
According to task PAYPAL-868

## Testing / Proof
Screenshot 2020-11-24 at 14 11 01

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,7 +329,6 @@ service.initializePayment({
     methodId: 'paypalcommerce',
     paypalcommerce: {
         container: '#container',
-// Callback for submitting payment form that gets called when a buyer approves PayPal payment
         submitForm: () => {
             service.submitOrder(
                 {
@@ -337,7 +336,6 @@ service.initializePayment({
                 }
             );
         },
-// Callback is used to define the state of the payment form, validate if it is applicable for submit.
         onValidate: (resolve, reject) => {
             const isValid = service.validatePaymentForm();
             if (isValid) {
@@ -345,7 +343,6 @@ service.initializePayment({
             }
             return reject();
         },
-// Callback that is called right before render of a Smart Payment Button. It gets called when a buyer is eligible for use of the particular PayPal method. This callback can be used to hide the standard submit button.
         onRenderButton: () => {
             service.hidePaymentSubmitButton();
         }
--- a/README.md
+++ b/README.md
@@ -318,6 +318,39 @@ await service.initializePayment({
         container: 'walletWidget',
     },
 });
+
+```
+Also, PayPal (also known as PayPal Commerce Platform) requires specific options to initialize the PayPal Smart Payment Button on checkout page that substitutes a standard submit button
+
+````
+<!-- This is where the PayPal button will be inserted -->
+<div id="container"></div>
+service.initializePayment({
+    methodId: 'paypalcommerce',
+    paypalcommerce: {
+        container: '#container',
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -318,6 +318,36 @@ await service.initializePayment({
...
+        },
+        onValidate: (resolve, reject) => {
+            const isValid = service.validatePaymentForm();
+            if (isValid) {
+                return resolve();
+            }
+            return reject();
+        },
+        onRenderButton: () => {
+            service.hidePaymentSubmitButton();
```
BC-ACherednichenko, line 347: Can you also add comments for `onRenderButton` and `onValidate`, and `submitForm` methods https://github.com/bigcommerce/checkout-sdk-js/pull/1017#discussion_r529058647
like 
```js
        //  A callback that is called right before render of a Smart Payment Button. It gets called when a buyer is eligible for use of the particular PayPal method. This callback can be used to hide the standard submit button.
        onRenderButton: () => {
        ...
        // this callback is used to define the state of the payment form, validate if it is applicable for submit
        onValidate: (resolve, reject) => {
        ...
        // A callback for submitting payment form that gets called when buyer approved PayPal payment
        submitForm: () => {
        ...
```

dist/checkout-button.d.ts
```diff
@@ -1,496 +0,0 @@
-import { Timeout } from '@bigcommerce/request-sender';
-import { createTimeout } from '@bigcommerce/request-sender';
```
BC-ACherednichenko, line 2: I believe all these changes are unintentional, can you revert these back?

BC-ACherednichenko (changes requested):

BC-ACherednichenko (approved):
LGTM

README.md
davidchin: Provider-specific instructions should be surfaced in the `docs` folder (e.g: https://github.com/bigcommerce/checkout-sdk-js/blob/master/docs/interfaces/paypalcommercepaymentinitializeoptions.md) rather than the top-level readme markdown file.

docs/interfaces/paypalcommercepaymentinitializeoptions.md
```diff
@@ -5,6 +5,7 @@
 A set of options that are required to initialize the PayPal Commerce payment
 method for presenting its PayPal button.
 
+Please note that the minimum version of checkout-sdk is 1.100
```
davidchin, line 8: This file is auto-generated. You shouldn't edit it directly as it'd get overridden when we regenerate the doc. Instead, you should edit this file instead - https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/payment/strategies/paypal-commerce/paypal-commerce-payment-initialize-options.ts#L8. We extract the comment and re-generate this markdown file every time we cut a new release.

davidchin (changes requested):

README.md
BC-ACherednichenko: @davidchin  do you suggest to move the entire section to the options?

davidchin (approved):

TomA-R:
in the future could we amend the original commit, or make fix commit messages meaningful? thanks!

PascalZajac:
This is especially important given this repository is public.

PR #1026, opened 2020-12-01
CHECKOUT-4571 Remove alpha tags from stable functionality

capsula4:
## What?
Remove alpha tags from stable functionality

## Why?
Functionality has been in use for a long period and proven to be stable.

## Testing / Proof
CI

@bigcommerce/checkout


ankurbigcomm (approved):
👍

PR #1038, opened 2020-12-10
feat(payments): PAYPAL-876 implement polling mechanism on front-end

andriiVitvitskyi1990:
## What?
Implement polling mechanism on front-end

## Why?
When we create order we can have drop off scenario, and we need to implement polling mechanism that sends request and ask for status and proceed with succeed flow if order status 'approved'

## Testing / Proof
Tested on Dev

### DIFF
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
@@ -33,7 +33,8 @@ export default class PaypalCommercePaymentProcessor {
 
     constructor(
         private _paypalScriptLoader: PaypalCommerceScriptLoader,
-        private _paypalCommerceRequestSender: PaypalCommerceRequestSender
+        private _paypalCommerceRequestSender: PaypalCommerceRequestSender,
+        private _orderId?: string
     ) {}
 
     async initialize(paramsScript: PaypalCommerceScriptParams, isProgressiveOnboardingAvailable?: boolean): Promise<PaypalCommerceSDK> {
@@ -79,6 +80,10 @@ export default class PaypalCommercePaymentProcessor {
         this._paypalButtons.render(container);
 
         return this._paypalButtons;
+    }
+
+    getOrderId() {
+        return this._orderId;
     }
 
     renderMessages(cartTotal: number, container: string): PaypalCommerceMessages {
@@ -145,7 +150,6 @@ export default class PaypalCommercePaymentProcessor {
 
     deinitialize() {
         this._paypalButtons?.close?.();
-
         this._paypal = undefined;
         this._paypalButtons = undefined;
         this._fundingSource = undefined;
@@ -155,6 +159,7 @@ export default class PaypalCommercePaymentProcessor {
     private async _setupPayment(cartId: string, params: ParamsForProvider = {}): Promise<string> {
         const paramsForProvider = { ...params, isCredit: this._fundingSource === 'credit' || this._fundingSource === 'paylater' };
         const { orderId } = await this._paypalCommerceRequestSender.setupPayment(cartId, paramsForProvider);
+        this._orderId = orderId;
 
         return orderId;
     }
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
@@ -30,11 +30,11 @@ export default class PaypalCommercePaymentProcessor {
     private _paypalMessages?: PaypalCommerceMessages;
     private _hostedFields?: PaypalCommerceHostedFields;
     private _fundingSource?: string;
+    private _orderId?: string;
 
     constructor(
         private _paypalScriptLoader: PaypalCommerceScriptLoader,
-        private _paypalCommerceRequestSender: PaypalCommerceRequestSender,
-        private _orderId?: string
+        private _paypalCommerceRequestSender: PaypalCommerceRequestSender
     ) {}
 
     async initialize(paramsScript: PaypalCommerceScriptParams, isProgressiveOnboardingAvailable?: boolean): Promise<PaypalCommerceSDK> {
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
@@ -19,7 +19,8 @@ import { ApproveDataOptions,
     PaypalCommerceRequestSender,
     PaypalCommerceScriptParams } from './index';
 
-const APPROVED = 'approved';
+const ORDER_STATUS_APPROVED = 'approved';
+const POLLING_INTERVAL = 3000;
 
 export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
     private _orderId?: string;
@@ -61,11 +62,11 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
             style: buttonStyle,
             onApprove: data => this._tokenizePayment(data, submitForm),
             onClick: async (_, actions) => {
-                this.setPollingMechanism(gatewayId, submitForm);
+                this._initializePollingMechanism(submitForm, gatewayId);
 
-                return  onValidate(actions.resolve, actions.reject);
+                return onValidate(actions.resolve, actions.reject);
             },
-            onCancel: () => clearInterval(this._pollingInterval),
+            onCancel: () => this._deinitializePollingMechanism(),
         };
 
         await this._paypalCommercePaymentProcessor.initialize(paramsScript);
@@ -111,27 +112,34 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
     }
 
     async deinitialize(): Promise<InternalCheckoutSelectors> {
-        clearInterval(this._pollingInterval);
+        this._deinitializePollingMechanism();
         this._orderId = undefined;
         this._paypalCommercePaymentProcessor.deinitialize();
 
         return Promise.resolve(this._store.getState());
     }
 
-    private setPollingMechanism(gatewayId: string | undefined, submitForm: any) {
+    private _initializePollingMechanism(submitForm: () => void, gatewayId?: string) {
         this._pollingInterval = setInterval(async () =>  {
             try {
                 if (gatewayId === PaymentStrategyType.PAYPAL_COMMERCE_ALTERNATIVE_METHODS) {
                     const res = await this._paypalCommerceRequestSender.getOrderStatus();
-                    if (res.status.toLowerCase() === APPROVED) {
-                        clearInterval(this._pollingInterval);
+                    if (res.status.toLowerCase() === ORDER_STATUS_APPROVED) {
+                        this._deinitializePollingMechanism();
                         this._tokenizePayment({orderID: this._paypalCommercePaymentProcessor.getOrderId()}, submitForm);
                     }
                 }
             } catch (e) {
-                clearInterval(this._pollingInterval);
+                this._deinitializePollingMechanism();
             }
-        }, 3000);
+        }, POLLING_INTERVAL);
+    }
+
+    private _deinitializePollingMechanism() {
+        if (this._pollingInterval) {
+            clearInterval(this._pollingInterval);
+            this._pollingInterval = null;
+        }
     }
 
     private _isPaypalCommerceOptionsPayments(options: PaypalCommercePaymentInitializeOptions | PaypalCommerceCreditCardPaymentInitializeOptions): options is PaypalCommercePaymentInitializeOptions {
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
@@ -21,6 +21,7 @@ import { ApproveDataOptions,
 
 const ORDER_STATUS_APPROVED = 'approved';
 const POLLING_INTERVAL = 3000;
+const POLLING_MAX_TIME = 600000;
 
 export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
     private _orderId?: string;
@@ -32,7 +33,8 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
         private _paypalCommercePaymentProcessor: PaypalCommercePaymentProcessor,
         private _paypalCommerceFundingKeyResolver: PaypalCommerceFundingKeyResolver,
         private _paypalCommerceRequestSender: PaypalCommerceRequestSender,
-        private _pollingInterval?: any
+        private _pollingInterval?: number,
+        private _pollingTimer?: number,
     ) {}
 
     async initialize({ gatewayId, methodId, paypalcommerce }: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
@@ -66,7 +68,10 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
 
                 return onValidate(actions.resolve, actions.reject);
             },
-            onCancel: () => this._deinitializePollingMechanism(),
+            onCancel: () => {
+                this._deinitializePollingTimer(this._pollingInterval);
+                this._deinitializePollingTimer(this._pollingTimer);
+                },
         };
 
         await this._paypalCommercePaymentProcessor.initialize(paramsScript);
@@ -112,7 +117,8 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
     }
 
     async deinitialize(): Promise<InternalCheckoutSelectors> {
-        this._deinitializePollingMechanism();
+        this._deinitializePollingTimer(this._pollingTimer);
+        this._deinitializePollingTimer(this._pollingInterval);
         this._orderId = undefined;
         this._paypalCommercePaymentProcessor.deinitialize();
 
@@ -120,25 +126,40 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
     }
 
     private _initializePollingMechanism(submitForm: () => void, gatewayId?: string) {
-        this._pollingInterval = setInterval(async () =>  {
+       this._initializePollingTimer();
+       this._pollingInterval = window.setTimeout(async () =>  {
             try {
                 if (gatewayId === PaymentStrategyType.PAYPAL_COMMERCE_ALTERNATIVE_METHODS) {
+                    console.log('%c POLLING TIMEOUT', 'color: green');
                     const res = await this._paypalCommerceRequestSender.getOrderStatus();
                     if (res.status.toLowerCase() === ORDER_STATUS_APPROVED) {
-                        this._deinitializePollingMechanism();
+                        this._deinitializePollingTimer(this._pollingTimer);
+                        this._deinitializePollingTimer(this._pollingInterval);
                         this._tokenizePayment({orderID: this._paypalCommercePaymentProcessor.getOrderId()}, submitForm);
+                    } else {
+                        this._initializePollingMechanism(submitForm, gatewayId)
                     }
                 }
             } catch (e) {
-                this._deinitializePollingMechanism();
+                this._deinitializePollingTimer(this._pollingInterval);
+                this._deinitializePollingTimer(this._pollingTimer);
             }
         }, POLLING_INTERVAL);
     }
 
-    private _deinitializePollingMechanism() {
-        if (this._pollingInterval) {
-            clearInterval(this._pollingInterval);
-            this._pollingInterval = null;
+    private _deinitializePollingTimer(timer?: number) {
+        if (timer) {
+            clearTimeout(timer);
+            timer = 0;
+        }
+    }
+
+    private _initializePollingTimer(): void {
+        if(!this._pollingTimer) {
+            this._pollingTimer = window.setTimeout(() => {
+                this._deinitializePollingTimer(this._pollingInterval);
+                this._deinitializePollingTimer(this._pollingTimer);
+            }, POLLING_MAX_TIME);
         }
     }
 
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
@@ -1,6 +1,6 @@
 import { Cart } from '../../../cart';
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
-import { InvalidArgumentError } from '../../../common/error/errors';
+import { InvalidArgumentError, TimeoutError } from '../../../common/error/errors';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
 import { PaymentArgumentInvalidError, PaymentMethodInvalidError } from '../../errors';
@@ -19,7 +19,8 @@ import { ApproveDataOptions,
     PaypalCommerceRequestSender,
     PaypalCommerceScriptParams } from './index';
 
-const ORDER_STATUS_APPROVED = 'approved';
+const ORDER_STATUS_APPROVED = 'APPROVED';
+const ORDER_STATUS_CREATED = 'CREATED';
 const POLLING_INTERVAL = 3000;
 const POLLING_MAX_TIME = 600000;
 
@@ -34,7 +35,7 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
         private _paypalCommerceFundingKeyResolver: PaypalCommerceFundingKeyResolver,
         private _paypalCommerceRequestSender: PaypalCommerceRequestSender,
         private _pollingInterval?: number,
-        private _pollingTimer?: number,
+        private _pollingTimer = 0
     ) {}
 
     async initialize({ gatewayId, methodId, paypalcommerce }: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
@@ -62,16 +63,18 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
         const paramsScript = this._getOptionsScript(initializationData, currencyCode);
         const buttonParams: ButtonsOptions = {
             style: buttonStyle,
-            onApprove: data => this._tokenizePayment(data, submitForm),
+            onApprove: data => {
+                this._deinitializePollingTimer(gatewayId);
+                this._tokenizePayment(data, submitForm);
+            },
             onClick: async (_, actions) => {
-                this._initializePollingMechanism(submitForm, gatewayId);
+                this._initializePollingMechanism(submitForm, gatewayId, methodId, paypalcommerce);
 
                 return onValidate(actions.resolve, actions.reject);
             },
             onCancel: () => {
-                this._deinitializePollingTimer(this._pollingInterval);
-                this._deinitializePollingTimer(this._pollingTimer);
-                },
+                this._deinitializePollingTimer(gatewayId);
+            },
         };
 
         await this._paypalCommercePaymentProcessor.initialize(paramsScript);
@@ -116,50 +119,50 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
         return Promise.reject(new OrderFinalizationNotRequiredError());
     }
 
-    async deinitialize(): Promise<InternalCheckoutSelectors> {
-        this._deinitializePollingTimer(this._pollingTimer);
-        this._deinitializePollingTimer(this._pollingInterval);
+    async deinitialize({gatewayId}: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        this._deinitializePollingTimer(gatewayId);
         this._orderId = undefined;
         this._paypalCommercePaymentProcessor.deinitialize();
 
         return Promise.resolve(this._store.getState());
     }
 
-    private _initializePollingMechanism(submitForm: () => void, gatewayId?: string) {
-       this._initializePollingTimer();
-       this._pollingInterval = window.setTimeout(async () =>  {
+    private _initializePollingMechanism(submitForm: () => void, gatewayId?: string, methodId?: any, paypalcommerce?: any ) {
+        if (gatewayId !== PaymentStrategyType.PAYPAL_COMMERCE_ALTERNATIVE_METHODS)  {
+            return;
+        }
+
+        this._pollingInterval = window.setTimeout(async () => {
             try {
-                if (gatewayId === PaymentStrategyType.PAYPAL_COMMERCE_ALTERNATIVE_METHODS) {
-                    console.log('%c POLLING TIMEOUT', 'color: green');
-                    const res = await this._paypalCommerceRequestSender.getOrderStatus();
-                    if (res.status.toLowerCase() === ORDER_STATUS_APPROVED) {
-                        this._deinitializePollingTimer(this._pollingTimer);
-                        this._deinitializePollingTimer(this._pollingInterval);
-                        this._tokenizePayment({orderID: this._paypalCommercePaymentProcessor.getOrderId()}, submitForm);
-                    } else {
-                        this._initializePollingMechanism(submitForm, gatewayId)
-                    }
+                this._pollingTimer += POLLING_INTERVAL;
+
+                const { status } = await this._paypalCommerceRequestSender.getOrderStatus();
+
+                if (status === ORDER_STATUS_APPROVED) {
+                    this._deinitializePollingTimer(gatewayId);
+                    this._tokenizePayment({ orderID: this._paypalCommercePaymentProcessor.getOrderId() }, submitForm);
+                } else if (status === ORDER_STATUS_CREATED  && this._pollingTimer < POLLING_MAX_TIME) {
+                    this._initializePollingMechanism(submitForm, gatewayId, methodId, paypalcommerce);
+                } else {
+                    this._reinitializeButtons({ gatewayId, methodId, paypalcommerce });
+                    throw new TimeoutError();
                 }
             } catch (e) {
-                this._deinitializePollingTimer(this._pollingInterval);
-                this._deinitializePollingTimer(this._pollingTimer);
+                this._deinitializePollingTimer(gatewayId);
+                paypalcommerce.onError?.(e);
             }
         }, POLLING_INTERVAL);
     }
 
-    private _deinitializePollingTimer(timer?: number) {
-        if (timer) {
-            clearTimeout(timer);
-            timer = 0;
-        }
+    private _reinitializeButtons({ gatewayId, methodId, paypalcommerce }: PaymentInitializeOptions) {
+        this.deinitialize({methodId, gatewayId});
+        this.initialize({ gatewayId, methodId, paypalcommerce });
     }
 
-    private _initializePollingTimer(): void {
-        if(!this._pollingTimer) {
-            this._pollingTimer = window.setTimeout(() => {
-                this._deinitializePollingTimer(this._pollingInterval);
-                this._deinitializePollingTimer(this._pollingTimer);
-            }, POLLING_MAX_TIME);
+    private _deinitializePollingTimer(gatewayId?: string) {
+        if (gatewayId === PaymentStrategyType.PAYPAL_COMMERCE_ALTERNATIVE_METHODS) {
+            clearTimeout(this._pollingInterval);
+            this._pollingTimer = 0;
         }
     }
 
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
@@ -4,11 +4,22 @@ import { InvalidArgumentError } from '../../../common/error/errors';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
 import { PaymentArgumentInvalidError, PaymentMethodInvalidError } from '../../errors';
+import { PaymentStrategyType } from '../../index';
 import PaymentActionCreator from '../../payment-action-creator';
 import { PaymentInitializeOptions, PaymentRequestOptions } from '../../payment-request-options';
 import PaymentStrategy from '../payment-strategy';
 
-import { ApproveDataOptions, ButtonsOptions, PaypalCommerceCreditCardPaymentInitializeOptions, PaypalCommerceFundingKeyResolver, PaypalCommerceInitializationData, PaypalCommercePaymentInitializeOptions, PaypalCommercePaymentProcessor, PaypalCommerceScriptParams } from './index';
+import { ApproveDataOptions,
+    ButtonsOptions,
+    PaypalCommerceCreditCardPaymentInitializeOptions,
+    PaypalCommerceFundingKeyResolver,
+    PaypalCommerceInitializationData,
+    PaypalCommercePaymentInitializeOptions,
+    PaypalCommercePaymentProcessor,
+    PaypalCommerceRequestSender,
+    PaypalCommerceScriptParams } from './index';
+
+const APPROVED = 'approved';
 
 export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
     private _orderId?: string;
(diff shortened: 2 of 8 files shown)


src/payment/strategies/paypal-commerce/paypal-commerce-request-sender.ts
```diff
@@ -38,4 +38,16 @@ export default class PaypalCommerceRequestSender {
 
         return res.body as OrderData;
     }
+
+    async askForOrderStatus(): Promise<OrderStatus> {
```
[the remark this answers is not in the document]
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
```diff
@@ -180,6 +185,8 @@ export interface PaypalCommerceInitializationData {
     clientId: string;
     merchantId?: string;
     intent?: 'capture' | 'authorize';
+    isDeveloperModeApplicable?: boolean;
```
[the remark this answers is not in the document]
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -89,16 +104,34 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
 
         return Promise.resolve(this._store.getState());
     }
 
+    private setPollingMechanism(gatewayId: string | undefined, submitForm: any) {
+        this._pollingInterval = setInterval(async () =>  {
+            try {
+                if (gatewayId === 'paypalcommercealternativemethods') {
+                    const res = await this._paypalCommerceRequestSender.askForOrderStatus();
+                    if (res.status.toLowerCase() === 'approved') {
```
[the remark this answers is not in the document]
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -89,16 +104,34 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
         this._orderId = undefined;
         this._paypalCommercePaymentProcessor.deinitialize();
 
         return Promise.resolve(this._store.getState());
     }
 
+    private setPollingMechanism(gatewayId: string | undefined, submitForm: any) {
+        this._pollingInterval = setInterval(async () =>  {
+            try {
+                if (gatewayId === 'paypalcommercealternativemethods') {
```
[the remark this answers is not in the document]
andriiVitvitskyi1990: fixed

src/payment/payment-strategy-type.ts
```diff
@@ -19,6 +19,7 @@ enum PaymentStrategyType {
     LEGACY = 'legacy',
     OFFLINE = 'offline',
     OFFSITE = 'offsite',
+    APPROVED = 'approved',
```
BC-ACherednichenko, line 22: There are no statuses in this type, I think `APPROVED = 'approved',` status should be in the paypalcommerce strategy

andriiVitvitskyi1990: fixed

BC-ACherednichenko (approved):
Nicely done!

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
```diff
@@ -1,6 +1,6 @@
 
 export interface ApproveDataOptions {
-    orderID: string;
+    orderID: string | undefined;
```
capsula4, line 3: we tend to use `orderID?: string` rather than `string | undefined`.

src/payment/strategies/paypal-commerce/paypal-commerce-request-sender.ts
```diff
@@ -38,4 +38,16 @@ export default class PaypalCommerceRequestSender {
...
+    async getOrderStatus(): Promise<OrderStatus> {
+        const url = `/api/storefront/initialization/paypalcommerce`;
+        const headers = {
+            'X-API-INTERNAL': INTERNAL_USE_ONLY,
+            'Content-Type': ContentType.Json,
+        };
+
+        const res = await this._requestSender.get(url, {headers});
+
+        return res.body as OrderStatus;
```
capsula4, line 51: instead of doing `as OrderStatus`, you should be able to set the type on line 49
```
const res = await this._requestSender.get< OrderStatus>(url, {headers});
```

src/payment/strategies/paypal-commerce/paypal-commerce-request-sender.ts
```diff
@@ -38,4 +38,16 @@ export default class PaypalCommerceRequestSender {
 
         return res.body as OrderData;
     }
+
+    async getOrderStatus(): Promise<OrderStatus> {
+        const url = `/api/storefront/initialization/paypalcommerce`;
```
capsula4, line 43: there's no variable in this string, you should be able to use simple quotes `'`

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +111,29 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
 
     async deinitialize(): Promise<InternalCheckoutSelectors> {
+        clearInterval(this._pollingInterval);
         this._orderId = undefined;
         this._paypalCommercePaymentProcessor.deinitialize();
 
         return Promise.resolve(this._store.getState());
     }
 
+    private setPollingMechanism(gatewayId: string | undefined, submitForm: any) {
```
capsula4, line 121: instead of `string | undefined` it's preferred to use `gatewayId?: string`
Also, can we have a type for `submitForm`? Using `any` is a bit like cheating.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +111,29 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+        this._pollingInterval = setInterval(async () =>  {
+            try {
+                if (gatewayId === PaymentStrategyType.PAYPAL_COMMERCE_ALTERNATIVE_METHODS) {
+                    const res = await this._paypalCommerceRequestSender.getOrderStatus();
+                    if (res.status.toLowerCase() === APPROVED) {
```
capsula4, line 126: what happens if the order is never approved? How long will this keep polling for? Is there any chance that a process got stuck in a "pre-approved" status? We have other payment methods where we redirect to the order confirmation page despite the order not being approved but "placed". We display a message to the shopper notifying that the order is still pending manual verification from the merchant. Maybe this is something you would want to explore?

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +111,29 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+                        clearInterval(this._pollingInterval);
+                        this._tokenizePayment({orderID: this._paypalCommercePaymentProcessor.getOrderId()}, submitForm);
+                    }
+                }
+            } catch (e) {
+                clearInterval(this._pollingInterval);
+            }
+        }, 3000);
```
capsula4, line 134: maybe this can be a const at the top of the file, like `APPROVED`

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +111,29 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
     }
 
     async deinitialize(): Promise<InternalCheckoutSelectors> {
+        clearInterval(this._pollingInterval);
```
capsula4, line 114: you have `clearInterval(this._pollingInterval);` in quite a few places.

Maybe you can have a private method `_deinitiazePollingMechanism()` ?
```ts
private _deinitiazePollingMechanism() {
  if (this._pollingInterval) {
    clearInterval(this._pollingInterval);
    this._pollingInterval = null;
  }
}
```

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -47,7 +60,12 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
         const buttonParams: ButtonsOptions = {
             style: buttonStyle,
             onApprove: data => this._tokenizePayment(data, submitForm),
-            onClick: async (_, actions) => onValidate(actions.resolve, actions.reject),
+            onClick: async (_, actions) => {
+                this.setPollingMechanism(gatewayId, submitForm);
+
+                return  onValidate(actions.resolve, actions.reject);
```
capsula4, line 66: extra space after `return `

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -4,11 +4,22 @@ import { InvalidArgumentError } from '../../../common/error/errors';
...
+    ButtonsOptions,
+    PaypalCommerceCreditCardPaymentInitializeOptions,
+    PaypalCommerceFundingKeyResolver,
+    PaypalCommerceInitializationData,
+    PaypalCommercePaymentInitializeOptions,
+    PaypalCommercePaymentProcessor,
+    PaypalCommerceRequestSender,
+    PaypalCommerceScriptParams } from './index';
+
+const APPROVED = 'approved';
```
capsula4, line 22: maybe the const can be more self-descriptive? is this `const ORDER_STATUS_APPROVED = 'approved'`

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-request-sender.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-request-sender.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -18,7 +30,9 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
         private _orderActionCreator: OrderActionCreator,
         private _paymentActionCreator: PaymentActionCreator,
         private _paypalCommercePaymentProcessor: PaypalCommercePaymentProcessor,
-        private _paypalCommerceFundingKeyResolver: PaypalCommerceFundingKeyResolver
+        private _paypalCommerceFundingKeyResolver: PaypalCommerceFundingKeyResolver,
+        private _paypalCommerceRequestSender: PaypalCommerceRequestSender,
+        private _pollingInterval?: any
```
capsula4, line 35: sorry I missed this in the first review, can you use `number` instead of `any` ?

capsula4 (commented):
Code in general looks good, thanks for addressing the feedback! The only outstanding concern is the one around the polling mechanism that could possibly never end. If the order never gets approved for whatever reason, the user might get stuck seeing a loading spinner forever. Might be worth setting a max waiting time? Maybe 5-10 seconds? Might need guidance from product/UX team for advice on how to handle this scenario with the final user.

BC-ACherednichenko:
@capsula4 thanks for the suggestion! Yeah, we are adding the polling time limitation, however, 5-10 seconds will not be enough, as in this timespan a buyer needs to:
1. Interact with PayPal popup
2. redirect to a 3rd-party APM in this popup 
3. fill out all of the details required by an APM (pass all fo the steps)
4. and then there is a small-time span for a payment session update (when PayPal updates the status of a payment session which status we try to define with a poller according to the result of APM actions)

Hence, I would say that we need not less than 10 minutes (it is an assumption), in the case when a buyer has some complexities there will not be enough time.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -47,7 +61,12 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
         const buttonParams: ButtonsOptions = {
             style: buttonStyle,
             onApprove: data => this._tokenizePayment(data, submitForm),
-            onClick: async (_, actions) => onValidate(actions.resolve, actions.reject),
+            onClick: async (_, actions) => {
+                this._initializePollingMechanism(submitForm, gatewayId);
+
+                return onValidate(actions.resolve, actions.reject);
+            },
+            onCancel: () => this._deinitializePollingMechanism(),
```
davidchin, line 69: Do you need to cancel the polling if `onApprove` is called?

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -47,7 +61,12 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
         const buttonParams: ButtonsOptions = {
             style: buttonStyle,
             onApprove: data => this._tokenizePayment(data, submitForm),
-            onClick: async (_, actions) => onValidate(actions.resolve, actions.reject),
+            onClick: async (_, actions) => {
+                this._initializePollingMechanism(submitForm, gatewayId);
```
davidchin, line 65: I assume polling is done because it's the only way to handle this situation? Also why do we still want to proceed with payment if the user explicitly cancels the transaction?

BC-ACherednichenko: yeah, this topic is partially covered in the [thread](https://threads.com/34389654791). If a buyer cancels a transaction, then the `onCancel` is triggered and we stop the initiated session (it is similar to the regular PayPal flow).

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
BC-ACherednichenko: Good point, thanks

andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +112,36 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
     }
 
+    private _initializePollingMechanism(submitForm: () => void, gatewayId?: string) {
+        this._pollingInterval = setInterval(async () =>  {
+            try {
+                if (gatewayId === PaymentStrategyType.PAYPAL_COMMERCE_ALTERNATIVE_METHODS) {
+                    const res = await this._paypalCommerceRequestSender.getOrderStatus();
+                    if (res.status.toLowerCase() === ORDER_STATUS_APPROVED) {
+                        this._deinitializePollingMechanism();
+                        this._tokenizePayment({orderID: this._paypalCommercePaymentProcessor.getOrderId()}, submitForm);
```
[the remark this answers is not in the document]
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
davidchin: I see, it's not "cancelling", but rather, it's about confirmation status being returned asynchronously due to the way APM works.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +121,51 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+            timer = 0;
+        }
+    }
+
+    private _initializePollingTimer(): void {
+        if (!this._pollingTimer) {
+            this._pollingTimer = window.setTimeout(() => {
+                this._deinitializePollingTimer(this._pollingInterval);
+                this._deinitializePollingTimer(this._pollingTimer);
+            }, POLLING_MAX_TIME);
```
davidchin, line 165: Is there a reason why you need another `setTimeout`? Is it because you want to stop polling after a minute? Wouldn't the polling be automatically cancelled if the order status becomes approved (or not)? What would happen if the polling is stopped before the approval? Would the user keep seeing the loading spinner?

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +121,51 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
         return Promise.resolve(this._store.getState());
     }
 
+    private _initializePollingMechanism(submitForm: () => void, gatewayId?: string) {
+        this._initializePollingTimer();
+        this._pollingInterval = window.setTimeout(async () =>  {
+            try {
+                if (gatewayId === PaymentStrategyType.PAYPAL_COMMERCE_ALTERNATIVE_METHODS) {
+                    const res = await this._paypalCommerceRequestSender.getOrderStatus();
+                    if (res.status.toLowerCase() === ORDER_STATUS_APPROVED) {
```
davidchin, line 138: Would you get a status other than pending? Like rejected? Would you keep retrying?

BC-ACherednichenko: The expected behaviour is that we have `CREATED` status (when the payment session created) and `APPROVED` status (when the buyer confirmed this session). 
However, it is an extremely legit point for the cases when something goes not by the flow we expect! It makes absolute sense to have the second check, let's say if the status `CREATED` and not approved `APPROVED` -> we have to continue on our poller. 
If the status `CREATED` and not approved `APPROVED` -> we have to continue on our poller.
If the status is not `CREATED` and not approved `APPROVED` -> we have to reject the session because it is not something we can handle.

Thanks for the suggestion!

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +119,58 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+    private _initializePollingMechanism(submitForm: () => void, gatewayId?: string, timer?: number) {
+        if (!timer) {
+            timer = 0;
+        }
+        this._pollingInterval = window.setTimeout(async () =>  {
+            try {
+                if (timer !== undefined) {
+                    timer += POLLING_INTERVAL;
+                }
+                if (gatewayId === PaymentStrategyType.PAYPAL_COMMERCE_ALTERNATIVE_METHODS) {
```
BC-ACherednichenko, line 138: I think it is worthy to move this `if` before `window.setTimeout` ( as this logic should not be started if use `PAYPAL_COMMERCE_ALTERNATIVE_METHODS`) or even in the places where it could be triggered
like `onClick`

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +119,58 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+    }
+
+    private _pollingTimerExpired() {
+        throw new TimeoutError();
+    }
+
+    private _deinitializePollingTimer(timer?: number) {
+        if (timer) {
+            clearTimeout(timer);
+        }
```
BC-ACherednichenko, line 171:
```suggestion
    private _deinitializePollingTimer() {
         clearTimeout(this._pollingInterval);
```

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +119,58 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
     async deinitialize(): Promise<InternalCheckoutSelectors> {
+        this._deinitializePollingTimer(this._pollingInterval);
         this._orderId = undefined;
         this._paypalCommercePaymentProcessor.deinitialize();
 
         return Promise.resolve(this._store.getState());
     }
 
+    private _initializePollingMechanism(submitForm: () => void, gatewayId?: string, timer?: number) {
+        if (!timer) {
```
BC-ACherednichenko, line 130: I believe it would be better to keep `timer` counter in the class property and initialize it with the default value (0). Along with the `_deinitializePollingTimer ` we can reset this value back to 0 if needed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -46,8 +62,18 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
         const paramsScript = this._getOptionsScript(initializationData, currencyCode);
         const buttonParams: ButtonsOptions = {
             style: buttonStyle,
-            onApprove: data => this._tokenizePayment(data, submitForm),
-            onClick: async (_, actions) => onValidate(actions.resolve, actions.reject),
+            onApprove: data => {
+                this._deinitializePollingTimer(this._pollingInterval);
```
BC-ACherednichenko, line 66: Not sure if causes any issue, but it is worthy to do that only for alternative method

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -4,11 +4,25 @@ import { InvalidArgumentError } from '../../../common/error/errors';
...
+    PaypalCommerceCreditCardPaymentInitializeOptions,
+    PaypalCommerceFundingKeyResolver,
+    PaypalCommerceInitializationData,
+    PaypalCommercePaymentInitializeOptions,
+    PaypalCommercePaymentProcessor,
+    PaypalCommerceRequestSender,
+    PaypalCommerceScriptParams } from './index';
+
+const ORDER_STATUS_APPROVED = 'approved';
+const ORDER_STATUS_CREATED = 'created';
```
golcinho, line 23: I see that `OrderStatus` interface has a `string` type for the `status` property, what if we make the `status` type more specific? I think it will be more clear and helps with understandability 

Current interface:
```
export interface OrderStatus {
    status: string;
}
```

New example:
```
export interface OrderStatus {
    status: 'approved' | 'created'; // don't foget to add all possible types here
}
```

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -18,7 +32,10 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
         private _orderActionCreator: OrderActionCreator,
         private _paymentActionCreator: PaymentActionCreator,
         private _paypalCommercePaymentProcessor: PaypalCommercePaymentProcessor,
-        private _paypalCommerceFundingKeyResolver: PaypalCommerceFundingKeyResolver
+        private _paypalCommerceFundingKeyResolver: PaypalCommerceFundingKeyResolver,
+        private _paypalCommerceRequestSender: PaypalCommerceRequestSender,
+        private _pollingInterval?: number,
+        private _pollingTimer: number = 0
```
golcinho, line 38: 🍹 ts should handle correct type according to default value
```suggestion
        private _pollingTimer = 0
```

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +122,48 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
         this._paypalCommercePaymentProcessor.deinitialize();
 
         return Promise.resolve(this._store.getState());
     }
 
+    private _initializePollingMechanism(submitForm: () => void, gatewayId?: string) {
+        if (gatewayId === PaymentStrategyType.PAYPAL_COMMERCE_ALTERNATIVE_METHODS) {
+            this._pollingInterval = window.setTimeout(async () => {
+                try {
+                    if (this._pollingTimer !== undefined) {
```
golcinho, line 136: This check is redundant because `_pollingTimer` will always be a `number` (`private _pollingTimer: number = 0`). 

Also, could you please check all same conditions?

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +122,48 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+                        this._pollingTimer += POLLING_INTERVAL;
+                    }
+                    const res = await this._paypalCommerceRequestSender.getOrderStatus();
+                    if (res.status.toLowerCase() !== ORDER_STATUS_CREATED && res.status.toLowerCase() !== ORDER_STATUS_APPROVED) {
```
golcinho, line 140: Looks like we can avoid using the `toLowerCase` method if we change the `status` constants according to actual status values.

for example: `'approved' => `APPROVED`` (suppose that status is uppercased)

andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-request-sender.ts
```diff
@@ -38,4 +38,16 @@ export default class PaypalCommerceRequestSender {
 
         return res.body as OrderData;
     }
+
+    async getOrderStatus(): Promise<OrderStatus> {
```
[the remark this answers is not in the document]
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +122,46 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+                    if (res.status === ORDER_STATUS_APPROVED) {
+                        this._deinitializePollingTimer();
+                        this._tokenizePayment({ orderID: this._paypalCommercePaymentProcessor.getOrderId() }, submitForm);
+                    } else {
+                        if (this._pollingTimer < POLLING_MAX_TIME) {
+                            this._initializePollingMechanism(submitForm, gatewayId);
+                        } else {
+                            this._deinitializePollingTimer();
+
+                            return;
```
davidchin, line 150: You don't need a `return` as there's no code after this line that can be executed in this code block.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +122,46 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+                        }
+                    }
+                } catch (e) {
+                    this._deinitializePollingTimer();
```
davidchin, line 154: I can see that the polling stops after 10min. But it also stops if there's any error within 10min. (i.e.: network error). Is this intentional? Do you think it's a good idea to retry for a few times instead? Also, do you want log the error somehow instead of swallowing it entirely? i.e.: You can add an `onError` callback so that the client app (UCO) can hook into it and report the error in Sentry?

BC-ACherednichenko: Thanks for the question, at the moment we are checking the way to stop the payment session entirely and show an error. 
Ideally, at the end of the timer, we want to close the PayPal popup and overlay and throw an error, it is possible to achieve via `paypal.close()` of the PayPal SDK method. However, in addition to the PayPal popup shut down, the iframe buttons will be deinitialized as well, hence we need to re-initialize them as well. We decided to move further with the testing of what we have now to make sure the poller works as expected and continue on the improvement with graceful finalization of the "timer finalized" scenario.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -33,7 +33,8 @@ export default class PaypalCommercePaymentProcessor {
 
     constructor(
         private _paypalScriptLoader: PaypalCommerceScriptLoader,
-        private _paypalCommerceRequestSender: PaypalCommerceRequestSender
+        private _paypalCommerceRequestSender: PaypalCommerceRequestSender,
+        private _orderId?: string
```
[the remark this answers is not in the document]
andriiVitvitskyi1990: removed from constructor

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
andriiVitvitskyi1990: removed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +122,46 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+        if (gatewayId === PaymentStrategyType.PAYPAL_COMMERCE_ALTERNATIVE_METHODS) {
+            this._pollingInterval = window.setTimeout(async () => {
+                try {
+                    this._pollingTimer += POLLING_INTERVAL;
+                    const res = await this._paypalCommerceRequestSender.getOrderStatus();
+                    if (res.status !== ORDER_STATUS_CREATED && res.status !== ORDER_STATUS_APPROVED) {
+                        this._deinitializePollingTimer();
+                    }
+                    if (res.status === ORDER_STATUS_APPROVED) {
+                        this._deinitializePollingTimer();
```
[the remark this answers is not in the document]
andriiVitvitskyi1990: We need do deinitialize timer in both cases.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -93,12 +122,46 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
     async deinitialize(): Promise<InternalCheckoutSelectors> {
+        this._deinitializePollingTimer();
         this._orderId = undefined;
         this._paypalCommercePaymentProcessor.deinitialize();
 
         return Promise.resolve(this._store.getState());
     }
 
+    private _initializePollingMechanism(submitForm: () => void, gatewayId?: string) {
+        if (gatewayId === PaymentStrategyType.PAYPAL_COMMERCE_ALTERNATIVE_METHODS) {
```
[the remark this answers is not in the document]
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -46,8 +63,20 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+                }
+                this._tokenizePayment(data, submitForm);
+            },
+            onClick: async (_, actions) => {
+                this._initializePollingMechanism(submitForm, gatewayId);
+
+                return onValidate(actions.resolve, actions.reject);
+            },
+            onCancel: () => {
+                this._deinitializePollingTimer();
```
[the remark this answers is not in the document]
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -92,13 +119,58 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+        this.deinitialize({methodId, gatewayId});
+        this.initialize({ gatewayId, methodId, paypalcommerce });
+    }
+
+    private _deinitializePollingTimer(gatewayId?: string) {
+        if (gatewayId !== PaymentStrategyType.PAYPAL_COMMERCE_ALTERNATIVE_METHODS) {
+            return;
+        }
+        clearTimeout(this._pollingInterval);
+        this._pollingTimer = 0;
```
golcinho, line 171: I don't really like inverse logic 😄 

```suggestion
        if (gatewayId === PaymentStrategyType.PAYPAL_COMMERCE_ALTERNATIVE_METHODS) {
            clearTimeout(this._pollingInterval);
            this._pollingTimer = 0;
        }
```

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
```diff
@@ -17,6 +17,10 @@ export interface OrderData {
     approveUrl: string;
 }
 
+export interface OrderStatus {
+    status: 'APPROVED' | 'CREATED';
```
golcinho, line 21: Looks like it could be then 2 statuses...

```suggestion
    status: 'APPROVED' | 'CREATED' | string;
```

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -92,13 +119,58 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+                if (res.status === ORDER_STATUS_APPROVED) {
+                    this._deinitializePollingTimer(gatewayId);
+                    this._tokenizePayment({ orderID: this._paypalCommercePaymentProcessor.getOrderId() }, submitForm);
+                } else {
+                    if (this._pollingTimer < POLLING_MAX_TIME) {
+                        this._initializePollingMechanism(submitForm, gatewayId, methodId, paypalcommerce);
+                    } else {
+                        this._reinitializeButtons({ gatewayId, methodId, paypalcommerce });
+                    }
+                }
```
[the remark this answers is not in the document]
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
andriiVitvitskyi1990: fixed

golcinho (approved):
Thank you for addressing all my comments! The current code looks good to me! Approved from my side!

Sorry but, I'm sure that you need to get approval from guys who familiar with the business logic across `checkout-sdk-js` before merge.

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -92,13 +119,54 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+            } catch (e) {
+                this._deinitializePollingTimer(gatewayId);
+            }
+        }, POLLING_INTERVAL);
+    }
+
+    private _reinitializeButtons({ gatewayId, methodId, paypalcommerce }: PaymentInitializeOptions) {
+        this._deinitializePollingTimer(gatewayId);
+        // @ts-ignore
```
davidchin, line 157: What compiler rule are you trying to ignore? Is it because `getButtons` can return `null`? And you're suppressing that null exception warning?

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -92,13 +119,54 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+        this._paypalCommercePaymentProcessor.getButtons().close();
```
davidchin, line 158: Why do you have to call `getButtons().close()` again, when `deinitialize` already calls `paypalButtons?.close?.()`?

andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
andriiVitvitskyi1990: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -92,13 +119,53 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+                    this._tokenizePayment({ orderID: this._paypalCommercePaymentProcessor.getOrderId() }, submitForm);
+                } else if (status === ORDER_STATUS_CREATED  && this._pollingTimer < POLLING_MAX_TIME) {
+                    this._initializePollingMechanism(submitForm, gatewayId, methodId, paypalcommerce);
+                } else {
+                    this._reinitializeButtons({ gatewayId, methodId, paypalcommerce });
+                    throw new TimeoutError();
+                }
+            } catch (e) {
+                this._deinitializePollingTimer(gatewayId);
+                paypalcommerce.onError(e);
```
BC-ACherednichenko, line 152: Can you add the `onError` method to the options interface https://github.com/bigcommerce/checkout-sdk-js/blob/6af83917b16e3257c19ac727368126162be0aec3/src/payment/strategies/paypal-commerce/paypal-commerce-payment-initialize-options.ts#L48
it should be optional param (we use it only for alternative methods, the regular paypal is not affected)

BC-ACherednichenko: `paypalcommerce.onError(e)` -> paypalcommerce.onError?.(e)

andriiVitvitskyi1990: fixed

BC-ACherednichenko (approved):
Nicely done! Thanks for addressing all comments!

davidchin (approved):

PR #1043, opened 2020-12-14
fix(provider): PAYPAL-767 Fix device data collection for Braintree stored credit cards

andrykonchin:
## What?

Fixed the issue with collection device data and sending it to Braintree when buyer uses credit cards vaulting.

Changes:
- don't initialize Braintree Hosted Fields SDK when there are no fields on the credit card form

## Why?

When buyer uses stored credit cards device data isn't sent to Braintree and transactions are rejected by anti-fraud mechanisms.

Screen Shot 2020-12-11 at 8 34 48 PM

The issue is that Hosted Fields Braintree JS SDK requires non empty `fields` option to initialize Hosted Fields ([documentation](https://braintree.github.io/braintree-web/current/module-braintree-web_hosted-fields.html#.create), [source](https://github.com/braintree/braintree-web/blob/3.59.0/src/hosted-fields/external/hosted-fields.js#L418)) and to declare credit card fields which are precent on the form. But there are some cases when there is no any field - e.g. when stored credit card chosen - and checkout-sdk-js passes empty `fields`.

It leads to exception raising and interrupting initialization process:

```json
{
  "code": "INSTANTIATION_OPTION_REQUIRED",
  "message": "options.fields is required when instantiating Hosted Fields.",
  "name": "BraintreeError",
  "type": "MERCHANT",
}
```

As a consequence the device data collector isn't initialized and device data aren't sent to Braintree.

## Testing / Proof

Device data is sending correctly to the BigPay backend

Screen Shot 2020-12-11 at 3 48 22 PM

Device data successfully stored on the Braintree side.

Screen Shot 2020-12-11 at 2 28 43 PM

### DIFF
--- a/src/payment/strategies/braintree/braintree-hosted-form.ts
+++ b/src/payment/strategies/braintree/braintree-hosted-form.ts
@@ -1,4 +1,4 @@
-import { isNil, omitBy, Dictionary } from 'lodash';
+import { isEmpty, isNil, omitBy, Dictionary } from 'lodash';
 
 import { Address } from '../../../address';
 import { NotInitializedError, NotInitializedErrorType } from '../../../common/error/errors';
@@ -26,15 +26,21 @@ export default class BraintreeHostedForm {
         private _braintreeSDKCreator: BraintreeSDKCreator
     ) {}
 
-    async initialize(options: BraintreeFormOptions): Promise<void> {
+    async initialize(options: BraintreeFormOptions): Promise<boolean> {
         this._formOptions = options;
 
         this._type = isBraintreeFormFieldsMap(options.fields) ?
             BraintreeHostedFormType.CreditCard :
             BraintreeHostedFormType.StoredCardVerification;
 
+        const fields = this._mapFieldOptions(options.fields);
+
+        if (isEmpty(fields)) {
+            return Promise.resolve(false);
+        }
+
         this._cardFields = await this._braintreeSDKCreator.createHostedFields({
-            fields: this._mapFieldOptions(options.fields),
+            fields,
             styles: options.styles && this._mapStyleOptions(options.styles),
         });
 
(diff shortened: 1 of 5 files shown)


src/payment/strategies/braintree/braintree-hosted-form.ts
```diff
@@ -26,15 +26,21 @@ export default class BraintreeHostedForm {
         private _braintreeSDKCreator: BraintreeSDKCreator
     ) {}
 
-    async initialize(options: BraintreeFormOptions): Promise<void> {
+    async initialize(options: BraintreeFormOptions): Promise<boolean> {
```
BC-ACherednichenko, line 29: I think that the change of the interface is not beneficial here, as we are able to define whether we are able to initialize the form or not.

src/payment/strategies/braintree/braintree-hosted-form.ts
```diff
@@ -26,15 +26,21 @@ export default class BraintreeHostedForm {
...
         this._formOptions = options;
 
         this._type = isBraintreeFormFieldsMap(options.fields) ?
             BraintreeHostedFormType.CreditCard :
             BraintreeHostedFormType.StoredCardVerification;
 
+        const fields = this._mapFieldOptions(options.fields);
+
+        if (isEmpty(fields)) {
+            return Promise.resolve(false);
```
davidchin, line 39: Just return `return false` is fine, because the method is marked as `async`, it'd automatically return a `false` wrapped in a `Promise`

src/payment/strategies/braintree/braintree-hosted-form.ts
```diff
@@ -53,6 +59,8 @@ export default class BraintreeHostedForm {
             this._cardNameField.on('focus', this._handleNameFocus);
             this._cardNameField.attach();
         }
+
+        return Promise.resolve(true);
```
davidchin, line 63: Same here, `return true` is sufficient.

davidchin (approved):

andrykonchin:
@davidchin  Thank you for review :bow:

PR #1045, opened 2020-12-15
CHECKOUT-5418 Add support for recaptcha for customer account creation

capsula4:
## What?
Add support for recaptcha for customer account creation

## Why?
Because when Storefont recaptcha is enabled, it's required for customer account creation

## Testing / Proof
unit

@bigcommerce/checkout

### DIFF
--- a/src/checkout/checkout-service.ts
+++ b/src/checkout/checkout-service.ts
@@ -560,13 +560,12 @@ export default class CheckoutService {
      * of development. Therefore the API is unstable and not ready for public
      * consumption.
      *
-     * @internal
      * @param customerAccount - The customer account data.
      * @param options - Options for creating customer account.
      * @returns A promise that resolves to the current state.
      */
     createCustomerAccount(customerAccount: CustomerAccountRequestBody, options?: RequestOptions): Promise<CheckoutSelectors> {
-        const action = this._customerActionCreator.createAccount(customerAccount, options);
+        const action = this._customerActionCreator.createCustomer(customerAccount, options);
 
         return this._dispatch(action);
     }
@@ -1205,7 +1204,7 @@ export default class CheckoutService {
      * @returns A promise that resolves to the current state.
      */
     executeSpamCheck(): Promise<CheckoutSelectors> {
-        const action = this._spamProtectionActionCreator.execute();
+        const action = this._spamProtectionActionCreator.verifyCheckoutSpamProtection();
 
         return this._dispatch(action, { queueId: 'spamProtection' });
     }
--- a/src/spam-protection/spam-protection-action-creator.ts
+++ b/src/spam-protection/spam-protection-action-creator.ts
@@ -1,5 +1,5 @@
 import { createAction, ThunkAction } from '@bigcommerce/data-store';
-import { concat, defer, empty, of } from 'rxjs';
+import { concat, defer, from, of } from 'rxjs';
 import { catchError, switchMap, take } from 'rxjs/operators';
 
 import { InternalCheckoutSelectors } from '../checkout';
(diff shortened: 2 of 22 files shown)


src/spam-protection/google-recaptcha.ts
```diff
@@ -54,13 +63,25 @@ export default class GoogleRecaptcha {
...
+        onError: (error: Error) => void = noop
+    ): Promise<void> {
+        const recaptcha = await this.googleRecaptchaScriptLoader.load();
+
+        this._event$ = this._memoized(recaptcha, sitekey, document.getElementById(containerId), size);
+        this._recaptcha = recaptcha;
+
+        this._event$.subscribe(result => {
+            if (result && result.token) {
+                onSuccess(result.token);
```
davidchin, line 80: I think there's some logic in `execute` that handles the scenario where reCaptcha puzzle is dismissed by the shopper before it is solved. That scenario won't trigger the error callbacks that the reCaptcha client provides, so we need some custom code to handle it. Do we not need to handle that scenario anymore?

src/checkout/checkout-service.ts
```diff
@@ -1128,6 +1128,25 @@ export default class CheckoutService {
...
+     *
+     * ```js
+     * await service.executeStorefrontSpamProtection({ containerId: 'recatpcha' });
+     * ```
+     *
+     * @param options - Options for initializing spam protection.
+     * @returns A promise that resolves to the current state.
+     */
+    executeStorefrontSpamProtection(options: SpamProtectionOptions): Promise<CheckoutSelectors> {
+        const action = this._spamProtectionActionCreator.initializeStorefront(options);
```
davidchin, line 1145: I think if we go down the route of invisible reCaptcha, we probably don't need a separate method to initialise reCaptcha. The initialisation and verification could just be done as a part of `createCustomerAccount` call. You also don't need to store it in `CheckoutStore` because you can just pass the token along in the promise chain.

src/spam-protection/google-recaptcha.ts
capsula4: The regular (not invisible catpcha) works differently, the execute method doesn't need to be called.

src/spam-protection/spam-protection-action-creator.ts
```diff
@@ -48,17 +44,37 @@ export default class SpamProtectionActionCreator {
...
+            const checkout = state.checkout.getCheckout();
 
-            if (!shouldExecuteSpamCheck) {
-                return empty();
+            if (!checkout) {
+                throw new MissingDataError(MissingDataErrorType.MissingCheckout);
             }
 
+            return from(this.execute()(store))
+                .pipe(last())
```
davidchin, line 57: Similarly, `last()` means taking the last emitted action so it'd filter out all the other actions that are emitted. Also, I'd emit `Requested` and `Failed` actions, not just `Succeeded`, because it is an async action creator. I'd do something like this instead:

```ts
return concat(
    createAction(SpamProtectionActionType.CheckoutVerifyRequested),
    from(this.execute()(store))
        .pipe(switchMap(action =>
            isSpamProtectionExecuteSucceededAction(action) && action.payload?.token ?
                from(this._requestSender.validate(checkout.id, action.payload.token))
                    .pipe(switchMap(({ body }) => [
                        action,
                        createAction(SpamProtectionActionType.CheckoutVerifySucceeded, body),
                    ])) :
                of(action)
        ))
).pipe(
    catchError(error => throwErrorAction(SpamProtectionActionType.CheckoutVerifyFailed, error))
);
```

src/spam-protection/spam-protection-action-creator.ts
```diff
@@ -48,17 +44,37 @@ export default class SpamProtectionActionCreator {
         );
     }
 
-    execute(): ThunkAction<SpamProtectionAction, InternalCheckoutSelectors> {
+    executeCheckoutSpamProtection(): ThunkAction<SpamProtectionAction> {
```
davidchin, line 47: `verifyCheckoutSpamProtection` so that it mirrors the name of the action you're emitting?

capsula4:
Thanks @davidchin  for the thorough review, I have updated based on your suggestions

davidchin (approved):
All good, nice work! 👍

PR #1025, opened 2020-12-01
feat(customer): CHECKOUT-4941 Add Customer Creation functionality

capsula4:
## What?
Add Create Customer Account functionality.

## Why?
So that customers can create an account during the checkout process.

## Testing / Proof
unit

@bigcommerce/checkout

PR #1047, opened 2020-12-15
Revert "feat(customer): CHECKOUT-4941 Add Customer Creation functionality"

helga1507:
Reverts bigcommerce/checkout-sdk-js#1025


capsula4:
Do you still think this needs to be reverted?

PR #1049, opened 2020-12-15
INT-2816 Fix unit test for googlepay-adyenv2-payment

capsula4:
## What?
Fix unit test for googlepay-adyenv2-payment. Removes time-offset assertion.

## Why?
 It depends on what machine is being executed, it could fail.

## Testing / Proof
unit


davidchin (approved):

davidchin (commented):
Actually, can we stub out the current date time instead?

capsula4:
Don't think it can be done? The object is instantiated in the function, don't think you can stub that? Feel free to update if you want, but at least this will stop the test from failing for now.

davidchin:
Yeah you need to do a bit of refactoring in the code to achieve that. Let's just get the test to pass for now. 👍

PR #1053, opened 2021-01-06
feat(checkout): INT-3675 Add Bolt's tracking script to checkout

FranciscoOnt:
## What? [INT-3675](https://jira.bigcommerce.com/browse/INT-3675)
Load Bolt's `track.js` when using Bolt's Fraud option along with lightbox's script.

## Why?
To make Bolt able to collect data for the best checkout experience.

## Testing / Proof
QA review

### DIFF
--- a/src/payment/strategies/bolt/bolt-script-loader.spec.ts
+++ b/src/payment/strategies/bolt/bolt-script-loader.spec.ts
@@ -34,12 +34,14 @@ describe('BoltScriptLoader', () => {
             await boltScriptLoader.load(publishableKey);
 
             expect(scriptLoader.loadScript).toHaveBeenCalledWith('//connect.bolt.com/connect-bigcommerce.js', expect.any(Object));
+            expect(scriptLoader.loadScript).toHaveBeenCalledWith('//connect.bolt.com/track.js');
         });
 
         it('loads the bolt script in test mode', async () => {
           await boltScriptLoader.load(publishableKey, true);
 
           expect(scriptLoader.loadScript).toHaveBeenCalledWith('//connect-sandbox.bolt.com/connect-bigcommerce.js', expect.any(Object));
+          expect(scriptLoader.loadScript).toHaveBeenCalledWith('//connect-sandbox.bolt.com/track.js');
       });
 
         it('returns the Bolt Client from the window', async () => {
--- a/src/payment/strategies/bolt/bolt-script-loader.ts
+++ b/src/payment/strategies/bolt/bolt-script-loader.ts
@@ -19,7 +19,11 @@ export default class BoltScriptLoader {
             },
         };
 
-        await this._scriptLoader.loadScript(`//connect${testMode ? '-sandbox' : ''}.bolt.com/connect-bigcommerce.js`, options);
+        await Promise.all([
+            this._scriptLoader.loadScript(`//connect${testMode ? '-sandbox' : ''}.bolt.com/connect-bigcommerce.js`, options),
+            this._scriptLoader.loadScript(`//connect${testMode ? '-sandbox' : ''}.bolt.com/track.js`),
+        ]);
+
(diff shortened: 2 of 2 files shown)


bc-RicCampo:
@FranciscoOnt do we need what information we will be forwarding to Bolt? Could I have access to this script, please?

FranciscoOnt:
@bc-RicCampo you can find the [script here (https://connect-sandbox.bolt.com/track.js)](https://connect-sandbox.bolt.com/track.js), however there is only the minified version.

bc-RicCampo:
@FranciscoOnt would it be possible to get in touch with Bolt and see if we can get the proper script?

bc-RicCampo:
@bc-RicCampo

bc-RicCampo:
@FranciscoOnt I have performed a review of the subject javascript, this consisted of a walkthrough by the Bolt team a scan of the script for any vulnerabilities. 

From the walkthrough, I understood that the track.js script's function is to track the iframe and status of the transaction but it does not send any PII or confidential information. I did not check whether this is the case as I have not a functioning store with the script on it, however, from the walkthrough, scans, and discussions with Bolt I consider the risk of data leakage to be low on the basis of the information available to me at the time of the audit.

Please note, that even though I did not find any vulnerabilities in the code or in Bolt's walkthrough, it is possible there are unknown vulnerabilities and measures should be taken to sandbox the script and ensure to the best of our ability that the script only access the relevant information.

mauricio-sg (approved):

capsula4 (commented):
Changes look good. From a security perspective, we are also loading connect-bigcommerce.js which is not hosted by us, so technically they could be loading other scripts themselves through that one.

I think we can consider our partner's scripts as trusted, as we have no control over updates on them. Maybe a question for the @bigcommerce/security  team, CC @abhijethbc

bc-RicCampo:
@capsula4, unfortunately, we cannot and should not under any circumstances consider a third party script as trusted, many organizations have been compromised by making this assumption. Even though I have reviewed and assessed track.js, that does not mean is trusted, only that it represents a low risk to our platform. I am happy to explain in more detail the difference if requried.

Regarding the connect-bigcommerce.js, which was mentioned during my meeting with Bolt but was not reviewed, this script will need to be reviewed like the track.js, are there any other scripts?

capsula4:
Is reviewing an external script enough? How can we track changes to a file if it's not hosted by us? We load external scripts from most payment providers.

bc-RicCampo:
@capsula4, I would say we need to ensure all external scripts have to be treated as untrusted, run on sandboxes and ensure that they cause the minimum amount of impact in case they are compromised or introduced a vulnerability. We should ensure all external scripts are monitored and review as part of the pentesting activities. We also need to ensure they are reviewed and security assessed before being deployed to our platform.

In an ideal system, we would not deploy a third-party script/application without being passed through the security certification process, also we would have a register of all the scripts being run, but as I imagine this could possibly be into the thousands, it would be impracticable within our current resources. Taking into account our limitations, what we need to ensure, within our current capability, is that we run third party scripts in as secure an environment as we can make it, using sandboxing, monitoring, security certification, security audits, and assessments on a regular basis or as regular as our resources allow us to make it.

I hope that answers your questions, I imagine this has gone a bit out of the purpose of the PR, could I suggest we take this conversation to slack or some other channel?

pgonzalezr:
Looks good from QE side 💚 

Tested:
- Completing orders from checkout page
- Canceling checkout

cmedinilla:
@capsula4 @bc-RicCampo any update about this PR?

capsula4 (approved):
Looks good from a code perspective 👍

cmedinilla:
@bc-RicCampo any concern to merge this PR?

bc-RicCampo:
@cmedinilla the risk of low has been noted in my previous comments, please proceed with the merging of this PR.

PR #1055, opened 2021-01-11
fix(provider): INT-3655 Avoid passing null nonce token to StripeV3

orecio-bc:
## What?
[INT-3655](https://jira.bigcommerce.com/browse/INT-3655): Stripe V3 - Some customers unable to checkout with Stripe

This issue hasn't been reproduced, at least without forcing it to happen, it seems that sometimes the nonce token is not loaded and is sent as null to BigPay, the strategy should not allow this, so we are checking it before sending the payload to BigPay

## Why?
Customers are presenting issues that prevent them to finishing the payment when using Stripe as payment method, which end in bad user experience 

## Testing / Proof
Tests
image

Message that the client will see
image

### DIFF
--- a/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
+++ b/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
@@ -87,6 +87,10 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
 
                 paymentIntent = await this._confirmVaultedPayment(clientSecret);
 
+                if (!paymentIntent.id) {
+                    throw new MissingDataError(MissingDataErrorType.MissingPaymentToken);
+                }
+
                 paymentPayload = {
                     methodId,
                     paymentData: {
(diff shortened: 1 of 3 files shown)


src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -87,6 +87,10 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
 
                 paymentIntent = await this._confirmVaultedPayment(clientSecret);
 
+                if (!paymentIntent.id) {
+                    throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
```
FranciscoOnt, line 91: Usually,  `PaymentNotInitialized` is used when something goes wrong with the initialize step of the strategy.
A `MissingDataError` might be more accurate for this situation 🍹

FranciscoOnt (approved):

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
orecio-bc: Thanks, which one do you think it suits bests?
**MissingDataErrorType.MissingPayment** or **MissingDataErrorType.MissingPaymentToken**

orecio-bc:
@FranciscoOnt ♻️ changes ready, text now reads: **'Unable to proceed because the token required to submit a payment is missing.'**

FranciscoOnt (approved):
LGTM

capsula4 (approved):

mmartinezintersys:
💚 TESTED

QA Details: Run a regression on payments to verify:

 - Pay with credit cards
 - Pay with vaulted credit cards
 - Authorize & Capture orders
 - Authorize Only orders
 - Capture an order's funds
 - Void an order
 - Total refund of an order
 - Partial refund of an order

PR #1056, opened 2021-01-12
feat(payment): INT-3702 added vaulted cc

JoseAlvarezHdz:
## What? [INT-3702](https://jira.bigcommerce.com/browse/INT-3702)
added vaulted methods in order to save cc, added vault_payment_intrument variables, and create conditions to mount or not mount mollie components 

## Why?
We want the user to be able to save cc and use them later when logged in 

## Testing / Proof

Screen Shot 2021-01-13 at 11 35 01

### DIFF
--- a/src/payment/strategies/mollie/mollie-initialize-options.ts
+++ b/src/payment/strategies/mollie/mollie-initialize-options.ts
@@ -1,7 +1,57 @@
+/**
+ * A set of options that are required to initialize the Mollie payment method.
+ *
+ * Once Mollie payment is initialized, credit card form fields are provided by the
+ * payment provider as IFrames, these will be inserted into the current page. These
+ * options provide a location and styling for each of the form fields.
+ *
+ * ```js
+ * service.initializePayment({
+ *      methodId: 'mollie',
+ *      mollie: {
+ *          containerId: 'container',
+ *          cardNumberId: '',
+ *          cardHolderId: '',
+ *          cardCvcId: '',
+ *          cardExpiryId: '',
+ *          styles : {
+ *              base: {
+ *                  color: '#fff'
+ *              }
+ *          }
+ *      }
+ * });
+ */
 export default interface MolliePaymentInitializeOptions {
+    /**
+     * ContainerId is use in Mollie for determined either its showing or not the
+     * container, because when Mollie has Vaulted Instruments it gets hide,
+     * and shows an error because can't mount Provider Components
+     */
+    containerId: string;
+
+    /**
+     * The location to insert Mollie Component
+     */
     cardNumberId: string;
+
+    /**
+     * The location to insert Mollie Component
+     */
     cardHolderId: string;
+
+    /**
+     * The location to insert Mollie Component
+     */
     cardCvcId: string;
+
+    /**
+     * The location to insert Mollie Component
+     */
     cardExpiryId: string;
+
+    /**
+     * A set of styles required for the mollie components
+     */
     styles: object;
 }
--- a/src/payment/strategies/mollie/mollie-payment-strategy.ts
+++ b/src/payment/strategies/mollie/mollie-payment-strategy.ts
@@ -1,4 +1,4 @@
-import { some } from 'lodash';
+import { each, some } from 'lodash';
 
 import { PaymentActionCreator } from '../..';
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
@@ -6,12 +6,18 @@ import { getBrowserInfo } from '../../../common/browser-info';
 import { InvalidArgumentError, MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType, RequestError } from '../../../common/error/errors';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
 import { PaymentArgumentInvalidError } from '../../errors';
+import isVaultedInstrument from '../../is-vaulted-instrument';
+import { HostedInstrument } from '../../payment';
 import { PaymentInitializeOptions, PaymentRequestOptions } from '../../payment-request-options';
 import PaymentStrategy from '../payment-strategy';
 
 import { MollieClient, MollieElement } from './mollie';
 import MolliePaymentInitializeOptions from './mollie-initialize-options';
 import MollieScriptLoader from './mollie-script-loader';
+
+export enum MolliePaymentMethodType {
+    creditcard = 'credit_card',
+}
 
 export default class MolliePaymentStrategy implements PaymentStrategy {
     private _initializeOptions?: MolliePaymentInitializeOptions;
@@ -51,7 +57,7 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
             throw new InvalidArgumentError('Unable to initialize payment because "merchantId" argument is not provided.');
         }
 
-        if (methodId === 'creditcard') {
+        if (methodId === MolliePaymentMethodType.creditcard) {
             this._mollieClient = await this._loadMollieJs(merchantId, storeConfig.storeProfile.storeLanguage, testMode);
             this._mountElements();
         }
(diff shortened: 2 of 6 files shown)


src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -83,6 +94,8 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
                             credit_card_token: {
                                 token,
                             },
+                            vault_payment_instrument: shouldSaveInstrument || null,
+                            set_as_default_stored_instrument: shouldSetAsDefaultInstrument || null,
```
mauricio-sg, line 98: Why it fallbacks to `null`? 🤔

mauricio-sg (approved):
LGTM

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -70,6 +75,12 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
             if (payment.methodId === 'creditcard') {
                 await this._store.dispatch(this._orderActionCreator.submitOrder(order, options));
 
+                if (paymentData && isVaultedInstrument(paymentData)) {
+                    return this._store.dispatch(this._paymentActionCreator.submitPayment({
```
capsula4, line 79: any reason why this is not simply:
```js
return this._store.dispatch(this._paymentActionCreator.submitPayment(payment)
```

src/payment/strategies/mollie/mollie-initialize-options.ts
```diff
@@ -1,7 +1,9 @@
 export default interface MolliePaymentInitializeOptions {
+    containerId: string;
```
capsula4, line 2: can you provide a docblock explaining what each attribute does? It's unclear to me why `containerId` is now a required property and before it wasnt

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -149,17 +162,22 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
...
-        const { cardNumberId, cardCvcId, cardExpiryId, cardHolderId, styles } = this._getInitializeOptions();
+        const { containerId, cardNumberId, cardCvcId, cardExpiryId, cardHolderId, styles } = this._getInitializeOptions();
+
+        const container = document.getElementById(containerId);
 
-        this._cardHolderElement = mollieClient.createComponent('cardHolder', { styles });
-        this._cardNumberElement = mollieClient.createComponent('cardNumber', { styles });
-        this._verificationCodeElement = mollieClient.createComponent('verificationCode', { styles });
-        this._expiryDateElement = mollieClient.createComponent('expiryDate', { styles });
+        if (container?.style.display !== 'none') {
```
capsula4, line 171: this behaviour might be worth documenting in the interface, what happens when containerId is provided vs when not

src/payment/strategies/mollie/mollie-initialize-options.ts
```diff
@@ -1,7 +1,65 @@
...
+ *          styles : {
+ *              base: {
+ *                  color: '#fff'
+ *              }
+ *          }
+ *      }
+ * });
+ * ```
+ *
+ * ContainerId is use in Mollie for determined either its showing or not the
```
capsula4, line 26: This paragraphp is not very clear to me. If Mollie has vaulted instruments, does that mean containerId is mandatory? Also, this explanation could go in the docblock below, replacing the
```
The location where Mollie Components are inserted
```

src/payment/strategies/mollie/mollie-initialize-options.ts
```diff
@@ -1,7 +1,65 @@
+/**
+ * A set of options that are required to initialize the Mollie payment method.
+ *
+ * Once Mollie payment is initialized, credit card form fields are provided by the
+ * payment provider as IFrames, these will be inserted into the current page. These
```
capsula4, line 5: Is the iframe inserted in `containerId` ?

capsula4 (approved):
Code looks good, just pending to confirm some documentation that is still unclear to me.

src/payment/strategies/mollie/mollie-initialize-options.ts
JoseAlvarezHdz: Nope

src/payment/strategies/mollie/mollie-initialize-options.ts
JoseAlvarezHdz: when mollie has vaulted instruments we shouldnt create MollieComponents because there is no where to mount them, but when you select add another card it should create and mount MollieComponents 

thats why i added the containerId so i can know when its hidden or shown

capsula4 (approved):

src/payment/strategies/mollie/mollie-initialize-options.ts
JoseAlvarezHdz: ContainerId is for the container that will have the DIVs that will contain the iframe for each of the 4 Mollie Components, 
and i need it because i have to create and mount each of this Mollie Components every time the user selects to "add a new card" when having vaulted instruments, when they select a vaulted instrument, ContainerId element has display property as none, but when they select to add a new card it change to 'block' and that when i want to create and mount the MollieComponents and thats what i use containerId for

JoseAlvarezHdz:
I added some new changes! can you validate them please? @capsula4

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -61,15 +67,22 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
 
     async execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
         const { payment , ...order} = payload;
+        const paymentData = payment?.paymentData;
+        const shouldSaveInstrument = paymentData && (paymentData as HostedInstrument).shouldSaveInstrument;
+        const shouldSetAsDefaultInstrument = paymentData && (paymentData as HostedInstrument).shouldSetAsDefaultInstrument;
```
capsula4, line 72: maybe do `(paymentData as HostedInstrument)?.shouldSetAsDefaultInstrument` instead

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -105,19 +120,40 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
...
+
+        for (const i in mollieComponents) {
+            if (mollieComponents.hasOwnProperty(i)) {
+                mollieComponents[i].remove();
+            }
+        }
+
+        const controllers = document.querySelectorAll('.mollie-components-controller');
+
+        for (const i in controllers) {
```
capsula4, line 146: you can use `Object.entries()` instead of `for ...in`, that whay you dont need to check hasOwnProperty.

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -131,12 +167,12 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
         return this._initializeOptions;
     }
 
-    private async _loadMollieJs(merchantId: string, locale: string, testmode: boolean = false) {
+    private _loadMollieJs(merchantId: string, locale: string, testmode: boolean = false) {
```
capsula4, line 170: would be good to add a return type to this method

src/payment/strategies/mollie/mollie-payment-strategy.ts
JoseAlvarezHdz: I get this error 

`Property 'entries' does not exist on type 'ObjectConstructor'.ts(2339)`

Screen Shot 2021-04-07 at 10 43 57

JoseAlvarezHdz:
I added the changes you request! can you validate them please? @capsula4

src/payment/strategies/mollie/mollie-payment-strategy.ts
capsula4: Not sure why isn't working. What I meant to use was`Object.entries(controllers).forEach((value => value.remove());`

Another option is `Object.keys(controllers).forEach(key => controllers[key].remove());` or lodash' values

JoseAlvarezHdz:
I added the changes you request! can you validate them please? @capsula4

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -105,19 +118,27 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
...
+
+        this.removeMollieComponents();
 
         return Promise.resolve(this._store.getState());
     }
 
+    private removeMollieComponents(): void {
+        const mollieComponents = document.querySelectorAll('.mollie-component');
+
+        each(mollieComponents, component => component.remove());
```
capsula4, line 131: optional, you could use the native version, `mollieComponents.forEach(component => component.remove());`

capsula4 (approved):

PR #1062, opened 2021-01-28
CHECKOUT-4726 Allow creating customer address

capsula4:
## What?
Allow creating customer address. Alpha functionality.

## Why?
To allow creating customer addresses during checkout, which can be used for the multishipping UI.

## Testing / Proof
unit

@bigcommerce/checkout

### DIFF
--- a/src/customer/customer-action-creator.ts
+++ b/src/customer/customer-action-creator.ts
@@ -1,5 +1,5 @@
 import { createAction, ThunkAction } from '@bigcommerce/data-store';
-import { concat, from, of } from 'rxjs';
+import { concat, defer, from, of, Observable } from 'rxjs';
 import { catchError, switchMap } from 'rxjs/operators';
 
 import { CheckoutActionCreator, InternalCheckoutSelectors } from '../checkout';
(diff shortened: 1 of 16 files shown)


src/customer/customer-action-creator.ts
```diff
@@ -54,6 +54,24 @@ export default class CustomerActionCreator {
...
+    createAddress(
+        customerAddress: CustomerAddressRequestBody,
+        options?: RequestOptions
+    ): ThunkAction<CreateCustomerAddressAction, InternalCheckoutSelectors> {
+        return store => concat(
+            of(createAction(CustomerActionType.CreateCustomerAddressRequested)),
+            from(this._customerRequestSender.createAddress(customerAddress, options))
+                .pipe(
+                    switchMap(() => concat(
+                        this._checkoutActionCreator.loadCurrentCheckout(options)(store),
```
davidchin, line 66: I wonder if the customer address endpoint itself can return all the newly created address, we can then map the response in the same shape as the checkout's customer sub resource and store it in memory. That way we don't have to make an extra call to hydrate the cart, which can sometimes be slow depending on its size.

davidchin (approved):

src/customer/customer-action-creator.ts
capsula4: Yeah, that would be great, or have a customer endpoint where we can retrieve addresses, avoiding having to reload the cart/checkout.

davidchin: cc @snaderiBC

snaderiBC: >That way we don't have to make an extra call to hydrate the cart

That's interesting idea 🤔

PR #1071, opened 2021-02-15
fix(checkout): INT-3665 3D Secure is declining the transaction in Braintree without any reason in some stores

julioabarca2:
## What? [INT-3665](https://jira.bigcommerce.com/browse/INT-3665)

Fix the 3D Secure in Braintree because sometimes the nonce is not created correctly and the flow process must be break to doesn't show 3D Secure dialog.

## Why?

Because the transaction is declined after the 3D Secure dialog is showed without any kind of message that what was wrong

## Testing / Proof

[Video](https://drive.google.com/file/d/1rwQ6O5C0JrSHkeAqrndgE3z4gtGgVRXu/view?usp=sharing)

The issue can't be reproduced locally, but we know that the creation of the nonce is the problem, so if the nonce is not created we break the flow of the payment process to doesn't show 3D Secure dialog.

Ping @bigcommerce/payments @bigcommerce/apex-integrations

### DIFF
--- a/src/payment/strategies/braintree/braintree-payment-processor.ts
+++ b/src/payment/strategies/braintree/braintree-payment-processor.ts
@@ -129,7 +129,7 @@ export default class BraintreePaymentProcessor {
         amount: number,
         nonce: string
     ): Promise<BraintreeVerifyPayload> {
-        if (!this._threeDSecureOptions) {
+        if (!this._threeDSecureOptions || !nonce) {
             throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
         }
 
@@ -171,6 +171,8 @@ export default class BraintreePaymentProcessor {
                         validate: false,
                     },
                     billingAddress: billingAddress && {
+                        countryCodeAlpha2: billingAddress.countryCode,
+                        locality: billingAddress.city,
                         countryName: billingAddress.country,
                         postalCode: billingAddress.postalCode,
                         streetAddress: billingAddress.address2 ?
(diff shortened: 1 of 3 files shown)


src/payment/strategies/braintree/braintree-payment-processor.ts
```diff
@@ -133,6 +133,10 @@ export default class BraintreePaymentProcessor {
             throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
         }
 
+        if (typeof nonce === 'undefined' || !nonce) {
```
mauricio-sg, line 136: Why not just `if (!nonce) {` ? 🤔

mauricio-sg (approved):
LGTM 👍

src/payment/strategies/braintree/braintree-payment-processor.ts
mauricio-sg: Also, you can use line 132 `if (!this._threeDSecureOptions || !nonce) {`

julioabarca2: > Why not just `if (!nonce) {` ? 🤔

because in typescript a string can be undefined too

julioabarca2: > Also, you can use line 132 `if (!this._threeDSecureOptions || !nonce) {`

I could merge that code

mmartinezintersys:
💚 TESTED

QA Details:

 - Verify valid 3DS transactions are not declined 
 - Sanity check for purchases wiothout 3DS enabled

capsula4 (approved):

PR #1061, opened 2021-01-26
fix(payment): INT-3611 Refresh the state in Braintree

roberbnd:
## What? [INT-3611](https://jira.bigcommerce.com/browse/INT-3611)
Refresh the values in store

## Why?
when the backend delete the **nonce** value in the front-end is not refreshed, then every time we want to pay we are using the same nonce value

## Testing / Proof
https://drive.google.com/file/d/126Y8zs07T9h5KelzlH7lTC4if85KwBem/view?usp=sharing

@bigcommerce/checkout @bigcommerce/payments @bigcommerce/apex-integrations

PR #1082, opened 2021-03-09
fix(payment): INT-3611 Refresh the state in Googlepay and Braintree r…

roberbnd:
…efactor

## What? [INT-3611](https://jira.bigcommerce.com/browse/INT-3611)
Refresh this._paymentMethod after use the nonce value with that if we want to pay one more time and if the nonce is empty we refresh the value.

The changes related with braintree is a little refactor I did to simplify the code with this same issue:  [PR](https://github.com/bigcommerce/checkout-sdk-js/pull/1061)

## Why?
Googlepay was keeping the same nonce value in the backend.

## Testing / Proof
https://drive.google.com/file/d/1KvdlRl46SMZL4FUAyPb4ND4pqIuO_tfa/view?usp=sharing

Revert this PR

## Dependencies
This PR has dependency with [fix(payments): INT-3611 Googlepay - Remove nonce value after use it](https://github.com/bigcommerce/bigcommerce/pull/39827)

### DIFF
--- a/src/payment/strategies/googlepay/googlepay-payment-strategy.ts
+++ b/src/payment/strategies/googlepay/googlepay-payment-strategy.ts
@@ -1,10 +1,11 @@
 import { CheckoutActionCreator, CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
 import { getBrowserInfo } from '../../../common/browser-info';
 import { InvalidArgumentError, MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType } from '../../../common/error/errors';
-import { bindDecorator as bind } from '../../../common/utility';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
+import { PaymentArgumentInvalidError } from '../../errors';
 import PaymentActionCreator from '../../payment-action-creator';
+import PaymentMethod from '../../payment-method';
 import PaymentMethodActionCreator from '../../payment-method-action-creator';
 import { PaymentInitializeOptions, PaymentRequestOptions } from '../../payment-request-options';
 import PaymentStrategyActionCreator from '../../payment-strategy-action-creator';
@@ -18,8 +19,9 @@ import GooglePayPaymentProcessor from './googlepay-payment-processor';
 
 export default class GooglePayPaymentStrategy implements PaymentStrategy {
     private _googlePayOptions?: GooglePayPaymentInitializeOptions;
-    private _methodId?: string;
     private _walletButton?: HTMLElement;
+    private _paymentMethod?: PaymentMethod;
+    private _buttonClickEventHandler?: (event: Event ) => Promise<InternalCheckoutSelectors>;
 
     constructor(
         private _store: CheckoutStore,
@@ -32,71 +34,84 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
         private _googlePayAdyenV2PaymentProcessor?: GooglePayAdyenV2PaymentProcessor
     ) {}
 
-    initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
-        this._methodId = options.methodId;
+    async initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
+        const { methodId } = options;
 
-        return this._googlePayPaymentProcessor.initialize(this._methodId)
-            .then(() => {
-                this._googlePayOptions = this._getGooglePayOptions(options);
+        const state = await this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId));
+        this._paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(methodId);
 
-                if (!this._googlePayOptions) {
-                    throw new InvalidArgumentError('Unable to initialize payment because "options.googlepay" argument is not provided.');
-                }
+        this._googlePayOptions = this._getGooglePayOptions(options);
 
-                const walletButton = this._googlePayOptions.walletButton && document.getElementById(this._googlePayOptions.walletButton);
+        this._buttonClickEventHandler = this._handleButtlonClickedEvent(methodId);
 
-                if (walletButton) {
-                    this._walletButton = walletButton;
-                    this._walletButton.addEventListener('click', this._handleWalletButtonClick);
-                }
+        if (this._paymentMethod.initializationData.nonce) {
+            return Promise.resolve(this._store.getState());
+        }
 
-                return this._store.getState();
-            });
+        await this._googlePayPaymentProcessor.initialize(methodId);
+
+        if (!this._googlePayOptions.walletButton) {
+            throw new InvalidArgumentError('walletButton argument is missing');
+        }
+
+        const walletButton = document.getElementById(this._googlePayOptions.walletButton);
+
+        if (!walletButton) {
+            throw new InvalidArgumentError('Unable to create wallet, walletButton ID could not be found');
+        }
+
+        this._walletButton = walletButton;
+        this._walletButton.addEventListener('click', this._buttonClickEventHandler);
+
+        return Promise.resolve(this._store.getState());
     }
 
     deinitialize(): Promise<InternalCheckoutSelectors> {
-        if (this._walletButton) {
-            this._walletButton.removeEventListener('click', this._handleWalletButtonClick);
+        if (this._walletButton && this._buttonClickEventHandler) {
+            this._walletButton.removeEventListener('click', this._buttonClickEventHandler);
         }
 
+        this._buttonClickEventHandler = undefined;
         this._walletButton = undefined;
 
         return this._googlePayPaymentProcessor.deinitialize()
             .then(() => this._store.getState());
     }
 
-    execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+    async execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
         if (!this._googlePayOptions) {
             throw new InvalidArgumentError('Unable to initialize payment because "options.googlepay" argument is not provided.');
         }
 
-        const {
-            onError = () => {},
-            onPaymentSelect = () => {},
-        } = this._googlePayOptions;
+        if (!payload.payment) {
+            throw new PaymentArgumentInvalidError(['payment']);
+        }
 
-        return Promise.resolve(this._getPayment())
-            .then(payment => {
-                if (!payment.paymentData.nonce || !payment.paymentData.cardInformation) {
-                    // TODO: Find a way to share the code with _handleWalletButtonClick method
-                    return this._googlePayPaymentProcessor.displayWallet()
-                        .then(paymentData => this._paymentInstrumentSelected(paymentData))
-                        .then(() => onPaymentSelect())
-                        .then(() => this._getPayment())
-                        .catch(error => {
-                            if (error.statusCode !== 'CANCELED') {
-                                onError(error);
-                            }
-                        });
-                }
+        const { methodId } = payload.payment;
 
-                return payment;
-            })
-            .then(() =>
-                this._store.dispatch(this._orderActionCreator.submitOrder({ useStoreCredit: payload.useStoreCredit }, options))
-                    .then(() => this._store.dispatch(this._paymentActionCreator.submitPayment(this._getPayment())))
-                    .catch(error => this._googlePayAdyenV2PaymentProcessor?.processAdditionalAction(error) || Promise.reject(error))
-            );
+        let payment = await this._getPayment(methodId);
+
+        if (!payment.paymentData.nonce || !payment.paymentData.cardInformation) {
+            const {
+                onError = () => {},
+                onPaymentSelect = () => {},
+            } = this._googlePayOptions;
+            await this._displayWallet(methodId, onPaymentSelect, onError);
+            payment = await this._getPayment(methodId);
+
+            if (!payment.paymentData.nonce) {
+                throw new MissingDataError(MissingDataErrorType.MissingPayment);
+            }
+        }
+
+        try {
+            await this._store.dispatch(this._orderActionCreator.submitOrder({ useStoreCredit: payload.useStoreCredit }, options));
+
+            return await this._store.dispatch(this._paymentActionCreator.submitPayment(payment));
+        } catch (error) {
+            this._googlePayAdyenV2PaymentProcessor?.processAdditionalAction(error);
+            throw error;
+        }
     }
 
     finalize(): Promise<InternalCheckoutSelectors> {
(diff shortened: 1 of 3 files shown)


src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -65,38 +74,40 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+        if (!payment.paymentData.nonce) {
+            throw new MissingDataError(MissingDataErrorType.MissingPayment);
+        }
+
+        try {
+            await this._store.dispatch(this._orderActionCreator.submitOrder({ useStoreCredit: payload.useStoreCredit }, options));
+
+            return await this._store.dispatch(this._paymentActionCreator.submitPayment(payment));
+        } catch (error) {
+
```
vmparra, line 108: Remove this line break

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -217,4 +197,39 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
             this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId)),
         ]);
     }
+
+    private _buttonClickEvent(methodId: string): any {
```
vmparra, line 201: Please do not return any.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -217,4 +197,39 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+                onPaymentSelect = () => {},
+            } = this._googlePayOptions;
+
+            return this._store.dispatch(this._paymentStrategyActionCreator.widgetInteraction(async () => {
+                return await this._displayWallet(methodId, onPaymentSelect, onError);
+            }, { methodId }), { queueId: 'widgetInteraction' });
+        };
+    }
+
+    private async _displayWallet(methodId: string, onPaymentSelect: any, onError: any ): Promise<void>  {
```
vmparra, line 223: Please make sure you type every argument.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -134,81 +145,50 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
-        if (!this._methodId) {
+    private async _paymentInstrumentSelected(paymentData: GooglePaymentData, methodId: string) {
+        if (!methodId) {
             throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
         }
 
-        const methodId = this._methodId;
-
         // TODO: Revisit how we deal with GooglePaymentData after receiving it from Google
         await this._googlePayPaymentProcessor.handleSuccess(paymentData);
```
vmparra, line 194: Remove this line break

vmparra (changes requested):
Hi @roberbnd, thanks for doing this. Actually, jira ticket INT-3611 refers Pay Pal only. We do appreciate proactivity but we also like to keep work as atomic as possible, in this case I would suggest to leave this PR for Pay Pal only, and then issue a ticket for Google Pay

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
roberbnd: it's needed before a `return` statement (eslint rules)

roberbnd:
@vmparra  Could you review, please? ♻️

vmparra (changes requested):
@roberbnd I love the refactor on Google Pay, just one last comment. Please tell me if you have questions.

roberbnd:
@vmparra  done

vmparra (approved):
LGTM 👍🏻

mmartinezintersys:
💚 TESTED

QA Details: regression test for Google Pay with Braintree

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -32,31 +34,38 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+        const state = await this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId));
+        this._paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(methodId);
 
-                if (!this._googlePayOptions) {
-                    throw new InvalidArgumentError('Unable to initialize payment because "options.googlepay" argument is not provided.');
-                }
+        try {
+            this._googlePayOptions = this._getGooglePayOptions(options);
+        } catch (error) {
+            return Promise.reject(error);
```
capsula4, line 46: I believe you don't need to do try/catch if you will simply return a rejected promise with the same caught error.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -32,31 +34,38 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+        }
 
-                const walletButton = this._googlePayOptions.walletButton && document.getElementById(this._googlePayOptions.walletButton);
+        this._buttonClickEventCurry = this._buttonClickEvent(methodId);
```
capsula4, line 49: maybe name this something like `this._buttonClickEventHandler` ?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -32,31 +34,38 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+        if (this._paymentMethod.initializationData.nonce) {
+            return Promise.resolve(this._store.getState());
+        }
 
-                return this._store.getState();
-            });
+        await this._googlePayPaymentProcessor.initialize(methodId);
+        const walletButton = this._googlePayOptions.walletButton && document.getElementById(this._googlePayOptions.walletButton);
+
+        if (walletButton) {
```
capsula4, line 58: if walletButton is null, should the promise be rejected on line 63?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -32,31 +34,38 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
     }
 
     deinitialize(): Promise<InternalCheckoutSelectors> {
-        if (this._walletButton) {
-            this._walletButton.removeEventListener('click', this._handleWalletButtonClick);
+        if (this._walletButton && this._buttonClickEventCurry) {
+            this._walletButton.removeEventListener('click', this._buttonClickEventCurry);
         }
 
         this._walletButton = undefined;
```
capsula4, line 71: should you wipe out `this._buttonClickEventCurry = null` too?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -65,38 +74,39 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+        if (!payment.paymentData.nonce) {
+            throw new MissingDataError(MissingDataErrorType.MissingPayment);
+        }
+
+        try {
+            await this._store.dispatch(this._orderActionCreator.submitOrder({ useStoreCredit: payload.useStoreCredit }, options));
+
+            return await this._store.dispatch(this._paymentActionCreator.submitPayment(payment));
+        } catch (error) {
+            return this._googlePayAdyenV2PaymentProcessor?.processAdditionalAction(error) || Promise.reject(error);
```
capsula4, line 108: rather than doing `||`, I think what you actually want is to execute both
```js
catch(error) {
   this._googlePayAdyenV2PaymentProcessor?.processAdditionalAction(error);
   throw error;
}
```

re-throwing the error should behave the same as `Promise.reject(error)` but simpler.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -134,81 +144,50 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
-
         const paymentData = {
-            method: this._methodId,
+            method: methodId,
             nonce,
-            cardInformation: paymentMethod.initializationData.card_information,
+            cardInformation: this._paymentMethod.initializationData.card_information,
         };
 
+        if (this._paymentMethod.initializationData.nonce) {
```
capsula4, line 175: shouldn't you be checking `nonce` rather than `this._paymentMethod.initializationData.nonce`

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -217,4 +196,39 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
             this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId)),
         ]);
     }
+
+    private _buttonClickEvent(methodId: string): (event: Event) => Promise<InternalCheckoutSelectors> {
```
capsula4, line 200: this is a method name, so it should contain a verb, maybe `_handleButtlonClickedEvent` ?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -217,4 +196,39 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+
+        return (event: Event) => {
+            if (event) {
+                event.preventDefault();
```
capsula4, line 204: `event?.preventDefault()` rather than `if(event) ...`
Also, if event might not be present, maybe better to mark it as optional in the function signature?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -217,4 +196,39 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+                throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+            }
+
+            const {
+                onError = () => {},
+                onPaymentSelect = () => {},
+            } = this._googlePayOptions;
+
+            return this._store.dispatch(this._paymentStrategyActionCreator.widgetInteraction(async () => {
+                return await this._displayWallet(methodId, onPaymentSelect, onError);
```
capsula4, line 217: this function is a one liner, so maybe it reads better without the curly braces?
```js
return this._store.dispatch(
  this._paymentStrategyActionCreator.widgetInteraction(
    async () => await this._displayWallet(methodId, onPaymentSelect, onError),
   { methodId }
  ),
  { queueId: 'widgetInteraction' }
);

File: src/payment/strategies/googlepay/googlepay-payment-strategy.ts [hunk 14]
```diff
@@ -217,4 +196,39 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+    private async _displayWallet(methodId: string, onPaymentSelect: () => void, onError: (onError: Error) => void ): Promise<void> {
+        try {
+            const paymentData = await this._googlePayPaymentProcessor.displayWallet();
+            await this._paymentInstrumentSelected(paymentData, methodId);
+
+            return onPaymentSelect();
+        } catch (error) {
+            if (error.statusCode !== 'CANCELED') {
+                onError(error);
+            }
```
capsula4, line 231: should you be re-throwing the error here? `throw error;` on line 232?

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
roberbnd: no, because in the case` if (methodId === 'googlepayadyenv2')` the `nonce` value couldn't be empty then it isn't going to call the backend to refresh the value, if `this._paymentMethod.initializationData.nonce` is empty

roberbnd:
@capsula4 Could you review, please? ♻️

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -32,71 +34,79 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+        const walletButton = this._googlePayOptions.walletButton && document.getElementById(this._googlePayOptions.walletButton);
+
+        if (walletButton) {
+            this._walletButton = walletButton;
+            this._walletButton.addEventListener('click', this._buttonClickEventHandler);
+
+            return Promise.resolve(this._store.getState());
+        }
+
+        return Promise.reject(new Error('walletButton is missing'));
```
capsula4, line 61: because this is an async method, you can simply `throw new Error('walletButton is missing');`

I think ideally you might want to differentiate the errors

```js
if (!this._googlePayOptions.walletButton) {
  throw new InvalidArgumentError('walletButton argument is missing'
}

const walletButton = document.getElementById(this._googlePayOptions.walletButton);

if (!walletButton) {
  throw new InvalidArgumentError('Unable to create wallet, walletButton ID could not be found');
}

File: src/payment/strategies/googlepay/googlepay-payment-strategy.ts (hunk 10 above)
Comment (capsula4) [reply src/payment/strategies/googlepay/googlepay-payment-strategy.ts:175]: I think this is quite confusing, you have `this._paymentMethod.initializationData.nonce` and `nonce` and in some cases they might be different things.

To make you can refactor this logic so it's easier to read by removing ambiguity around what `nonce` actually holds.

```js
const { nonce } = this._paymentMethod.initializationData;

if (nonce) {
  state = await this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId));
  this._paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(methodId);
}

const { card_information: cardInformation } = this._paymentMethod.initializationData;

return {
  methodId,
  paymentData: {
     method: methodId,
     cardInformation,
     nonce: this._getNonce(this._paymentMethod),
  },
};
```
```js
/* omitted types for simplicty */
function _getNonce({ methodId, initializationData: { nonce }}) {
 if (methodId === 'googlepayadyenv2') {
   return JSON.stringify({
                type: AdyenPaymentMethodType.GooglePay,
                googlePayToken: nonce,
                browser_info: getBrowserInfo(),
            });
  }

  return nonce;

}
```

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -217,4 +196,42 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
             this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId)),
         ]);
     }
+
+    private _handleButtlonClickedEvent(methodId: string): (event: Event) => Promise<InternalCheckoutSelectors> {
```
capsula4, line 200: the return type should be
```ts
(event?: Event) => Promise<InternalCheckoutSelectors>

**Review by capsula4 — COMMENTED** [review]
Looking great! Great refactoring so far, just some small minor comments remaining.

**Comment by roberbnd** [comment]
@capsula4 is ok?

File: src/payment/strategies/googlepay/googlepay-payment-strategy.ts [hunk 17]
```diff
@@ -32,71 +34,88 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+            throw new InvalidArgumentError('walletButton argument is missing');
+        }
+
+        const walletButton = document.getElementById(this._googlePayOptions.walletButton);
+
+        if (!walletButton) {
+            throw new InvalidArgumentError('Unable to create wallet, walletButton ID could not be found');
+        }
+
+        if (walletButton) {
```
capsula4, line 63: you don't need this `if` condition here. walletButton will always be truthy. You are already checking on line 59 for the value of this variable.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -32,71 +34,88 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+        if (!payment.paymentData.nonce || !payment.paymentData.cardInformation) {
+            const {
+                onError = () => {},
+                onPaymentSelect = () => {},
+            } = this._googlePayOptions;
+            await this._displayWallet(methodId, onPaymentSelect, onError);
+            payment = await this._getPayment(methodId);
+        }
+
+        if (!payment.paymentData.nonce) {
```
capsula4, line 107: this `if` condition makes sense to happen inside the previous `if` block. This is to be explicit around how the logic will be executed

```js
        if (!payment.paymentData.nonce || !payment.paymentData.cardInformation) {
            const {
                onError = () => {},
                onPaymentSelect = () => {},
            } = this._googlePayOptions;
            await this._displayWallet(methodId, onPaymentSelect, onError);
            payment = await this._getPayment(methodId);

            if (!payment.paymentData.nonce) {
              throw new MissingDataError(MissingDataErrorType.MissingPayment);
            }
        }
```

roberbnd:
@capsula4  heelloo! 😃  (poke you) ♻️  please

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -32,71 +34,84 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+        const walletButton = document.getElementById(this._googlePayOptions.walletButton);
+
+        if (!walletButton) {
+            throw new InvalidArgumentError('Unable to create wallet, walletButton ID could not be found');
+        }
+
+        this._walletButton = walletButton;
+        this._walletButton.addEventListener('click', this._buttonClickEventHandler);
+
+        return Promise.resolve(this._store.getState());
```
capsula4, line 66: whenever you are working with `async` methods, you don't need to wrap the return value in `Promise.resolve()`.

Async methods automatically wrap returned values in a promise, and thrown errors are wrapped in a `Promise.reject()`

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -32,71 +34,84 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
 
-                const walletButton = this._googlePayOptions.walletButton && document.getElementById(this._googlePayOptions.walletButton);
+        this._buttonClickEventHandler = this._handleButtlonClickedEvent(methodId);
 
-                if (walletButton) {
-                    this._walletButton = walletButton;
-                    this._walletButton.addEventListener('click', this._handleWalletButtonClick);
-                }
+        if (this._paymentMethod.initializationData.nonce) {
+            return Promise.resolve(this._store.getState());
```
capsula4, line 48: same as I commented below

capsula4 (approved):

PR #1085, opened 2021-03-13
fix(provider): INT-3571 Google Pay [StripeV3] - Billing address is missing

julioabarca2:
## What?

The Billing address is missing when payment with google pay in Stripe V3

## Why?

We are validating the billing address when checkout-sdk-js is retrieving for first time

## Testing / Proof

Screen Shot 2021-03-12 at 5 53 49 p m

Ping @bigcommerce/payments


src/payment/strategies/googlepay/googlepay-payment-processor.ts
```diff
@@ -217,6 +217,18 @@ export default class GooglePayPaymentProcessor {
         }
 
         const googlePayAddressMapped = this._mapGooglePayAddressToBillingAddress(paymentData, remoteBillingAddress.id);
+        const firstName = 'firstName';
```
vmparra, line 220: I will rather suggest to move this block to move this code to `_mapGooglePayAddressToBillingAddress` function.

Please include a tes case for such scenario too.

julioabarca2: ok

PR #1089, opened 2021-03-17
fix(provider): INT-3571 Google Pay [StripeV3] - Billing address is missing

julioabarca2:
## What?

The Billing address is missing when payment with google pay in Stripe V3

## Why?

We are validating the billing address when is retrieving for first time and throw a exception if does not exist

## Testing / Proof

Screen Shot 2021-03-12 at 5 53 49 p m

Ping @bigcommerce/payments

### DIFF
--- a/src/payment/strategies/googlepay/googlepay-payment-processor.ts
+++ b/src/payment/strategies/googlepay/googlepay-payment-processor.ts
@@ -156,18 +156,28 @@ export default class GooglePayPaymentProcessor {
     }
 
     private _mapGooglePayAddressToBillingAddress(paymentData: GooglePaymentData, id: string): BillingAddressUpdateRequestBody {
+        const firstName = paymentData.paymentMethodData.info.billingAddress.name.split(' ').slice(0, -1).join(' ');
+        const address1 =  paymentData.paymentMethodData.info.billingAddress.address1;
+        const city =  paymentData.paymentMethodData.info.billingAddress.locality;
+        const postalCode =  paymentData.paymentMethodData.info.billingAddress.postalCode;
+        const countryCode =  paymentData.paymentMethodData.info.billingAddress.countryCode;
+
+        if (!firstName || !address1 || !city || !postalCode || !countryCode) {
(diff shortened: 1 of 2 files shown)


julioabarca2:
Hi @vmparra I already did the changes

vmparra (approved):
LGTM 👍🏻

src/payment/strategies/googlepay/googlepay-payment-processor.spec.ts
```diff
@@ -232,10 +232,15 @@ describe('GooglePayPaymentProcessor', () => {
...
-            await processor.initialize('googlepay');
-            await processor.handleSuccess(googlePaymentDataMock);
+            try {
+                await processor.initialize('googlepay');
+                await processor.handleSuccess(googlePaymentDataMock);
 
-            expect(googlePayInitializer.parseResponse).toHaveBeenCalled();
+                expect(googlePayInitializer.parseResponse).toHaveBeenCalled();
+            } catch (error) {
+                expect(error).toBeInstanceOf(MissingDataError);
```
capsula4, line 241: using try/catch in a unit tests is discouraged because there's no guarantee the exception is actually being thrown.

instead, you should use
```js
expect(() => functionToTest()).toThrow(new MissingDataError(MissingDataErrorType.MissingBillingAddress))
```

julioabarca2: hi @capsula4 I did the changes. the unit tests where billing address does not exist, already exists in the file, so i removed the try catch, from the googlepay-payment-processor.spec.ts, i just modify the googlepay.mock.ts adding lastname to the "name" because its needed.

capsula4 (approved):

mmartinezintersys:
💚 TESTED

QA details:

 - Verify that an error is displayed to the shopper if the billing address in empty when checking out with Google Pay 
 - Sanity check for Auth Only and Authorize and Capture with Google Pay
 - Sanity check for Capture, Void & Refund with Google Pay

PR #1103, opened 2021-04-06
CHECKOUT-5740: Post to and receive messages from `www` subdomain

davidchin:
## What?
Post to and receive messages from `www` subdomain.

## Why?
Before this change, hosted payment forms do not load for stores that have WWW redirect setting set as "no preference" (meaning it doesn't do any kind of redirect to force shoppers to be on the `ShopPath` domain).

## Testing / Proof
Unit + manual

The test store is set up with WWW redirect setting as "no preference"

www_test

### DIFF
--- a/src/common/iframe/iframe-event-listener.ts
+++ b/src/common/iframe/iframe-event-listener.ts
@@ -1,4 +1,4 @@
-import { parseUrl } from '../url';
+import { appendWww, parseUrl } from '../url';
 import { bindDecorator as bind } from '../utility';
 
 import { IframeEventMap } from './iframe-event';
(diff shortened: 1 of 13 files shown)


src/common/iframe/iframe-event-listener.ts
```diff
@@ -75,7 +78,7 @@ export default class IframeEventListener<TEventMap extends IframeEventMap<keyof
 
     @bind
     private _handleMessage(event: MessageEvent): void {
-        if ((event.origin !== this._sourceOrigin) ||
+        if ((this._sourceOrigins.indexOf(event.origin) === -1) ||
```
capsula4, line 81: 🍹  can use `!this._sourceOrigins.includes(event.origin)`for better readability

src/common/iframe/iframe-event-poster.ts
```diff
@@ -30,18 +32,20 @@ export default class IframeEventPoster<TEvent> {
...
-        if (!this._targetWindow) {
+        if (!targetWindow) {
             throw new Error('Unable to post message because target window is not set.');
         }
 
         const result = options && fromEvent<MessageEvent>(window, 'message')
             .pipe(
                 filter(event =>
-                    event.origin === this._targetOrigin &&
+                    (this._targetOrigins.indexOf(event.origin) !== -1) &&
```
capsula4, line 48: same as above

capsula4 (approved):

src/common/iframe/iframe-event-listener.ts
davidchin: I just realised I can't do this because we don't polyfill `Array` for the NPM release. 😞

capsula4 (approved):

PR #1104, opened 2021-04-07
fix(payment): INT-4063 Avoid holding inventory if a payment intent co…

mauricio-sg:
…nfirmation fails for StripeV3

## What? [INT-4063](https://jira.bigcommerce.com/browse/INT-4063)
Allow reaching the backend when a payment intent confirmation fails. Allowing inventory to be returned to stock immediately.

## Why?
To avoid inventory be held for several minutes due to a failed payment.

## Testing / Proof
Watch the video >>[here](https://drive.google.com/file/d/14KQHRTmkRlvRHCNYJplAfcaLVNzelY9b/view?usp=sharing)<<

## Sibling PR
https://github.com/bigcommerce/bigpay/pull/3674

### DIFF
--- a/src/payment/strategies/stripev3/stripev3-payment-strategy.spec.ts
+++ b/src/payment/strategies/stripev3/stripev3-payment-strategy.spec.ts
@@ -28,7 +28,7 @@ import { getErrorPaymentResponseBody } from '../../payments.mock';
 import { StripeElement, StripeElements, StripeElementType, StripePaymentMethodType, StripeV3Client } from './stripev3';
 import StripeV3PaymentStrategy from './stripev3-payment-strategy';
 import StripeV3ScriptLoader from './stripev3-script-loader';
-import { getConfirmPaymentResponse, getFailingStripeV3JsMock, getPaymentMethodResponse, getStripeBillingAddress, getStripeBillingAddressWithoutPhone, getStripePaymentMethodOptionsWithGuestUserWithoutAddress, getStripeV3InitializeOptionsMock, getStripeV3InitializeOptionsMockSingleElements, getStripeV3JsMock, getStripeV3OrderRequestBodyMock, getStripeV3OrderRequestBodyVaultMock, getWrongPaymentMethodResponse, getWrongPaymentResponse } from './stripev3.mock';
+import { getConfirmPaymentResponse, getFailingStripeV3JsMock, getPaymentMethodResponse, getStripeBillingAddress, getStripeBillingAddressWithoutPhone, getStripePaymentMethodOptionsWithGuestUserWithoutAddress, getStripeV3InitializeOptionsMock, getStripeV3InitializeOptionsMockSingleElements, getStripeV3JsMock, getStripeV3OrderRequestBodyMock, getStripeV3OrderRequestBodyVaultMock } from './stripev3.mock';
 
 describe('StripeV3PaymentStrategy', () => {
     let checkoutMock: Checkout;
@@ -532,6 +532,46 @@ describe('StripeV3PaymentStrategy', () => {
                     }
                 });
 
+                it('throws an error that is not a RequestError', async () => {
+                    const errorResponse = new Error();
+
+                    jest.spyOn(paymentActionCreator, 'submitPayment')
+                        .mockReturnValue(of(createErrorAction(PaymentActionType.SubmitPaymentFailed, errorResponse)));
+
+                    await strategy.initialize(options);
+
+                    await expect(strategy.execute(getStripeV3OrderRequestBodyMock())).rejects.toThrow(Error);
+
+                    expect(orderActionCreator.submitOrder).toHaveBeenCalled();
+                    expect(paymentActionCreator.submitPayment).toHaveBeenCalledTimes(1);
+                    expect(stripeV3JsMock.createPaymentMethod).toHaveBeenCalled();
+                });
+
+                it('throws when user closes the auth modal or auth fails', async () => {
+                    const errorResponse = new RequestError(getResponse({
+                        ...getErrorPaymentResponseBody(),
+                        errors: [
+                            {code: 'three_d_secure_required'},
+                        ],
+                        three_ds_result: {
+                            token: 'token',
+                        },
+                        status: 'error',
+                    }));
+
+                    jest.spyOn(paymentActionCreator, 'submitPayment')
+                        .mockReturnValueOnce(of(createErrorAction(PaymentActionType.SubmitPaymentFailed, errorResponse)));
+
+                    stripeV3JsMock.handleCardAction = jest.fn(() => Promise.resolve({ error: 'foo' }));
+
+                    await strategy.initialize(options);
+                    await strategy.execute(getStripeV3OrderRequestBodyVaultMock());
+
+                    expect(orderActionCreator.submitOrder).toHaveBeenCalled();
+                    expect(paymentActionCreator.submitPayment).toHaveBeenCalledTimes(2);
+                    expect(stripeV3JsMock.handleCardAction).toHaveBeenCalled();
+                });
+
                 it('throws when stored instrument requires SCA and then shopper successfully authenticates', async () => {
                     const errorResponse = new RequestError(getResponse({
                         ...getErrorPaymentResponseBody(),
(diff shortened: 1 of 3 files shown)


src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -159,7 +121,24 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
         return Promise.resolve(this._store.getState());
     }
 
-    private async _processAdditionalAction(error: StripeAdditionalActionError, methodId: string, shouldSaveInstrument: boolean): Promise<any> {
+    private _buildPaymentPayload(methodId: string, formattedPayload: any, shouldSetAsDefaultInstrument: boolean): any {
```
clopezh, line 124: Maybe the return type is Payment? since the sign of submitPayment(payment: Payment)

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -159,7 +121,24 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
...
+
+        return { methodId, paymentData };
+    }
+
+    private async _processAdditionalAction(
+        error: StripeAdditionalActionError,
+        methodId: string,
+        shouldSaveInstrument: boolean,
+        shouldSetAsDefaultInstrument: boolean
+    ): Promise<any> {
```
clopezh, line 137: it cannot be Promise<any>, maybe you can find a proper type

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -248,27 +223,23 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
             }
 
             if (error) {
-                reject(error);
+                return reject(error);
```
clopezh, line 226: Is this return needed? at the end you have this code wrapped by the Promise, so if you reject a response, then the Promise understand that is a returned response

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -248,27 +223,23 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
...
             }
 
-            if (paymentMethod.method === StripeElementType.CreditCard) {
-                resolve(stripePaymentMethod);
+            if (stripePaymentMethod && paymentMethod.method === StripeElementType.CreditCard) {
+                return resolve(stripePaymentMethod);
```
clopezh, line 230: Same comment as above

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -248,27 +223,23 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
...
             }
 
-            resolve(paymentIntent);
+            if (paymentIntent) {
+                return resolve(paymentIntent);
```
clopezh, line 234: same comment as above about the return

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -248,27 +223,23 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
...
+            }
         });
     }
 
-    private async _confirmVaultedPayment(clientSecret: string): Promise<PaymentIntent> {
-        return new Promise(async (resolve, reject) => {
-            const { error, paymentIntent } = await this._getStripeJs().handleCardAction(clientSecret);
+    private async _handleCustomerAction(clientSecret: string): Promise<PaymentIntent> {
+        const { paymentIntent } = await this._getStripeJs().handleCardAction(clientSecret);
```
clopezh, line 240: Is it not possible that this call answers an error as before?

src/payment/strategies/stripev3/stripev3-payment-strategy.spec.ts
```diff
@@ -575,48 +575,6 @@ describe('StripeV3PaymentStrategy', () => {
                     expect(stripeV3JsMock.confirmCardPayment).not.toHaveBeenCalled();
                 });
 
-                it('throws an error when paymentIntent.id is not set properly into payload and is Vaulted', async () => {
```
clopezh, line 578: I see that you remove some tests, and you add new code, why did not you add a tests that covers the new code?

clopezh (changes requested):
Hi @mauricio-sg I have some questions/comments, let me know if you have a question with this

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -55,16 +57,15 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
 
     async execute(orderRequest: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
         const { payment, ...order } = orderRequest;
-        let paymentResult: PaymentIntent | StripePaymentMethod;
-        let paymentPayload: any;
+        let formattedPayload: any;
```
vmparra, line 60: Any?

Maybe you can at least use something like `let formattedPayload: { }`; But I would rather recommend to create an interface for it.

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
vmparra, line 124: It's better to avoid the use `any` instead you can use `unknown` as it is less permissive

vmparra (changes requested):
Hi @mauricio-sg So far it looks real solid, a few minor comments. Please let me know if you have questions.

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
mauricio-sg: It's a good practice. This is better explained here: https://stackoverflow.com/questions/32536049/do-i-need-to-return-after-early-resolve-reject/

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
mauricio-sg: That's the PR's point. I removed on purpose the logic you mention because we need to reach the backend when an error happens in order to return the inventory.

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
mauricio-sg: If you don't mind, we'll raise a new ticket to address this because `formattedPayload` has a strange structure and needs further investigation and testing. 😬 @cbandam

clopezh (approved):
LGTM 👍🏻  Thanks to addressed my comments

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
clopezh: I think Victor referred in a create an specific type for this scenario, not for all formattedPayloads, maybe something like StripeV3FormattedPayload or some similar of that, Or Do I have a different idea @vmparra  ?

vmparra (approved):
LGTM 👍🏻

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -248,27 +224,23 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
...
 
             if (error) {
-                reject(error);
+                return reject(error);
             }
 
-            if (paymentMethod.method === StripeElementType.CreditCard) {
-                resolve(stripePaymentMethod);
+            if (stripePaymentMethod && paymentMethod.method === StripeElementType.CreditCard) {
+                return resolve(stripePaymentMethod);
```
capsula4, line 231: I just realised this method is `async`, so you don't really need to wrap it inside a Promise and use `resolve/reject`. Async methods automatically wrap the returned value in a promise. If you want to reject a value, you should throw an exception.

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -159,7 +122,24 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
...
+    }
+
+    private async _processAdditionalAction(
+        error: StripeAdditionalActionError,
+        methodId: string,
+        shouldSaveInstrument: boolean,
+        shouldSetAsDefaultInstrument: boolean
+    ): Promise<InternalCheckoutSelectors | never> {
+        if (!(error instanceof RequestError)) {
+            return Promise.reject(error);
```
capsula4, line 140: `throw error;` is preferred.

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -73,79 +75,40 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
...
-                        formattedPayload: {
-                            credit_card_token: {
-                                token: paymentResult.id,
-                            },
-                            confirm: true,
-                        },
-                    },
-                };
-            }
+            // tslint:disable-next-line: variable-name
```
capsula4, line 78: why is this required?

capsula4:
Looking good, just some minor syntactic issues.

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
mauricio-sg: Because linter complains about `credit_card_number_confirmation`
and `verification_value`. Do you want me to remove it?

mauricio-sg:
@capsula4 ♻️ 🙇

capsula4 (approved):

JoseBobadilla20:
💚 Tested
QA Details: I ran a regression for Stripe V3 to validate empty payment notification to check that works correctly.

PR #1105, opened 2021-04-08
feat(payment): PAYPAL-976 Order placement v2

vneutrino:
## What?
Added Paypal APM fields on checkout

## Why?
Due to https://jira.bigcommerce.com/browse/PAYPAL-976

## Testing / Proof
Tested manually
image

### DIFF
--- a/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.spec.ts
+++ b/src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.spec.ts
@@ -5,6 +5,7 @@ import { EventEmitter } from 'events';
 import { Cart } from '../../../cart';
 import { getCart } from '../../../cart/carts.mock';
 import { NotImplementedError } from '../../../common/error/errors';
+import { PaymentMethodClientUnavailableError } from '../../errors';
 
 import { ButtonsOptions, ParamsRenderHostedFields, PaypalCommerceHostedFields, PaypalCommerceHostedFieldsApprove, PaypalCommercePaymentProcessor, PaypalCommerceRequestSender, PaypalCommerceScriptLoader, PaypalCommerceScriptParams, PaypalCommerceSDK } from './index';
 import { getPaypalCommerceMock } from './paypal-commerce.mock';
@@ -23,6 +24,7 @@ describe('PaypalCommercePaymentProcessor', () => {
     let cart: Cart;
     let containers: HTMLElement[];
     let render: () => void;
+    let renderApmFields: (container: string) => void;
     let submit: () => (PaypalCommerceHostedFieldsApprove);
     let orderID: string;
     let fundingSource: string;
@@ -72,7 +74,12 @@ describe('PaypalCommercePaymentProcessor', () => {
         jest.spyOn(paypalCommerceRequestSender, 'setupPayment')
             .mockImplementation(jest.fn().mockReturnValue(Promise.resolve({ body: orderID })));
 
-        render = jest.spyOn(paypal, 'Buttons')
+        renderApmFields = jest.fn();
+        jest.spyOn(paypal, 'Fields')
+            .mockImplementation(jest.fn().mockReturnValue({ render: renderApmFields }));
+
+        render = jest.fn();
+        jest.spyOn(paypal, 'Buttons')
             .mockImplementation((options: ButtonsOptions) => {
                 eventEmitter.on('onClick', () => {
                     if (options.onClick) {
@@ -93,7 +100,7 @@ describe('PaypalCommercePaymentProcessor', () => {
                 });
 
                 return {
-                    render: () => {},
+                    render,
                     isEligible: () => true,
                 };
             });
@@ -174,7 +181,7 @@ describe('PaypalCommercePaymentProcessor', () => {
             await paypalCommercePaymentProcessor.initialize(initOptions);
             await paypalCommercePaymentProcessor.renderButtons(cart.id, 'container');
 
-            expect(render).toHaveBeenCalled();
+            expect(render).toHaveBeenCalledWith('container');
         });
 
         it('throws error if unable to setting PayPalCommerce button', async () => {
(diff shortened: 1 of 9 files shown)


src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
```diff
@@ -59,13 +59,45 @@ export interface PaypalButtonStyleOptions {
 export interface ButtonsOptions {
     style?: PaypalButtonStyleOptions;
     fundingSource?: string;
-    createOrder?(): Promise<string>;
+    createOrder?(_: any, actions: any): Promise<string>;
```
andriiVitvitskyi1990, line 62: Is there any way to specify type instead of any?

andriiVitvitskyi1990 (approved):

src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
vneutrino: fixed

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -96,12 +107,32 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
         this._paypalCommercePaymentProcessor.renderButtons(cartId, container, buttonParams, {
             onRenderButton,
-            fundingKey: this._paypalCommerceFundingKeyResolver.resolve(methodId, gatewayId),
+            fundingKey,
             paramsForProvider: {isCheckout: true},
         });
 
+        if (this._isAPM)  {
+            /**
+             * FALLBACK for old checkouts that didn't pass apmFieldsContainer id
```
BC-ACherednichenko, line 120: I think we should go with the existing approach (when the payment widget is next to the radio button) https://github.com/bigcommerce/checkout-js/blob/051363765fb15c03d818dcb09d0668be574f490a/src/app/payment/paymentMethod/HostedWidgetPaymentMethod.tsx#L269, rather than inject widget before the button container

src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts
```diff
@@ -96,12 +107,32 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+             */
+            const apmFieldsContainerId = apmFieldsContainer || `${container}apmFieldsContainer`;
+            const fullName = `${firstName} ${lastName}`;
+
+            if (!document.querySelector(apmFieldsContainerId)) {
+                const apmFieldsContainerElement = document.createElement('div');
+                apmFieldsContainerElement.setAttribute('id', apmFieldsContainerId.substring(1));
+
+                const buttonContainer = document.querySelector(container);
+                buttonContainer?.insertBefore(apmFieldsContainerElement, buttonContainer.firstChild);
```
BC-ACherednichenko, line 130: I think we should go with the existing approach (when the payment widget is next to the radio button) https://github.com/bigcommerce/checkout-js/blob/051363765fb15c03d818dcb09d0668be574f490a/src/app/payment/paymentMethod/HostedWidgetPaymentMethod.tsx#L269, rather than inject widget before the button container

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.spec.ts
```diff
@@ -249,6 +256,60 @@ describe('PaypalCommercePaymentProcessor', () => {
         });
     });
 
+    describe('render apmFields', () => {
+        it('setting PaypalCommerce apmFields without styles and fields predifined values', async () => {
```
capsula4, line 260: for test names, it's good to read them some as a sentence, e.g.
`it('sets PaypalCommerce')` as in `it sets PaypalCommerce`

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.spec.ts
```diff
@@ -249,6 +256,60 @@ describe('PaypalCommercePaymentProcessor', () => {
...
+        });
+
+        it('render PaypalCommerce apmFields', async () => {
+            await paypalCommercePaymentProcessor.initialize(initOptions);
+            await paypalCommercePaymentProcessor.renderFields('#fieldsContainer', 'P24');
+
+            expect(renderApmFields).toHaveBeenCalledWith('#fieldsContainer');
+        });
+
+        it('throws error if unable to setting PayPalCommerce fields', async () => {
```
capsula4, line 285: minor typo, `if unable to set`

src/payment/strategies/paypal-commerce/paypal-commerce-payment-processor.ts
```diff
@@ -85,6 +87,36 @@ export default class PaypalCommercePaymentProcessor {
         return this._paypalButtons;
     }
 
+    renderFields(apmFieldsContainer: string, fundingKey: keyof PaypalCommerceSDKFunding, apmFieldsStyles?: PaypalFieldsStyleOptions, fullName?: string, email?: string): PaypalCommerceFields {
```
capsula4, line 90: this method seems to have a bit too many fields. I would recommend creating an interface
`RenderApmFieldsParams` and pass a single object rather than a lot of arguments.

you could then do something like

```js
renderFields({
  apmFieldsContainer,
  fundingKey,
  apmFieldsStyles,
  ...
}: RenderApmFieldsParams): PaypalCommerceFields

File: src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts (hunk 2 above)
Comment (capsula4) [inline src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts:120]: rather than adding this comment, I would recommend adding a unit test not passing `apmFieldsContainer` param, which would ensure the logic is not accidentally removed.

File: src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts (hunk 3 above)
Comment (capsula4) [reply src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts:130]: what if `buttonContainer` is empty? Should we be throwing an error? Otherwise, it might just fail silently and we won't have any visibility.

**Review by capsula4 — COMMENTED** [review]
Code mostly looking good. I would recommend more descriptive commit messages and also PR descriptions, especially considering that commit messages are what we export to the changelog, and that's what SDK users check to see what's new. 

https://github.com/bigcommerce/checkout-sdk-js/blob/master/CHANGELOG.md

File: src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts (hunk 3 above)
Comment (vneutrino) [reply src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts:130]: To be 100% sure, I’ll add check if there is container in DOM on initialization of PAYPAL. To avoid problem with button and fields initialization.

File: src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts (hunk 2 above)
Comment (vneutrino) [reply src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts]: There was unit test for this case :)

**Review by capsula4 — APPROVED** [review]
LGTM, thanks for addressing the feedback! 🙇  
🚢  it

File: src/payment/strategies/paypal-commerce/paypal-commerce-payment-strategy.ts [hunk 7]
```diff
@@ -96,12 +107,39 @@ export default class PaypalCommercePaymentStrategy implements PaymentStrategy {
...
+            fundingKey,
             paramsForProvider: {isCheckout: true},
         });
 
+        if (this._isAPM)  {
+            const apmFieldsContainerId = apmFieldsContainer || `${container}apmFieldsContainer`;
+            const fullName = `${firstName} ${lastName}`;
+
+            if (!document.querySelector(apmFieldsContainerId)) {
+                const apmFieldsContainerElement = document.createElement('div');
```
BC-ACherednichenko, line 128: Per our discussion, it is worthy to remove behavior that implicitly creates a container in the case when it is not given in options. The APM feature should be fully controllable by the options of strategy and in the case if consumers with custom checkout update `checkout-SDK, they will need to follow our docs in order to make APM feature work in their custom checkout implementation

PR #1108, opened 2021-04-12
[POC] PPSDK Pseudo-code integration

leeBigCommerce:
**N.B. Not meant for merging, ever**

## What?

- Pseudo code of a possible PPSDK solution

## Why?

- In-situ code to spur conversation and evaluate direction

## Testing / Proof

- None, this should never be merged


leeBigCommerce:
It's not super clear from the code, so here's an illustration of example payment method objects that would trigger each strategy:

```

// Example 1: A payment method which will init the SDK HostedFormStrategy
{
    gateway: 'SDK',
    method: 'card',
    type: 'PAYMENT_TYPE_API',
    // ...plus other properties that don't impact strategy selection
}

// Example 2: A payment method which will init the SDK Strategy
{
    gateway: 'SDK',
    method: 'anything_other_than_card',
    type: 'PAYMENT_TYPE_API',
    // ...plus other properties that don't impact strategy selection
}

// Example 3: A payment method which will init a future 'non-standard' SDK Strategy
{
    gateway: 'SDK-CUSTOM',
    method: 'something_specific_to_this_provier_and_method',
    type: 'PAYMENT_TYPE_API',
    // ...plus other properties that don't impact strategy selection
}
```

Does the job, but still feels a little clunky to me - suggestions welcome!

hx (commented):
Thanks @leeBigCommerce for putting this together. The FE side of this project has mostly felt abstract and theoretical until now. It's really nice to have a glance into how it could all fit together.

There are quite a few aspects that I simply don't understand as a non-FE person. I have a couple of comments in relation to the overall process, which I trust you can filter through the more accepted FE mannerisms for them to make sense.

As usual I'm happy to follow up on Slack or a call if there's more to discuss than a few comments allow.

src/payment/payment-strategy-registry.ts
```diff
@@ -33,6 +33,29 @@ export default class PaymentStrategyRegistry extends Registry<PaymentStrategy, P
...
+                which have their own client side code?
+
+            We don't want these accidentally falling into either of these two strategies
+        */
+        if (paymentMethod.gateway === 'SDK') {
+
+            if (paymentMethod.method === 'card') {
+                return PaymentStrategyType.SDK_HOSTED;
+            } else {
+                // N.B. this is dangerous, we need an escape hatch for non standard SDK methods
```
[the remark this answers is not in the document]
leeBigCommerce: Moving to a singular strategy is definitely doable, we'd just need to move away from using the current `strategy-registry`.

If I'm reading the current code right (please correct me @davidchin @capsula4), the payment method that comes over the wire right now doesn't make it _into_ the execution scope of strategies. It is only used by the registry to _pick_ the strategy.

In which case, we'd need to introduce an SDK/non-SDK branch earlier than I've currently got it (maybe here https://github.com/bigcommerce/checkout-sdk-js/blob/67b1e0a818f16ba94b729683ea31362907980fad/src/payment/payment-strategy-action-creator.ts#L48 ?)

That way we could pick a singular `SDK` strategy, while still passing in the Payment Method object (or key bits of it).
Details from the Payment Method object could then be used to pick the correct initialiser.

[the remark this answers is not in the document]
leeBigCommerce: My original comment about this bit being 'dangerous' was more pointing out that we don't to accidentally pick the wrong strategy/initialisation by accident. Right now I believe we fallback to standard Credit Card behaviour if `checkout-js` doesn't recognised the Payment Method coming over the wire.

src/payment/strategies/sdk/handler/handler.ts
```diff
@@ -0,0 +1,52 @@
...
+import { handleFailedResponse, isFailedResponse } from './failed';
+import { handleSuccessResponse, isSuccessResponse } from './success';
+
+/*
+    This will likely need some local state
+    for storing payment method data only available at time of instantiation
+    so will probably need to become a class
+*/
+
+export const handler = async (response: Response): Promise<void> => {
```
davidchin, line 13: I'd probably make those functions injectable so they can unit tested without having to use `jest.mock` to hijack `import/require`.

src/payment/strategies/sdk/hosted-form-strategy.ts
```diff
@@ -0,0 +1,97 @@
...
+export class HostedFormStrategy implements PaymentStrategy {
+    protected _hostedForm?: HostedForm;
+
+    constructor(
+        protected _store: CheckoutStore,
+        protected _orderActionCreator: OrderActionCreator,
+        protected _hostedFormFactory: HostedFormFactory
+    ) {}
+
+    initialize(options?: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
```
davidchin, line 49: How come for CC methods, we don't need to do `POST /payment/provider.method` to know we need to show a CC form. But we do with other methods, like redirect?

src/payment/strategies/sdk/handler/handler.ts
```diff
@@ -0,0 +1,52 @@
...
+            which destroys our JS runtime
+
+            However, all others won't
+
+            As such they could be infinitely chained
+
+            To handle this, our handleContinueResponse function will need
+            to take a reference for the parent handler
+            and be coded for recursion
+            (possibly using observables?)
```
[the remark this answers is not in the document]
davidchin: I imagine it'd just call the `handler` function again. The `handler` will have an exit condition, i.e.: there are no more steps or the transaction is aborted etc...

src/payment/strategies/sdk/strategy.ts
```diff
@@ -0,0 +1,76 @@
...
+        if (!options?.gatewayId || !options?.methodId) {
+            // throw? reject?
+            return Promise.reject();
+        }
+
+        const { gatewayId: paymentProvider, methodId } = options;
+
+        const jsonBody = JSON.stringify(payload);
+
+        const initialResponse = await fetch(`/payment/${paymentProvider}.${methodId}`, {
```
davidchin, line 62: In this the new payment endpoint? If so, does it require an order to be created first? If it is not the new payment endpoint, what is it? And at what point do we call the new payment endpoint to start/continue a payment transaction?

davidchin (commented):
Thank you @leeBigCommerce for putting this together to start the discussion!

src/payment/strategies/sdk/hosted-form-strategy.ts
leeBigCommerce: Here we were making that decision based off the `paymentMethod.method === 'card'` check that had already occurred in order to choose this strategy `PaymentStrategyType.SDK_HOSTED`.

Might be a moot point after I try unifying the strategies, stay tuned 😄

src/payment/strategies/sdk/handler/handler.ts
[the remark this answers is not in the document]
leeBigCommerce: Observables are a part of `rxjs` which is already used throughout `checkout-sdk` for handling events. I wasn't sure if the checkouters would prefer we made use of them here.

But it sounds like we're happy just to pass around the `handler` function and deal with the recursion that way anyhow for now.

src/payment/strategies/sdk/strategy.ts
leeBigCommerce: This is a new endpoint, one that would allow us to get the next bit of orchestration from the BE. (https://github.com/bigcommerce/api-schema/pull/446#discussion_r603002341)

You're right, it would need an order to be created first.

I think I may be misusing it here, will refactor into a singular Strategy and take another look

src/payment/payment-strategy-registry.ts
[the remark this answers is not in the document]
leeBigCommerce: @hx @davidchin 

This PR has now been updated to use a singular SDK Strategy, please take a look when you get chance

In terms of how integrators bring their own JS to the party, I think we're right to move that discussion out of scope if we can. Though please do call out if we see any potential blockers in extending this solution in that direction in future.

src/payment/strategies/sdk/hosted-form-strategy.ts
leeBigCommerce: The non-hosted 'branch' has been effectively removed from this PR, hopefully making the comparison moot and this bit of conversation resolvable 😄

src/payment/strategies/sdk/strategy.ts
leeBigCommerce: The non-hosted 'branch' has been effectively removed from this PR, hopefully making the comparison moot and this bit of conversation resolvable 😄

src/payment/strategies/sdk/strategy.ts
```diff
@@ -0,0 +1,155 @@
...
+        /*
+            This is actually called first, before initialize
+            We can check if there was a payment left for us to continue
+            i.e. at the return of a redirect
+
+            N.B. we will need a different mechanism for continue steps in the future
+            'finalize' runs and dead-ends prior to rendering payment methods
+            some future steps will want to run with the method choice visually open
+            we'll likely need a new 'continueOrderIfNeeded' feature in checkout-js
+        */
```
icatalina, line 118: Considering how these new methods operate I don't think we could use `finalize` in the way it was intended previously, we'll just have to instantiate the same method and keep running `continue`s... 🤷

We might be able to change finalize to allow it to keep running and be just some sort of _pick it up where it was left_ for SDK providers... 😕

leeBigCommerce: Yeah, this is bending `finalize` purpose quite a bit. The plan is to use it only for phase 2.1, then to make an alternative lifecycle method within `checkout-js` that is better suited to 'continuing' and use that instead.

src/payment/strategies/sdk/handler/handler.ts
```diff
@@ -0,0 +1,52 @@
...
+import { handleContinueResponse, isContinueResponse } from './continue';
+import { handleErrorResponse } from './error';
+import { handleFailedResponse, isFailedResponse } from './failed';
+import { handleSuccessResponse, isSuccessResponse } from './success';
+
+/*
+    This will likely need some local state
+    for storing payment method data only available at time of instantiation
+    so will probably need to become a class
+*/
```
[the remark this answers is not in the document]
leeBigCommerce: Yep 👍

src/payment/strategies/sdk/strategy.ts
```diff
@@ -0,0 +1,155 @@
...
+    protected _paymentProcessor?: PaymentProcessor;
+
+    constructor(
+        protected _store: CheckoutStore,
+        protected _orderActionCreator: OrderActionCreator
+    ) {}
+
+    async initialize(options?: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
+        if (!options || !options.methodId) {
+            // TODO: Can we make this object and attribute mandatory?
```
capsula4, line 40: this is mandatory but it's always provided automatically by the registry, so there's nothing to really about at this stage. If you are talking about not making it mandatory at the entry point (checkoutService.submitOrder), how would the SDK be aware of what payment strategy it should use?

src/payment/strategies/sdk/strategy.ts
```diff
@@ -0,0 +1,155 @@
...
+        /*
+            TODO: probably too brittle to just use methodId going forward
+
+            if methodId is 'card', how do we differentiate between standard and "has own JS" implementations?
+
+            N.B. looking for a "has own JS" initializer, not finding one,
+            then falling back to a default one for that methodId
+            could produce some unpredictable (and likely broken) UX during execute
+        */
+        const initializer = initializers[methodId];
```
capsula4, line 55: rather than have a different range of initializer, maybe it's worth having different PPSDK strategies for each "type" of PPSDK payment.

We could have a `methodId: 'PPSDK'` and `submethodId: 'card'`. This way if methodId is PPSDK, it could have it's own `PPSDKStrategyRegistry` (if we foresee many different strategies will be required)

src/payment/strategies/sdk/strategy.ts
```diff
@@ -0,0 +1,155 @@
...
+
+        // Submitting the order should let us restore it post redirect in finalize
+        const submitOrder = () => this._store.dispatch(this._orderActionCreator.submitOrder(order, options));
+        const getPaymentStep = () => paymentProcessor(payment);
+        const loadOrder = () => this._store.dispatch(this._orderActionCreator.loadCurrentOrder());
+
+        // TODO: move to async/await format?
+        return submitOrder()
+            // Very key differences to a typical strategy:
+            .then(getPaymentStep)
```
capsula4, line 103: I wonder if there could be multiple payments step, in which case after submitting the order, you might need some sort of recursive call that processes all the steps until it reaches the end.

Basically something like
```js
function processStep() {
  step = getPaymentStep();
  if (handler(step)) {
    return;
  } else {
    processStep();
  }

}

File: src/payment/strategies/sdk/handler/continue/continue.ts [hunk 11]
```diff
@@ -0,0 +1,35 @@
...
+    /*
+        Future considerations:
+
+        Some future continue steps may need to control things outside of this execution scope
+        either still inside checkout-sdk's codebase
+        or also outside in checkout-js's codebase
+
+        currently checkout-sdk doesn't know checkout-js exists (orchestration runs the other way)
+
+        for either scenario we're probably going to need some kind of pub/sub/eventing system
```
capsula4, line 26: Currently, Checkout JS can subscribe via the Checkout SDK to changes in the checkout state. That could be sufficient to start with? However, if it's related to how a payment should be processed, maybe a more specific mechanism would be more appropriate (which could still be based on the current `subscribe` mechanism).

src/payment/strategies/sdk/strategy.ts
leeBigCommerce: I had started a strategies per initialisation type, much like you're suggesting. Then the argument was made for unifying them into a singular SDK strategy https://github.com/bigcommerce/checkout-sdk-js/pull/1108#discussion_r612884341 

As you point out, this does lead to a split then being needed at the 'initialiser' level of that singular strategy instead.

The downside being that the singular strategy is even more multifaceted.
The plus side being that we don't have a bunch of strategies that differ only at this initialisation step. 

I think it's something we'll need to keep a close eye on as we start adding real code and real tests.

You're right about the likely need for a registry either way 👍

src/payment/strategies/sdk/strategy.ts
leeBigCommerce: I was wanting to make it mandatory (right now the type has it as optional: https://github.com/bigcommerce/checkout-sdk-js/blob/67b1e0a818f16ba94b729683ea31362907980fad/src/payment/strategies/payment-strategy.ts#L10 )

But I guess if other consumers are already using it without passing in options, I might be stuck with my "not undefined" checks. 

I think I'll organise a meeting to clear up what this proposal needs from the Payment Method object, as the use of `gateway` vs `provider` vs `type` is all a little confusing.

src/payment/strategies/sdk/strategy.ts
leeBigCommerce: Yep, we'll definitely have something like that to handle the multiple steps.

https://github.com/bigcommerce/checkout-sdk-js/pull/1108/files#diff-34830b76d352231339db788439c1039d6d412c2ee859dca4b4f77bfbf09703bcR42

The current thinking is that the 'handler' would pass itself to any function that carries out a step.
So that if/when that step's executions ends in another response from a processing endpoint, that response is passed once again to the handler.

I guess from the point of the Strategy file, this recursion is hidden within the internals of the handler.

src/payment/strategies/sdk/handler/continue/continue.ts
leeBigCommerce: Yeah, I can imagine quite a bit of research is going to be needed once we come to that phase.

What I wanted to make sure with this comment is that I'm not doing anything to impede such mechanisms when we later come to add them.

leeBigCommerce:
Closed as this POC PR has done it's job

PR #1112, opened 2021-04-14
Revert "feat(payment): INT-3702 added vaulted cc for Mollie"

TomA-R:
Reverts bigcommerce/checkout-sdk-js#1056

Seems to break checkout-js - see https://github.com/bigcommerce/checkout-js/pull/554


capsula4 (approved):
Apparently, this introduced a breaking change, requiring a containerId. It should have been optional and support backwards compatibiltiy.

PR #1114, opened 2021-04-15
feat(cart): CHECKOUT-5747 Add cart changed error data

capsula4:
## What?
Provide the previous and updated cart as part of `CartChangedError`

## Why?
So that we can log the contents in case of issues to related to this error.

## Testing / Proof
unit

Tested also in combination with Checkout JS changes
Screen Shot 2021-04-15 at 3 09 48 pm

### DIFF
--- a/src/checkout/checkout-store-error-selector.ts
+++ b/src/checkout/checkout-store-error-selector.ts
@@ -1,5 +1,6 @@
 import { memoizeOne } from '@bigcommerce/memoize';
 
+import { CartChangedError } from '../cart/errors';
 import { RequestError } from '../common/error/errors';
 import { createSelector, createShallowEqualSelector } from '../common/selector';
 import { Omit } from '../common/types';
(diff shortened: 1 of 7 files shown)


src/checkout/checkout-store-error-selector.ts
```diff
@@ -36,7 +37,7 @@ export default interface CheckoutStoreErrorSelector {
      *
      * @returns The error object if unable to submit, otherwise undefined.
      */
-    getSubmitOrderError(): Error | undefined;
+    getSubmitOrderError(): Error | CartChangedError | undefined;
```
capsula4, line 40: this was done in order to get the type `CartChangedError` exported in `checkout-sdk.d.ts`

davidchin (approved):
Thanks Luis. I think the filtering should be done on the logging side though (UCO). Because the data you’re throwing here can be used for things other than logging. Just that some data can’t be logged so the condition should be applied there instead.

PR #1118, opened 2021-04-22
feat(payments): INT-4150 changing zip strategy flow

deov31:
Parent ticket: [INT-3825](https://jira.bigcommerce.com/browse/INT-3825)
[INT-4150](https://jira.bigcommerce.com/browse/INT-4150)

## What?
In order to implement new refactor for zip flow I added a new validation to submit payment and let backend handles the verification of payment status.

## Why?
Due to security concerns previously discussed this refactor is needed.

## Testing / Proof
Demo:
https://drive.google.com/file/d/1IJaeEQxmunYbUaI9XTNmNsTqmQMER_fU/view?usp=sharing

Screen Shot 2021-04-20 at 22 06 52

Sibling PR CheckoutJS: https://github.com/bigcommerce/checkout-js/pull/560

### DIFF
--- a/src/payment/strategies/zip/zip-payment-strategy.spec.ts
+++ b/src/payment/strategies/zip/zip-payment-strategy.spec.ts
@@ -1,5 +1,5 @@
 import { createClient as createPaymentClient } from '@bigcommerce/bigpay-client';
-import { createAction, Action } from '@bigcommerce/data-store';
+import { createAction, createErrorAction, Action } from '@bigcommerce/data-store';
 import { createRequestSender, RequestSender } from '@bigcommerce/request-sender';
 import { createScriptLoader, ScriptLoader } from '@bigcommerce/script-loader';
 import { of, Observable } from 'rxjs';
@@ -7,7 +7,8 @@ import { of, Observable } from 'rxjs';
 import { getCartState } from '../../../cart/carts.mock';
 import { createCheckoutStore, Checkout, CheckoutRequestSender, CheckoutStore, CheckoutValidator } from '../../../checkout';
 import { getCheckout, getCheckoutState } from '../../../checkout/checkouts.mock';
-import { NotInitializedError } from '../../../common/error/errors';
+import { NotInitializedError, RequestError } from '../../../common/error/errors';
+import { getResponse } from '../../../common/http-request/responses.mock';
 import { getConfigState } from '../../../config/configs.mock';
 import { getCustomerState } from '../../../customer/customers.mock';
 import { OrderActionCreator, OrderActionType, OrderRequestBody } from '../../../order';
@@ -25,6 +26,7 @@ import { PaymentActionType } from '../../payment-actions';
 import PaymentMethodRequestSender from '../../payment-method-request-sender';
 import { PaymentInitializeOptions } from '../../payment-request-options';
 import PaymentRequestTransformer from '../../payment-request-transformer';
+import { getErrorPaymentResponseBody } from '../../payments.mock';
 import PaymentStrategy from '../payment-strategy';
 
 import { Zip } from './zip';
@@ -102,8 +104,7 @@ describe('ZipPaymentStrategy', () => {
             .mockReturnValue(submitPaymentAction);
         jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod')
             .mockResolvedValue(store.getState());
-        jest.spyOn(store, 'dispatch')
-            .mockResolvedValue(store.getState());
+        jest.spyOn(store, 'dispatch');
         jest.spyOn(store.getState().paymentMethods, 'getPaymentMethod')
             .mockReturnValue(paymentMethodMock);
         jest.spyOn(store.getState().checkout, 'getCheckoutOrThrow')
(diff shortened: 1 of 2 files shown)


src/payment/strategies/zip/zip-payment-strategy.spec.ts
```diff
@@ -102,8 +104,7 @@ describe('ZipPaymentStrategy', () => {
             .mockReturnValue(submitPaymentAction);
         jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod')
             .mockResolvedValue(store.getState());
-        jest.spyOn(store, 'dispatch')
-            .mockResolvedValue(store.getState());
+        jest.spyOn(store, 'dispatch');
```
mauricio-sg, line 107: 😭 👍

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -75,6 +75,21 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
             throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
         }
 
+        if (this._redirectFlowIsTrue()) {
+            const nonce = JSON.parse(this._paymentMethod.clientToken);
+            await this._prepareForReferredRegistration(payment.methodId, nonce.id);
+            try {
```
mauricio-sg, line 81: 🍹 
```suggestion
            let nonce: { id: string };
            try {
                nonce = JSON.parse(this._paymentMethod.clientToken);
            } catch (error) {
                throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
            }

            await this._prepareForReferredRegistration(payment.methodId, nonce.id);

            try {
                ...
```

mauricio-sg (approved):
LGTM 👍

src/payment/strategies/zip/zip-payment-strategy.ts
deov31: Thanks!

JoseBobadilla20:
💚 Tested
QA Details: I ran a regression for Zip, to check that everything works correctly.

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -75,6 +75,28 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
...
+                throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+            }
+
+            await this._prepareForReferredRegistration(payment.methodId, nonce.id);
+
+            try {
+                return await this._store.dispatch(this._paymentActionCreator.submitPayment({methodId: payment.methodId, paymentData: { nonce: nonce.id }}));
+            } catch (error) {
+                if (error instanceof RequestError && error.body.status === 'additional_action_required') {
+                    return new Promise(() => {
```
capsula4, line 92: this method is `async` so you don't need to return a promise. You can simply do
```js
const { redirect_url } = error.body.additional_action_required.data;
window.location.replace(redirect_url);

return;
```
 
Also, object destructuring is recommended to improve readability.

deov31: I kept the returned promised due to types, If I don't return the promise it will create a mismatch error with the expected `Promise<InternalCheckoutSelectors>` of the execute function.

capsula4: That's fine, you can keep the return type `Promise<InternalCheckoutSelectors>`. When the method is `async`, it will automatically wrap any returned value in a Promise, so it's redundant to manually wrap the returned value in a Promise.

deov31: Following your recommendation i would have to delete `return` because it returns void which causes a mismatch with the return type of execute, so if I delete it and keep only the window.location.replace it will throw this error before doing the actual redirect. 

Screen Shot 2021-04-26 at 20 34 37

Any suggestion related to that?

capsula4: Right, I think I understand. What you are actually returning here is `Promise<never>` which is a promise that never resolves.

one thing you could do but doesn't really matter much is
```js
const { redirect_url } = error.body.additional_action_required.data;
window.location.replace(redirect_url);

return new Promise();
```

There's really no need to wrap that code around a promise.

capsula4 (approved):

src/payment/strategies/zip/zip-payment-strategy.ts
deov31: I will keep it right now as it is, If a return the empty promise I still can get errors
 ```ERROR in /Users/estefaniaocampo/dev/cloud-dev-vm/bcappvm/codebases/microapps/checkout-sdk-js/src/payment/strategies/zip/zip-payment-strategy.ts
[tsl] ERROR in /Users/estefaniaocampo/dev/cloud-dev-vm/bcappvm/codebases/microapps/checkout-sdk-js/src/payment/strategies/zip/zip-payment-strategy.ts(99,28)
      TS2554: Expected 1 arguments, but got 0.```

Thenks for the review.

PR #1121, opened 2021-04-27
feat(payment): INT-4071 Clearpay - Create custom strategy

roberbnd:
## What? [INT-4071](https://jira.bigcommerce.com/browse/INT-4071)
Adding clearpay payment in the registry as external payment

## Why?
enable the payment method from checkout page

## Testing / Proof
https://drive.google.com/file/d/1PxBA4NdAhkWWVNQ3_bHal7-KmKoS10UG/view?usp=sharing

## Dependencies
[https://github.com/bigcommerce/checkout-js/pull/565](https://github.com/bigcommerce/checkout-js/pull/565)

### DIFF
--- a/src/payment/strategies/clearpay/clearpay-payment-strategy.spec.ts
+++ b/src/payment/strategies/clearpay/clearpay-payment-strategy.spec.ts
@@ -1,0 +1,250 @@
+import { createClient as createPaymentClient } from '@bigcommerce/bigpay-client';
+import { createAction, createErrorAction, Action } from '@bigcommerce/data-store';
+import { createRequestSender } from '@bigcommerce/request-sender';
+import { createScriptLoader } from '@bigcommerce/script-loader';
+import { merge } from 'lodash';
+import { of, Observable } from 'rxjs';
+
+import { createCheckoutStore, CheckoutRequestSender, CheckoutStore, CheckoutValidator } from '../../../checkout';
+import { getCheckout, getCheckoutPayment, getCheckoutStoreState } from '../../../checkout/checkouts.mock';
+import { MissingDataError, NotInitializedError } from '../../../common/error/errors';
+import { OrderActionCreator, OrderActionType, OrderRequestBody, OrderRequestSender } from '../../../order';
+import { getOrderRequestBody } from '../../../order/internal-orders.mock';
+import { createSpamProtection, PaymentHumanVerificationHandler } from '../../../spam-protection';
+import { StoreCreditActionCreator, StoreCreditActionType, StoreCreditRequestSender } from '../../../store-credit';
+import PaymentActionCreator from '../../payment-action-creator';
+import { PaymentActionType } from '../../payment-actions';
+import PaymentMethod from '../../payment-method';
+import PaymentMethodActionCreator from '../../payment-method-action-creator';
+import { PaymentMethodActionType } from '../../payment-method-actions';
+import PaymentMethodRequestSender from '../../payment-method-request-sender';
+import { getClearpay } from '../../payment-methods.mock';
+import PaymentRequestSender from '../../payment-request-sender';
+import PaymentRequestTransformer from '../../payment-request-transformer';
+
+import ClearpayPaymentStrategy from './clearpay-payment-strategy';
+import ClearpayScriptLoader from './clearpay-script-loader';
+
+describe('ClearpayPaymentStrategy', () => {
+    let checkoutValidator: CheckoutValidator;
+    let checkoutRequestSender: CheckoutRequestSender;
+    let loadPaymentMethodAction: Observable<Action>;
+    let orderActionCreator: OrderActionCreator;
+    let orderRequestSender: OrderRequestSender;
+    let payload: OrderRequestBody;
+    let paymentActionCreator: PaymentActionCreator;
+    let paymentMethod: PaymentMethod;
+    let paymentMethodActionCreator: PaymentMethodActionCreator;
+    let storeCreditActionCreator: StoreCreditActionCreator;
+    let scriptLoader: ClearpayScriptLoader;
+    let submitOrderAction: Observable<Action>;
+    let submitPaymentAction: Observable<Action>;
+    let store: CheckoutStore;
+    let strategy: ClearpayPaymentStrategy;
+
+    const clearpaySdk = {
+        initialize: () => {},
+        redirect: () => {},
+    };
+
+    beforeEach(() => {
+        orderRequestSender = new OrderRequestSender(createRequestSender());
+        store = createCheckoutStore(getCheckoutStoreState());
+        paymentMethodActionCreator = new PaymentMethodActionCreator(new PaymentMethodRequestSender(createRequestSender()));
+        checkoutRequestSender = new CheckoutRequestSender(createRequestSender());
+        checkoutValidator = new CheckoutValidator(checkoutRequestSender);
+        orderActionCreator = new OrderActionCreator(orderRequestSender, checkoutValidator);
+        paymentActionCreator = new PaymentActionCreator(
+            new PaymentRequestSender(createPaymentClient()),
+            orderActionCreator,
+            new PaymentRequestTransformer(),
+            new PaymentHumanVerificationHandler(createSpamProtection(createScriptLoader()))
+        );
+        storeCreditActionCreator = new StoreCreditActionCreator(
+            new StoreCreditRequestSender(createRequestSender())
+        );
+        scriptLoader = new ClearpayScriptLoader(createScriptLoader());
+        strategy = new ClearpayPaymentStrategy(
+            store,
+            checkoutValidator,
+            orderActionCreator,
+            paymentActionCreator,
+            paymentMethodActionCreator,
+            storeCreditActionCreator,
+            scriptLoader
+        );
+
+        paymentMethod = getClearpay();
+
+        payload = merge({}, getOrderRequestBody(), {
+            payment: {
+                methodId: paymentMethod.id,
+                gatewayId: paymentMethod.gateway,
+            },
+        });
+
+        loadPaymentMethodAction = of(createAction(
+            PaymentMethodActionType.LoadPaymentMethodSucceeded,
+            { ...paymentMethod, id: 'clearpay' },
+            { methodId: paymentMethod.gateway }
+        ));
+
+        submitOrderAction = of(createAction(OrderActionType.SubmitOrderRequested));
+        submitPaymentAction = of(createAction(PaymentActionType.SubmitPaymentRequested));
+
+        payload = merge({}, getOrderRequestBody(), {
+            payment: {
+                methodId: paymentMethod.id,
+                gatewayId: paymentMethod.gateway,
+            },
+        });
+
+        jest.spyOn(store, 'dispatch');
+
+        jest.spyOn(checkoutValidator, 'validate')
+            .mockReturnValue(new Promise(resolve => resolve()));
+
+        jest.spyOn(orderActionCreator, 'submitOrder')
+            .mockReturnValue(submitOrderAction);
+
+        jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod')
+            .mockReturnValue(loadPaymentMethodAction);
+
+        jest.spyOn(storeCreditActionCreator, 'applyStoreCredit')
+            .mockReturnValue(of(createAction(StoreCreditActionType.ApplyStoreCreditSucceeded)));
+
+        jest.spyOn(paymentActionCreator, 'submitPayment')
+            .mockReturnValue(submitPaymentAction);
+
+        jest.spyOn(scriptLoader, 'load')
+            .mockReturnValue(Promise.resolve(clearpaySdk));
+
+        jest.spyOn(clearpaySdk, 'initialize')
+            .mockImplementation(() => {});
+
+        jest.spyOn(clearpaySdk, 'redirect')
+            .mockImplementation(() => {});
+    });
+
+    describe('#initialize()', () => {
+        it('loads script when initializing strategy', async () => {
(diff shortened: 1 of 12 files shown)


josesonora (changes requested):

roberbnd:
@josesonora  could you review, please? ♻️

src/payment/strategies/clearpay/clearpay-script-loader.ts
```diff
@@ -0,0 +1,36 @@
+import { ScriptLoader } from '@bigcommerce/script-loader';
+
+import PaymentMethod from '../../payment-method';
+
+import ClearpaySdk from './clearpay-sdk';
+import ClearpayWindow from './clearpay-window';
+
+interface ClearpayScripts {
```
josesonora, line 8: this is not necessary

src/payment/strategies/clearpay/clearpay-window.ts
```diff
@@ -0,0 +1,5 @@
+import ClearpaySdk from './clearpay-sdk';
+
+export default interface ClearpayWindow extends Window {
+    AfterPay: ClearpaySdk;
```
josesonora, line 4:
```suggestion
    ClearPay: ClearpaySdk;
```

src/payment/strategies/clearpay/clearpay-script-loader.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    const scriptLoader = createScriptLoader();
+    const clearpayScriptLoader = new ClearpayScriptLoader(scriptLoader);
+
+    beforeEach(() => {
+        jest.spyOn(scriptLoader, 'loadScript')
+            .mockReturnValue(Promise.resolve(new Event('load')));
+    });
+
+    it('loads widget script', () => {
+        const method = getAfterpay();
```
josesonora, line 18:
```suggestion
        const method = getClearpay();
```

src/payment/strategies/clearpay/clearpay-script-loader.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+
+        clearpayScriptLoader.load(method);
+
+        expect(scriptLoader.loadScript).toHaveBeenCalledWith(
+            '//portal.clearpay.co.uk/afterpay-async.js'
+        );
+    });
+
+    it('loads sandbox widget script if in test mode', () => {
+        const method = merge({}, getAfterpay(), { config: { testMode: true } });
```
josesonora, line 28: same here

josesonora (changes requested):

src/payment/strategies/clearpay/clearpay-window.ts
roberbnd: The global variable in the clearpay's script is called Afterpay for this reason I leave this name.

roberbnd:
@josesonora  Could you review, please? ♻️

josesonora (approved):

src/payment/strategies/clearpay/clearpay-payment-strategy.ts
```diff
@@ -0,0 +1,104 @@
...
+    async execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        const paymentId = payload.payment && payload.payment.gatewayId;
+
+        if (!paymentId) {
+            throw new PaymentArgumentInvalidError(['payment.gatewayId']);
+        }
+
+        const { isStoreCreditApplied: useStoreCredit } = this._store.getState().checkout.getCheckoutOrThrow();
+        let state = this._store.getState();
+        if (useStoreCredit !== undefined) {
```
mauricio-sg, line 52:
```suggestion
        let state = this._store.getState();

        if (useStoreCredit !== undefined) {
```

src/payment/strategies/clearpay/clearpay-payment-strategy.ts
```diff
@@ -0,0 +1,104 @@
...
+         // Validate nothing has changed before redirecting to Clearpay checkout page
+        await this._checkoutValidator.validate(state.checkout.getCheckout(), options);
+
+        state = await this._store.dispatch(
+            this._paymentMethodActionCreator.loadPaymentMethod(paymentId, options)
+        );
+
+        this._redirectToClearpay(state.paymentMethods.getPaymentMethod(paymentId));
+
+        return Promise.resolve(this._store.getState());
```
mauricio-sg, line 65:
```suggestion
        return new Promise<never>(noop)
```

src/payment/strategies/clearpay/clearpay-payment-strategy.ts
```diff
@@ -0,0 +1,104 @@
...
+
+        return this._store.dispatch(this._paymentActionCreator.submitPayment(paymentPayload));
+    }
+
+    private _redirectToClearpay(paymentMethod?: PaymentMethod): void {
+        if (!this._clearpaySdk || !paymentMethod || !paymentMethod.clientToken) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+        }
+
+        this._clearpaySdk.initialize({countryCode: this._getCountryCode()});
```
mauricio-sg, line 96: why not just...
```suggestion
        this._clearpaySdk.initialize({ countryCode: 'GB' });
```
IDK if clearpay will work in another countries, so you could keep the `_mapCountryToISO2` from afterpay instead

src/payment/strategies/clearpay/clearpay-script-loader.ts
```diff
@@ -0,0 +1,31 @@
...
+        private _scriptLoader: ScriptLoader
+    ) {}
+
+    /**
+     * Loads the appropriate Clearpay SDK depending on the payment method data.
+     * @param PaymentMethod the payment method data
+     */
+    load(method: PaymentMethod): Promise<ClearpaySdk> {
+        return this._scriptLoader.loadScript(this._getScript(method.config.testMode))
+            .then(() => (window as unknown as ClearpayWindow).AfterPay);
```
mauricio-sg, line 25:
```suggestion
        await this._scriptLoader.loadScript(this._getScript(method.config.testMode))
        
        if (!this._window.AfterPay) {
            throw new PaymentMethodClientUnavailableError();
        }

        return this._window.AfterPay;
```

src/payment/strategies/clearpay/clearpay-script-loader.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    private _getScript(testMode?: boolean): string {
```
mauricio-sg, line 28:
```suggestion
    private _getScriptUrl(testMode?: boolean): string {
```

mauricio-sg (changes requested):
just a couple of comments 😁

roberbnd:
@mauricio-sg  Could you review, please? ♻️

src/payment/strategies/clearpay/clearpay-payment-strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        this._clearpaySdk.initialize({countryCode: this._mapCountryToISO2()});
+        this._clearpaySdk.redirect({ token: paymentMethod.clientToken });
+    }
+
+    // Maps the country code to be passed to the sdk initialization
+    private _mapCountryToISO2(): string {
+        return 'GB';
+    }
```
mauricio-sg, line 105: IMO this function is not necessary 😕 just hardcode the value 😄

mauricio-sg (approved):
LGTM

src/payment/strategies/clearpay/clearpay-payment-strategy.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+
+            expect(paymentActionCreator.submitPayment).toHaveBeenCalledWith({
+                methodId: paymentMethod.id,
+                paymentData: { nonce },
+            });
+        });
+
+        it('throws error if unable to finalize order due to missing data', async () => {
+            try {
+                await strategy.finalize({ methodId: paymentMethod.id, gatewayId: paymentMethod.gateway });
```
capsula4, line 244: for testing exceptions, the following approach is preferred as it ensures the assertion is executed:
```js
expect(() => await strategy.finalize({ methodId: paymentMethod.id, gatewayId: paymentMethod.gateway)
  .toThrow(MissingDataError)
```

src/payment/strategies/clearpay/clearpay-payment-strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+        await this._checkoutValidator.validate(state.checkout.getCheckout(), options);
+
+        state = await this._store.dispatch(
+            this._paymentMethodActionCreator.loadPaymentMethod(paymentId, options)
+        );
+
+        this._redirectToClearpay(state.paymentMethods.getPaymentMethod(paymentId));
+
+        // Clearpay will handle the rest of the flow so return a promise that doesn't really resolve
+        return new Promise<never>(() => {});
```
capsula4, line 67: I don't think you need `<never>` ?

src/payment/strategies/clearpay/clearpay-payment-strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+
+        return this._store.dispatch(this._paymentActionCreator.submitPayment(paymentPayload));
+    }
+
+    private _redirectToClearpay(paymentMethod?: PaymentMethod): void {
+        if (!this._clearpaySdk || !paymentMethod || !paymentMethod.clientToken) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+        }
+
+        this._clearpaySdk.initialize({countryCode: this._mapCountryToISO2()});
```
capsula4, line 98: This is not a mapping function so the name is misleading. I wonder if we should use the country code comes from the billing address? And if it's not UK show an error message?

capsula4, line 98: This is not a mapping function so the name is misleading. I wonder if we should use the country code comes from the billing address? And if it's not UK show an error message?

src/payment/strategies/clearpay/clearpay-script-loader.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import ClearpaySdk from './clearpay-sdk';
+import ClearpayWindow from './clearpay-window';
+
+const SCRIPTS_DEFAULT = {
+    PROD: '//portal.clearpay.co.uk/afterpay-async.js',
+    SANDBOX: '//portal.sandbox.clearpay.co.uk/afterpay-async.js',
+};
+
+/** Class responsible for loading the Clearpay SDK */
```
capsula4, line 14: I don't think the comment is necessary. The class name is descriptive enough.

src/payment/strategies/clearpay/clearpay-script-loader.ts
```diff
@@ -0,0 +1,38 @@
...
+export default class ClearpayScriptLoader {
+    constructor(
+        private _scriptLoader: ScriptLoader,
+        public _window: ClearpayWindow = window
+    ) { }
+
+    /**
+     * Loads the appropriate Clearpay SDK depending on the payment method data.
```
capsula4, line 22: You don't need this docblock, we already have types provided by the SDK.

We only add docblocks for public SDK methods, so devs using the SDK know how to use them. This is an internal method of the SDK, so there's no need to document it.

roberbnd:
@capsula4  Could you review, please?

src/payment/strategies/clearpay/clearpay-payment-strategy.ts
capsula4: I think the problem here is that a shopper could have selected a transactional currency that is no GBP, and that would mean that suddenly if they choose Clearpay, everything will be in GBP.

Even if Clearpay doesn't allow it, we should cover that scenario and handle it properly, rather than hardcoding the country ourselves. This will help developers using the SDK in case they make a mistake.

src/payment/strategies/clearpay/clearpay-payment-strategy.ts
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        const { isStoreCreditApplied: useStoreCredit } = this._store.getState().checkout.getCheckoutOrThrow();
+        let state = this._store.getState();
+
+        if (useStoreCredit !== undefined) {
+            state = await this._store.dispatch(this._storeCreditActionCreator.applyStoreCredit(useStoreCredit));
+        }
+
+         // Validate nothing has changed before redirecting to Clearpay checkout page
```
capsula4, line 60: rather than having this as a comment, you can remove it since you already have a unit test for it. You could rename the unit test from
```it('validates the checkout')```
to
```it('validates nothing has changed before redirecting to Clearpay checkout page`)```

File: src/payment/strategies/clearpay/clearpay-payment-strategy.ts [hunk 17]
```diff
@@ -0,0 +1,113 @@
...
+    private _redirectToClearpay(countryCode: string, paymentMethod?: PaymentMethod): void {
+        if (!this._clearpaySdk || !paymentMethod || !paymentMethod.clientToken) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+        }
+
+        this._clearpaySdk.initialize({countryCode});
+        this._clearpaySdk.redirect({ token: paymentMethod.clientToken });
+    }
+
+    private _validateBillingAddress( countryCode: string): boolean {
```
capsula4, line 110: minor, this method is not validating the billing address. Maybe rename it to `_isCountrySupported`

roberbnd:
@capsula4  could you review, please?

capsula4 (approved):

josesonora:
💚  Tested:

- Shoppers can complete a payment by using Clearpay as payment option

PR #1122, opened 2021-04-28
fix(payment): INT-4266 Googlepay - fixing where nonce value is going …

roberbnd:
…to be reloaded

## What? [INT-4266](https://jira.bigcommerce.com/browse/INT-4266)
change where the nonce value is going to be reloaded

## Why?
the nonce value was being reloaded before be used then the value was empty, triggering one more time the popup to select the credit card
this happens because the backend delete the nonce value after be used: [Link](https://github.com/bigcommerce/bigcommerce/pull/39827/files#diff-ec498346f9c15c814cc5db89dd9abe81d9509baad8b0085124ed355238e145c8R438)

the function `loadPaymentMethod` calls the backend (deleting the nonce value) then I have to be careful where is called

## Testing / Proof
Happy path, it works form checkout and the cart section
https://drive.google.com/file/d/1vfEomqIo-Y_Ob0FYEMbXvl0yrTcZEh8x/view?usp=sharing

the original issue with the nonce value is fixed: [INT-3611](https://jira.bigcommerce.com/browse/INT-3611)
https://drive.google.com/file/d/1jF8xMUZkK_sInv_W2G5QNRELB-CtUaTB/view?usp=sharing

### DIFF
--- a/src/payment/strategies/googlepay/googlepay-payment-strategy.ts
+++ b/src/payment/strategies/googlepay/googlepay-payment-strategy.ts
@@ -37,16 +37,12 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
     async initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
         const { methodId } = options;
 
-        const state = await this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId));
+        const state = this._store.getState();
         this._paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(methodId);
 
         this._googlePayOptions = this._getGooglePayOptions(options);
 
         this._buttonClickEventHandler = this._handleButtlonClickedEvent(methodId);
-
-        if (this._paymentMethod.initializationData.nonce) {
-            return Promise.resolve(this._store.getState());
-        }
 
         await this._googlePayPaymentProcessor.initialize(methodId);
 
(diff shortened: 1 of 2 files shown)


mauricio-sg (approved):
LGTM 👍

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -153,41 +156,39 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
-                cardInformation,
-                nonce: this._getNonce(methodId, this._paymentMethod),
+                cardInformation: this._paymentMethod.initializationData.card_information,
+                nonce: await this._getNonce(methodId),
             },
         };
     }
 
-    private _getNonce(methodId: string, { initializationData: { nonce }}: PaymentMethod) {
+    private async _getNonce(methodId: string) {
```
capsula4, line 173: maybe add the missing return type to the method signature?

capsula4 (commented):
Code changes look good however I can see no unit tests were updated. I guess this fixes an issue that wasn't unit tested. I would encourage adding one to avoid re-introducing this bug accidentally.

roberbnd:
@capsula4  could you review please? ♻️

capsula4 (approved):

rabinshr:
@vmparra tested this fix and confirmed it was working 💚

PR #1123, opened 2021-04-29
Revert "fix(payment): INT-3611 Refresh the state in Googlepay and Braintree r…"

capsula4:
Reverts bigcommerce/checkout-sdk-js#1082


icatalina:
I'm going to assume this is no longer needed.

PR #1128, opened 2021-05-05
feat(payment): PAYMENTS-6806 add dummy PPSDK strategy

leeBigCommerce:
https://jira.bigcommerce.com/browse/PAYMENTS-6806

BCApp Experiment PR: https://github.com/bigcommerce/bigcommerce/pull/40715

## What?
- Add the start of a PPSDK Strategy behind an experiment flag: `PAYMENTS-6806.enable_ppsdk_strategy`
- The initial strategy is just a dummy class with no functionality

## Why?
- To safely begin building out the ability to consume PPSDK payment methods

## Testing / Proof
- Passing tests
- No outward changes - due to work in progress being gated behind flag

### DIFF
--- a/src/payment/payment-strategy-registry.spec.ts
+++ b/src/payment/payment-strategy-registry.spec.ts
@@ -1,9 +1,11 @@
+import { set } from 'lodash';
+
 import { createCheckoutStore, CheckoutStore, InternalCheckoutSelectors } from '../checkout';
 import { getConfigState } from '../config/configs.mock';
 import { getFormFieldsState } from '../form/form.mock';
 import { OrderFinalizationNotRequiredError } from '../order/errors';
 
-import { getAdyenAmex, getAmazonPay, getBankDeposit, getBraintree, getBraintreePaypal, getCybersource } from './payment-methods.mock';
+import { getAdyenAmex, getAmazonPay, getBankDeposit, getBraintree, getBraintreePaypal, getCybersource, getPPSDK } from './payment-methods.mock';
 import PaymentStrategyRegistry from './payment-strategy-registry';
 import PaymentStrategyType from './payment-strategy-type';
 import { PaymentStrategy } from './strategies';
@@ -49,6 +51,9 @@ describe('PaymentStrategyRegistry', () => {
     // tslint:disable-next-line:max-classes-per-file
     class AmazonPayPaymentStrategy extends BasePaymentStrategy {}
 
+    // tslint:disable-next-line:max-classes-per-file
+    class PPSDKPaymentStrategy extends BasePaymentStrategy {}
+
     beforeEach(() => {
         store = createCheckoutStore({
             config: getConfigState(),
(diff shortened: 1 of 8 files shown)


src/payment/payment-strategy-registry.spec.ts
```diff
@@ -47,7 +49,10 @@ describe('PaymentStrategyRegistry', () => {
     class OffsitePaymentStrategy extends BasePaymentStrategy {}
 
     // tslint:disable-next-line:max-classes-per-file
-    class AmazonPayPaymentStrategy extends BasePaymentStrategy {}
+    class AmazonPayPaymentStrategy extends BasePaymentStrategy { }
```
Tharaae, line 52: Please get rid of the extra space 🙏 
```suggestion
    class AmazonPayPaymentStrategy extends BasePaymentStrategy {}
```

capsula4 (approved):
Looks good! 🐑  it

src/payment/payment-strategy-registry.ts
```diff
@@ -41,6 +41,15 @@ export default class PaymentStrategyRegistry extends Registry<PaymentStrategy, P
     }
 
     private _getToken(paymentMethod: PaymentMethod): PaymentStrategyType {
+        const ppsdkFeatureOn =
+            this._store.getState()
+                .config.getStoreConfig()
+                ?.checkoutSettings.features['PAYMENTS-6806.enable_ppsdk_strategy'];
+
+        if (ppsdkFeatureOn && paymentMethod.type === PaymentStrategyType.PPSDK) {
```
capsula4, line 49: what happens if `paymentMethod.type === PaymentStrategyType.PPSDK` and `!ppsdkFeatureOn` ? Maybe throw an exception explaining that is not available?

Tharaae (commented):
There should be a sibling bcapp PR to add the experiment and update the features list.
Please add a link to the bcapp PR to this PR description.

src/payment/payment-strategy-registry.ts
Tharaae: I think if `!ppsdkFeatureOn`, we want it to return `PaymentStrategyType.CREDIT_CARD` which currently happens.
Is that what we want @leeBigCommerce ?

capsula4: Would be good to add a unit test for that scenario then, to ensure the expected behaviour is tested and maintained.

leeBigCommerce: Yep, this is largely an extra precaution to gate off in progress code.

I've got a test to prove that it isn't selected in that scenario: https://github.com/bigcommerce/checkout-sdk-js/pull/1128/files#diff-7ab1b1fea01093913fc2ae689daa95fb1326449a6eaa1ca2d51d9fd13b50b354R76

It doesn't assert which strategy is chosen, as I didn't want to hardcode it to expect a specific default (which is tested elsewhere).

src/payment/payment-strategy-registry.spec.ts
leeBigCommerce: done 👍

PR #1140, opened 2021-05-18
feat(payment): INT-4170 added hosted fields on Mollie verification field

JoseAlvarezHdz:
What? INT-4170

Added hosted fields on verification fields on Mollie. 

Why?

Currently, we have some payment methods that are not using hosted credit card fields for stored instrument verifications (card number and CVV). This presents a security vulnerability as malicious JS can easily scrape credit card information.

## Testing / Proof

Screen Shot 2021-05-20 at 16 07 38

### DIFF
--- a/src/payment/strategies/mollie/mollie-payment-strategy.ts
+++ b/src/payment/strategies/mollie/mollie-payment-strategy.ts
@@ -4,7 +4,8 @@ import { PaymentActionCreator } from '../..';
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
 import { getBrowserInfo } from '../../../common/browser-info';
 import { InvalidArgumentError, MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType, RequestError } from '../../../common/error/errors';
-import { OrderActionCreator, OrderRequestBody } from '../../../order';
+import { HostedForm, HostedFormFactory, HostedFormOptions } from '../../../hosted-form';
+import { OrderActionCreator, OrderPaymentRequestBody, OrderRequestBody } from '../../../order';
 import { PaymentArgumentInvalidError } from '../../errors';
 import isVaultedInstrument from '../../is-vaulted-instrument';
 import { HostedInstrument } from '../../payment';
@@ -27,7 +28,10 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
     private _verificationCodeElement?: MollieElement;
     private _expiryDateElement?: MollieElement;
 
+    private _hostedForm?: HostedForm;
+
     constructor(
+        private _hostedFormFactory: HostedFormFactory,
         private _store: CheckoutStore,
         private _mollieScriptLoader: MollieScriptLoader,
         private _orderActionCreator: OrderActionCreator,
@@ -35,11 +39,16 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
     ) { }
 
     async initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
-        const { mollie, methodId } = options;
+        const { mollie, methodId, gatewayId } = options;
 
         if (!mollie) {
             throw new InvalidArgumentError('Unable to initialize payment because "options.mollie" argument is not provided.');
         }
+
+        if (!methodId || !gatewayId) {
+            throw new InvalidArgumentError('Unable to initialize payment because "methodId" and/or "gatewayId" argument is not provided.');
+        }
+
         const state = this._store.getState();
         const storeConfig = state.config.getStoreConfig();
 
@@ -50,14 +59,16 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
         this._initializeOptions = mollie;
 
         const paymentMethods = state.paymentMethods;
-        const paymentMethod = paymentMethods.getPaymentMethodOrThrow(options.methodId);
+        const paymentMethod = paymentMethods.getPaymentMethodOrThrow(methodId, gatewayId);
         const { config: { merchantId, testMode } } = paymentMethod;
 
         if (!merchantId) {
             throw new InvalidArgumentError('Unable to initialize payment because "merchantId" argument is not provided.');
         }
 
-        if (methodId === MolliePaymentMethodType.creditcard) {
+        if (this.isCreditCard(methodId) && mollie.form  && this.shouldShowTSVHostedForm(methodId, gatewayId)) {
+            this._hostedForm = await this._mountCardVerificationfields(mollie.form);
+        } else if (this.isCreditCard(methodId)) {
             this._mollieClient = await this._loadMollieJs(merchantId, storeConfig.storeProfile.storeLanguage, testMode);
             this._mountElements();
         }
@@ -68,55 +79,23 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
     async execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
         const { payment , ...order} = payload;
         const paymentData = payment?.paymentData;
-        const shouldSaveInstrument = (paymentData as HostedInstrument)?.shouldSaveInstrument;
-        const shouldSetAsDefaultInstrument = (paymentData as HostedInstrument)?.shouldSetAsDefaultInstrument;
 
-        if (!payment || !paymentData) {
-            throw new PaymentArgumentInvalidError([ 'payment' ]);
+        if (!payment || !payment.gatewayId || !paymentData) {
+            throw new PaymentArgumentInvalidError([ 'payment', 'gatewayId', 'paymentData' ]);
         }
 
         try {
-            if (payment.methodId === MolliePaymentMethodType.creditcard) {
-                await this._store.dispatch(this._orderActionCreator.submitOrder(order, options));
+            await this._store.dispatch(this._orderActionCreator.submitOrder(order, options));
 
-                if (paymentData && isVaultedInstrument(paymentData)) {
-                    return this._store.dispatch(this._paymentActionCreator.submitPayment(payment));
-                }
-
-                const { token, error } = await this._getMollieClient().createToken();
-
-                if (error) {
-                    return Promise.reject(error);
-                }
-
-                return await this._store.dispatch(this._paymentActionCreator.submitPayment({
-                    ...payment,
-                    paymentData: {
-                        formattedPayload: {
-                            credit_card_token: {
-                                token,
-                            },
-                            vault_payment_instrument: shouldSaveInstrument,
-                            set_as_default_stored_instrument: shouldSetAsDefaultInstrument,
-                            browser_info: getBrowserInfo(),
-                        },
-                    },
-                }));
-            } else {
-                await this._store.dispatch(this._orderActionCreator.submitOrder(order, options));
-
-                const issuer = 'issuer' in paymentData ? paymentData.issuer : '';
-
-                return await this._store.dispatch(this._paymentActionCreator.submitPayment({
-                    ...payment,
-                    paymentData: {
-                        ...paymentData,
-                        formattedPayload: {
-                            issuer,
-                        },
-                    },
-                }));
+            if (isVaultedInstrument(paymentData)) {
+                return await this.executeWithVaulted(payment);
             }
+
+            if (this.isCreditCard(payment.methodId)) {
+                return await this.executeWithCC(payment);
+            }
+
+            return await this.executeWithAPM(payment);
         } catch (error) {
 
             return this._processAdditionalAction(error);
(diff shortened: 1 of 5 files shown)


src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -80,7 +88,21 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
...
+                        if (!form) {
+                            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+                        }
+
+                        try {
+                            await form.validate();
+
+                            return this._store.dispatch(this._paymentActionCreator.submitPayment(payment));
+                        } catch (error) {
+                            throw new Error(error.message);
```
capsula4, line 103: this seems to simply rethrow the same error. Do you ned this try/catch block at all?

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -135,6 +157,55 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
...
 
+    private _isHostedFieldAvailable(options?: PaymentInitializeOptions): boolean {
+        if (!options) {
+            throw new InvalidArgumentError();
+        }
+
+        return (options.mollie?.containerId) ? true : false;
+    }
+
+    private  _mountCardVerificationfields(options: any): Promise<any> {
```
capsula4, line 168: can you please replace `any` with the specific type?

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -135,6 +157,55 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
...
+        const { config } = this._store.getState();
+        const { paymentSettings: { bigpayBaseUrl: host = '' } = {} } = config.getStoreConfig() || {};
+
+        if (!formOptions) {
+            throw new InvalidArgumentError();
+        }
+
+        const form = this._hostedFormFactory.create(host, formOptions);
+
+        return new Promise(async (resolve, reject) => {
```
capsula4, line 183: by making the method async, you won't need to create this Promise wrap.

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -135,6 +157,55 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
...
+            if (options.mollie.form) {
```
capsula4, line 184: I don't think you need to check this again, you have this heck on line 177.

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -135,6 +157,55 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
...
+                }
+            }
+
+            resolve(form);
+        });
+    }
+
+    private _isHostedPaymentFormEnabled(methodId?: string, gatewayId?: string): boolean {
+        if (!methodId) {
+            return false;
```
capsula4, line 200: methodId should be mandatory. Or if it's missing, this should throw an error rather than return false.

src/payment/strategies/mollie/mollie-payment-strategy.ts
JoseAlvarezHdz: Yes, for some reason (i think its because the type of the Error) it just console log the error it doesnt throw it!

JoseAlvarezHdz: its for form.validate error

JoseAlvarezHdz:
@capsula4  i have made some changes! can you review again please!

JoseBobadilla20:
💚 Tested
QA Details: I tested TSV for Mollie to check works correctly.

JoseAlvarezHdz:
@capsula4 i have made some changes! can you review again please!

src/payment/strategies/mollie/mollie-payment-strategy.ts
capsula4: We don't do this for other strategies, e.g. `credit-card-payment-strategy` so I'm still not convinced why it's needed here.

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -135,6 +166,33 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
         return Promise.resolve(this._store.getState());
     }
 
+    private async _mountCardVerificationfields(formOptions: HostedFormOptions) {
+        if (!formOptions) {
+            throw new InvalidArgumentError();
+        }
+
+        const { config } = this._store.getState();
+        const { paymentSettings: { bigpayBaseUrl: host = '' } = {} } = config.getStoreConfig() || {};
```
capsula4, line 175: should you be throwing an error if `host` is an empty string?

JoseAlvarezHdz: When is empty it already show this error
Screen Shot 2021-05-26 at 13 30 42

JoseAlvarezHdz:
@capsula4 i have made some changes! can you review again please!

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -71,16 +85,29 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
...
+                        if (!form) {
+                            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+                        }
+
+                        await form.validate();
+                        await form.submit(payment);
+
+                        return this._store.dispatch(this._orderActionCreator.loadCurrentOrder());
+                    } else {
+                        return await this._store.dispatch(this._paymentActionCreator.submitPayment(payment));
```
icatalina, line 109: How come we `await` for the last `distpatch` but not the first one? Do we need to `await` for any of them... 🤷

JoseAlvarezHdz:
@icatalina i've address your comments, can you review again please!

icatalina (commented):
One of the comments was resolved with no answer or code change. Can we please make sure all comments are addressed before posting the PR for re-review?

src/payment/strategies/mollie/mollie-payment-strategy.ts
JoseAlvarezHdz: @icatalina 
if we remove the "dummy" value, its showing an alert about not providing a default value

Screen Shot 2021-06-02 at 8 49 24

icatalina: Hey @JoseAlvarezHdz!

I was thinking on something more like this:

```ts
const bigpayBaseUrl = config.getStoreConfig()?.paymentSettings.bigpayBaseUrl;

if (!bigpayBaseUrl) {
}

const form = this._hostedFormFactory.create(bigpayBaseUrl, formOptions);
```

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -50,14 +59,19 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
...
 
         if (!merchantId) {
             throw new InvalidArgumentError('Unable to initialize payment because "merchantId" argument is not provided.');
         }
 
         if (methodId === MolliePaymentMethodType.creditcard) {
+
+            if (mollie.form && this._isHostedPaymentFormEnabled(methodId, gatewayId) && this._isHostedFieldAvailable()) {
+                await this._mountCardVerificationfields(mollie.form);
+            }
```
icatalina, line 73: hey @JoseAlvarezHdz

This is a bit confusing. I was under the impression this PR was meant to replace the standard mollie form for a hosted field version. The screenshot shown in the PR description shows a saved credit card form, not the default one.

Also, in line https://github.com/bigcommerce/checkout-sdk-js/pull/1140/files#diff-ef86963fe7cb5db6143559804e7e2f0dc81a48ab8562e81b03ff1968b12f314aR76 we seem to be "mounting" the mollie form after we've already attached the hosted one.

I could be misreading this PR, but could you provide an screenshot where you show the actual hosted form being shown? I'm under the impression this code will try and render both, the hosted fields and mollie ones alongside each other.

icatalina (changes requested):

src/payment/strategies/mollie/mollie-payment-strategy.ts
JoseAlvarezHdz: @icatalina 

This ticket is about **adding Verification Fields as Hosted Fields for TSV**, it does not REPLACE anything as Mollie didn't have verification Fields! 

This is how it looks with **Hosted Field for Verification** meaning the hosted fields were added for TSV not for cc form as Mollie **IS ALREADY HOSTED FORM**

Screen Shot 2021-06-04 at 10 43 52

This is how it looks when adding a new CC when the client has a vaulted cc

Screen Shot 2021-06-04 at 10 45 35


And this is how it looks just as the CC FORM
Screen Shot 2021-06-04 at 10 48 15

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -27,19 +28,27 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
     private _verificationCodeElement?: MollieElement;
     private _expiryDateElement?: MollieElement;
 
+    private _hostedForm?: HostedForm;
+    private _shouldRenderHostedForm?: boolean;
```
icatalina, line 32: Since we are setting these 2 in tandem, should we just use the `_hostedForm`? If present, that means render it.

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -135,6 +162,34 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
...
+
+        if (!bigpayBaseUrl) {
+            throw new MissingDataError(MissingDataErrorType.MissingCheckoutConfig);
+        }
+
+        const form = this._hostedFormFactory.create(bigpayBaseUrl, formOptions);
+
+        await form.attach();
+        this._shouldRenderHostedForm = true;
+        this._hostedForm = form;
```
icatalina, line 177: It is generally a good practice to return values out of functions. That way it is easier to follow what's going on in the caller.

```ts
this._hostedForm = await this._mountCardVerificationfields(mollie.form);
```

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -71,16 +85,29 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
...
+                            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+                        }
+
+                        await form.validate();
+                        await form.submit(payment);
+
+                        return await this._store.dispatch(this._orderActionCreator.loadCurrentOrder());
+                    } else {
+                        return await this._store.dispatch(this._paymentActionCreator.submitPayment(payment));
+                    }
```
icatalina, line 110: 🍹 This might be a good point to split the different paths into different `execute` methods. I can see we have at least 4 different paths in here: `executeCreditCard`/`executeWithInstrumentHosted`/`executeOther`/`executeWithInstrument`. What do you think?

That way this method could just route to the actual method performing the action:

```ts
try {
	if (payment.methodId !== MolliePaymentMethodType.creditcard) {
		return this._executeOther();
    }

	if (!paymentData || !isVaultedInstrument(paymentData) {
		return this._executeCreditCard();
	}

	if (this._isHostedPaymentFormEnabled(payment.methodId, payment.gatewayId) && this._shouldRenderHostedForm) {
		return this._executeWithInstrumentHosted();
    }

    return this._executeWithInstrument();
} catch (error) {
	// rest of the code...
}
```

icatalina, line 110: 🍹 This might be a good point to split the different paths into different `execute` methods. I can see we have at least 4 different paths in here: `executeCreditCard`/`executeWithInstrumentHosted`/`executeOther`/`executeWithInstrument`. What do you think?

That way this method could just route to the actual method performing the action:

```ts
try {
	if (payment.methodId !== MolliePaymentMethodType.creditcard) {
		return this._executeOther();
    }

	if (!paymentData || !isVaultedInstrument(paymentData) {
		return this._executeCreditCard();
	}

	if (this._isHostedPaymentFormEnabled(payment.methodId, payment.gatewayId) && this._shouldRenderHostedForm) {
		return this._executeWithInstrumentHosted();
    }

    return this._executeWithInstrument();
} catch (error) {
	// handle error
}
```

src/payment/strategies/mollie/mollie-payment-strategy.ts
JoseAlvarezHdz: Yes, kind of. but i have made a change on it

src/payment/strategies/mollie/mollie-payment-strategy.ts
JoseAlvarezHdz: yeap sure

src/payment/strategies/mollie/mollie-payment-strategy.ts
JoseAlvarezHdz: Done!

src/payment/strategies/mollie/mollie-payment-strategy.ts
JoseAlvarezHdz: Can you review my approach, i did it i bit different let me know what you think

JoseAlvarezHdz:
@icatalina can you take a look again please!

src/payment/strategies/mollie/mollie-payment-strategy.ts
```diff
@@ -68,55 +79,23 @@ export default class MolliePaymentStrategy implements PaymentStrategy {
...
-                            browser_info: getBrowserInfo(),
-                        },
-                    },
-                }));
-            } else {
-                await this._store.dispatch(this._orderActionCreator.submitOrder(order, options));
-
-                const issuer = 'issuer' in paymentData ? paymentData.issuer : '';
+            if (isVaultedInstrument(paymentData)) {
+                return this.executeWithVaulted(payment);
```
icatalina, line 91: I think you need to `await` these for the `try`/`catch` to work.

icatalina (approved):
Just the issue with the awaits. Other than that, it looks ok.

src/payment/strategies/mollie/mollie-payment-strategy.ts
JoseAlvarezHdz: Yes you were right! thanks!

PR #1146, opened 2021-05-26
fix(checkout): INT-3604 Omit error on googlepay and amazon pay when container is invalid

FranciscoOnt:
## What?
...

## Why?
...

## Testing / Proof
...


FranciscoOnt:
Approach is not working as expected

PR #1148, opened 2021-05-27
fix(checkout): PAYPAL-1055 changed expected methodId from przelewy24 to p24 at paypal commerce

vneutrino:
## What?
PAYPAL-1055 changed expected methodId from przelewy24 to p24 at paypal commerce
Removed buyer-country property from paypal commerce initialization because it's not used anymore.

## Why?
due to the https://jira.bigcommerce.com/browse/PAYPAL-1055

## Testing / Proof
Tested manually

### DIFF
--- a/src/payment/payment-methods.mock.ts
+++ b/src/payment/payment-methods.mock.ts
@@ -107,30 +107,6 @@ export function getPaypalCommerce(): PaymentMethod {
     };
 }
 
-export function getPaypalCommerceTestModeOn(): PaymentMethod {
-    return {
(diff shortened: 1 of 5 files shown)


icatalina (changes requested):
The code change is ok, I still have some concerns with respect to the overall change though.

1. Is this payment method live?
2. Will this change break current consumers?
3. Are there sibling PRs in other codebases?

icatalina (approved):

icatalina:
Conversation was resolved offline. It seems like the payment method is currently not working, this fix is necessary.

PR #1149, opened 2021-05-31
fix(payments): Renew the nonce for Google Payment when the payment fails.

EduardCerv:
## What? [INT-4266](https://jira.bigcommerce.com/browse/INT-4266)
Renew the nonce when the payment fails for GooglePay

## Why?
If the payment fails we don't renew the nonce so the user used the same nonce to pay and never can pay again until the user delete his order and create a new one with a new nonce.

## Testing / Proof
- [x] Manual Verification () => [VIDEO](https://drive.google.com/file/d/1ofP_cVPm2424YeDXjpAaQQDFLjOAp5zU/view?usp=sharing)
- [x] Unit Test.
image

### DIFF
--- a/src/payment/strategies/googlepay/googlepay-payment-strategy.spec.ts
+++ b/src/payment/strategies/googlepay/googlepay-payment-strategy.spec.ts
@@ -297,6 +297,34 @@ describe('GooglePayPaymentStrategy', () => {
             expect(googlePayPaymentProcessor.displayWallet).toBeCalled();
         });
 
+        it('gets again the payment information and get a new nonce', async () => {
+            jest.spyOn(googlePayPaymentProcessor, 'handleSuccess').mockReturnValue(Promise.resolve());
+            jest.spyOn(googlePayPaymentProcessor, 'displayWallet').mockReturnValue(getGooglePaymentDataMock());
+            jest.spyOn(store.getState().paymentMethods, 'getPaymentMethodOrThrow').mockReturnValueOnce({
+                initializationData: {
+                    nonce: '',
+                    card_information: {
+                        type: 'type',
+                        number: 'number',
+                    },
+                },
+            }).mockReturnValue({
+                initializationData: {
+                    nonce: 'newNonce',
+                    card_information: {
+                        type: 'type',
+                        number: 'number',
+                    },
+                },
+            });
+
+            await strategy.initialize(googlePayOptions);
+            await strategy.execute(getGoogleOrderRequestBody());
+
+            expect(store.getState().paymentMethods.getPaymentMethodOrThrow).toHaveBeenCalledTimes(3);
+            expect(googlePayPaymentProcessor.displayWallet).toBeCalled();
+        });
+
         it('gets again the payment information and user closes widget', async () => {
             jest.spyOn(googlePayPaymentProcessor, 'handleSuccess').mockReturnValue(Promise.resolve());
             jest.spyOn(googlePayPaymentProcessor, 'displayWallet').mockReturnValue(
--- a/src/payment/strategies/googlepay/googlepay-payment-strategy.ts
+++ b/src/payment/strategies/googlepay/googlepay-payment-strategy.ts
@@ -42,7 +42,7 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
 
         this._googlePayOptions = this._getGooglePayOptions(options);
 
-        this._buttonClickEventHandler = this._handleButtlonClickedEvent(methodId);
+        this._buttonClickEventHandler = this._handleButtonClickedEvent(methodId);
 
         if (this._paymentMethod.initializationData.nonce) {
             return Promise.resolve(this._store.getState());
@@ -89,6 +89,11 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
 
         const { methodId } = payload.payment;
 
+        if (this._paymentMethod?.initializationData.nonce !== '') {
+            const state = this._store.getState();
+            this._paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(methodId);
+        }
+
         let payment = await this._getPayment(methodId);
 
         if (!payment.paymentData.nonce || !payment.paymentData.cardInformation) {
@@ -97,11 +102,11 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
                 onPaymentSelect = () => {},
             } = this._googlePayOptions;
             await this._displayWallet(methodId, onPaymentSelect, onError);
-            payment = await this._getPayment(methodId);
+            payment = await this._getPayment(methodId, true);
+        }
 
-            if (!payment.paymentData.nonce) {
-                throw new MissingDataError(MissingDataErrorType.MissingPayment);
-            }
+        if (!payment.paymentData.nonce) {
+            throw new MissingDataError(MissingDataErrorType.MissingPayment);
         }
 
         try {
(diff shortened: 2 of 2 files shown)


JoseAlvarezHdz (approved):
LGTM

roberbnd (approved):
LGTM 👍

vmparra:
@BC-ISilva, could you 🙏🏻  make sure you validate this ticket on following gateways:

- Stripe
- StripeV3
- AdyenV2
- CybersourceV2
- Checkout.com
- Authorize.Net
- Braintree

src/payment/strategies/googlepay/googlepay-payment-strategy.spec.ts
```diff
@@ -291,6 +291,32 @@ describe('GooglePayPaymentStrategy', () => {
                 },
             });
 
+            await strategy.initialize(googlePayOptions);
+            await expect(strategy.execute(getGoogleOrderRequestBody())).rejects.toThrow(MissingDataError);
+            expect(googlePayPaymentProcessor.displayWallet).toBeCalled();
+        });
+
+        it('gets again the payment information and get a new nonce', async () => {
```
icatalina, line 299: I don't see a rejected promise or any hint that this is checking for a failed payment method as the PR suggests. Is the test checking for the correct thing?

src/payment/strategies/googlepay/googlepay-payment-strategy.spec.ts
```diff
@@ -291,6 +291,32 @@ describe('GooglePayPaymentStrategy', () => {
...
+            }).mockReturnValue({
+                initializationData: {
+                    nonce: 'newNonce',
+                    card_information: {
+                        type: 'type',
+                        number: 'number',
+                    },
+                },
+            });
+
```
icatalina, line 319: This checks for `displayWallet` to have been called. That will happen unconditionally when we call execute with an empty `nonce`. I'm not sure what this test is trying to prove.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -153,32 +154,30 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
-    private _getNonce(methodId: string, { initializationData: { nonce }}: PaymentMethod) {
+    private async _getNonce(methodId: string) {
+        if (!this._paymentMethod) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+        }
+        const nonce = this._paymentMethod.initializationData.nonce;
+
+        if (this._paymentMethod.initializationData.nonce) {
+            this._paymentMethod = { ...this._paymentMethod,  initializationData: { nonce: ''} };
+        }
```
icatalina, line 179: I don't think this method should be cleaning the used nonce. That should probably live somewhere else. The caller maybe?

icatalina (changes requested):
This approach is a bit odd. We seem to be waiting for a button click to load the payment method and get the nonce, it seems like a better approach will be to renew the nonce on failure. We could update the `remote.php` endpoint with an empty nonce and refetch the payment information or something on those lines.

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -37,16 +37,12 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+        const state = this._store.getState();
         this._paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(methodId);
 
         this._googlePayOptions = this._getGooglePayOptions(options);
 
-        this._buttonClickEventHandler = this._handleButtlonClickedEvent(methodId);
-
-        if (this._paymentMethod.initializationData.nonce) {
-            return Promise.resolve(this._store.getState());
-        }
```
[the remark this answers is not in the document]
EduardCerv: I will restore this functionality, is not necessary to delete it.

src/payment/strategies/googlepay/googlepay-payment-strategy.spec.ts
EduardCerv: I try to check the flow when the `_getPayment` returns a empty nonce and enter into the function to display the wallet

EduardCerv:
Hi @icatalina thanks for you review I tried to track the flow in the `[remote.php](https://github.com/bigcommerce/bigcommerce/blob/master/app/controllers/Storefront/RemoteController.php)` but I cannot receive the any call so I update my pr with your recommendations, If you have another concern please let me know or give another approach to follow 🙏🏼

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
```diff
@@ -233,6 +243,11 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
             const paymentData = await this._googlePayPaymentProcessor.displayWallet();
             await this._paymentInstrumentSelected(paymentData, methodId);
 
+            if (this._paymentMethod?.initializationData.nonce) {
+                const state = await this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId));
+                this._paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(methodId);
+            }
+
```
icatalina, line 250: Trying to fetch a new nonce should not be a responsibility of `_displayWallet`. The `_displayWallet` method knows nothing about the nonce nor uses it. 

After a quick scan of the code, it looks like the logic that refetches the nonce should either go in `_getPayment` or `_getNonce`, which are the 2 methods that know and care about the nonce.

src/payment/strategies/googlepay/googlepay-payment-strategy.spec.ts
icatalina: The expect in this test is `displayWallet` to have been called. Display wallet is called with and without nonce, the only reason this test fails when the nonce is not present is because some code _after_ display wallet blows up.

We shouldn't be relying on other parts of the code for the check, we should be validating the _nonce_ used is the one provided as `mockReturnValue`: `newNonce`

EduardCerv: I add a new expect when the  when we have this error the method `getPaymentMethodOrThrow` should be called 3 times, this assert just happen when the payment method fails and we need to show the wallet again, let me know if this assertion sounds more sense for you or we need to test some else @icatalina 🙏🏼

src/payment/strategies/googlepay/googlepay-payment-strategy.ts
EduardCerv: I add a new expect when the  when we have this error the method `getPaymentMethodOrThrow` should be called 3 times, this assert just happen when the payment method fails and we need to show the wallet again, let me know if this assertion sounds more sense for you or we need to test some else @icatalina 🙏🏼

EduardCerv:
♻️  Could you help me with another review @icatalina? 🙏🏼

icatalina (approved):
Approving because this works and there's some pressure to fix this bug.

Talked over this offline, the ideal solution will be to move all the code related to the nonce to `_getPayment` and have the body of `execute` completely agnostic of it. 

We should probably refactor all this area to make the methods/responsibilities a bit more clear.

JoseBobadilla20:
💚 Tested 
QA Details: I tested the flow for these payment providers to check that works correctly: 
- Stripe
- StripeV3
- AdyenV2
- CybersourceV2
- Checkout.com
- Authorize.Net
- Braintree

PR #1150, opened 2021-06-02
feat(payment): PAYMENTS-6807 extend PPSDK payment strategy

leeBigCommerce:
**N.B.** All of this work is protected by a Feature Flag which prevents the PPSDK Payment Strategy from ever being selected.

## What?

- Flesh out PPSDK payment strategy to begin to handle our method's with an initialisation strategy of `NONE`
- Dead-ends on dummy processing API response handling (`stepHandler` - to be grown in future tickets)
- Extend `PaymentMethod` type to contain `initializationStrategy`
- Create `PPSDKPaymentMethod` sub type to enforce required PPSDK method attributes
- Not covered: Continue the journey from a returning redirect

## Why?

- Incrementally moving towards support for an offsite, redirecting PPSDK method

## Testing / Proof

- New tests
- Current tests passing


src/payment/payment-method.ts
```diff
@@ -12,5 +11,5 @@ export default interface PaymentMethod {
     nonce?: string;
     initializationData?: any;
     returnUrl?: string;
-    initializationStrategy?: InitialisationStrategies;
+    initializationStrategy?: unknown;
```
leeBigCommerce, line 14: As the API grows over time `initializationStrategy` will effectively be a union of different interfaces.

At this point we do not know exactly how many or in what variety, so we've chosen to type as `unknown` and to defensively type-guard in later code to filter supported `initializationStrategy`(s).

src/payment/strategies/ppsdk/ppsdk-payments-api-response.ts
```diff
@@ -0,0 +1,3 @@
+import { Response } from '@bigcommerce/request-sender';
+
+export type PaymentsAPIResponse = Response<unknown>;
```
leeBigCommerce, line 3: Similar to `initializationStrategy` - this API response will eventually be a large union of various responses. To defend against this `unknown` is being used to force defensive type guarding prior to any  real use.

src/payment/strategies/ppsdk/ppsdk-strategy.spec.ts
```diff
@@ -0,0 +1,78 @@
+import { createRequestSender } from '@bigcommerce/request-sender';
```
leeBigCommerce, line 1: Given that this strategy currently dead-ends to dummy handlers (`stepHandler` and `PaymentProcessor`) and the overall feature is still a WIP, this test seemed like enough to cover the entire strategy file for now.

src/payment/strategies/ppsdk/step-handler/step-handler.ts
```diff
@@ -0,0 +1,5 @@
+import { PaymentsAPIResponse } from '../ppsdk-payments-api-response';
```
leeBigCommerce, line 1: Dummy code which will be expanded in on a future ticket: https://jira.bigcommerce.com/browse/PAYMENTS-6805

This hander will then discern response types and act accordingly.

src/payment/strategies/ppsdk/ppsdk-strategy.ts
```diff
@@ -1,23 +1,50 @@
...
 
-    initialize(_options?: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
+    initialize(options?: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
+        const paymentMethod = getPPSDKMethod(this._store, options);
+
+        if (!paymentMethod) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+        }
+
+        this._paymentProcessor = getPaymentProcessor(paymentMethod);
```
davidchin, line 42: Prefer to inject dependencies to produce side effects. `getPaymentProcessor` is directly imported here so it's not possible to pass a different implementation in your tests besides using Jest mock. Generally speaking, we discourage the use of Jest mock because it requires you to peek into the implementation of your test subject in order to do the mocking; where as via constructor or method parameters, you can tell what to mock by looking at their interfaces.

src/payment/strategies/ppsdk/ppsdk-payment-processor.ts
```diff
@@ -0,0 +1,18 @@
+import { OrderPaymentRequestBody } from '../../../order';
+import { PPSDKPaymentMethod } from '../../ppsdk-payment-method';
+
+import { initializeNone, isNone } from './initializers';
+
+export type PaymentProcessor = (payment: OrderPaymentRequestBody | undefined) => Promise<void>;
+
+type GetPaymentProcessor = (paymentMethod: PPSDKPaymentMethod) => PaymentProcessor | undefined;
+
+export const getPaymentProcessor: GetPaymentProcessor = paymentMethod => {
```
davidchin, line 10: It seems to me that this is just another resolver, which mean you can probably just use the existing `Registry` class to achieve what you want instead of creating another concrete implementation. e.g.: You can create a registry of all the PPSDK processors/handlers and resolve them based on their PPSDK strategy type.

davidchin (commented):
Looking good! I just have some comments and suggestions for you to think about :)

src/payment/payment-method.ts
leeBigCommerce: Unfortunately, `initializationStrategy` won't be a string, but a union of various objects. Something like:

```
interface BaseStrategy {
    type: string;
}

interface StrategyA extends BaseStrategy {
    // attributes and values specific to only StrategyA
}

interface StrategyB extends BaseStrategy {
    // attributes and values specific to only StrategyB
}

interface StrategyC extends BaseStrategy {
    // attributes and values specific to only StrategyC
}

type initializationStrategy = StrategyA | StrategyB | StrategyC;
```

And we don't know what the final union or shapes will be yet.

---

Typing it as union of what we know now, e.g.:

`type initializationStrategy = StrategyA | StrategyB`

Would lead to buggy behaviour once the same code also started receiving `StrategyC` from the updated (but not versioned) API.

src/payment/strategies/ppsdk/ppsdk-payments-api-response.ts
leeBigCommerce: Similarly to https://github.com/bigcommerce/checkout-sdk-js/pull/1150/files#r649623606, we don't know what the final union or competent shapes will be. 

`unknown` seemed safest, forcing the use of type guards later in the consuming code.

src/payment/strategies/ppsdk/initializers/none.ts
```diff
@@ -0,0 +1,23 @@
...
+
+interface None {
+    type: 'NONE';
+}
+
+export const isNone = (unknownStrategy: unknown): unknownStrategy is None =>
+    isEqual(unknownStrategy, { type: 'NONE' });
+
+export const initializeNone = (paymentMethod: PPSDKPaymentMethod): PaymentProcessor => {
+    const requestSender = createRequestSender();
```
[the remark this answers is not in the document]
leeBigCommerce: Done 👍  

injected: https://github.com/bigcommerce/checkout-sdk-js/pull/1150/files#diff-d42088af7e8a0106b9a187cf956d965bc6d677dfa26f573607c994dbd3a75a62R395
consumed: https://github.com/bigcommerce/checkout-sdk-js/pull/1150/files#diff-33de44235f9dcd9db1207d66aad23ba5747c08b4a53725f1726265a71b65afd8R8

src/payment/strategies/ppsdk/initializers/none.ts
```diff
@@ -0,0 +1,23 @@
...
+    const paymentMethodId = `${paymentMethod.id}.${paymentMethod.method}`;
+    const body = JSON.stringify({ payment_method_id: paymentMethodId });
```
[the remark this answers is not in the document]
leeBigCommerce: Done: https://github.com/bigcommerce/checkout-sdk-js/pull/1150/files#diff-33de44235f9dcd9db1207d66aad23ba5747c08b4a53725f1726265a71b65afd8R16

src/payment/strategies/ppsdk/initializers/none.ts
```diff
@@ -0,0 +1,23 @@
...
+import { stepHandler } from '../step-handler';
+
+interface None {
+    type: 'NONE';
+}
+
+export const isNone = (unknownStrategy: unknown): unknownStrategy is None =>
+    isEqual(unknownStrategy, { type: 'NONE' });
+
+export const initializeNone = (paymentMethod: PPSDKPaymentMethod): PaymentProcessor => {
```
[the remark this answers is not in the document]
leeBigCommerce: `stepHandler` will be used by each different initialisation strategy - it will consume the polymorphic `/payments` API responses and turn them into necessary action (redirecting, portraying success/failure etc).

Let me know if it makes more sense now things have been reorganised 👍

src/payment/strategies/ppsdk/ppsdk-payment-processor.ts
```diff
@@ -0,0 +1,18 @@
+import { OrderPaymentRequestBody } from '../../../order';
+import { PPSDKPaymentMethod } from '../../ppsdk-payment-method';
+
+import { initializeNone, isNone } from './initializers';
+
+export type PaymentProcessor = (payment: OrderPaymentRequestBody | undefined) => Promise<void>;
```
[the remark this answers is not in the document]
leeBigCommerce: It is a bit of a catch all.

Each payment processor will kick off the initial call to the `/payments` processing API, then handle its response(s) (via the `stepHandler`). Once all the responses have been dealt with it will terminate in either success (resolving) or error/failure (rejecting).

Given that success signifies that BigPay has updated the order to 'paid' (side-effect), we don't really have anything to return.

src/payment/strategies/ppsdk/step-handler/step-handler.ts
```diff
@@ -0,0 +1,5 @@
...
+
+export const stepHandler = (_stepResponse: PaymentsAPIResponse): Promise<void> => {
```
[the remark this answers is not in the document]
leeBigCommerce: 👍  bit of a left-over

src/payment/strategies/ppsdk/ppsdk-strategy.ts
leeBigCommerce: Done: https://github.com/bigcommerce/checkout-sdk-js/pull/1150/files#diff-d42088af7e8a0106b9a187cf956d965bc6d677dfa26f573607c994dbd3a75a62R395

src/payment/strategies/ppsdk/ppsdk-payment-processor.ts
leeBigCommerce: Done: https://github.com/bigcommerce/checkout-sdk-js/pull/1150/files#diff-6df6aa7ed80a40f81baf653e8cd4c02262704158ca6dc151b357c4fdd4db8a39R8

src/payment/strategies/ppsdk/create-ppsdk-payment-processor-registry.ts
```diff
@@ -0,0 +1,22 @@
...
+import { PaymentProcessorRegistry } from './ppsdk-payment-processor-registry';
+import { PaymentProcessorType } from './ppsdk-payment-processor-type';
+import { StepHandler } from './step-handler';
+
+export const createPaymentProcessorRegistry = (requestSender: RequestSender, stepHandler: StepHandler) => {
+    const registry = new PaymentProcessorRegistry();
+
+    registry.register(
+        PaymentProcessorType.NONE,
+        () => new NonePaymentProcessor(requestSender, stepHandler)
```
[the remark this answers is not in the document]
leeBigCommerce: For posterity: any initialiser could receive any of the possible responses from the processing API.

As such there will be one `StepHandler` used by all initialisers. The structure of that `StepHandler` will be fleshed out in future tickets.

src/payment/strategies/ppsdk/payment-processors/none-payment-processor.ts
```diff
@@ -0,0 +1,21 @@
...
+    constructor(
+        private _requestSender: RequestSender,
+        private _stepHandler: StepHandler
+    ) {}
+
+    process(paymentMethod: PPSDKPaymentMethod) {
+        const paymentMethodId = `${paymentMethod.id}.${paymentMethod.method}`;
+        const body = { payment_method_id: paymentMethodId };
+
+        return this._requestSender.post<PaymentsAPIResponse>('/payments', { body })
```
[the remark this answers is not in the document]
leeBigCommerce: Good catch, it's now updated to use the bigpay endpoint.

Used: https://github.com/bigcommerce/checkout-sdk-js/pull/1150/files#diff-33de44235f9dcd9db1207d66aad23ba5747c08b4a53725f1726265a71b65afd8R17
Passed in from the strategy: https://github.com/bigcommerce/checkout-sdk-js/pull/1150/files#diff-5b451c15e917a03094dd5d38df6c638a374c0b98d83522b1934dfac06652f659R21

src/payment/strategies/ppsdk/initializers/none.ts
```diff
@@ -0,0 +1,23 @@
...
+    const requestSender = createRequestSender();
+    const paymentMethodId = `${paymentMethod.id}.${paymentMethod.method}`;
+    const body = JSON.stringify({ payment_method_id: paymentMethodId });
+
+    return () => requestSender.post<PaymentsAPIResponse>('/payments', { body })
```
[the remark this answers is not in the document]
leeBigCommerce: Fixed: https://github.com/bigcommerce/checkout-sdk-js/pull/1150#discussion_r652423178

src/payment/payment-method.ts
leeBigCommerce: I've gone for 

```
interface UnknownObject {
    [key: string]: unknown;
}
export default interface InitializationStrategy extends Partial<UnknownObject> {
    type: string;
}
```

Which gives us the benefit of having `type` ensured, but also represents the 'yet to be clarified' nature of any other attributes. The latter paints a truer picture of the API and enforces later Typeguard use. (Without a union or `unknown` or `any` Typeguards would be out of place - only explained with domain knowledge).

It also means typeguards can interrogate possible properties on a strategy without type errors or needing to be overly reliant on just the `type` attribute. The one draw back was a slight 'hack' of using `pick` on the None typeguard (https://github.com/bigcommerce/checkout-sdk-js/pull/1150/files#diff-4dbdf09949352589db42a49b9d883d5c1f3713df9d0c4de312ba575dd8cd5cf8R6), but this should be the only exception.

What do you recon? I'm on the pessimistic/defensive side of Typing 😄

src/payment/strategies/ppsdk/ppsdk-payment-processor-registry.ts
```diff
@@ -0,0 +1,25 @@
...
+import { PaymentProcessorType } from './ppsdk-payment-processor-type';
+
+const getToken = cond([
+    [isNone, constant(PaymentProcessorType.NONE)],
+    [stubTrue, constant(PaymentProcessorType.UNSUPPORTED)],
+]);
+
+export class PaymentProcessorRegistry extends Registry<PaymentProcessor | undefined, PaymentProcessorType> {
+    constructor() {
+        super();
```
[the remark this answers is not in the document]
leeBigCommerce: Fixed 👍

davidchin (approved):
Thanks for working though all the points above with me. LGTM! 👍

PR #1151, opened 2021-06-08
feat(payment): INT-4222 Add vaulting compatibility to Moneris

FranciscoOnt:
## What? [(INT-4222)](https://jira.bigcommerce.com/browse/INT-4222)
Update strategy to support vaulted instruments and add Moneris to supported instruments.

## Why?
So vaulted instruments can be used with Moneris.

## Testing / Proof
UInit Testing

### DIFF
--- a/src/payment/strategies/moneris/moneris-payment-initialize-options.ts
+++ b/src/payment/strategies/moneris/moneris-payment-initialize-options.ts
@@ -1,10 +1,17 @@
+import { HostedFormValidationOptions } from '../../../hosted-form';
+
 import MonerisStylingProps from './moneris';
 
-export default interface MonerisaymentInitializeOptions {
+export default interface MonerisPaymentInitializeOptions {
     /**
      * The ID of a container where the Moneris iframe component should be mounted
      */
     containerId: string;
 
     style?: MonerisStylingProps;
+
+    /**
+     * Hosted Form Validation Options
+     */
+    form?: HostedFormValidationOptions;
 }
--- a/src/payment/strategies/moneris/moneris-payment-strategy.ts
+++ b/src/payment/strategies/moneris/moneris-payment-strategy.ts
@@ -2,42 +2,61 @@ import { map } from 'lodash';
 
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
 import { InvalidArgumentError , MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType } from '../../../common/error/errors';
-import { OrderActionCreator, OrderRequestBody } from '../../../order';
+import { HostedForm, HostedFormFactory, HostedFormValidationOptions } from '../../../hosted-form';
+import { OrderActionCreator, OrderPaymentRequestBody, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
 import { StoreCreditActionCreator } from '../../../store-credit';
 import { PaymentArgumentInvalidError } from '../../errors';
+import isVaultedInstrument from '../../is-vaulted-instrument';
+import { HostedInstrument } from '../../payment';
 import PaymentActionCreator from '../../payment-action-creator';
 import { PaymentInitializeOptions } from '../../payment-request-options';
 import PaymentStrategy from '../payment-strategy';
 
 import MonerisStylingProps,  { MoneriesHostedFieldsQueryParams, MonerisInitializationData, MonerisResponseData } from './moneris';
+import MonerisPaymentInitializeOptions from './moneris-payment-initialize-options';
 
 const IFRAME_NAME = 'moneris-payment-iframe';
 const RESPONSE_SUCCESS_CODE = '001';
 
 export default class MonerisPaymentStrategy implements PaymentStrategy {
     private _iframe?: HTMLIFrameElement;
+    private _initializeOptions?: MonerisPaymentInitializeOptions;
     private _windowEventListener?: (response: MessageEvent) => void;
 
+    private _hostedForm?: HostedForm;
+
     constructor(
+        private _hostedFormFactory: HostedFormFactory,
         private _store: CheckoutStore,
         private _orderActionCreator: OrderActionCreator,
         private _paymentActionCreator: PaymentActionCreator,
         private _storeCreditActionCreator: StoreCreditActionCreator
     ) {}
 
-    initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
+    async initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
         const state = this._store.getState();
 
-        const { moneris: monerisOptions } = options;
-        const { config, initializationData } = state.paymentMethods.getPaymentMethodOrThrow(options.methodId);
+        const { moneris: monerisOptions, methodId } = options;
+
+        if (!methodId) {
+            throw new InvalidArgumentError('Unable to initialize payment because "methodId" argument is not provided.');
+        }
 
         if (!monerisOptions) {
             throw new InvalidArgumentError('Unable to initialize payment because "options.moneris" argument is not provided.');
         }
 
+        this._initializeOptions = monerisOptions;
+
+        const { config, initializationData } = state.paymentMethods.getPaymentMethodOrThrow(methodId);
+
         if (!initializationData?.profileId) {
             throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        if (monerisOptions.form  && this._shouldShowTSVHostedForm(methodId)) {
+            this._hostedForm = await this._mountCardVerificationfields(monerisOptions.form);
         }
 
         if (!this._iframe) {
@@ -49,14 +68,10 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
 
     async execute(payload: OrderRequestBody, options?: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
         const { payment , ...order } = payload;
-        const { paymentMethods: { getPaymentMethodOrThrow } } = this._store.getState();
 
         if (!payment) {
             throw new PaymentArgumentInvalidError(['payment']);
         }
-
-        const paymentMethod = getPaymentMethodOrThrow(payment.methodId);
-        const testMode = paymentMethod.config.testMode;
 
         const { isStoreCreditApplied: useStoreCredit } = this._store.getState().checkout.getCheckoutOrThrow();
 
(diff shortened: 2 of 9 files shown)


JoseBobadilla20:
💚 Tested 
QA Details: I check that the Moneris vaulting flow works correctly.

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -55,6 +57,7 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
 
         const paymentMethod = getPaymentMethodOrThrow(payment.methodId);
         const testMode = paymentMethod.config.testMode;
+        const shouldSaveInstrument = payment.paymentData && (payment.paymentData as NonceInstrument).shouldSaveInstrument;
```
vmparra, line 60: You should first validate that instrument is in fact a nonce. See https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/payment/is-nonce-like.ts

vmparra (changes requested):
Hi @FranciscoOnt So far looks real good, just left one comment. Please let me know if you have questions.

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -87,7 +94,7 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
         if (nonce !== undefined) {
             return this._store.dispatch(this._paymentActionCreator.submitPayment({
                 methodId: payment.methodId,
-                paymentData: { nonce },
+                paymentData: { nonce, shouldSaveInstrument },
```
JoseAlvarezHdz, line 97: i think you are missing the `shouldSetAsDefaultInstrument` param 🤔

src/payment/strategies/moneris/moneris-payment-strategy.ts
FranciscoOnt: I don't think that will work, the `shouldSaveIntrument` is received when the payment is not yet a nonce instrument and since the `isNonceLike` validates the `nonce` field which is missing from this initial payload it will alway return false in the Moneris scenario then making the `shouldSaveIntrument`always undefined.

FranciscoOnt: Just to clarify that, I changed to the `HostedInstrument` Type

JoseAlvarezHdz (approved):
LGTM 👍🏾

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -47,22 +66,55 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
...
+
+        return Promise.resolve(this._store.getState());
+    }
+
+    protected async executeWithCC(payment: OrderPaymentRequestBody): Promise<InternalCheckoutSelectors> {
+        const { paymentMethods: { getPaymentMethodOrThrow } } = this._store.getState();
+        const paymentMethod = getPaymentMethodOrThrow(payment.methodId);
+
+        const testMode = paymentMethod.config.testMode;
+        const shouldSaveInstrument = payment.paymentData && (payment.paymentData as HostedInstrument)?.shouldSaveInstrument;
```
vmparra, line 116: You're still missing the type guard. You should also cast `paymentData` to HostedInstrument only once.

vmparra (changes requested):
It looks real good however is would be wise to have a type guard for HostedInstrument and then cast the object.

vmparra (approved):
Great work @FranciscoOnt 

image

JoseAlvarezHdz (approved):
LGTM! 👍🏾  nice!

JoseBobadilla20:
💚 Tested
QA Details: I check that the Moneris vaulting flow and TVS flow works correctly.

Tharaae (commented):
Generally LGTM, Good job 👍 
But please add the testing proof to the PR description while I have another more detailed look.

Tharaae (approved):
Checked the testing proof on the sibling checkout-js PR. LGTM 👍

Tharaae (approved):

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -89,29 +141,74 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
...
+            return await this._store.dispatch(this._orderActionCreator.loadCurrentOrder());
+        } else {
+            return await this._store.dispatch(this._paymentActionCreator.submitPayment(payment));
         }
+    }
 
-        if (this._iframe && this._iframe.parentNode) {
-            this._iframe.parentNode.removeChild(this._iframe);
-            this._iframe = undefined;
+    private shouldShowTSVHostedForm(methodId: string): boolean {
```
animesh1987, line 168: This method should be prefixed with a `_`.

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -49,22 +68,55 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
...
+
+        if (this._iframe && this._iframe.parentNode) {
+            this._iframe.parentNode.removeChild(this._iframe);
+            this._iframe = undefined;
+        }
+
+        return Promise.resolve(this._store.getState());
+    }
+
+    protected async executeWithCC(payment: OrderPaymentRequestBody): Promise<InternalCheckoutSelectors> {
```
animesh1987, line 113: Why is this a protected method and not a private? and can we please add a `_` as a prefix.

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -89,29 +141,74 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
...
             }));
         }
 
         return this._store.getState();
     }
 
-    finalize(): Promise<InternalCheckoutSelectors> {
-        return Promise.reject(new OrderFinalizationNotRequiredError());
-    }
+    protected async executeWithVaulted(payment: OrderPaymentRequestBody): Promise<InternalCheckoutSelectors> {
```
animesh1987, line 151: Why is this a protected method and not a private? and can we please add a _ as a prefix.

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -89,29 +141,74 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
...
-            window.removeEventListener('message', this._windowEventListener);
-            this._windowEventListener = undefined;
+            if (!form) {
+                throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+            }
+
+            await form.validate();
+            await form.submit(payment);
+
+            return await this._store.dispatch(this._orderActionCreator.loadCurrentOrder());
```
animesh1987, line 162: Why do we need to load the current order here?

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -89,29 +141,74 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
...
+    private _getInitializeOptions(): MonerisPaymentInitializeOptions {
+        if (!this._initializeOptions) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
         }
 
-        return Promise.resolve(this._store.getState());
+        return this._initializeOptions;
+    }
+
+    private _mountCardVerificationfields(formOptions: HostedFormOptions): Promise<HostedForm> {
```
animesh1987, line 193: The new promise construction is redundant, you can move `async` keyword on `_mountCardVerificationfields ` function and it should be good.

src/payment/strategies/moneris/moneris-payment-initialize-options.ts
```diff
@@ -1,10 +1,17 @@
...
      * The ID of a container where the Moneris iframe component should be mounted
      */
     containerId: string;
 
     style?: MonerisStylingProps;
+
+    /**
+     * Hosted Form Validation Options
+     */
+    form?: HostedFormOptions;
```
animesh1987, line 16: Seems like the change you are doing here is to use the hosted form fields for vaulted instrument verification and the `HostedFormOptions` accepts parameters more than just instrument verification. For reference you can see this https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/hosted-form/hosted-form-options.ts#L20. As far as I can tell if the client passes those options it will not work. But here you are looking for https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/hosted-form/hosted-form-options.ts#L29 this particular interface to be accepted.

So maybe you need to describe a new interface to check for the supplied parameters?

src/payment/strategies/moneris/moneris-payment-strategy.ts
FranciscoOnt: It's used to update the state of the checkout and check if the payment for the order has been submitted, neither `form.validate` and `form.submit` update the state.
This is what the base credit-card strategy does.

src/payment/strategies/moneris/moneris-payment-strategy.ts
FranciscoOnt: if the promise isn't marked as async, the line with `await form.attach();`  will be marked as an error, since the arrow function make it be in a different scope that would no longer be async.

FranciscoOnt:
@animesh1987 Could you please re-review ♻️

src/payment/strategies/moneris/moneris-payment-strategy.ts
animesh1987: Have you tried it removing the promise and putting `async` next to `_mountCardVerificationfields `?

FranciscoOnt: yes, that's what I tried, but it didn't work

PascalZajac:
Looks like this still needs final review from @bigcommerce/checkout before it's actually ready to merge.

src/payment/strategies/moneris/moneris-payment-strategy.ts
animesh1987: Have you tried this 
```
private async _mountCardVerificationfields(formOptions: HostedFormOptions): Promise<HostedForm> {
        try {
            const { config } = this._store.getState();
            const bigpayBaseUrl = config.getStoreConfig()?.paymentSettings.bigpayBaseUrl;

            if (!bigpayBaseUrl) {
            }

            const form = this._hostedFormFactory.create(bigpayBaseUrl, formOptions);

            await form.attach();

            return form;
        } catch (error) {
            return error;
        }
    }
```

FranciscoOnt: that does, work, I've added it to the PR

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -89,29 +141,72 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
...
+    private _getInitializeOptions(): MonerisPaymentInitializeOptions {
+        if (!this._initializeOptions) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
         }
 
-        return Promise.resolve(this._store.getState());
+        return this._initializeOptions;
+    }
+
+    private async _mountCardVerificationfields(formOptions: HostedFormOptions): Promise<HostedForm> {
```
animesh1987, line 193: Here we accept `HostedFormOptions` but `monerisOptions.form` is a different interface `HostedFormValidationOptions`. So maybe we need to adjust the types in all relevant places.

src/payment/strategies/moneris/moneris-payment-strategy.ts
animesh1987: Thanks for updating that.

animesh1987 (approved):

PR #1155, opened 2021-06-15
CHECKOUT-5873: Deploy compiled assets to GCS

davidchin:
## What?
Deploy compiled assets to GCS using the microapp deployment strategy rather than using the container-based approach.

## Why?
To reduce the overhead of deploying and running a container. This switch will also solve some of the browser and edge caching problems that we experienced with the recent change in CDN provider.

## Testing / Proof
I used the payment step to test this change because it loads Checkout SDK via CDN.

### Chrome
Screen Shot 2021-06-15 at 2 11 32 pm

### IE11
Screen Shot 2021-06-15 at 2 11 05 pm


.circleci/config.yml
```diff
@@ -139,3 +167,26 @@ workflows:
           requires:
             - approve
             - release
+      - build_cdn:
```
davidchin, line 170: I'm keeping `build_server` step and the code related to it until everything is tested and ready to be switched over from the old deployment strategy to the new.

davidchin (commented):
@bigcommerce/checkout this is ready for review. I haven't been able to test it end-to-end yet considering some of those steps require this PR to be merged to master. I'll try to figure out how to do it in the meantime.

.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ aliases:
 version: 2.1
 
 orbs:
-  ci: bigcommerce/internal@volatile
+  ci: bigcommerce/internal@dev:PR-127
```
davidchin, line 20: https://github.com/bigcommerce/circle-orbs/pull/127

.circleci/config.yml
```diff
@@ -139,3 +157,23 @@ workflows:
           requires:
             - approve
             - release
+      - build_cdn:
+          filters:
+            <<: *release_filter
```
icatalina, line 162: Do we want to filter by branch? I was under the impression launchbay will take care of it from now onwards...

icatalina (approved):

.circleci/config.yml
davidchin: Hrm I supposed we don't need the filter here because this task requires `release` task to complete, which already has a branch filter. But the idea is that it `release` and `build_cdn` can be skipped on PR build.

icatalina: Oh! because you want to keep the release numbers. So you only want to release certain versions. This is different from the rest of our apps. I guess there's no problem in doing it this way 🤷

davidchin: Cool. Anyway, I've decided to cleanup the filter regardless. 🙂

PR #1156, opened 2021-06-16
feat(payment): INT-4423 Set labels to Moneris iframe creation process

cmedinilla:
## What?
 - Add labels to Moneris iframe creation process
 -  Pass translated labels to Moneris iframe

## Why?
 - Labels must be translated

## Sibling PRs
~~https://github.com/bigcommerce/bigcommerce/pull/41468~~

## Testing / Proof
Unit Test

### DIFF
--- a/src/payment/strategies/moneris/moneris-payment-strategy.spec.ts
+++ b/src/payment/strategies/moneris/moneris-payment-strategy.spec.ts
@@ -156,7 +156,7 @@ describe('MonerisPaymentStrategy', () => {
 
             const iframe = document.getElementById(iframeId) as HTMLIFrameElement;
             expect(iframe).toBeTruthy();
-            expect(iframe.src).toMatch(/www3.moneris.com/);
+            expect(iframe.src).toContain('www3.moneris.com');
         });
 
         it('initializes moneris iframe and sets src to the test environment', async () => {
@@ -165,7 +165,62 @@ describe('MonerisPaymentStrategy', () => {
 
             const iframe = document.getElementById(iframeId) as HTMLIFrameElement;
             expect(iframe).toBeTruthy();
-            expect(iframe.src).toMatch(/esqa.moneris.com/);
+            expect(iframe.src).toContain('esqa.moneris.com');
+        });
+
+        it('initialize moneris iframe and sets data labels from initialization data', async () => {
+            paymentMethodMock.config.testMode = true;
+            await expect(strategy.initialize(initializeOptions)).resolves.toEqual(store.getState());
+
+            const iframe = document.getElementById(iframeId) as HTMLIFrameElement;
+            expect(iframe).toBeTruthy();
+            expect(iframe.src).toContain('pan_label=Credit%20Card&exp_label=Expiration%20Date&cvd_label=CVV');
+        });
+
+        it('initialize moneris iframe and sets data labels if initialization data is null', async () => {
+            paymentMethodMock.config.testMode = true;
+            paymentMethodMock.initializationData = { profileId: 'ABC123' };
+            await expect(strategy.initialize(initializeOptions)).resolves.toEqual(store.getState());
+
+            const iframe = document.getElementById(iframeId) as HTMLIFrameElement;
+            expect(iframe).toBeTruthy();
+            expect(iframe.src).toContain('pan_label=Credit%20Card%20Number&exp_label=Expiration&cvd_label=CVD');
+        });
+
+        it('initialize moneris iframe and sets expiry field', async () => {
+            paymentMethodMock.config.testMode = true;
+            await expect(strategy.initialize(initializeOptions)).resolves.toEqual(store.getState());
+
+            const iframe = document.getElementById(iframeId) as HTMLIFrameElement;
+            expect(iframe).toBeTruthy();
+            expect(iframe.src).toContain('enable_exp=1');
+        });
+
+        it('initialize moneris iframe and sets cvv field', async () => {
+            paymentMethodMock.config.testMode = true;
+            await expect(strategy.initialize(initializeOptions)).resolves.toEqual(store.getState());
+
+            const iframe = document.getElementById(iframeId) as HTMLIFrameElement;
+            expect(iframe).toBeTruthy();
+            expect(iframe.src).toContain('enable_cvd=1');
+        });
+
+        it('initialize moneris iframe and sets labels enabled', async () => {
+            paymentMethodMock.config.testMode = true;
+            await expect(strategy.initialize(initializeOptions)).resolves.toEqual(store.getState());
+
+            const iframe = document.getElementById(iframeId) as HTMLIFrameElement;
+            expect(iframe).toBeTruthy();
+            expect(iframe.src).toContain('display_labels=1');
+        });
+
+        it('initialize moneris iframe and sets hosted fields css', async () => {
+            paymentMethodMock.config.testMode = true;
+            await expect(strategy.initialize(initializeOptions)).resolves.toEqual(store.getState());
+
+            const iframe = document.getElementById(iframeId) as HTMLIFrameElement;
+            expect(iframe).toBeTruthy();
+            expect(iframe.src).toContain('css_body=background:white;&css_textbox=border-width:2px;&css_textbox_pan=width:140px;&css_textbox_exp=width:40px;&css_textbox_cvd=width:40px');
         });
 
         it('fails to initialize moneris strategy when initialization options are not provided', () => {
--- a/src/payment/strategies/moneris/moneris-payment-strategy.ts
+++ b/src/payment/strategies/moneris/moneris-payment-strategy.ts
@@ -1,3 +1,5 @@
+import { map } from 'lodash';
+
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
 import { InvalidArgumentError , MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType } from '../../../common/error/errors';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
@@ -8,7 +10,7 @@ import PaymentActionCreator from '../../payment-action-creator';
 import { PaymentInitializeOptions } from '../../payment-request-options';
 import PaymentStrategy from '../payment-strategy';
 
-import MonerisStylingProps,  { MonerisResponseData } from './moneris';
+import MonerisStylingProps,  { MoneriesHostedFieldsQueryParams, MonerisInitializationData, MonerisResponseData } from './moneris';
 
 const IFRAME_NAME = 'moneris-payment-iframe';
 const RESPONSE_SUCCESS_CODE = '001';
@@ -39,7 +41,7 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
         }
 
         if (!this._iframe) {
-            this._iframe = this._createIframe(monerisOptions.containerId, initializationData.profileId, !!config.testMode);
+            this._iframe = this._createIframe(monerisOptions.containerId, initializationData, !!config.testMode);
         }
 
         return Promise.resolve(this._store.getState());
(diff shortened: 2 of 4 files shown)


FranciscoOnt (approved):
LGTM 👍

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -39,7 +39,13 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
...
 
         if (!this._iframe) {
-            this._iframe = this._createIframe(monerisOptions.containerId, initializationData.profileId, !!config.testMode);
+            const monerisLabels: MonerisHostedFieldsLabels = {
+                panLabel: initializationData.creditCardLabel,
+                expLabel: initializationData.expiryDateLabel,
+                cvdLabel: initializationData.cvdLabel,
+            };
+
+            this._iframe = this._createIframe(monerisOptions.containerId, initializationData.profileId, !!config.testMode, monerisLabels);
```
icatalina, line 48: This seems a bit of extra work, we are creating an object here and then creating the defaults inside the `_createIframe` function. Considering how much of the `initializationData` is needed in `_createIframe`, why don't we just pass it all in and do all the transformations inside it?

src/payment/strategies/moneris/moneris-payment-strategy.spec.ts
```diff
@@ -168,6 +168,15 @@ describe('MonerisPaymentStrategy', () => {
             expect(iframe.src).toMatch(/esqa.moneris.com/);
         });
 
+        it('initialize moneris iframe and sets initailization data labels', async () => {
+            paymentMethodMock.config.testMode = true;
+            await expect(strategy.initialize(initializeOptions)).resolves.toEqual(store.getState());
+
+            const iframe = document.getElementById(iframeId) as HTMLIFrameElement;
+            expect(iframe).toBeTruthy();
+            expect(iframe.src).toMatch(/&display_labels=1&pan_label=Credit%20Card%20Number&exp_label=Expiration&cvd_label=CVV/);
```
icatalina, line 177: This doesn't really uses a regex. If we want to do partial matching we could use `expect.stringContaining(string)`: https://jestjs.io/docs/expect#expectstringcontainingstring

```ts
expect(iframe).toHaveAttribute('src', expect.stringContaining('&display_labels=1&pan_label=Credit%20Card%20Number&exp_label=Expiration&cvd_label=CVV'));
```

I noticed we use to match in other tests, I feel like using an exact string match is better in this case.

cmedinilla: I changed tu use `.tocontain()`
https://jestjs.io/docs/expect#tocontainitem

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -127,10 +133,15 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
...
+        const panLabel = labels?.panLabel ?? 'Credit Card Number';
+        const expLabel = labels?.expLabel ?? 'Expiration';
+        const cvdLabel = labels?.cvdLabel ?? 'CVV';
+        const hostedFieldsLabels = `display_labels=1&pan_label=${panLabel}&exp_label=${expLabel}&cvd_label=${cvdLabel}`;
+
         iframe.name = IFRAME_NAME;
         iframe.id = IFRAME_NAME;
 
-        iframe.src = `${ this._monerisURL(testMode) }?id=${profileId}&pmmsg=true&css_body=${cssBody}&css_textbox=${cssTextbox}&css_textbox_pan=${csstextboxPan}&enable_exp=1&css_textbox_exp=${cssTextboxExpiry}&enable_cvd=1&css_textbox_cvd=${csstexboxCvd}&display_labels=1`;
+        iframe.src = `${ this._monerisURL(testMode) }?id=${profileId}&pmmsg=true&css_body=${cssBody}&css_textbox=${cssTextbox}&css_textbox_pan=${csstextboxPan}&enable_exp=1&css_textbox_exp=${cssTextboxExpiry}&enable_cvd=1&css_textbox_cvd=${csstexboxCvd}&${hostedFieldsLabels}`;
```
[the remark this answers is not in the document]
cmedinilla: Using `Object.entries` is returning a compiler error so i changed to use `Object.keys` instead

src/payment/strategies/moneris/moneris-payment-strategy.spec.ts
icatalina: oh! I missed this, I thought it was only for arrays.
> `.toContain` can also check whether a string is a substring of another string.

src/payment/strategies/moneris/moneris-payment-strategy.ts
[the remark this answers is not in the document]
icatalina: That works as well 🙇 The purpose of this change is to make the code more readable, both `Object.keys` and `Object.entries` should help with that.

cmedinilla:
♻️  @icatalina  ready for another round, i added a negative test too

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -127,10 +127,20 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
         const cssTextboxExpiry = style?.cssTextboxExpiry ?? 'width:40px;';
         const csstexboxCvd = style?.csstexboxCvd ?? 'width:40px';
 
+        const monerisLabels: MonerisHostedFieldsLabels = {
+            pan_label: initializationData?.creditCardLabel || 'Credit Card Number',
+            exp_label: initializationData?.expiryDateLabel || 'Expiration',
+            cvd_label: initializationData?.cvdLabel || 'CVD',
```
icatalina, line 133: Should these only be optional as options? Based on how the query param was written, I would assume they are all required in order to generate the query param. Otherwise we could end up with things like `pan_label=undefined`.

If so, we can force all the previously optional values to be mandatory by wrapping `MonerisHostedFieldsLabels` in `Required`: `Required<MonerisHostedFieldsLabels>` so no value can be `undefined`.

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -127,10 +127,20 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
...
+        };
+
+        const labelsQueryString = Object.keys(monerisLabels)
+            .map(key => `${key}=${monerisLabels[key as keyof MonerisHostedFieldsLabels]}`)
+            .join('&');
```
icatalina, line 138: Sorry, I didn't realise using `Object.keys` will make `as keyof` required. Maybe use `lodash.map`?

```ts
import { map } from 'lodash';

// ...

const labelsQueryString = map(monerisLabels, (value, key) => `${key}=${value}`).join('&');
```

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -127,10 +127,20 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
...
+
         iframe.name = IFRAME_NAME;
         iframe.id = IFRAME_NAME;
 
-        iframe.src = `${ this._monerisURL(testMode) }?id=${profileId}&pmmsg=true&css_body=${cssBody}&css_textbox=${cssTextbox}&css_textbox_pan=${csstextboxPan}&enable_exp=1&css_textbox_exp=${cssTextboxExpiry}&enable_cvd=1&css_textbox_cvd=${csstexboxCvd}&display_labels=1`;
+        iframe.src = `${ this._monerisURL(testMode) }?id=${profileId}&pmmsg=true&css_body=${cssBody}&css_textbox=${cssTextbox}&css_textbox_pan=${csstextboxPan}&enable_exp=1&css_textbox_exp=${cssTextboxExpiry}&enable_cvd=1&css_textbox_cvd=${csstexboxCvd}&${labelsQueryString}&display_labels=1`;
```
icatalina, line 143: The goal with the change above (using `Object.entries` or `map` was to remove this string and make them all just an object. Can we move the rest of the entries in this list to a config object? We can also create a type to make sure the entries correspond with what's expected in the url.

```ts
interface MoneriesQueryParams {
}

const queryParams = {
	id: profileId,
    pmmsg: true,
    // ...
    display_labels: 1,
};

const queryString = map(queryParams, (value, key) => `${key}=${value}`).join('&');

iframe.src = `${ this._monerisURL(testMode) }?${ queryString }`;
```

cmedinilla:
♻️  @icatalina

src/payment/strategies/moneris/moneris.ts
```diff
@@ -6,6 +6,22 @@ export default interface MonerisStylingProps {
...
+    css_textbox: string;
+    css_textbox_pan: string;
+    enable_exp: number;
+    css_textbox_exp: string;
+    enable_cvd: number;
+    css_textbox_cvd: string;
+    display_labels: number;
+    pan_label?: string;
+    exp_label?: string;
+    cvd_label?: string;
```
icatalina, line 22: These are actually not optional in `MoneriesHostedFieldsQueryParams`. Now that we've defined a type just for the query params, there's no need to use `Required`, just make them mandatory here.

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -112,25 +114,36 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
         return Promise.resolve(this._store.getState());
     }
 
-    private _createIframe(containerId: string, profileId: string, testMode: boolean, style?: MonerisStylingProps): HTMLIFrameElement {
+    private _createIframe(containerId: string, profileId: string, testMode: boolean, initializationData: any, style?: MonerisStylingProps): HTMLIFrameElement {
```
icatalina, line 117: Why are we using `any` as initialization data? can we do better than that and provide some type that better matches what we expect?

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -112,25 +114,36 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
...
+            cvd_label: initializationData?.cvdLabel || 'CVD',
+        };
+
+        const queryString = map(monerisQueryParams, (value, key) => `${key}=${value}`).join('&');
 
         iframe.name = IFRAME_NAME;
         iframe.id = IFRAME_NAME;
 
-        iframe.src = `${ this._monerisURL(testMode) }?id=${profileId}&pmmsg=true&css_body=${cssBody}&css_textbox=${cssTextbox}&css_textbox_pan=${csstextboxPan}&enable_exp=1&css_textbox_exp=${cssTextboxExpiry}&enable_cvd=1&css_textbox_cvd=${csstexboxCvd}&display_labels=1`;
+        iframe.src = `${ this._monerisURL(testMode) }?${queryString}`;
```
icatalina, line 146: Thanks 🙇 this looks much neater.

cmedinilla:
♻️  @icatalina

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -112,25 +114,36 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
...
         }
 
         const iframe = document.createElement('iframe');
-        // Example CSS styling as provided by Moneris' documentation
-        const cssBody = style?.cssBody ?? 'background:white;';
-        const cssTextbox = style?.cssTextbox ?? 'border-width:2px;';
-        const csstextboxPan = style?.cssTextboxPan ?? 'width:140px;';
-        const cssTextboxExpiry = style?.cssTextboxExpiry ?? 'width:40px;';
-        const csstexboxCvd = style?.csstexboxCvd ?? 'width:40px';
+        const monerisQueryParams: Required<MoneriesHostedFieldsQueryParams> = {
```
icatalina, line 125: Since all the options are now mandatory in `MoneriesHostedFieldsQueryParams`, there's no need to wrap it in `Required`.

```suggestion
        const monerisQueryParams: MoneriesHostedFieldsQueryParams = {
```

icatalina (approved):
Hi @cmedinilla!

I've pre-approved the PR. Please address the last comment and make sure you squash all the commits before merging. 🙇 

alternatively we can just use _Squash and merge_:

image

PR #1159, opened 2021-06-18
fix(analytics): DATA-7883 Fix Segment + GAEE issue

bc-leprik:
## What? [DATA-7883](https://jira.bigcommerce.com/browse/DATA-7883)
Fix bug with Segment + GAEE.

## Why?
When the experiment `DATA-6891.missing_orders_within_GA` is ON there are issues with data being sent to Google Analytics when a merchant is using Segment.com to setup Google Analytics.

## Testing / Proof
Manually / Unit tests 

1) 34 items in the Cart, $850 total
image

2) now we send large orders in a transaction mode (each line_item per request)
image

3) proof from Google Analytics
image

### DIFF
--- a/src/analytics/analytics-step-tracker.ts
+++ b/src/analytics/analytics-step-tracker.ts
@@ -8,6 +8,7 @@ import { Coupon } from '../coupon';
 import { Order } from '../order';
 import { ShippingOption } from '../shipping';
 
+import { isGoogleAnalyticsAvailable, isPayloadSizeLimitReached, sendGoogleAnalytics } from './analytics-tracker-ga';
 import { AnalyticsTracker } from './analytics-tracker-window';
 import StepTracker from './step-tracker';
 
@@ -128,7 +129,7 @@ export default class AnalyticsStepTracker implements StepTracker {
             return;
         }
 
-        const isEcommerceGAEnabled = this.checkoutService.getState().data
+        const isMissingOrdersExperimentEnabled = this.checkoutService.getState().data
             .getConfig()?.checkoutSettings.features['DATA-6891.missing_orders_within_GA'];
 
         const payload = this.getTrackingPayload({
@@ -142,28 +143,34 @@ export default class AnalyticsStepTracker implements StepTracker {
             lineItems,
         });
 
-        if (isEcommerceGAEnabled && this.analytics.hasPayloadLimit(payload)) {
-            this.analytics.hit('transaction', {
-                '&ti': payload.orderId,
-                '&ta': payload.affiliation,
-                '&tr': payload.revenue,
-                '&ts': payload.shipping,
-                '&tt': payload.tax,
-                '&tcc': payload.coupon,
-                '&cu': payload.currency,
-            });
-
-            payload.products.map(product => {
-                this.analytics.hit('item', {
+        if (isMissingOrdersExperimentEnabled && isGoogleAnalyticsAvailable() && isPayloadSizeLimitReached(payload)) {
+            sendGoogleAnalytics(
+                'transaction',
+                {
                     '&ti': payload.orderId,
-                    '&in': product.name,
-                    '&ic': product.sku,
-                    '&iv': `${product.category}`,
-                    '&ip': product.price,
-                    '&iq': product.quantity,
-                });
+                    '&ta': payload.affiliation,
+                    '&tr': payload.revenue,
+                    '&ts': payload.shipping,
+                    '&tt': payload.tax,
+                    '&tcc': payload.coupon,
+                    '&cu': payload.currency,
+                }
+            );
+            payload.products.forEach(product => {
+                sendGoogleAnalytics(
+                    'item',
+                    {
+                        '&ti': payload.orderId,
+                        '&in': product.name,
+                        '&ic': product.sku,
+                        '&iv': `${product.category}`,
+                        '&ip': product.price,
+                        '&iq': product.quantity,
+                    }
+                );
             });
 
+            // TODO: decide how to send large orders to Segment without sending to GA again
             return this.clearExtraItemData(cartId);
         }
 
--- a/src/analytics/analytics-tracker-window.ts
+++ b/src/analytics/analytics-tracker-window.ts
@@ -1,7 +1,5 @@
 export interface AnalyticsTracker {
     track(step: string, data: any): void;
-    hit(hitName: string, data: any): void;
-    hasPayloadLimit(data: any): boolean;
 }
 
 export default interface AnalyticsTrackerWindow extends Window {
--- a/src/analytics/analytics-step-tracker.spec.ts
+++ b/src/analytics/analytics-step-tracker.spec.ts
@@ -1,14 +1,23 @@
+import { getGiftCertificateItem } from '../cart/line-items.mock';
 import { createCheckoutService, CheckoutService } from '../checkout';
 import { getCheckoutWithCoupons } from '../checkout/checkouts.mock';
 import { InvalidArgumentError } from '../common/error/errors';
 import { ShopperCurrency } from '../config';
 import { getConfig } from '../config/configs.mock';
 import { Order } from '../order';
+import { getPhysicalItem } from '../order/line-items.mock';
 import { getOrder } from '../order/orders.mock';
 import { getPaymentMethod } from '../payment/payment-methods.mock';
 import { getShippingOption } from '../shipping/shipping-options.mock';
 
 import AnalyticsStepTracker, { AnalyticStepId, AnalyticStepType } from './analytics-step-tracker';
+import { isGoogleAnalyticsAvailable, isPayloadSizeLimitReached, sendGoogleAnalytics } from './analytics-tracker-ga';
+
+jest.mock('./analytics-tracker-ga', () => ({
+    isGoogleAnalyticsAvailable: jest.fn(),
+    isPayloadSizeLimitReached: jest.fn(),
+    sendGoogleAnalytics: jest.fn(),
+}));
 
 describe('AnalyticsStepTracker', () => {
     let checkoutService: CheckoutService;
@@ -32,8 +41,6 @@ describe('AnalyticsStepTracker', () => {
     beforeEach(() => {
         analytics = {
             track: jest.fn(),
-            hit: jest.fn(),
-            hasPayloadLimit: jest.fn(),
         };
 
         sessionStorage = {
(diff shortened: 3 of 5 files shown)


src/analytics/analytics-tracker-window.ts
```diff
@@ -1,9 +1,56 @@
...
+    const _window = window as unknown as AnalyticsTrackerWindow;
+
+    if (!isFunction(_window?.ga)) {
+        return ;
+    }
+
+    _window?.ga('send', type, {
+        ...payload,
+        nonInteraction: false,
+    });
```
solofeed, line 22: 🍹 

```suggestion
    if (isFunction(_window?.ga)) {
        _window.ga('send', type, {
            ...payload,
            nonInteraction: false,
        });
    }
```

solofeed: 🍹  Or more crazy option :)

```suggestion
    _window?.ga?.('send', type, {
	       ...payload,
	       nonInteraction: false,
	  });
```

solofeed (approved):
LGTM

src/analytics/analytics-tracker-window.ts
bc-leprik: very complicated, let's leave with isFunction :)

bc-leprik: done

solofeed (approved):

src/analytics/analytics-tracker-window.ts
```diff
@@ -1,9 +1,60 @@
...
-    hasPayloadLimit(data: any): boolean;
 }
 
 export default interface AnalyticsTrackerWindow extends Window {
     analytics: AnalyticsTracker;
+    ga(command: string, eventName: string, payload: any): void;
 }
+
+export const sendGoogleAnalytics = (type: string, payload: any) => {
+    const _window = window as unknown as AnalyticsTrackerWindow;
```
davidchin, line 13: Instead of trying to cast it to the type that you want, write a type guard and do a type check instead. e.g.:

```ts
function isAnalyticsTrackerWindow(window: window | AnalyticsTrackerWindow): window is AnalyticsTrackerWindow {
    return 'ga' in window && typeof window.ga === 'function';
}

if (isAnalyticsTrackerWindow(window)) {
    window.ga('send', type, {
        // ...
    });
}
```

src/analytics/analytics-tracker-window.ts
```diff
@@ -1,9 +1,60 @@
...
+};
+
+/**
+ * Max size of the payload for the Google Analytics module
+ * if the limit will be succeeded, the GA throwing a silent error,
+ * and only in debug mode you can see it
+ */
+export const ANALYTICS_MAX_URI_LENGTH = 8096;
+
+export const isPayloadSizeLimitReached = (obj: any) => {
```
davidchin, line 36: Do you have to use `any`?

src/analytics/analytics-tracker-window.ts
```diff
@@ -1,9 +1,60 @@
...
+    return serializeAnalyticsEventPayload(obj).length >= ANALYTICS_MAX_URI_LENGTH;
+};
+
+const serializeAnalyticsEventPayload = (obj: Record<string, any>): string => {
```
davidchin, line 40: Do you have to use any? Perhaps `unknown` or `string|number` is a better type.

src/analytics/analytics-tracker-window.ts
```diff
@@ -1,9 +1,60 @@
...
+    return Object.keys(obj).reduce((acc: string[], key) => {
+        const type: string = typeof obj[key];
```
davidchin, line 42: You don't need `: string`, it can be implicitly inferred.

src/analytics/analytics-tracker-window.ts
```diff
@@ -1,9 +1,60 @@
...
+
+        if (type === 'string' || type === 'number') {
+            return [
+                ...acc,
+                `${key}=${obj[key]}`,
+            ];
+        }
+
+        if (type === 'object') {
```
davidchin, line 51: `null` is also an `object`. You might want to guard against that.

src/analytics/analytics-tracker-window.ts
```diff
@@ -1,9 +1,60 @@
+import { isFunction } from 'lodash';
+
 export interface AnalyticsTracker {
     track(step: string, data: any): void;
-    hit(hitName: string, data: any): void;
```
davidchin, line 3: What about the non-segment.com trackers that depend on `hit` and `hasPayloadLimit`?

src/analytics/analytics-step-tracker.ts
```diff
@@ -142,28 +142,34 @@ export default class AnalyticsStepTracker implements StepTracker {
...
+                        '&in': product.name,
+                        '&ic': product.sku,
+                        '&iv': `${product.category}`,
+                        '&ip': product.price,
+                        '&iq': product.quantity,
+                    }
+                );
             });
 
+            // TODO: decide how to send large orders to Segment without sending to GA again
```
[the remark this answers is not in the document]
bc-leprik: We will work on it on a separate PR.

src/analytics/analytics-tracker-window.ts
bc-leprik: This PR removes dependency on this methods. After merging this PR we also remove `hit` and `hasPayloadLimit` methods from analytics.js added in DATA-6891.

src/analytics/analytics-tracker-window.ts
bc-leprik: we can create AnalyticsPayload in analytics-step-tracker.ts and use instead of any

src/analytics/analytics-tracker-window.ts
bc-leprik: This function with recursive call and type always different.

src/analytics/analytics-tracker-window.ts
bc-leprik: good point, done

src/analytics/analytics-tracker-window.ts
bc-leprik: changed to AnalyticsPayload type

bc-leprik: Changed to AnalyticPayload

src/analytics/analytics-tracker-window.ts
```diff
@@ -1,9 +1,67 @@
...
+            return [
+                ...acc,
+                `${key}=${obj[key]}`,
+            ];
+        }
+
+        if (type === 'object' && obj[key] !== null) {
+            return [
+                ...acc,
+                serializeAnalyticsEventPayload(obj[key] as AnalyticPayload),
```
solofeed, line 61: @davidchin We have added `as AnalyticPayload` because TS doesn't understand that `number` and `string` cases are already handled.

Let us know if there is a better solution 🙇

src/analytics/analytics-tracker-window.ts
bc-leprik: @davidchin done!

src/analytics/analytics-tracker-window.ts
bc-leprik: I tried to refactor this but finally decided that’s a wrong way and we can produce more bugs.

bc-leprik: Also we will continue working on this for Segment Beta.

src/analytics/analytics-tracker-window.ts
davidchin: You can't remove `hit` and `hasPayloadLimit` from `analytics.js` due to the fact that there are clients using an older version of the SDK that references those methods.

src/analytics/analytics-tracker-window.ts
davidchin: In that case, you can use `unknown`. You're already doing `typeof` checks so it should just work. The reason why I'm pushing people not to use `any` is to force engineers to do explicit type checkint whenever we're dealing with unknow types.

src/analytics/analytics-tracker-window.ts
```diff
@@ -1,9 +1,63 @@
...
-    hit(hitName: string, data: any): void;
-    hasPayloadLimit(data: any): boolean;
 }
 
 export default interface AnalyticsTrackerWindow extends Window {
     analytics: AnalyticsTracker;
 }
+
+interface AnalyticsTrackerWindowGA extends AnalyticsTrackerWindow {
+    ga(command: string, eventName: string, payload: any): void;
```
davidchin, line 12: Can you provide a rough shape of the payload, like `Record<string, string>`?

src/analytics/analytics-tracker-window.ts
```diff
@@ -1,9 +1,63 @@
...
+
+function isAnalyticsTrackerWindowGA(window: Window | AnalyticsTrackerWindowGA): window is AnalyticsTrackerWindowGA {
+    return window && 'ga' in window && typeof window.ga === 'function';
+}
+
+export const isGoogleAnalyticsAvailable = () => {
+    return isAnalyticsTrackerWindow(window) && isAnalyticsTrackerWindowGA(window);
+};
+
+export const sendGoogleAnalytics = (type: string, payload: any) => {
```
davidchin, line 23: Same here.

src/analytics/analytics-tracker-window.ts
bc-leprik: Fair enough, so... when we can delete this methods from analytics.js? How can we check how many merchants use custom checkout?

bc-leprik: But when we disable or remove experiment DATA-6891.missing_orders_within_GA this methods will not reached anymore. Or we still can get an error in interface AnalyticsTracker?

if (isEcommerceGAEnabled ~~&& this.analytics.hasPayloadLimit(payload)~~) {

bc-leprik: Ok we will not drop this methods in `analytics.js`.

src/analytics/analytics-tracker-window.ts
bc-leprik: done

src/analytics/analytics-tracker-window.ts
bc-leprik: done

src/analytics/analytics-tracker-window.ts
bc-leprik: done

src/analytics/analytics-tracker-window.ts
bc-leprik: changed to AnalyticPayload interface

src/analytics/analytics-tracker-window.ts
```diff
@@ -1,9 +1,67 @@
...
+
+function isAnalyticsTrackerWindowGA(window: Window | AnalyticsTrackerWindowGA): window is AnalyticsTrackerWindowGA {
+    return window && 'ga' in window && typeof window.ga === 'function';
+}
+
+export const isGoogleAnalyticsAvailable = () => {
+    return isAnalyticsTrackerWindow(window) && isAnalyticsTrackerWindowGA(window);
+};
+
+export const sendGoogleAnalytics = (type: string, payload: AnalyticPayload) => {
```
davidchin, line 23: All these exported functions should be placed in their own files rather than under `analytics-tracker-window`, which is only ever intended to contain `AnalyticsTrackerWindow` as the name of the file suggests.

bc-leprik: @davidchin done!

src/analytics/analytics-tracker-window.ts
```diff
@@ -1,9 +1,67 @@
...
+        });
+    }
+};
+
+/**
+ * Max size of the payload for the Google Analytics module
+ * if the limit will be succeeded, the GA throwing a silent error,
+ * and only in debug mode you can see it
+ */
+export const ANALYTICS_MAX_URI_LENGTH = 8096;
```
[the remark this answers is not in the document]
bc-leprik: correct!

davidchin (approved):
All good, thanks for doing the updates, LGTM! 👍

bc-leprik:
> All good, thanks for doing the updates, LGTM! 👍

Thank you for your comments!

bc-junbai:
I've verified that when a merchant link Segment to their store, and set up Google Analytics via Segment site, large line item orders are sent to Google Analytics successfully with experiment DATA-6891.missing_orders_within_GA on. 

LGTM.

Screenshots:

PR #1158, opened 2021-06-18
[draft, wip] feat(payment): BOLT-19 added BoltCheckout implementation, added new c…

Sireks:
…ustomer continue strategy

## What?
Added new customer continue strategy to open BoltCheckout window before 'sign in', 'sign up' and 'continue as guest' events (before passing Customer step on checkout)

## Why?
Because of the task: https://jira.bigcommerce.com/browse/BOLT-19

## Sibling pull-request
checkout-js: https://github.com/bigcommerce/checkout-js/pull/616

## Testing / Proof
Some gifs to figure out how it looks like:
continue_as_customer
sign_in
sign_up

@bigcommerce/checkout @bigcommerce/payments

PR #1163, opened 2021-06-24
feat(payment): BOLT-19 added BoltCheckout implementation for customer…

Sireks:
…s with bolt account

## What?
Added BoltCheckout implementation for customers with bolt account.

## Why?
We need to add custom flow for _continue as guest_, _sign in_ and  _sign up_ events, to show BoltCheckout modal if the customer has BoltAccount

Task: https://jira.bigcommerce.com/browse/BOLT-19

## Sibling pull-request
checkout-js: https://github.com/bigcommerce/checkout-js/pull/619

## Testing / Proof
Some gifs to figure out how it looks like:
continue_as_customer
sign_in
sign_up

There how it will looks like if the user have default flow or customer hasn't bolt account:
continue_as_guest_without_bolt_account

### DIFF
--- a/src/customer/customer-strategy-reducer.spec.ts
+++ b/src/customer/customer-strategy-reducer.spec.ts
@@ -125,6 +125,49 @@ describe('customerStrategyReducer()', () => {
         });
     });
 
+    it('returns the pending flag as true if the customer continues as a guest', () => {
+        const action = createAction(
+            CustomerStrategyActionType.ContinueAsGuestRequested,
+            undefined,
+            { methodId: 'foobar' }
+        );
+
+        expect(customerStrategyReducer(initialState, action).statuses).toEqual({
+            continueAsGuestMethodId: 'foobar',
+            isContinuingAsGuest: true,
+        });
+    });
+
+    it('returns the pending flag as false if the customer has continued as guest successfully', () => {
+        const action = createAction(
+            CustomerStrategyActionType.ContinueAsGuestSucceeded,
+            undefined,
+            { methodId: 'foobar' }
+        );
+
+        expect(customerStrategyReducer(initialState, action).statuses).toEqual({
+            continueAsGuestMethodId: undefined,
+            isContinuingAsGuest: false,
+        });
+
+        expect(customerStrategyReducer(initialState, action).statuses).toEqual({
+            isContinuingAsGuest: false,
+        });
+    });
+
+    it('returns error if customer has failed to continue as guest', () => {
+        const action = createErrorAction(
+            CustomerStrategyActionType.ContinueAsGuestFailed,
+            new Error(),
+            { methodId: 'foobar' }
+        );
+
+        expect(customerStrategyReducer(initialState, action).errors).toEqual({
+            continueAsGuestMethodId: 'foobar',
+            continueAsGuestError: action.payload,
+        });
+    });
+
     it('returns pending flag as true if signing in customer', () => {
         const action = createAction(
             CustomerStrategyActionType.SignInRequested,
@@ -208,6 +251,49 @@ describe('customerStrategyReducer()', () => {
 
         expect(customerStrategyReducer(initialState, action).statuses).toEqual({
             isSigningOut: false,
+        });
+    });
+
+    it('returns pending flag as true if signing up customer', () => {
+        const action = createAction(
+            CustomerStrategyActionType.SignUpRequested,
+            undefined,
+            { methodId: 'foobar' }
+        );
+
+        expect(customerStrategyReducer(initialState, action).statuses).toEqual({
+            signUpMethodId: 'foobar',
+            isSigningUp: true,
+        });
+    });
+
+    it('returns pending flag as false if customer has signed up successfully', () => {
+        const action = createAction(
+            CustomerStrategyActionType.SignUpSucceeded,
+            undefined,
+            { methodId: 'foobar' }
+        );
+
+        expect(customerStrategyReducer(initialState, action).statuses).toEqual({
+            signUpMethodId: undefined,
+            isSigningUp: false,
+        });
+
+        expect(customerStrategyReducer(initialState, action).statuses).toEqual({
+            isSigningUp: false,
+        });
+    });
+
+    it('returns error if customer has failed to sign up', () => {
+        const action = createErrorAction(
+            CustomerStrategyActionType.SignUpFailed,
+            new Error(),
+            { methodId: 'foobar' }
+        );
+
+        expect(customerStrategyReducer(initialState, action).errors).toEqual({
+            signUpMethodId: 'foobar',
+            signUpError: action.payload,
         });
     });
 
--- a/src/customer/strategies/bolt/bolt-customer-strategy.ts
+++ b/src/customer/strategies/bolt/bolt-customer-strategy.ts
@@ -1,0 +1,129 @@
+import { BillingAddressActionCreator } from '../../../billing';
+import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
+import { InvalidArgumentError, MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType, UnhandledExternalError } from '../../../common/error/errors';
+import { PaymentMethodActionCreator } from '../../../payment';
+import { BoltCheckout, BoltScriptLoader } from '../../../payment/strategies/bolt';
+import CustomerAccountRequestBody from '../../customer-account';
+import CustomerActionCreator from '../../customer-action-creator';
+import CustomerCredentials from '../../customer-credentials';
+import { CustomerRequestOptions } from '../../customer-request-options';
+import { GuestCredentials } from '../../guest-credentials';
+import CustomerStrategy from '../customer-strategy';
+
+export default class BoltCustomerStrategy implements CustomerStrategy {
+    private _boltClient?: BoltCheckout;
+
+    constructor(
+        private _store: CheckoutStore,
+        private _boltScriptLoader: BoltScriptLoader,
+        private _paymentMethodActionCreator: PaymentMethodActionCreator,
+        private _billingAddressActionCreator: BillingAddressActionCreator,
+        private _customerActionCreator: CustomerActionCreator
+    ) {}
+
+    async initialize(options: CustomerRequestOptions): Promise<InternalCheckoutSelectors> {
+        const { methodId } = options;
+
+        if (!methodId) {
+            throw new InvalidArgumentError('Unable to proceed because "methodId" argument is not provided.');
+        }
+
+        const state = await this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethods());
+
+        const paymentMethod = state.paymentMethods.getPaymentMethod(methodId);
+
+        if (!paymentMethod || !paymentMethod.initializationData.publishableKey) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        const { developerConfig, publishableKey } = paymentMethod.initializationData;
+
+        this._boltClient = await this._boltScriptLoader.load(publishableKey, paymentMethod.config.testMode, developerConfig);
+
+        return this._store.getState();
+    }
+
+    async deinitialize(): Promise<InternalCheckoutSelectors> {
+        this._boltClient = undefined;
+
+        return this._store.getState();
+    }
+
+    async continueAsGuest(credentials: GuestCredentials, options?: CustomerRequestOptions): Promise<InternalCheckoutSelectors> {
+        const { methodId } = options || {};
+
+        if (!methodId) {
+            throw new InvalidArgumentError('Unable to proceed because "methodId" argument is not provided.');
+        }
+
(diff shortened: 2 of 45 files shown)


BC-ACherednichenko (commented):
Nicely done! I like this idea more than than the first revision https://github.com/bigcommerce/checkout-sdk-js/pull/1158, we re-use the existing strategy without creating additional one that behaves pretty much in the same way 

@davidchin @icatalina can you take a look at these two possible options and let us know your thoughts. The possible way that we will choose more likely correspond to the philosophy of the existing customer strategy, hence your input will be invaluable

src/customer/strategies/bolt/bolt-customer-strategy.ts
```diff
@@ -0,0 +1,119 @@
...
+                    };
+
+                    this._boltClient.openCheckout(email, callbacks)
+                        .then(result => !result && resolve());
+                } else {
+                    resolve();
+                }
+            } catch (error) {
+                resolve();
+                throw error;
```
BC-ACherednichenko, line 115: per our discussion, it worth to just log an error in order to not interrupt the default flow (will check if `throw` fully interrupts the default flow). continueWith strategy is not as vital for checkout as a strategy of a payment provider. we can remain the default flow if we have any issue

BC-ACherednichenko (approved):
Looks good to me! There is only one minor concern

@davidchin can take a look?

vneutrino (approved):
LGTM, except some tests names. I've sent my propositions to @Sireks in slack message.

src/customer/strategies/bolt/bolt-customer-strategy.ts
Sireks: I checked it and it is working as expected. If we somehow get an error, we will continue with default flow without any problems

src/customer/customer-strategy-reducer.spec.ts
```diff
@@ -125,6 +125,49 @@ describe('customerStrategyReducer()', () => {
...
+            undefined,
+            { methodId: 'foobar' }
+        );
+
+        expect(customerStrategyReducer(initialState, action).statuses).toEqual({
+            continueAsGuestMethodId: undefined,
+            isContinuingAsGuest: false,
+        });
+
+        expect(customerStrategyReducer(initialState, action).statuses).toEqual({
```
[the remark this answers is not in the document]
Sireks: Yes, it is.

I am already fixed all those propositions that you sent me in slack.

src/customer/strategies/bolt/bolt-customer-strategy.ts
```diff
@@ -0,0 +1,119 @@
...
+        if (!methodId) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        return this._store.dispatch(this._customerActionCreator.createCustomer(customerAccount, options))
+            .then(() => this.openCustomBoltCheckoutWidget(customerAccount.email, methodId))
+            .then(() => this._store.getState());
+    }
+
+    private openCustomBoltCheckoutWidget(email: string, methodId: string): Promise<InternalCheckoutSelectors> {
```
davidchin, line 89: We have a convention of prefixing private members, including methods, with an underscore. This is because when TS is compiled to JS, people know that they shouldn't be accessing those members.

src/customer/strategies/bolt/bolt-customer-strategy.ts
```diff
@@ -0,0 +1,119 @@
...
+                    throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+                }
+
+                if (paymentMethod.initializationData?.customContinueFlowEnabled) {
+                    const callbacks = {
+                        close: resolve,
+                    };
+
+                    this._boltClient.openCheckout(email, callbacks)
+                        .then(result => !result && resolve());
```
davidchin, line 109: Can `openCheckout` return an error / rejected promise. If so, we need to handle it accordingly,.

src/customer/strategies/bolt/bolt-customer-strategy.ts
```diff
@@ -0,0 +1,119 @@
...
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        return this._store.dispatch(this._customerActionCreator.createCustomer(customerAccount, options))
+            .then(() => this.openCustomBoltCheckoutWidget(customerAccount.email, methodId))
+            .then(() => this._store.getState());
+    }
+
+    private openCustomBoltCheckoutWidget(email: string, methodId: string): Promise<InternalCheckoutSelectors> {
+        return new Promise(resolve => {
```
davidchin, line 90: I don't think you need to manually construct a `Promise` as `openCheckout` already returns one. So you can just chain to it and return a Promise that way.

src/customer/strategies/bolt/bolt-customer-strategy.ts
davidchin: I think the intention is good. But I think what we should do, from SDK library's perspective, is to return an error and let the client to catch and decide what to do next as a recovery. The client may choose to silently ignore the error and log it behind the scene. Or it may choose to display a message informing that the shopper that something went wrong but they can still progress.

src/customer/strategies/bolt/bolt-customer-strategy.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    signIn(credentials: CustomerCredentials, options?: CustomerRequestOptions): Promise<InternalCheckoutSelectors> {
+        const { methodId } = options || {};
+
+        if (!methodId) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        return this._store.dispatch(this._customerActionCreator.signInCustomer(credentials, options))
```
davidchin, line 68: Doesn't this mean that if a shopper chooses to sign into their BC account, and then refresh the page before they get a chance to complete whatever that's needed in the Bolt widget, they won't be prompted to complete the Bolt widget again (because the shopper is already signed in so they will be prompted to complete the next/shipping step instead)?

tslint.json
```diff
@@ -4,6 +4,7 @@
         "max-line-length": false,
         "no-empty": false,
         "no-shadowed-variable": false,
-        "object-literal-sort-keys": false
+        "object-literal-sort-keys": false,
+        "cyclomatic-complexity": false
```
davidchin, line 8: You should try to resolve the cyclomatic complexity by refactoring the code rather than disabling the rule. If the refactoring takes too much effort and you'd rather tackle it separately, then you can disable the rule locally for the file only rather than disabling it globally for all files.

davidchin (changes requested):
Thank you @Sireks. It's looking good no major issue with the code itself. However, as I mentioned in your other PR, I'm blocking this for now until we get some clarification from product.

src/customer/strategies/bolt/bolt-customer-strategy.ts
Sireks: I will check this one

src/customer/strategies/bolt/bolt-customer-strategy.ts
Sireks: The logged in shopper will have Bolt checkout modal open each time when open Checkout page and if the shopper has Bolt Account

davidchin: I think we have to be careful on this because you're assuming that the shopper wants to use Bolt. It might be that they intentionally close the Bolt modal. But they are prompted again everytime they reopen checkout or go back to the customer step.

Sireks: But we don't know why the user closed Bolt Checkout modal, maybe shopper want to get more stuff to make the order or so. We will open the modal only for those customer that has their own Bolt Account. Otherwise the modal will not be showed.

We discussed this case a couple times... And we made the decision to make it like this.

src/customer/strategies/bolt/bolt-customer-strategy.ts
Sireks: I will change the method to make it more readable, but I need to keep `new Promise(resolve => ...)` to make all the code that left run only if the modal was closed. 

If Bolt Checkout modal was closed, it's ok, we will run default customer flow. But if shopper will continue with Bolt Checkout Modal, he/she will be redirected to Success page in the end of the flow.

src/customer/strategies/bolt/bolt-customer-strategy.ts
Sireks: It is already wrapped in `try...catch`.
I will remove that `then` chaining, because it is unnecessary anymore

src/customer/strategies/bolt/bolt-customer-strategy.ts
davidchin: I understand your team has made a decision. And I'm sorry if you feel like we're revisiting that decision. But considering that decision was made without consulting with our team, I think it's reasonable for us to question it and perhaps make adjustments to it if we both agree they are appropriate.

Anyway the reason why I brought this up is because you're making an assumption that shoppers with an existing Bolt account really want to use Bolt. Having a Bolt account does not mean the shopper wants to use it. It just means they have an account which they may choose to use it. Anyway, I guess you can say the shopper can just close it again if they really don't want use Bolt. That's fair enough. But what if the store has PayPal enabled and the shopper (with an existing Bolt account) clicks on the PayPal button on the cart page, authorises the payment on PayPal side and then get redirected to checkout, would they see the Bolt modal open automatically? If that's the case, I think we have a problem. At least, we have to make sure that if there's a payment in progress, don't try to encourage the shopper to abandon it and switch to use Bolt.

src/customer/strategies/bolt/bolt-customer-strategy.ts
davidchin: The `try/catch` block has no effect because you're using the `then` syntax. It only works if you use the `async/await` syntax.

tslint.json
Sireks: Fixed!

src/customer/strategies/bolt/bolt-customer-strategy.ts
Sireks: Thanks @davidchin!

It's ok. We will discuss the issue and I or someone from my team will tell you about what we will doing next.

But it's a question for another task and the case has not to be a blocker for current task.

src/customer/strategies/bolt/bolt-customer-strategy.ts
Sireks: Fixed!

src/customer/strategies/bolt/bolt-customer-strategy.ts
```diff
@@ -0,0 +1,119 @@
...
+        const paymentMethod = state.paymentMethods.getPaymentMethod(methodId);
+
+        if (!paymentMethod || !paymentMethod.initializationData.publishableKey) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        const { developerConfig, publishableKey } = paymentMethod.initializationData;
+
+        return this._boltScriptLoader.load(publishableKey, paymentMethod.config.testMode, developerConfig)
+            .then(boltCheckout => this._boltClient = boltCheckout)
```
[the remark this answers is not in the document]
Sireks: Fixed

src/customer/strategies/bolt/bolt-customer-strategy.ts
```diff
@@ -0,0 +1,119 @@
...
+        private _boltScriptLoader: BoltScriptLoader,
+        private _billingAddressActionCreator: BillingAddressActionCreator,
+        private _customerActionCreator: CustomerActionCreator
+    ) {}
+
+    initialize(options: CustomerRequestOptions): Promise<InternalCheckoutSelectors> {
+        const { methodId } = options;
+
+        if (!methodId) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
```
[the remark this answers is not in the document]
Sireks: Fixed

src/customer/strategies/bolt/bolt-customer-strategy.ts
BC-ACherednichenko: > I understand your team has made a decision. And I'm sorry if you feel like we're revisiting that decision. But considering that decision was made without consulting with our team, I think it's reasonable for us to question it and perhaps make adjustments to it if we both agree they are appropriate.

@davidchin this is what these 2 PRs about, as long as the flow is a bit tricky, we made some decisions and started conversation with these 2 POCs

>I think we have to be careful on this because you're assuming that the shopper wants to use Bolt. It might be that they intentionally close the Bolt modal. But they are prompted again everytime they reopen checkout or go back to the customer step.

To my knowdledge, we have decided to introduce some flag in checkout session (the research on this feature is in the progress), that will be set whenever a buyer mananges popup, so in the scope of single quote they will not be bombarded with the popup over and overagain

Sireks:
@davidchin 
I added some fixed. Can you check if everything all right?

PR #1168, opened 2021-07-01
feat(payment): INT-3946 added locale from browser on masterpass SRC

BC-AxelMalagon:
## What? [INT-3946](https://jira.bigcommerce.com/browse/INT-3946)
Added locale from browser on masterpass SRC.

## Why?
In order to be able to see masterpass or SRC button based on shopper locale.

## Testing / Proof
Screen Shot 2021-06-30 at 6 53 52 PM

### DIFF
--- a/src/checkout-buttons/strategies/masterpass/masterpass-button-strategy.ts
+++ b/src/checkout-buttons/strategies/masterpass/masterpass-button-strategy.ts
@@ -13,7 +13,8 @@ export default class MasterpassButtonStrategy implements CheckoutButtonStrategy 
     constructor(
         private _store: CheckoutStore,
         private _checkoutActionCreator: CheckoutActionCreator,
-        private _masterpassScriptLoader: MasterpassScriptLoader
+        private _masterpassScriptLoader: MasterpassScriptLoader,
+        private _locale: string
     ) {}
 
     initialize(options: CheckoutButtonInitializeOptions): Promise<void> {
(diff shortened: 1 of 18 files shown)


src/payment/strategies/masterpass/masterpass-payment-strategy.ts
```diff
@@ -26,21 +26,18 @@ export default class MasterpassPaymentStrategy implements PaymentStrategy {
 
     initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
         const { methodId } = options;
-        const state = this._store.getState();
-        const storeConfig = state.config.getStoreConfig();
+        const { language: localeLanguage } = navigator;
```
davidchin, line 29: You should read from the locale provided by the storefront rather than the browser. This is because going forward, it is a value that can be configured by merchants. The value should already be passed into `CheckoutService` so you should be able to pass it along to the various strategies for Masterpass (https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/payment/create-payment-strategy-registry.ts#L76).

davidchin (changes requested):

BC-AxelMalagon:
@davidchin  could you give another view ♻️ ?

src/checkout-buttons/strategies/masterpass/masterpass-button-strategy.ts
```diff
@@ -142,4 +139,11 @@ export default class MasterpassButtonStrategy implements CheckoutButtonStrategy
 
         this._masterpassClient.checkout(this._createMasterpassPayload());
     }
+
+    private formatLocale( localeLanguage: string): string {
+        const locale = localeLanguage.replace(/[-_]/g, '_');
+        const formatedLocale = locale.includes('_') ? `${locale}` : `${locale}_${locale}`;
```
davidchin, line 145: So if the locale is `en`, you want to return `en_en`? I think `en_en` is not a valid locale.

vmparra:
Closing as @BC-AxelMalagon no longer works for BC

PR #1169, opened 2021-07-01
fix(payment): PAYPAL-1026 fixed button height

andriiVitvitskyi1990:
## What?
Fixed button height for Braintree payment provider

## Why?
According to task https://jira.bigcommerce.com/browse/PAYPAL-1026
## Testing / Proof
Tested on dev and integration

### DIFF
--- a/src/checkout-buttons/strategies/braintree/braintree-paypal-button-strategy.ts
+++ b/src/checkout-buttons/strategies/braintree/braintree-paypal-button-strategy.ts
@@ -12,9 +12,11 @@ import { BraintreeError,
     BraintreeTokenizePayload,
     PaypalClientInstance,
     RenderButtonsData } from '../../../payment/strategies/braintree';
-import { PaypalAuthorizeData, PaypalHostWindow } from '../../../payment/strategies/paypal';
+import { PaypalAuthorizeData, PaypalButtonStyleOptions, PaypalHostWindow } from '../../../payment/strategies/paypal';
 import { CheckoutButtonInitializeOptions } from '../../checkout-button-options';
 import CheckoutButtonStrategy from '../checkout-button-strategy';
+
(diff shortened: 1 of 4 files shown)


src/checkout-buttons/strategies/braintree/braintree-paypal-button-strategy.ts
```diff
@@ -85,7 +85,8 @@ export default class BraintreePaypalButtonStrategy implements CheckoutButtonStra
                     style: {
                         shape: 'rect',
                         label: this._offerCredit ? 'credit' : undefined,
-                        ...pick(paypalOptions.style, 'layout', 'size', 'color', 'label', 'shape', 'tagline', 'fundingicons'),
+                        ...pick(paypalOptions.style, 'layout', 'size', 'color', 'label', 'shape', 'tagline', 'fundingicons', 'height'),
+                        height: Number(paypalOptions.style?.height),
```
animesh1987, line 89: Do we need to do this? The pick method above should do this for us already

andriiVitvitskyi1990: We get string type value for height in paypalOptions but need number, so that is why i did that

animesh1987: Then in that case you should be able to remove it height from pick statement.

andriiVitvitskyi1990: We changed type of height in bcapp so i removed  standalone height property

animesh1987 (approved):

animesh1987 (approved):

bc-fetisov (approved):
LGTM

PR #1170, opened 2021-07-02
feat(payment): BOLT-30 updated finalize method for Bolt payment strategy

Sireks:
## What?
Updated executeBoltFullCheckout method for Bolt payment strategy.

## Why?
From current implementation the fact that an order is created after the Bolt Payment happens in some cases (issue on Bolt side or slow connection) cause of the issues with orphan orders on Bolt side. For minimising this issues we some changes to executeBoltFullCheckout method more useful and Bolt will send us transaction reference. When Bolt creates transaction reference the are understand that we have an order created and I they will have some troubles on their side, we will throw an error to let user know, that something went wrong.

## Testing / Proof
All unit tests passed locally:
Screenshot 2021-07-02 at 10 09 15

### DIFF
--- a/src/payment/strategies/bolt/bolt-payment-strategy.ts
+++ b/src/payment/strategies/bolt/bolt-payment-strategy.ts
@@ -1,10 +1,10 @@
 import { isNonceLike } from '../..';
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
-import { MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType } from '../../../common/error/errors';
+import { InvalidArgumentError, MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType } from '../../../common/error/errors';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
 import { StoreCreditActionCreator } from '../../../store-credit';
-import { PaymentArgumentInvalidError, PaymentMethodCancelledError } from '../../errors';
+import { PaymentArgumentInvalidError, PaymentMethodCancelledError, PaymentMethodInvalidError } from '../../errors';
 import { NonceInstrument } from '../../payment';
 import PaymentActionCreator from '../../payment-action-creator';
 import PaymentMethodActionCreator from '../../payment-method-action-creator';
@@ -30,20 +30,22 @@ export default class BoltPaymentStrategy implements PaymentStrategy {
     async initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
         const { bolt, methodId } = options;
 
-        this._useBoltClient = !!(bolt && bolt.useBigCommerceCheckout);
-
-        if (this._useBoltClient) {
-            const state = this._store.getState();
-            const paymentMethod = state.paymentMethods.getPaymentMethod(methodId);
-
-            if (!paymentMethod || !paymentMethod.initializationData.publishableKey) {
-                throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
-            }
-
-            const { developerConfig, publishableKey } = paymentMethod.initializationData;
-
-            this._boltClient = await this._boltScriptLoader.load(publishableKey, paymentMethod.config.testMode, developerConfig);
+        if (!methodId) {
+            throw new InvalidArgumentError('Unable to initialize payment because "options.methodId" argument is not provided.');
         }
+
+        this._useBoltClient = bolt?.useBigCommerceCheckout || false;
+
+        const state = this._store.getState();
+        const paymentMethod = state.paymentMethods.getPaymentMethod(methodId);
+
+        if (!paymentMethod || !paymentMethod.initializationData.publishableKey) {
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        const { developerConfig, publishableKey } = paymentMethod.initializationData;
+
+        this._boltClient = await this._boltScriptLoader.load(publishableKey, paymentMethod.config.testMode, developerConfig);
 
         return Promise.resolve(this._store.getState());
     }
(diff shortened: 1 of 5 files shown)


vneutrino (approved):
LGTM

icatalina:
Hey @davidchin, can you have a look at this PR, this seems like a pretty fundamental change into how `finalizeOrder` works, I don't feel comfortable reviewing it.

BC-ACherednichenko (approved):
LGTM

src/checkout/checkout-service.ts
```diff
@@ -292,7 +292,7 @@ export default class CheckoutService {
      * @throws `OrderFinalizationNotRequiredError` error if order finalization
      * is not required for the current order at the time of execution.
      */
-    finalizeOrderIfNeeded(options?: RequestOptions): Promise<CheckoutSelectors> {
+    finalizeOrderIfNeeded(options?: PaymentFinalizeOptions): Promise<CheckoutSelectors> {
```
davidchin, line 295: This is a significant change to the public interface of the library. Order finalisation, as the name of the method suggests, can only happen after an order is created and payment details are received. Therefore, it does not make much sense for this method to accept payment details as one of its parameters. Payment details should therefore only be passed through the `submitOrder` method.

src/payment/strategies/bolt/bolt-payment-strategy.ts
```diff
@@ -62,10 +62,48 @@ export default class BoltPaymentStrategy implements PaymentStrategy {
...
+        if (!paymentData || !isNonceLike(paymentData)) {
+            throw new MissingDataError(MissingDataErrorType.MissingPayment);
+        }
+
+        const paymentPayload = {
+            methodId,
+            paymentData,
+        };
+
+        return this._store.dispatch(this._paymentActionCreator.submitPayment(paymentPayload));
```
davidchin, line 96: I'm not sure how it is even possible for you to submit payment without a valid payment token, which is only made available after submitting an order. If that's the case, you're asking the SDK client to call `checkoutService.submitOrder` and then `checkoutService.finalizeOrderIfNeeded` immediately afterwards. What you're effectively doing is changing the expectation that `finalizeOrderIfNeeded` can be called in isolation (i.e.: after redirecting from a hosted payment provider back to checkout). Although you're not technically making a non-additive change to the contract of `finalizeOrderIfNeeded`, you're in effect changing its observable behaviour therefore can be considered a breaking change.

davidchin (changes requested):
Thank you @Sireks. I just have some questions for you to think about 🙂

src/payment/strategies/bolt/bolt-payment-strategy.ts
Sireks: Bolt sometimes has some troubles on their side cause of the issues with orphan orders on their side and they don't want to let us submit an order before Bolt payment will pass. They will trigger `checkoutService.finalizeOrderIfNeeded` after Bolt payment will pass, that will trigger our submitPayment request

Sireks: @BC-TKozub can give to @davidchin some more information about the case?

BC-TKozub: Sure. @davidchin, the main issue with current Bolt integration is submitting BC order after checkout is finished on Bolt side. (Bolt multi-step checkout). In some cases they have an issue with submitting BC order (connection issue, issues on shopper side, etc). As a result shopper has successfully completed checkout process but merchant does not have an order in control panel (orphan Bolt orders). The only way to fix current bihavior is do submit order via checkout-sdk before Bolt authorize payment and then finalize an order (perform submitPayment).

src/checkout/checkout-service.ts
Sireks: I need to add an ability for Bolt team to pass paymentData to the `checkoutService.finalizeOrderIfNeeded` method, to pass `bolt transaction reference` and submit payment.

src/payment/strategies/bolt/bolt-payment-strategy.ts
davidchin: Yes, I'm aware of the orphan order situation and I was the one who suggested to them to create BC order first. But what I was picturing is to do that as a part of `execute`, not `finalize`. If you do that, you wouldn't need to change the interface of `finalizeOrderIfNeeded`.

vitaliyreznikov: Hi @davidchin! From bolt perspective, we need to have two different checkout SDK calls: one to create pre-order and the second to create a payment.
In the current PR proposed solution where we use **execute** for pre-order and and **finalize** for payment. If you think we need to split two actions in another way please let us know.

davidchin: Hi @vitaliyreznikov, I'm slightly confused. Is there any reason why you can't create the order and submit payment details (in that sequence) within `execute`? Also, we're the client of the Bolt library that they provided, they shouldn't care whether we have one method or two because they shouldn't have a reference to our Checkout SDK library.

vitaliyreznikov: Hi @davidchin 

**1.** 
> Also, we're the client of the Bolt library that they provided
Yes and no. We have two different implementations in this file:
- _executeWithBoltClient: used in bolt product "fraud only", Checkout SDK calls Bolt library
- _executeWithBoltCheckout: used in bolt product "full checkout", bolt code calls Checkout SDK.

This PR changes _executeWithBoltCheckout only, so we are not going to add any cyclical dependency

**2. Why we need this**

Current flow for _executeWithBoltCheckout (full checkout):

```
1. Bolt creates transaction (bolt API call)
2. Bolt creates bigc order and payment (checkout SDK call)
```

desired flow:
```
1. Bolt creates bigc order (checkout SDK call)
2. Bolt creates transaction (bolt API call)
3. Bolt creates bigc payment (checkout SDK call)
```
In this flow, if (1) failed we are going to show an error instead of payment creation. This fixed the orphan order issue.

3. We can use another approach: instead of splitting checkout SDK call into two calls we can call bolt code in the middle. So after line 147 https://github.com/bigcommerce/checkout-sdk-js/blob/a46540bfcf50a313ebfe624dda59153f0ae1f983/src/payment/strategies/bolt/bolt-payment-strategy.ts#L147 add something like:

```
        const transactionReference = await this._boltClient.getTransactionReference();
        if (transactionReference === undefined) {
            return state;
        }
        paymentData.nonce = transactionReference;
```
It also solves issue because on bolt side we can actually start transaction only when we see getTransactionReference call.
Two advantages:
- code is very similar to existing code in _executeWithBoltClient implementation: https://github.com/bigcommerce/checkout-sdk-js/blob/a46540bfcf50a313ebfe624dda59153f0ae1f983/src/payment/strategies/bolt/bolt-payment-strategy.ts#L99
- no need to change finalizeOrderIfNeeded signature.

Please let me know if you think that this approach works better for you than the current PR.

Also if you have time I am happy to show you both implementation (_executeWithBoltClient and _executeWithBoltCheckout), how they look on client sites.

Sireks: Hello @davidchin, can you check last comment from @vitaliyreznikov, please?

davidchin: Thank you @vitaliyreznikov and @Sireks for coming up with a solution that doesn't involve changing the signature of `finalizeOrderIfNeeded`. It looks like a better option to me. 👍

Sireks:
@davidchin I already updated the pr, can you look at it one more time, please?

src/payment/strategies/bolt/bolt-payment-strategy.ts
```diff
@@ -144,11 +157,30 @@ export default class BoltPaymentStrategy implements PaymentStrategy {
             throw new MissingDataError(MissingDataErrorType.MissingPayment);
         }
 
-        await this._store.dispatch(this._orderActionCreator.submitOrder(order, _options));
+        await this._store.dispatch(this._orderActionCreator.submitOrder(order, options));
+
+        const transactionReference = await boltClient.getTransactionReference();
+
+        if (!transactionReference) {
+            return Promise.resolve(this._store.getState());
```
davidchin, line 165: Are you sure you want to resolve? I think that'd imply to the client that everything is okay and it can redirect the shopper to the order confirmation page.

Sireks: Oh... My bad. I need to throw an error here to let customer know that something went wrong. I will fix it soon

Sireks: @davidchin Fixed.

I am throwing an error if transactionReference will be empty. In this case, the customer will understand that something went wrong.

Thanks for your comments! 🙏
Now I understand that I need to think more about our customers experience

davidchin (approved):
@Sireks LGTM, thank you for your patience, much appreciated!

PR #1175, opened 2021-07-08
fix(payment): INT-4594 [Afterpay] display correct error message when amount is out of limits

orecio-bc:
## What? [INT-4594](https://jira.bigcommerce.com/browse/INT-4594)
Display the correct error message when trying to purchase with Afterpay and the amount is outside the limits provided by Afterpay (due to a Gift Certificate, Coupon code, Store credit)

## Why?
When the amount is outside limits Afterpay will respond with an error code when trying to fetch the client Token for the transaction, we are catching that error to display a custom error message

## Testing / Proof
Unit test
Video https://drive.google.com/file/d/15jMDT0uv4SsSrSgy62sfKWRMKyJt4lLb/view?usp=sharing

## Depends on
https://github.com/bigcommerce/bigcommerce/pull/41787
https://github.com/bigcommerce/bigpay/pull/4070

Revert this PR

### DIFF
--- a/src/payment/strategies/afterpay/afterpay-payment-strategy.spec.ts
+++ b/src/payment/strategies/afterpay/afterpay-payment-strategy.spec.ts
@@ -7,7 +7,8 @@ import { of, Observable } from 'rxjs';
 
 import { createCheckoutStore, CheckoutRequestSender, CheckoutStore, CheckoutValidator } from '../../../checkout';
 import { getCheckout, getCheckoutPayment, getCheckoutStoreState } from '../../../checkout/checkouts.mock';
-import { MissingDataError, NotInitializedError } from '../../../common/error/errors';
+import { InvalidArgumentError, MissingDataError, NotInitializedError, RequestError } from '../../../common/error/errors';
+import { getErrorResponse } from '../../../common/http-request/responses.mock';
 import { OrderActionCreator, OrderActionType, OrderRequestBody, OrderRequestSender } from '../../../order';
 import { getOrderRequestBody } from '../../../order/internal-orders.mock';
 import { createSpamProtection, PaymentHumanVerificationHandler } from '../../../spam-protection';
@@ -196,15 +197,59 @@ describe('AfterpayPaymentStrategy', () => {
         it('throws error if trying to execute before initialization', async () => {
             await strategy.deinitialize();
 
-            try {
-                await strategy.execute(payload);
-            } catch (error) {
-                expect(error).toBeInstanceOf(NotInitializedError);
-            }
+            await expect(strategy.execute(payload)).rejects.toThrow(NotInitializedError);
+        });
+
+        it('throws error if amount is outside the Afterpay limits', async () => {
+            store = createCheckoutStore(merge(getCheckoutStoreState(), {
+                checkout: { data: { outstandingBalance: 5 } },
+            }));
+
+            strategy = new AfterpayPaymentStrategy(
+                store,
+                checkoutValidator,
+                orderActionCreator,
+                paymentActionCreator,
+                paymentMethodActionCreator,
+                storeCreditActionCreator,
+                scriptLoader
+            );
+
+            await strategy.initialize({ methodId: paymentMethod.id, gatewayId: paymentMethod.gateway });
+
+            await expect(strategy.execute(payload)).rejects.toThrow(InvalidArgumentError);
+        });
+
+        it('throws InvalidArgumentError if payment method is not found', async () => {
+            const errorResponse = merge(
+                getErrorResponse(), {
+                    body: {
+                        status: 404,
+                    },
+                });
+
+            jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod').mockImplementation(() => {
+                throw new RequestError(errorResponse);
+            });
+
+            expect(store.dispatch).toHaveBeenCalledWith(loadPaymentMethodAction);
+
+            await expect(strategy.execute(payload)).rejects.toThrow(InvalidArgumentError);
+        });
+
+        it('throws RequestError if loadPaymentMethod fails', async () => {
+            jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod').mockImplementation(() => {
+                throw new RequestError(getErrorResponse());
+            });
+
+            expect(store.dispatch).toHaveBeenCalledWith(loadPaymentMethodAction);
+
+            await expect(strategy.execute(payload)).rejects.toThrow(RequestError);
         });
 
         it('loads payment client token', () => {
-            expect(paymentMethodActionCreator.loadPaymentMethod).toHaveBeenCalledWith(paymentMethod.gateway, undefined);
+            expect(paymentMethodActionCreator.loadPaymentMethod)
+                .toHaveBeenCalledWith(`${paymentMethod.gateway}?method=${paymentMethod.id}`, undefined);
             expect(store.dispatch).toHaveBeenCalledWith(loadPaymentMethodAction);
         });
     });
--- a/src/payment/strategies/afterpay/afterpay-payment-strategy.spec.ts
+++ b/src/payment/strategies/afterpay/afterpay-payment-strategy.spec.ts
@@ -7,8 +7,7 @@ import { of, Observable } from 'rxjs';
 
 import { createCheckoutStore, CheckoutRequestSender, CheckoutStore, CheckoutValidator } from '../../../checkout';
 import { getCheckout, getCheckoutPayment, getCheckoutStoreState } from '../../../checkout/checkouts.mock';
-import { InvalidArgumentError, MissingDataError, NotInitializedError, RequestError } from '../../../common/error/errors';
-import { getErrorResponse } from '../../../common/http-request/responses.mock';
+import { MissingDataError, NotInitializedError } from '../../../common/error/errors';
 import { OrderActionCreator, OrderActionType, OrderRequestBody, OrderRequestSender } from '../../../order';
 import { getOrderRequestBody } from '../../../order/internal-orders.mock';
 import { createSpamProtection, PaymentHumanVerificationHandler } from '../../../spam-protection';
@@ -198,33 +197,6 @@ describe('AfterpayPaymentStrategy', () => {
             await strategy.deinitialize();
 
             await expect(strategy.execute(payload)).rejects.toThrow(NotInitializedError);
-        });
-
-        it('throws InvalidArgumentError if payment method is not found', async () => {
-            const errorResponse = merge(
-                getErrorResponse(), {
-                    body: {
-                        status: 404,
-                    },
-                });
-
-            jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod').mockImplementation(() => {
-                throw new RequestError(errorResponse);
-            });
-
-            expect(store.dispatch).toHaveBeenCalledWith(loadPaymentMethodAction);
-
-            await expect(strategy.execute(payload)).rejects.toThrow(InvalidArgumentError);
-        });
-
-        it('throws RequestError if loadPaymentMethod fails', async () => {
-            jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod').mockImplementation(() => {
-                throw new RequestError(getErrorResponse());
-            });
-
-            expect(store.dispatch).toHaveBeenCalledWith(loadPaymentMethodAction);
-
-            await expect(strategy.execute(payload)).rejects.toThrow(RequestError);
         });
 
         it('loads payment client token', () => {
--- a/src/payment/strategies/afterpay/afterpay-payment-strategy.spec.ts
+++ b/src/payment/strategies/afterpay/afterpay-payment-strategy.spec.ts
@@ -200,26 +200,6 @@ describe('AfterpayPaymentStrategy', () => {
             await expect(strategy.execute(payload)).rejects.toThrow(NotInitializedError);
         });
 
-        it('throws error if amount is outside the Afterpay limits', async () => {
-            store = createCheckoutStore(merge(getCheckoutStoreState(), {
-                checkout: { data: { outstandingBalance: 5 } },
-            }));
-
-            strategy = new AfterpayPaymentStrategy(
-                store,
-                checkoutValidator,
-                orderActionCreator,
-                paymentActionCreator,
-                paymentMethodActionCreator,
-                storeCreditActionCreator,
-                scriptLoader
-            );
-
-            await strategy.initialize({ methodId: paymentMethod.id, gatewayId: paymentMethod.gateway });
-
-            await expect(strategy.execute(payload)).rejects.toThrow(InvalidArgumentError);
-        });
-
         it('throws InvalidArgumentError if payment method is not found', async () => {
             const errorResponse = merge(
                 getErrorResponse(), {
--- a/src/payment/strategies/afterpay/afterpay-payment-strategy.ts
+++ b/src/payment/strategies/afterpay/afterpay-payment-strategy.ts
@@ -1,6 +1,5 @@
 import { CheckoutStore, CheckoutValidator, InternalCheckoutSelectors } from '../../../checkout';
-import { InvalidArgumentError, MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType, RequestError } from '../../../common/error/errors';
-import { RequestOptions } from '../../../common/http-request';
+import { InvalidArgumentError, MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType } from '../../../common/error/errors';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
 import { StoreCreditActionCreator } from '../../../store-credit';
 import { PaymentArgumentInvalidError } from '../../errors';
@@ -15,7 +14,7 @@ import AfterpaySdk from './afterpay-sdk';
 
 export default class AfterpayPaymentStrategy implements PaymentStrategy {
     private _afterpaySdk?: AfterpaySdk;
-    private _invalidAmountErrorMessage = 'The amount for your order is not supported by Afterpay, please select another payment method';
+    private _isSupportedAmount?: boolean;
 
     constructor(
         private _store: CheckoutStore,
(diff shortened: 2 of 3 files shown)


src/payment/strategies/afterpay/afterpay-payment-strategy.ts
```diff
@@ -101,7 +104,11 @@ export default class AfterpayPaymentStrategy implements PaymentStrategy {
         return this._store.dispatch(this._paymentActionCreator.submitPayment(paymentPayload));
     }
 
-    private _redirectToAfterpay(countryName: string, paymentMethod?: PaymentMethod): void {
+    private _redirectToAfterpay(countryName: string, amount: number, paymentMethod?: PaymentMethod): void {
+        if (this._outsideLimits(amount, paymentMethod)) {
```
josesonora, line 108: please move this validation out of this method and call it first (before calling this one)

src/payment/strategies/afterpay/afterpay-payment-strategy.ts
```diff
@@ -125,4 +132,8 @@ export default class AfterpayPaymentStrategy implements PaymentStrategy {
             return 'AU';
         }
     }
+
+    private _outsideLimits(amount: number, paymentMethod?: PaymentMethod): boolean {
+        return amount < paymentMethod?.initializationData.min || amount > paymentMethod?.initializationData.max;
```
josesonora, line 137: please take in mind that order min can be empty and not a number

src/payment/strategies/afterpay/afterpay-payment-strategy.ts
```diff
@@ -72,7 +72,10 @@ export default class AfterpayPaymentStrategy implements PaymentStrategy {
             this._paymentMethodActionCreator.loadPaymentMethod(paymentId, options)
         );
 
-        await this._redirectToAfterpay(storeCountryName, state.paymentMethods.getPaymentMethod(paymentId));
+        const checkout = state.checkout.getCheckout();
+        const amount = checkout?.outstandingBalance || 0;
```
josesonora, line 76: just to make sure, this balance also considers store credit, right?

josesonora (changes requested):

src/payment/strategies/afterpay/afterpay-payment-strategy.ts
orecio-bc: yes it works, I will upload a video showing both

src/payment/strategies/afterpay/afterpay-payment-strategy.ts
```diff
@@ -72,6 +72,13 @@ export default class AfterpayPaymentStrategy implements PaymentStrategy {
...
         );
 
+        const checkout = state.checkout.getCheckout();
+        const amount = checkout?.outstandingBalance || 0;
+
+        if (this._outsideLimits(amount, state.paymentMethods.getPaymentMethod(paymentId))) {
+            throw new InvalidArgumentError('INCORRECT AMOUNT');
+        }
+
         await this._redirectToAfterpay(storeCountryName, state.paymentMethods.getPaymentMethod(paymentId));
```
josesonora, line 82:
```suggestion
       const paymentMethod = state.paymentMethods.getPaymentMethod(paymentId);
       
        if (this._outsideLimits(amount, paymentMethod)) {
            throw new InvalidArgumentError('INCORRECT AMOUNT');
        }

        await this._redirectToAfterpay(storeCountryName, paymentMethod);
```

src/payment/strategies/afterpay/afterpay-payment-strategy.ts
```diff
@@ -102,6 +109,7 @@ export default class AfterpayPaymentStrategy implements PaymentStrategy {
     }
 
     private _redirectToAfterpay(countryName: string, paymentMethod?: PaymentMethod): void {
+
```
josesonora, line 112: this line is not necessary

src/payment/strategies/afterpay/afterpay-payment-strategy.spec.ts
```diff
@@ -127,7 +133,7 @@ describe('AfterpayPaymentStrategy', () => {
         it('loads script when initializing strategy', async () => {
             await strategy.initialize({ methodId: paymentMethod.id, gatewayId: paymentMethod.gateway });
 
-            expect(scriptLoader.load).toHaveBeenCalledWith(paymentMethod, 'US');
+            expect(scriptLoader.load).toHaveBeenCalled();
```
josesonora, line 136: why did you change this?

josesonora (changes requested):

src/payment/strategies/afterpay/afterpay-payment-strategy.ts
```diff
@@ -125,4 +134,24 @@ export default class AfterpayPaymentStrategy implements PaymentStrategy {
...
         }
     }
+
+    private _outsideLimits(amount: number, paymentMethod?: PaymentMethod): boolean {
+        const $minimum = paymentMethod?.initializationData?.min || 0;
+
+        return amount < $minimum || amount > paymentMethod?.initializationData?.max;
+    }
+
+    private async _loadPaymentMethod(gatewayId: string, methodId: string, options?: RequestOptions) {
```
orecio-bc, line 144: When purchasing outside the Afterpay limits, bigpay will not return Afterpay while fetching the paymentMethods, we want to catch that error and replace it with an `_invalidAmountErrorMessage`, any other error will be re threw.

josesonora (approved):

src/payment/strategies/afterpay/afterpay-payment-strategy.ts
```diff
@@ -125,4 +136,24 @@ export default class AfterpayPaymentStrategy implements PaymentStrategy {
...
```
roberbnd, line 146: it's missing the returned value in the signature function.

src/payment/strategies/afterpay/afterpay-payment-strategy.ts
```diff
@@ -68,11 +73,17 @@ export default class AfterpayPaymentStrategy implements PaymentStrategy {
 
         await this._checkoutValidator.validate(state.checkout.getCheckout(), options);
 
-        state = await this._store.dispatch(
-            this._paymentMethodActionCreator.loadPaymentMethod(paymentId, options)
-        );
+        state = await this._loadPaymentMethod(gatewayId, methodId, options);
```
roberbnd, line 76: I would change the name because in a glance I feel a little confused,  Are we working with the state or with the payment method?
```suggestion
        state = await this._loadState(gatewayId, methodId, options);
```

roberbnd (changes requested):

src/payment/strategies/afterpay/afterpay-payment-strategy.ts
orecio-bc: Even if we are returning the state, the purpose of this method is to reload the payment method, that's why I decided to call it that way

roberbnd (approved):

src/payment/strategies/afterpay/afterpay-payment-strategy.ts
```diff
@@ -125,4 +136,24 @@ export default class AfterpayPaymentStrategy implements PaymentStrategy {
...
         }
     }
+
+    private _outsideLimits(amount: number, paymentMethod?: PaymentMethod): boolean {
+        const $minimum = paymentMethod?.initializationData?.min || 0;
+
+        return amount < $minimum || amount > paymentMethod?.initializationData?.max;
+    }
+
+    private async _loadPaymentMethod(gatewayId: string, methodId: string, options?: RequestOptions): Promise<InternalCheckoutSelectors> {
```
orecio-bc, line 146: When purchasing outside the Afterpay limits, bigpay will not return Afterpay while fetching the paymentMethods, we want to catch that error (404 Not Found) and replace it with a custom message, any other error will be re threw.

leeBigCommerce: I'm probably missing some info, but how is it possible to reach this scenario?

Does the `payments` endpoint method not filter out Afterpay if the payable value is too big/small?
If it did, this strategy would never be initialised and wouldn't need to handle this eventuality.

orecio-bc: The problem occurs when the shopper adds a coupon/gift certificate in the last step of the checkout, at this point the payments endpoint returned Afterpay as payment method because the payable value is within the limits, but when the shopper adds the coupon that new payable amount may not be within the limits and when we try to get the initialization token (necessary to initialize their SDK) the payments endpoint will return an Not Found error because now the payment method does not meet the requirements around the order amount and it is not supported as payment method

orecio-bc: @leeBigCommerce The problem occurs when the shopper adds a coupon/gift certificate in the last step of the checkout, at this point the payments endpoint returned Afterpay as payment method because the payable value is within the limits, but when the shopper adds the coupon that new payable amount may not be within the limits and when we try to get the initialization token (necessary to initialize their SDK) the payments endpoint will return an Not Found error because now the payment method does not meet the requirements around the order amount and it is not supported as payment method

leeBigCommerce: Sounds like we should be refetching available payment method options whenever the payable amount changes (including coupons and gift certificates), but I'm guessing this would be a bigger change to plan in @davidchin ?

If that is too big for now, can we change the endpoint to return a `400` or `422` in this scenario (perhaps with a more specific error key in the body)?

Currently this is a pretty big stretch of what `404` is meant to represent

orecio-bc: @leeBigCommerce We think it's a good idea to change the returned status to 422 (it will be easier to filter the error code), but we believe that this change has to be part of a new ticket since all the providers use this endpoint, we'll first need to analyze what could be the impacted areas around all the platform and then run regression tests in several providers to ensure functionality is preserved across the platform, for this reason, we suggest to extract these changes you are proposing to a new ticket.

leeBigCommerce: Sounds good 👍 

Maybe add some inline comments around the 404 uses to explain the reason and temporary nature to developers?

This test in particular makes it seem like desired behaviour: https://github.com/bigcommerce/checkout-sdk-js/pull/1175/files#diff-9dc05fcb1acb1dc5a9246aaa802472601f28f2eab3516bd09d0e1cb856ac605cR202 - not mentioning that the 404 implicitly represents a 'invalid amount' issue.

orecio-bc: @leeBigCommerce I've been looking at the error code and where and why is raised as 404, is raised in BCApp whenever the payment provider is present in BCApp but not in BigPay, I will depend on the provider and why it can't be initialized in this step of the checkout process, for Afterpay it's because of the amount, but another provider could fail for other reason.

I believe tthat the 404 is the correct error code, the provider was not found, and the strategy should explain why.

leeBigCommerce: the Payments team is going to review the use of `404` meaning both 'not found' and 'unavailable' outside of this ticket as it seems to be a bit of a wider issue

src/payment/strategies/afterpay/afterpay-payment-strategy.ts
```diff
@@ -68,11 +73,17 @@ export default class AfterpayPaymentStrategy implements PaymentStrategy {
...
-        state = await this._store.dispatch(
-            this._paymentMethodActionCreator.loadPaymentMethod(paymentId, options)
-        );
+        state = await this._loadPaymentMethod(gatewayId, methodId, options);
+
+        const checkout = state.checkout.getCheckout();
+        const amount = checkout?.outstandingBalance || 0;
+        const paymentMethod = state.paymentMethods.getPaymentMethod(methodId, gatewayId);
+
+        if (this._outsideLimits(amount, paymentMethod)) {
```
leeBigCommerce, line 82: If the amount was outside of the limits, would we not have thrown the error at line `76` already?

orecio-bc: @leeBigCommerce This case covers an issue with the store credit, they are initializing in a different way, that's why we went with this double validation

orecio-bc:
@leeBigCommerce any other questions regarding the changes?

leeBigCommerce:
Apologies for the delay, I've been struggling to get store credit working on my dev machine to try things out for myself.

I think it'd be good to get @bigcommerce/checkout eyes on this as having to do 'double validation' does sound a bit odd.

josesonora:
@leeBigCommerce Hi, we are discussing the changes for the store credit here https://github.com/bigcommerce/bigcommerce/pull/42022#discussion_r678785961, as you can see this is taking longer than we expected, our concern is that we are on certification phase and this is a beta-blocker for Afterpay, that’s why we want to ask if we can approve and merge this PR in its current state and when we get merged the changes for the store credit, make the necessary changes in this strategy to make it work with the final approach

animesh1987:
Hi @josesonora is there a reason to do a check in the FE when the API already returns an exception which FE can interpret and present to the user ? As @leeBigCommerce  already stated above `Sounds like we should be refetching available payment method options whenever the payable amount changes (including coupons and gift certificates)`

orecio-bc:
@animesh1987 
We need to do both verifications because they represent 2 different scenarios, the first one is when we have a coupon code, the API will return a **Not Found** error because the payment method is no longer available [here](https://github.com/bigcommerce/checkout-sdk-js/pull/1175/files#diff-286f495ea51236f037f02176c26017285e3f630dbe31bcf1e1a399f0f2c03e80R141)

The second verification is when we use store credit because the API receives the full amount when retrieving the available payment methods (so Afterpay will be there), it will fail and return an empty token that will throw an error and it will be interpreted and presented to the user [here](https://github.com/bigcommerce/checkout-sdk-js/pull/1175/files#diff-286f495ea51236f037f02176c26017285e3f630dbe31bcf1e1a399f0f2c03e80R82)

We're also working on the solution for the problem with store credit [here](https://github.com/bigcommerce/bigcommerce/pull/42022), as you can see we're on a discussion with Payments and Checkout teams trying to implement the best solution, but this is affecting this PR since there is a high dependency with it, that's why we're doing the double-check in the FE, to remove that dependency with the other PR, we want to do this because Afterpay is blocking the certification/release until we fix the problem this PR is trying to solve.

We propose to get approval for this PR and merge it and once we define the solution regarding the store credit in its PR and get it merged then open another PR to delete the validation/check on this strategy because it won't be needed anymore, but for now it is a quick solution that will allow us to continue with the certification while we wait for a resolution in the PR related to the store credit. 

On the other hand, regarding what @leeBigCommerce has stated, the re-fetching of the available payment method will not work because of the same cause with the store credit that we're working [here](https://github.com/bigcommerce/bigcommerce/pull/42022) also, I don't think that change has to be done as part of this PR since it will affect all providers and other flows and it may cause side effects ...

I agree with @leeBigCommerce, he has stated in the same paragraph
> but I'm guessing this would be a bigger change to plan in David Chin/Checkout Team

davidchin (changes requested):
I agree with @leeBigCommerce, 404 does not mean "The amount for your order is not supported by Afterpay". The API should be explicit about why a request cannot be fulfilled and provide an explanation for the failure. The reason why it's not a good idea to conflate the meaning of 404 with this scenario is because clients won't be able to differentiate between the scenario of a store not having AfterPay enabled versus a store having AfterPay enabled but not available to pay for a particular order.

It is important to decide on the API contract now and commit to it because we won't be able to change it in the future, as returning a different status code would effectively mean a breaking change for the client. Could you please take a look at revising the proposed change? Thanks!

orecio-bc:
@davidchin 
I think we should not change the API because that is the expected behavior, in other words, that endpoint is used to return the data for a specific available payment method and if the requested payment method is not available then it returns a 404 Not Found code.

Talking about this strategy (Afterpay), at this point in the process, Afterpay was already presented to the shopper as an available payment method but if the shopper adds a coupon/store credit it may be no longer available, this is the only part when a 404 is returned by the API (for this part of the flow), so we can confidently say that a 404 when calling `this._paymentMethodActionCreator.loadPaymentMethod` will be a result from that verification only, which means AfterPay is enabled but not available to use it to pay for the order.

On the other hand, we can go with the proposed approach of reloading all the payment methods whenever the amount changes but despite the following two possible solutions, we stand by the last change submitted yesterday, as we discuss in the first two paragraphs 404 is the correct error code for this, at least within the context of Afterpay.

We have these 2 possible solutions if we would want to go with the reloading of all payment methods:
1. The first one is to reload the checkout page whenever the amount is outside the limits provided from Afterpay, which is the easiest solution but we believe that the user experience would be affected.
2. The second one is to subscribe to any changes on the state and verify if the amount was changed, if so then reload the payment methods using
`this._paymentMethodActionCreator.loadPaymentMethods(options)`
but this would require some changes on the **PaymentStrategyActionCreator#deinitialize** method, 
this method currently verifies only the presence of the payment_method which is not present because it was not returned by the call to **loadPaymentMethods** and will throw an error resulting in a bad user experience.
https://github.com/bigcommerce/checkout-sdk-js/blob/91f3a3be6da4f5e83c37bbeb2e12af9efaeafeaf/src/payment/payment-strategy-action-creator.ts#L124-L126

That's why we also propose to verify if the payment_method was previously initialized (because it was initialized the first time it was displayed to shoppers), in this way we won't throw an error when this happens and the payment method would not appear on the list of available payment methods to the shopper and it'll be possible to correctly deinitialize the strategy that may be initialized
```
if (!method && !state.paymentStrategies.isInitialized(methodId)) {
}
```
This [video](https://drive.google.com/file/d/1B62aZG5jBlYpyNVy0sJW1zDFf2Lx1GdJ/view?usp=sharing) shows how it will work after applying the proposed changes

Is any of these two approaches correct for you? or do you have another better solution in mind?

orecio-bc:
@davidchin 
We were reviewing the possibility to reload all the payment methods and we concluded that there are some concerns around it.
This is because if we implement the listener, that we proposed yesterday, in the Afterpay strategy (or any other strategy), once it's **deinitialized** it won't load again the payment methods when we remove the coupon or store credit, therefore we should implement a listener to reload all payment methods when the amount changes not in a specific strategy but a general context (maybe checkout-js), but this will have some side effects:

1. There are some strategies (e.g. Klarna and Digital River) that rely on the same method of subscribing and watching for changes on the amount, this reloads their widgets, and their functionality/user experience may be directly affected.
https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/payment/strategies/digitalriver/digitalriver-payment-strategy.ts
https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/payment/strategies/klarnav2/klarnav2-payment-strategy.ts

2. Because reloading all the payment methods may remove an already listed payment method then the **deinitialized** will be executed unexpectedly (assuming we already fixed the issue related to this method that we mentioned yesterday), so the cost of testing every strategy will be huge.

After considering other options, we decided to keep the validation within the Afterpay context and to achieve this, we're validating if the amount is between the limits but this time we're subscribing to the amount changes so this way we remove the validation that we did around the 404 and we no longer trust it.
You can see the new changes here https://github.com/bigcommerce/checkout-sdk-js/pull/1175/commits/34d967ee37cad98e749fb5260c43e13000c3ba22

Here is a [video](https://drive.google.com/file/d/1QJhJl7PriUPiPqrtxYQd0FG_PpTeazFe/view?usp=sharing) showing how this approach will work

Is this approach correct for you? or do you have a better solution in mind?

davidchin:
@orecio-bc Thanks for your detailed explanation. I appreciate you spending the time finding an alternative solution. However, it still has this problem where the client has the logic to determine if AfterPay is available for use (not available if outstanding balance is under the instalment amount). Instead it should delegate such decision to the API provider instead. The reason is, if it’s important for the client to know the specific reason why payment method is unavailable, the API should inform the client about it. If that’s put in place, the client can simply reflect that to end users.

I'm not too opinionated about exactly which status code to return (404 or 422). I just want to make sure that the client is not drawing a specific conclusion from a generic status code. Perhaps you can reach out to me directly on Slack and we can discuss how you can make the aforementioned API change. Thank you.

PR #1176, opened 2021-07-08
feat(payment): PAYMENTS-6814 add NonePaymentProcessor test

leeBigCommerce:
## What?

- Add missing test to the `NonePaymentProcessor`

## Why?

- Ensures the payments endpoint it called correctly
- Ensures that the result of that result is passed to the `stepHandler`
- Ensures that the `stepHandler`'s result is returned from the processor's `process` method

## Testing / Proof

- New test

### DIFF
--- a/src/payment/strategies/ppsdk/ppsdk-payment-processor.ts
+++ b/src/payment/strategies/ppsdk/ppsdk-payment-processor.ts
@@ -3,7 +3,7 @@ import { PPSDKPaymentMethod } from '../../ppsdk-payment-method';
 
 export interface ProcessorSettings {
     paymentMethod: PPSDKPaymentMethod;
-    payment: OrderPaymentRequestBody | undefined;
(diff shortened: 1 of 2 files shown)


src/payment/strategies/ppsdk/ppsdk-payment-processor.ts
```diff
@@ -3,7 +3,7 @@ import { PPSDKPaymentMethod } from '../../ppsdk-payment-method';
 
 export interface ProcessorSettings {
     paymentMethod: PPSDKPaymentMethod;
-    payment: OrderPaymentRequestBody | undefined;
+    payment?: OrderPaymentRequestBody;
```
leeBigCommerce, line 6: Little type fix to stop `undefined` from having to be passed in explicitly as a param

animesh1987 (approved):

PR #1177, opened 2021-07-09
feat(payment): INT-4456 Add CA/FR to klarna countries list

ejesinskyBC:
https://jira.bigcommerce.com/browse/INT-4456
https://jira.bigcommerce.com/browse/INT-4712

## What?
Add FR/CA to klarna

## Why?
To support them

## Testing / Proof
...

## Siblings
https://github.com/bigcommerce/bigpay/pull/4067
https://github.com/bigcommerce/bigcommerce/pull/41800

### DIFF
--- a/src/payment/strategies/klarnav2/klarna-supported-countries.ts
+++ b/src/payment/strategies/klarnav2/klarna-supported-countries.ts
@@ -1,2 +1,2 @@
-export const supportedCountries = ['AT', 'BE', 'CH', 'DE', 'DK', 'ES', 'FI', 'GB', 'IT', 'NL', 'NO', 'NZ', 'SE'];
+export const supportedCountries = ['AT', 'BE', 'CA', 'CH', 'DE', 'DK', 'ES', 'FI', 'FR', 'GB', 'IT', 'NL', 'NO', 'NZ', 'SE'];
 export const supportedCountriesRequiringStates = ['AU'];


cmedinilla (approved):
LGTM!

ejesinskyBC:
> Is this a list of countries that Klarna supports?
> 
> Or a list of countries that this BC specific integration of Klarna supports?
> 
> Either way, is there an option to get this list from an API rather than manually hardcoding it?

@leeBigCommerce this is what we support specifically, it does not necessarily match what Klarna offers
As for the API call not that I'm aware of, their documentation is here https://developers.klarna.com/documentation/klarna-payments/in-depth-knowledge/puchase-countries-currencies-locales/ and actually doesn't always mention what they support, for example it's missing CA, FR, and NZ so must not be updated

leeBigCommerce:
Thanks @ejesinskyBC - if it's a BC notion of supported countries, do our BE services have their own notion of this list?

My point being that it's a bit odd to hardcode such config in FE code. Each time there's a change, we needed to make a PR such as this one - leading country support to be tied to an arbitrary release versions of the SDK.

If instead we got this data from the BE, countries could be added and removed without needing a new release. It would also keep the notion of supported countries perfect in sync between FE and BE code.

Maybe something to explore in a future ticket?

leeBigCommerce (approved):

ejesinskyBC:
> Thanks @ejesinskyBC - if it's a BC notion of supported countries, do our BE services have their own notion of this list?
> 
> My point being that it's a bit odd to hardcode such config in FE code. Each time there's a change, we needed to make a PR such as this one - leading country support to be tied to an arbitrary release versions of the SDK.
> 
> If instead we got this data from the BE, countries could be added and removed without needing a new release. It would also keep the notion of supported countries perfect in sync between FE and BE code.
> 
> Maybe something to explore in a future ticket?

@leeBigCommerce 100% agree with you there, if you look at the related PRs it's just updating a bunch of lists, which is sad I concur and not too uncommon an occurrence between bigpay/bcapp/checkout, it would be nice to refactor some of these things in the future

TomA-R (approved):

PR #1180, opened 2021-07-13
feat(payment): INT-3946 added locale from browser on masterpass SRC

vmparra:
## What? [INT-3946](https://jira.bigcommerce.com/browse/INT-3946)
Added locale from browser on masterpass SRC.

## Why?
In order to be able to see masterpass or SRC button based on shopper locale.

## Testing / Proof
Screen Shot 2021-06-30 at 6 53 52 PM


vmparra:
Replica of #1168

animesh1987:
Can we get a screen shot for another locale than English please?

PR #1185, opened 2021-07-15
feat(payment): INT-4585 - Refactor Zip uri data for redirect

dannharo:
## What?[INT-4585](https://jira.bigcommerce.com/browse/INT-4585)
Refactor in the zip-payment-strategy to get the `redirect_url` from the `initializeData` instead of the `additional_action_required.data ` response.

## Why?
ticket[INT-4585](https://jira.bigcommerce.com/browse/INT-4585)

## Testing / Proof
[Video](https://drive.google.com/file/d/1wvP2GrLgHZXhDRT7FnYqa8uZpfecXr77/view?usp=sharing)

## Dependency of
[bigpay: INT-4585](https://github.com/bigcommerce/bigpay/pull/4091)

## Depends on
[Bigcommerce INT-4585](https://github.com/bigcommerce/bigcommerce/pull/42037)

### DIFF
--- a/src/payment/strategies/zip/zip-payment-strategy.ts
+++ b/src/payment/strategies/zip/zip-payment-strategy.ts
@@ -1,3 +1,5 @@
+import { noop } from 'lodash';
+
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
 import { InvalidArgumentError, MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType, RequestError } from '../../../common/error/errors';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
@@ -74,6 +76,12 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
         }
 
         if (this._redirectFlowIsTrue()) {
+            const redirectUrl = this._paymentMethod?.initializationData?.redirectUrl;
+
+            if (!redirectUrl) {
+                throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+            }
+
             let nonce: { id: string };
             try {
                 nonce = JSON.parse(this._paymentMethod.clientToken);
@@ -87,10 +95,9 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
                 return await this._store.dispatch(this._paymentActionCreator.submitPayment({methodId: payment.methodId, paymentData: { nonce: nonce.id }}));
             } catch (error) {
                 if (error instanceof RequestError && error.body.status === 'additional_action_required') {
-                    return new Promise(() => {
-                        const { redirect_url } = error.body.additional_action_required.data;
-                        window.location.replace(redirect_url);
-                    });
+                    window.location.replace(redirectUrl);
+
+                    return new Promise(noop);
                 }
                 throw error;
             }
--- a/src/payment/strategies/zip/zip-payment-strategy.spec.ts
+++ b/src/payment/strategies/zip/zip-payment-strategy.spec.ts
@@ -7,7 +7,7 @@ import { of, Observable } from 'rxjs';
 import { getCartState } from '../../../cart/carts.mock';
 import { createCheckoutStore, Checkout, CheckoutRequestSender, CheckoutStore, CheckoutValidator } from '../../../checkout';
 import { getCheckout, getCheckoutState } from '../../../checkout/checkouts.mock';
-import { NotInitializedError, RequestError } from '../../../common/error/errors';
+import { MissingDataError, NotInitializedError, RequestError } from '../../../common/error/errors';
 import { getResponse } from '../../../common/http-request/responses.mock';
 import { getConfigState } from '../../../config/configs.mock';
 import { getCustomerState } from '../../../customer/customers.mock';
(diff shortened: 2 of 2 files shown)


ggarcia-bc (approved):
LGTM

JoseBobadilla20:
💚 Tested
QA Details: I made a small regression for Zip

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -74,7 +74,7 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
         }
 
         if (this._redirectFlowIsTrue()) {
-            let nonce: { id: string };
+            let nonce: { id: string; uri: string };
             try {
                 nonce = JSON.parse(this._paymentMethod.clientToken);
```
animesh1987, line 79: Why is the `uri` part of `clientToken` here? Also why is the API returning redirect_url as part of the client token?

dannharo: Hello, the same answer for this provider , We are using this approach due to that the zip/quadpay dev team has removed the `uri` data from the `Retrieve checkout` endpoint(the `get` call to the ` /checkouts/{Id}`), and the only way to get this redirect url is from the initialization call `POST` to `/checkouts`, so, we are passing this `uri` data through the Payment initialization call(where we are getting the `client_token`) to get the `uri` data and then redirect to the external zip/quadpay checkout.

animesh1987: I understand why this is implemented. But this seems like a breaking change, since older versions of SDK will still expect it as part of the older contract.

Is there a way to change the implementation so as to it doesn't break older versions of SDK.

ggarcia-bc (approved):
LGTM

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -86,10 +86,9 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
             try {
                 return await this._store.dispatch(this._paymentActionCreator.submitPayment({methodId: payment.methodId, paymentData: { nonce: nonce.id }}));
             } catch (error) {
-                if (error instanceof RequestError && error.body.status === 'additional_action_required') {
+                if (error instanceof RequestError && error.body.status === 'additional_action_required' && this._paymentMethod?.initializationData) {
```
mauricio-sg, line 89:
```suggestion
                if (error instanceof RequestError && error.body.status === 'additional_action_required') {
```

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -86,10 +86,9 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
...
                     return new Promise(() => {
-                        const { redirect_url } = error.body.additional_action_required.data;
-                        window.location.replace(redirect_url);
+                        window.location.replace(this._paymentMethod?.initializationData?.redirectUrl);
```
mauricio-sg, line 91:
```suggestion
                        window.location.replace(redirectUrl);
```

mauricio-sg (changes requested):
```
        if (this._redirectFlowIsTrue()) {
            const redirectUrl = this._paymentMethod?.initializationData?.redirectUrl;

            if (!redirectUrl) {
            }
            ...
```

src/payment/strategies/zip/zip-payment-strategy.ts
animesh1987: Don't think this will work, since the line ` const { redirect_url } = error.body.additional_action_required.data;` is removed.

mauricio-sg (approved):
LGTM 👍

dannharo:
♻️  @mauricio-sg, @animesh1987 thanks for your comments and suggestions, I already attended the comments, could you check the changes please?

@animesh1987 I changed the approach, Now we get the `uri` data from the `InitializationData`, changes on bcapp in this PR: https://github.com/bigcommerce/bigcommerce/pull/42037

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -74,6 +74,12 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
         }
 
         if (this._redirectFlowIsTrue()) {
+            const redirectUrl = this._paymentMethod?.initializationData?.redirectUrl;
+
+            if (!redirectUrl) {
+                throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
```
animesh1987, line 80: Do we need a new Error for this case?

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -88,8 +94,7 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
             } catch (error) {
                 if (error instanceof RequestError && error.body.status === 'additional_action_required') {
                     return new Promise(() => {
-                        const { redirect_url } = error.body.additional_action_required.data;
-                        window.location.replace(redirect_url);
+                        window.location.replace(redirectUrl);
```
animesh1987, line 97: Can we do similar here https://github.com/bigcommerce/checkout-sdk-js/blob/e348c5b688fc6512b4a25b17efd9680a324f4395/src/payment/strategies/quadpay/quadpay-payment-strategy.ts#L72

dannharo:
♻️  @animesh1987 Thanks for your comments and suggestions, I've attended them, could you check it again please?

JoseBobadilla20:
💚 Tested
QA Details: I made a full regression for ZIP.

src/payment/strategies/zip/zip-payment-strategy.ts
```diff
@@ -88,8 +96,9 @@ export default class ZipPaymentStrategy implements PaymentStrategy {
             } catch (error) {
                 if (error instanceof RequestError && error.body.status === 'additional_action_required') {
                     return new Promise(() => {
```
animesh1987, line 98: You can get rid of this as well

dannharo: ♻️  @animesh1987 thanks for your comments, I've changed this with your suggestions.

animesh1987 (approved):

PR #1192, opened 2021-07-21
fix(payment): INT-4672 Allow shoppers to change payment method when order finalization fails

josesonora:
## What? [INT-4672](https://jira.bigcommerce.com/browse/INT-4672)
Add a new error type, **OrderFinalizationNotCompletedError** and modify the Afterpay strategy to catch any errors during the order finalization, within the **finalize** method, and reject the promise with an OrderFinalizationNotCompletedError.

Add support for the **forgetProvider** endpoint, and use it to forget the remote checkout provider which is applied when shoppers initialize a purchase by using an offsite provider.

Modify the Afterpay strategy to apply the necessary changes to fix the issue: catch the error from the order completion, forget the provider and refresh the state with all the supported payment methods.

## Why?
Currently if we get an error after calling the **finalize** method in the strategy and the error is other than **OrderFinalizationNotRequiredError** then shoppers see a modal informing them that their payment was declined but at this point the payment methods list do not load again and the "loading" animation spin indefinitely, so shoppers cannot select another payment method or retry the payment.

We found that the issue is shared with all the providers that implemented the offsite flow (like BluesnapV2, AdyenV1, Afterpay and others), the problem occurs when, for some reason, the **finalize** step of the strategy cannot be completed correctly, for example when the provider returns that the payment validation was not valid, if something like this happens then shoppers see the error informing them that their payment was declined but the payment methods do not load again and the "loading" animation spin indefinitely.

Here are some videos showing the issue (Bluesnap and Adyen):
https://drive.google.com/file/d/1bbtOiL3-3qxV1487k1v5iAxug9DT483E/view?usp=sharing
https://drive.google.com/file/d/1pBDl0t3Hh9pfcr5HnVq--C4fUivLO4oj/view?usp=sharing

By applying these changes we can make the **checkout-sdk** to catch the error and refresh the payment method lists to allow shoppers to select another payment method.

## Testing / Proof
https://drive.google.com/file/d/15aMngXeY-Rq1oA3iE4kh0hjxBz5g6BQF/view?usp=sharing

## Depends on
[bigpay](https://github.com/bigcommerce/bigpay/pull/4070)
[bcapp](https://github.com/bigcommerce/bigcommerce/pull/42144)

## Merge process
1. Merge this dependency https://github.com/bigcommerce/bigpay/pull/4070
2. Merge this PR

### DIFF
--- a/src/payment/strategies/afterpay/afterpay-payment-strategy.ts
+++ b/src/payment/strategies/afterpay/afterpay-payment-strategy.ts
@@ -1,6 +1,8 @@
 import { CheckoutStore, CheckoutValidator, InternalCheckoutSelectors } from '../../../checkout';
 import { MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType } from '../../../common/error/errors';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
+import { OrderFinalizationNotCompletedError } from '../../../order/errors';
+import { RemoteCheckoutRequestSender } from '../../../remote-checkout';
 import { StoreCreditActionCreator } from '../../../store-credit';
 import { PaymentArgumentInvalidError } from '../../errors';
 import PaymentActionCreator from '../../payment-action-creator';
@@ -21,6 +23,7 @@ export default class AfterpayPaymentStrategy implements PaymentStrategy {
         private _orderActionCreator: OrderActionCreator,
         private _paymentActionCreator: PaymentActionCreator,
         private _paymentMethodActionCreator: PaymentMethodActionCreator,
+        private _remoteCheckoutRequestSender: RemoteCheckoutRequestSender,
         private _storeCreditActionCreator: StoreCreditActionCreator,
         private _afterpayScriptLoader: AfterpayScriptLoader
     ) {}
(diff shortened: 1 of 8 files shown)


roberbnd (approved):
LGTM

mauricio-sg (approved):
LGTM 👍

src/payment/strategies/afterpay/afterpay-payment-strategy.ts
```diff
@@ -98,7 +101,13 @@ export default class AfterpayPaymentStrategy implements PaymentStrategy {
 
         await this._store.dispatch(this._orderActionCreator.submitOrder({}, options));
 
-        return this._store.dispatch(this._paymentActionCreator.submitPayment(paymentPayload));
+        return this._store.dispatch(this._paymentActionCreator.submitPayment(paymentPayload)).catch(error =>
```
davidchin, line 104: Can you use `async/await`? Also, does this error handling flow apply to all kinds of error?

josesonora: @davidchin thanks for your feedback 🙇 
I've applied the suggested changes, and as for the question above ... yes, this error handling applies to all errors that might come from submitting the payment

src/payment/strategies/afterpay/afterpay-payment-strategy.ts
```diff
@@ -98,7 +101,14 @@ export default class AfterpayPaymentStrategy implements PaymentStrategy {
...
         await this._store.dispatch(this._orderActionCreator.submitOrder({}, options));
 
-        return this._store.dispatch(this._paymentActionCreator.submitPayment(paymentPayload));
+        try {
+            return await this._store.dispatch(this._paymentActionCreator.submitPayment(paymentPayload));
+        } catch (error) {
+            await this._remoteCheckoutRequestSender.forgetCheckout();
+            await this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethods());
+
+            return Promise.reject(new OrderFinalizationNotCompletedError(error.body?.errors?.[0]?.message));
```
davidchin, line 110:
```suggestion
            throw new OrderFinalizationNotCompletedError(error.body?.errors?.[0]?.message)
```

davidchin (approved):
LGTM :)

JuanTorres077:
💚 Passed Test

QA Details: 
Payment flow tested, working as intended.  Please take a look at below link for further details:
[1192.pdf](https://github.com/bigcommerce/checkout-sdk-js/files/7010409/1192.pdf)

PR #1195, opened 2021-07-26
feat(payment): BOLT-23 embedded Bolt payment method

Sireks:
## What?
Added Bolt embedded fields implementation into Bolt payment strategy.

## Why?
Because of the task: https://jira.bigcommerce.com/browse/BOLT-23

## Testing / Proof
Here is test result:
checkout-sdk-test

### DIFF
--- a/src/payment/strategies/bolt/bolt-payment-initialize-options.ts
+++ b/src/payment/strategies/bolt/bolt-payment-initialize-options.ts
@@ -1,13 +1,57 @@
+/**
+ * A set of options that are required to initialize the Bolt payment method with:
+ *
+ * 1) Bolt Full Checkout:
+ *
+ * If the customer chooses to pay with Bolt, he will be asked to
+ * enter his payment details via Bolt Full Checkout.
+ *
+ * ```js
+ * service.initializePayment({
+ *     methodId: 'bolt',
+ * });
+ * ```
+ *
+ * 2) Bolt Client:
+ *
+ * If the customer chooses to pay with Bolt in payment section of Checkout page,
+ * the Bolt Payment Modal will be shown, and the customer will be asked
+ * to enter payment details via Bolt Modal
+ *
+ * ```js
+ * service.initializePayment({
+ *     methodId: 'bolt',
+ *     bolt: {
+ *         useBigCommerceCheckout: true,
+ *     }
+ * });
+ * ```
+ *
+ * 3) Bolt Embedded:
+ *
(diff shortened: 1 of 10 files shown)


BC-ACherednichenko (approved):
Nicely done!

src/payment/strategies/bolt/bolt-payment-initialize-options.ts
```diff
@@ -1,13 +1,56 @@
...
+ * 2) Bolt Client:
+ *
+ * If the customer chooses to pay with Bolt, he will be asked to
+ * enter his payment details via Bolt modal.
+ *
+ * ```js
+ * service.initializePayment({
+ *     methodId: 'bolt',
+ *     bolt: {
+ *         useBigCommerceCheckout: true,
```
animesh1987, line 24: Just one question, is this always `true`. The `useBigCommerceCheckout: true` implies that the customer is going to use Bigcommerce checkout but from description feels like the customer will be using Bolt checkout.

Sireks: Yes, for this case it always true. 

Bolt payment provider has 3 different flows:
1) Full Checkout - customer pass checkout on the Bolt side;
2) Fraud protection only - customer pass the BC checkout but has bolt payment modal open when he select bolt in payments section;
3) Embedded one Click - is a flow with embedded fields

Sireks: Updated the description to make it easier to understand

src/payment/strategies/bolt/bolt-payment-strategy.ts
```diff
@@ -11,12 +11,15 @@ import PaymentMethodActionCreator from '../../payment-method-action-creator';
...
 import PaymentStrategy from '../payment-strategy';
 
-import { BoltCheckout, BoltTransaction } from './bolt';
+import { BoltCheckout, BoltEmbedded, BoltTransaction } from './bolt';
 import BoltScriptLoader from './bolt-script-loader';
 
 export default class BoltPaymentStrategy implements PaymentStrategy {
     private _boltClient?: BoltCheckout;
+    private _boltEmbedded?: BoltEmbedded;
+    private _embeddedField?: any;
```
animesh1987, line 20: Would like to understand the use of `any` here?

Sireks: Oh, my bad 😥 
Already fixed this thing

PR #1196, opened 2021-07-26
fix(payment): INT-4087 Catch stripe exception to process 3ds when payment intent is updated on the backend - StripeV3

BC-AOrtiz:
[INT-4087](https://jira.bigcommerce.com/browse/INT-4087)

## What?
- Catch exception when try to  execute the handle card action to execute the confirm card peyment when 3ds auth is required on stripev3

## Why?
 To prevent duplicate payment intent entries, we reuse the payment intent created on the initialization step and update its data and confirm on the backend.
When use a 3ds card the response of that confirmation tell us that need a 3ds auth and send that action to the front  where we use the  handleCardAction to process it, but it will try a exception with the its message say that use confirmCardPayment  instead.

This is the two case that covers this changes:

1.- With new payment Intent: 
 - Use handleCardAction to resolve 3ds
 - Send to the back confirm=true
 
2.- With payment intent udpated and confirmed:
- Use handleCardAction to resolve 3ds, it will throw a exception
- Catch the exception and use confirmCardPayment
- Send to the back confirm=false (confirmation is not need it)
 
## Testing / Proof
.
image
image

## Depends on
https://github.com/bigcommerce/bigcommerce/pull/42559

### DIFF
--- a/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
+++ b/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
@@ -32,6 +32,7 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
     private _stripeCardElements?: StripeCardElements;
     private _useIndividualCardFields?: boolean;
     private _hostedForm?: HostedForm;
+    private _reusePaymentIntent?: boolean;
 
     constructor(
         private _store: CheckoutStore,
@@ -54,10 +55,11 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
         this._initializeOptions = stripev3;
 
         const paymentMethod = this._store.getState().paymentMethods.getPaymentMethodOrThrow(methodId);
-        const { initializationData: { stripePublishableKey, stripeConnectedAccount, useIndividualCardFields } } = paymentMethod;
+        const { initializationData: { stripePublishableKey, stripeConnectedAccount, useIndividualCardFields, reusePaymentIntent } } = paymentMethod;
         const form = this._getInitializeOptions().form;
 
         this._useIndividualCardFields = useIndividualCardFields;
+        this._reusePaymentIntent = reusePaymentIntent;
 
         if (this._isCreditCard(methodId) && this._shouldShowTSVHostedForm(methodId, gatewayId) && form) {
             this._hostedForm = await this._mountCardVerificationFields(form);
@@ -113,7 +115,7 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
                 confirm: false,
             };
 
-            if (method === StripeElementType.CreditCard) {
+            if (method === StripeElementType.CreditCard && this._reusePaymentIntent === true) {
                 formattedPayload.client_token = clientToken;
             }
 
@@ -471,7 +473,16 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
 
         if (isThreeDSecureRequiredError) {
             const clientSecret = error.body.three_ds_result.token;
-            const result = await this._getStripeJs().handleCardAction(clientSecret);
+            let result;
+            let needsConfirm = false;
+
+            if (this._reusePaymentIntent) {
+                result = await this._getStripeJs().confirmCardPayment(clientSecret);
+            } else {
+                result = await this._getStripeJs().handleCardAction(clientSecret);
+                needsConfirm = true;
+            }
+
             const { id: token } = result.paymentIntent || { id: '' };
 
             if (result.error) {
@@ -484,7 +495,7 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
             const formattedPayload = {
                 credit_card_token: { token },
                 vault_payment_instrument: shouldSaveInstrument,
-                confirm: true,
+                confirm: needsConfirm,
             };
 
             const paymentPayload = this._buildPaymentPayload(methodId, formattedPayload, shouldSetAsDefaultInstrument);
--- a/src/payment/payment-methods.mock.ts
+++ b/src/payment/payment-methods.mock.ts
@@ -619,7 +619,7 @@ export function getQuadpay(): PaymentMethod {
     };
 }
 
-export function getStripeV3(method: string = 'card', shouldUseIndividualCardFields: boolean = false, isHostedFormEnabled: boolean = false): PaymentMethod {
+export function getStripeV3(method: string = 'card', shouldUseIndividualCardFields: boolean = false, isHostedFormEnabled: boolean = false, shouldReusePaymentIntent: boolean = false): PaymentMethod {
     return {
         id: method,
         logoUrl: '',
@@ -634,6 +634,7 @@ export function getStripeV3(method: string = 'card', shouldUseIndividualCardFiel
         initializationData: {
             stripePublishableKey: 'key',
             useIndividualCardFields: shouldUseIndividualCardFields,
+            reusePaymentIntent: shouldReusePaymentIntent,
         },
         type: 'PAYMENT_TYPE_API',
         clientToken: 'clientToken',
--- a/src/payment/strategies/stripev3/stripev3-payment-strategy.spec.ts
+++ b/src/payment/strategies/stripev3/stripev3-payment-strategy.spec.ts
@@ -399,6 +399,9 @@ describe('StripeV3PaymentStrategy', () => {
                 });
 
                 it('submit payment with credit card and passes back the client token', async () => {
+                    jest.spyOn(store.getState().paymentMethods, 'getPaymentMethodOrThrow')
+                        .mockReturnValue(getStripeV3('card', false, false, true));
+
                     await strategy.initialize(options);
                     const response = await strategy.execute(getStripeV3OrderRequestBodyMock());
 
@@ -868,7 +871,7 @@ describe('StripeV3PaymentStrategy', () => {
                     const errorResponse = new RequestError(getResponse({
                         ...getErrorPaymentResponseBody(),
                         errors: [
-                            { code: 'three_d_secure_required' },
+                            {code: 'three_d_secure_required'},
                         ],
                         three_ds_result: {
                             token: 'token',
@@ -891,7 +894,7 @@ describe('StripeV3PaymentStrategy', () => {
                     expect(stripeV3JsMock.handleCardAction).toHaveBeenCalled();
                 });
 
-                it('throws stripe error when call handleCardAction and then call confirmCardPayment to shopper auth and complete the payment', async () => {
+                it('call confirmCardPayment to shopper auth and complete the payment when reUsePaymentIntent is anabled', async () => {
                     const errorResponse = new RequestError(getResponse({
                         ...getErrorPaymentResponseBody(),
                         errors: [
@@ -902,14 +905,13 @@ describe('StripeV3PaymentStrategy', () => {
                         },
                         status: 'error',
                     }));
-                    const stripeErrorMessage = 'Stripe Exception';
+
+                    jest.spyOn(store.getState().paymentMethods, 'getPaymentMethodOrThrow')
+                        .mockReturnValue(getStripeV3('card', false, false, true));
 
                     jest.spyOn(paymentActionCreator, 'submitPayment')
                         .mockReturnValueOnce(of(createErrorAction(PaymentActionType.SubmitPaymentFailed, errorResponse)));
 
-                    stripeV3JsMock.handleCardAction = jest.fn(
-                        () => Promise.reject(new Error(stripeErrorMessage))
-                    );
                     stripeV3JsMock.confirmCardPayment = jest.fn(
                         () => Promise.resolve(getConfirmPaymentResponse())
                     );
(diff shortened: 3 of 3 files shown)


clopezh (approved):
LGTM 👍🏻

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -170,13 +170,20 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
 
         if (isThreeDSecureRequiredError) {
             const clientSecret = error.body.three_ds_result.token;
-            const result = await this._getStripeJs().handleCardAction(clientSecret);
+            let result;
+            let needsConfirm = false;
+            try {
+                result = await this._getStripeJs().handleCardAction(clientSecret);
+                needsConfirm = true;
+            } catch (errorCardAction) {
```
leeBigCommerce, line 178: What scenario causes `isThreeDSecureRequiredError` to return `true`

and

`handleCardAction` to throw (indicating that 3DS is _not_ needed)?

BC-AOrtiz: That first scenario is when try to complete the payment  and stripe response back that the payment needs a 3ds auth, bigpay send  back to the front a error with the code  three_d_secure_required and setup isThreeDSecureRequiredErrror as true.

We use the handleCardAction to show the modal for 3ds auth to the shopper, but that will Throw a error is the payment was previously updated on the back end, in that case we use confirmCardPayment to show the the 3ds auth modal.

The only way that 3ds in not needed is when the card do not suppport 3ds.

leeBigCommerce: Sorry, I'm not sure that I understand.

> but that will Throw a error is the payment was previously updated on the back end

What update is our back end making which will cause `handleCardAction` to throw?

Is it something we can avoid doing on the back end, rather than trying to handle here?

JelopezBC:
PR has been validated following the next escenarios:

- Auth & Capture
- Auth Only
- Refund
- Void
- Manual orders
- Apple pay
- Google Pay
- APIs
- 3DS
- Updates in Sandbox

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
BC-AOrtiz: There are two flows, the first one (this is the actual):
In the initialization we create a payment intent to get the client token, when place a order in the back a new payment intent is created with the payment data and stripe tell us that need a 3ds auth and send the three_d_secure_required error to the front to display the 3ds auth modal via handleCardAction to the shopper the shopper authorize and complete the payment.

In the second flow (added in this PR):
We get the client token from the payment intent on the initializtion step, when place the order we send back the client token to the back to get the previous payment intent instead to create a new one and update it with the payment data, in the response for that update stripe tell us that the payment intent needs no be confirmed and we send the confirm to stripe and after stripe response that the payment intent needs a 3ds auth and send three_d_secure_required to the front, in the front we use handleCardAction to show the 3ds modal, but in this case is when handleCardAction throws a exception that indacates whe need to use confirCardPayment to show the 3ds modal.

I support the 2 flows because I need this changes to be merged and hit production before the bigpay one

leeBigCommerce: Am I right in understanding that we're wanting to deploy a breaking change to Bigpay, to change how it handles Stripe V3 requests?

And to cover that change, you're altering checkout-sdk to work with both the old _and_ new flows?
(I assume with the idea of removing the old flow after the Bigpay change has been deployed)

BC-AOrtiz: That's correct

leeBigCommerce: The problem with that rollout is that consumers of existing older `checkout-sdk` versions will suddenly suffer from a broken Stripe V3 experience after the changes to Bigpay are deployed.

Typically we do not want to make breaking API changes for that reason. We want to maintain backwards compatibility as much as possible.

As such, the preferred rollout would be:

1. Keep the current `checkout-sdk` Stripe strategy and current Bigpay API functionality
2. Create a new and seperate Stripe strategy which consumes new and and seperate Bigpay API functionality
3. Work to migrate existing Stripe users from the old to the new version
4. Remove the old version functionality from both `checkout-sdk` and Bigpay when everyone is safely migrated

(Sound right to @bigcommerce/checkout ?)

BC-AOrtiz: @leeBigCommerce  this changes support the current and new functionality of bigpay, but I need this changes be on production before the bigpay ones

leeBigCommerce: Maybe I'm missing something 😄 

1. If Bigpay already works the new and old way
2. Then we update Checkout-SDK (with this PR) to work the new _and_ old way
3. We release this updated Checkout-SDK version
4. Then we remove the old way from Bigpay 
5. All merchants who are _not_ on the latest Checkout-SDK version (#3 on this list) will no longer be able to use Stripe

Are the Bigpay changes backwards compatible in a way that I'm not understanding?
Or do we have a different solution for merchants using Stripe on pre-existing Checkout-SDK versions?

BC-AOrtiz: let me try to explain better

The new way is trigger when the front send back to  bigpay the stripe client token ( this functionality is already in checkout SDK), but actually is ignored by bigpay the code that process the new way is in this PR https://github.com/bigcommerce/bigpay/pull/4135 but It can't be merged because it will fail with 3ds card, that problem is resolved by this PR for that reason this changes support the old and new.

Basically the step are
 1. Merge check sdk with old and new ways
 2. Wait th sdk chagnges reach production
 3. Merge the bigpay changes
 4. After that we can remove the old ways in sdk and bigpay.
 
all of this was caused because previously  merged a two PRs one in sdk that send back the client token and one in bigpay that update the payment intent related to tht client token, but we got some error  with 3ds cards and decided to revert the bigpay one and preserve the sdk one,this because if the bigpay changes are not ther the client token send it by the front is ignored and the old way is executed.

leeBigCommerce: After your 4 steps, will _past_ versions of Checkout-SDK still have working Stripe V3 flows?

BC-AOrtiz: If we want to support it, we need to omit the step  4 and preserve the 2 flows

BC-AOrtiz:
♻️  @leeBigCommerce comments have been addressed, please take another look.

leeBigCommerce:
Hi @BC-AOrtiz 

If I'm reading the code correctly, we're still trying to handle 3DS in two different ways in the same strategy: https://github.com/bigcommerce/checkout-sdk-js/pull/1196/files#diff-64769e0f245efac8cc1c139b49f4605cdc64f84951f12bf6906719bee001c72bR175


Is the plan still the following?

> Basically the step are
>
> 1. Merge check sdk with old and new ways
> 2. Wait th sdk chagnges reach production
> 3. Merge the bigpay changes
> 4. After that we can remove the old ways in sdk and bigpay.

BC-AOrtiz:
> Hi @BC-AOrtiz
> 
> If I'm reading the code correctly, we're still trying to handle 3DS in two different ways in the same strategy: https://github.com/bigcommerce/checkout-sdk-js/pull/1196/files#diff-64769e0f245efac8cc1c139b49f4605cdc64f84951f12bf6906719bee001c72bR175
> 
> Is the plan still the following?
> 
> > Basically the step are
> > 
> > 1. Merge check sdk with old and new ways
> > 2. Wait th sdk chagnges reach production
> > 3. Merge the bigpay changes
> > 4. After that we can remove the old ways in sdk and bigpay.

That's right

BC-AOrtiz:
@leeBigCommerce I added a condition to enable/disable this functionality to support the merchant that have an old version of SDK, could you review this PR again.

leeBigCommerce:
Thanks @BC-AOrtiz , just want to make sure I'm understanding things.

How are we planning on using the toggle to migrate merchants from the old to new flows?

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -170,13 +172,27 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
...
             const clientSecret = error.body.three_ds_result.token;
-            const result = await this._getStripeJs().handleCardAction(clientSecret);
+            let result;
+            let needsConfirm = false;
+
+            if (this._reusePaymentIntent) {
+                try {
+                    result = await this._getStripeJs().handleCardAction(clientSecret);
+                    needsConfirm = true;
+                } catch (errorCardAction) {
```
leeBigCommerce, line 182: If `this._reusePaymentIntent` being `true` represents the BE working in the 'new flow', why do we need this `catch`? 

I was expecting something like:


```
            if (this._reusePaymentIntent) {
                // new flow...
                result = await this._getStripeJs().confirmCardPayment(clientSecret);
            } else {
                // old flow...
                result = await this._getStripeJs().handleCardAction(clientSecret);
                needsConfirm = true;
            }
```

BC-AOrtiz: @leeBigCommerce I made the changes, I have some problems with the test after a rebase, but it is ready

BC-AOrtiz:
♻️ @leeBigCommerce changes have been addressed, please take another look

leeBigCommerce:
@BC-AOrtiz the code changes look good, my only remaining questions is

> How are we planning on using the toggle [`reusePaymentIntent`] to migrate merchants from the old to new flows?

BC-AOrtiz:
> @BC-AOrtiz the code changes look good, my only remaining questions is
> 
> > How are we planning on using the toggle [`reusePaymentIntent`] to migrate merchants from the old to new flows?

I created a bcapp PR to add an experiment to switch between the two flows, this is the PR

leeBigCommerce:
Is the idea that the toggle will live there forever, 

or do we have a plan to migrate merchants from old to new? If we do, how will the toggle be used in that plan?

BC-AOrtiz:
> Is the idea that the toggle will live there forever,
> 
> or do we have a plan to migrate merchants from old to new? If we do, how will the toggle be used in that plan?

By default the experiment will be disabled and  it will enabled first for those stores that report the transactions duplications and after we  could start to enabled tier by tier, and if a merchant report an issue related we can disabled on his store an review the checkout version and if it is necessary suggest to upgrade the version.
When the experiment is rolled out on all tiers and all store have the checkout upgraded we can create a ticket to start to remove the old way

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -102,7 +104,7 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
                 confirm: false,
             };
 
-            if (method === StripeElementType.CreditCard) {
+            if (method === StripeElementType.CreditCard && this._reusePaymentIntent) {
```
leeBigCommerce, line 107: Just double checking this is right.

`this._reusePaymentIntent === true` represent our new flow, as such we're turning this current behaviour _off_ 

(also removing `&& this._reusePaymentIntent` does not break any tests, so I think there's a scenario there that we haven't covered)

JelopezBC:
💚 Tested!
PR has been validated following the next escenarios:

- Enable/Disable experiment
- Manual Orders 
- Credit card
    - Auth + Capture 
    - Auth Only 
    - Capture funds
    - Refund 
    - Void 
    - Order update status in Sandbox 
- Apple Pay 
- Google Pay 

Please review attached document with validation details.
 https://docs.google.com/document/d/1MKJHQ8lgDOjvhkxKPhNdXIMJlBB1bvjt/edit?usp=sharing&ouid=117149947636046892400&rtpof=true&sd=true

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
BC-AOrtiz: the validation of this._reusePaymentIntent complemente the validation of the method, if you remove it test doesn't fail this because only ignore the parameter reusePaymentIntent received from the back and only validate the method

leeBigCommerce (approved):
👍  thanks for all the hard work on this one

leeBigCommerce:
@bigcommerce/checkout code owner review please 😄

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -471,7 +473,16 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
 
         if (isThreeDSecureRequiredError) {
             const clientSecret = error.body.three_ds_result.token;
-            const result = await this._getStripeJs().handleCardAction(clientSecret);
+            let result;
+            let needsConfirm = false;
+
+            if (this._reusePaymentIntent === true) {
```
animesh1987, line 479:
```suggestion
            if (this._reusePaymentIntent) {
```

animesh1987 (approved):

PR #1203, opened 2021-08-02
feat(payment): INT-4222 Add vaulting compatibility to Moneris

FranciscoOnt:
## What? [(INT-4222)](https://jira.bigcommerce.com/browse/INT-4222)
Revert and fix the problem that caused https://github.com/bigcommerce/checkout-sdk-js/pull/1201 
Update strategy to support vaulted instruments and add Moneris to supported instruments.

Also revert to default interfaces in initialize options.

## Why?
So vaulted instruments can be used with Moneris.
  
The original PR caused a build error due to a Weak Typed interface that attempted to remove unused fields which left  only optional fields on the interface, causing the build error.
  
If attempted to fix on the checkout-js side will cause a dependency loop causing a the build to fail.
- if that fix is merged before the bump, it will cause a import error ([example PR](https://github.com/bigcommerce/checkout-js/pull/660))
- if bumped first it will cause a type error as happened before the revert ([logs](https://app.circleci.com/pipelines/github/bigcommerce/checkout-js/3279/workflows/bf2f005a-23a3-4b1b-81b2-318fb9e81964/jobs/8228))

## Testing / Proof
UInit Testing

### DIFF
--- a/src/payment/is-hosted-intrument-like.spec.ts
+++ b/src/payment/is-hosted-intrument-like.spec.ts
@@ -1,0 +1,54 @@
+import isHostedInstrumentLike from './is-hosted-intrument-like';
+import { getHostedInstrument, getPartialHostedInstrument } from './payments.mock';
+
+describe('isHostedInstrumentLike', () => {
+    it('returns true if the object has both matching properties as undefined', () => {
+        const paymentData = getHostedInstrument(undefined, undefined);
+        expect(paymentData && isHostedInstrumentLike(paymentData)).toBeTruthy();
+    });
+
+    it('returns true if the object has any matching properties as undefined', () => {
+      const paymentData = getHostedInstrument(true, undefined);
+      expect(paymentData && isHostedInstrumentLike(paymentData)).toBeTruthy();
+    });
+
+    it('returns true if the object has any matching properties as undefined', () => {
+      const paymentData = getHostedInstrument(undefined, false);
+      expect(paymentData && isHostedInstrumentLike(paymentData)).toBeTruthy();
+    });
+
+    it('returns true if the object has both boolean values as false', () => {
+      const paymentData = getHostedInstrument(false, false);
+      expect(paymentData && isHostedInstrumentLike(paymentData)).toBeTruthy();
+    });
+
+    it('returns true if the object has both boolean values as true', () => {
+      const paymentData = getHostedInstrument(true, true);
+      expect(paymentData && isHostedInstrumentLike(paymentData)).toBeTruthy();
+    });
+
+    it('returns true if the object has both boolean values', () => {
+      const paymentData = getHostedInstrument(false, true);
+      expect(paymentData && isHostedInstrumentLike(paymentData)).toBeTruthy();
+    });
+
+    it('returns true if the object has at least one matching property', () => {
+      const paymentData = getPartialHostedInstrument();
+      expect(paymentData && isHostedInstrumentLike(paymentData)).toBeTruthy();
+    });
+
+    it('returns true if the object is object-like', () => {
+      const paymentData = {};
+      expect(paymentData && isHostedInstrumentLike(paymentData)).toBeTruthy();
+    });
+
+    it('returns false if the object is null', () => {
+      const paymentData = null;
+      expect(paymentData && isHostedInstrumentLike(paymentData)).toBeFalsy();
+    });
+
+    it('returns false if the object is undefined', () => {
+      const paymentData = undefined;
+      expect(paymentData && isHostedInstrumentLike(paymentData)).toBeFalsy();
+    });
+});
--- a/src/payment/is-hosted-intrument-like.spec.ts
+++ b/src/payment/is-hosted-intrument-like.spec.ts
@@ -1,54 +1,77 @@
 import isHostedInstrumentLike from './is-hosted-intrument-like';
-import { getHostedInstrument, getPartialHostedInstrument } from './payments.mock';
 
 describe('isHostedInstrumentLike', () => {
     it('returns true if the object has both matching properties as undefined', () => {
-        const paymentData = getHostedInstrument(undefined, undefined);
-        expect(paymentData && isHostedInstrumentLike(paymentData)).toBeTruthy();
+      expect(
+        isHostedInstrumentLike({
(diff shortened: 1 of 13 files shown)


cmedinilla (approved):
LGTM!

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -89,29 +141,72 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
...
+                throw new MissingDataError(MissingDataErrorType.MissingCheckoutConfig);
+            }
+
+            const form = this._hostedFormFactory.create(bigpayBaseUrl, formOptions);
+
+            await form.attach();
+
+            return form;
+        } catch (error) {
+            return error;
```
leeBigCommerce, line 208: This `error` is not of type `HostedForm` - Typescript is being fooled by it being an `any`.

This could lead to use accidentally trying to execute `HostedForm` methods on what is really an error object: https://github.com/bigcommerce/checkout-sdk-js/pull/1203/files#diff-d755dc9987b2df0ced566e63258108d7af6892bc592ed1cccd9f4ddbe119dfc9R159

Do we want to throw this error rather than return it? In which case we don't need the `try/catch` block.

leeBigCommerce (commented):
More comments to come, I'm still in the process of reviewing.

src/payment/strategies/moneris/moneris-payment-strategy.spec.ts
```diff
@@ -290,6 +341,83 @@ describe('MonerisPaymentStrategy', () => {
...
+
+            expect(form.validate)
+                .toHaveBeenCalled();
+        });
+
+        it('does not submit payment data with hosted form if validation fails', async () => {
+            jest.spyOn(form, 'validate')
+                .mockRejectedValue(new Error());
+
+            try {
```
leeBigCommerce, line 411: We can avoid the `try/catch`, simplifying this to:

```
            await strategy.initialize(initializeOptions);

            await expect(strategy.execute(getOrderRequestBodyVaultedCC())).rejects.toThrow();
            expect(form.submit).not.toHaveBeenCalled();
```

FranciscoOnt:
@leeBigCommerce could you please re-review?

src/payment/strategies/moneris/moneris-payment-strategy.spec.ts
```diff
@@ -117,17 +118,18 @@ describe('MonerisPaymentStrategy', () => {
         jest.spyOn(storeCreditActionCreator, 'applyStoreCredit')
             .mockReturnValue(applyStoreCreditAction);
 
-        jest.spyOn(window, 'addEventListener')
-            .mockImplementation((type, listener) => {
-                return eventEmitter.addListener(type, listener);
-            });
+        jest.spyOn(window, 'addEventListener');
```
leeBigCommerce, line 121: Is this spy still required?

FranciscoOnt:
@leeBigCommerce , I've made the changes to the type guard and added the tests

src/payment/is-hosted-intrument-like.spec.ts
```diff
@@ -0,0 +1,54 @@
+import isHostedInstrumentLike from './is-hosted-intrument-like';
+import { getHostedInstrument, getPartialHostedInstrument } from './payments.mock';
+
+describe('isHostedInstrumentLike', () => {
+    it('returns true if the object has both matching properties as undefined', () => {
+        const paymentData = getHostedInstrument(undefined, undefined);
+        expect(paymentData && isHostedInstrumentLike(paymentData)).toBeTruthy();
```
leeBigCommerce, line 7: It's best to only have the unit under test executed with an `expect` block.

`expect(isHostedInstrumentLike(paymentData)).toBe(true)`

Your existence check on `paymentData` is redundant anyhow as Typescript guarantees a truthy output from your `getHostedInstrument` function.

It's also best to use as explicit an expectation as possible. I.e. `toBe(true)` instead of `toBeTruthy()` 

Lastly, the `getHostedInstrument` is an over optimisation. With tests it's better to be explicit and avoid indirection, rather than worrying about DRY.

I.e.
```
      expect(
        isHostedInstrumentLike({
          shouldSaveInstrument: undefined,
          shouldSetAsDefaultInstrument: undefined,
        })
      ).toBe(true);
```

src/payment/is-hosted-intrument-like.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+    it('returns true if the object is object-like', () => {
+      const paymentData = {};
+      expect(paymentData && isHostedInstrumentLike(paymentData)).toBeTruthy();
+    });
+
+    it('returns false if the object is null', () => {
+      const paymentData = null;
+      expect(paymentData && isHostedInstrumentLike(paymentData)).toBeFalsy();
```
leeBigCommerce, line 47: This is a good example of why it's best to have any the unit under test within an `expect`.

`paymentData && isHostedInstrumentLike(paymentData)` here is effectively `null && isHostedInstrumentLike(paymentData)` which will evaluate to just `null`. `isHostedInstrumentLike` won't be executed and won't be evaluated in the assertion.

We can fix this up with:
`expect(isHostedInstrumentLike(null)).toBe(false);`

src/payment/is-hosted-intrument-like.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+    it('returns false if the object is undefined', () => {
+      const paymentData = undefined;
+      expect(paymentData && isHostedInstrumentLike(paymentData)).toBeFalsy();
```
leeBigCommerce, line 52: Much the same as above

src/payment/payments.mock.ts
```diff
@@ -35,6 +35,19 @@ export function getCreditCardInstrument(): CreditCardInstrument {
     };
 }
 
+export function getHostedInstrument(defaultSave?: boolean, defaultSet?: boolean): HostedInstrument {
```
leeBigCommerce, line 38: We can probably remove these helpers

leeBigCommerce (approved):

TomA-R (approved):

PR #1205, opened 2021-08-07
feat(payment): CHECKOUT-5906 Payment step "Name on Card" of checkout requires label for screen readers

bc-NikitaPuzanenko:
## What?
Adding optional field "accessibilityLabel" to BraintreeFormField Options and setting aria-label attribute on card name input.
## Why?
Currently Braintree does not make use of the accessibility label made available [here](https://github.com/bigcommerce/checkout-js/pull/512)

This means that currently users relying on screen readers to checkout have a hard time understanding what part of the credit card form they are currently in.

## Testing / Proof
Before:
Screen Shot 2021-08-05 at 5 59 13 PM

After:
Screen Shot 2021-08-05 at 5 58 31 PM

### DIFF
--- a/src/payment/strategies/braintree/braintree-payment-options.ts
+++ b/src/payment/strategies/braintree/braintree-payment-options.ts
@@ -148,6 +148,7 @@ export interface BraintreeStoredCardFieldsMap {
 }
 
 export interface BraintreeFormFieldOptions {
+    accessibilityLabel?: string;
     containerId: string;
     placeholder?: string;
 }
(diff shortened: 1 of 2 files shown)


PascalZajac:
Are there any unit tests for these components we should update?

bc-NikitaPuzanenko:
The only test I could find is [withHostedCreditCardFieldset](https://github.com/bigcommerce/checkout-js/blob/master/src/app/payment/hostedCreditCard/withHostedCreditCardFieldset.spec.tsx) but that appears to be specific to Auth.net. I cannot find anything that tests the `BraintreeRegularField` class explicitly.

@PascalZajac

PascalZajac (approved):
Thanks for checking.

PR #1208, opened 2021-08-11
feat(payment): BOLT-19 added Bolt Embedded One Click Checkout impleme…

Sireks:
feat(payment): BOLT-19 added Bolt Embedded One Click Checkout implementation and add generic executePaymentProviderCheckout to customer strategy

## What?
Add Bolt Embedded One Click Checkout implementation.

## Why?
Because of the task: https://jira.bigcommerce.com/browse/BOLT-19

## What was done?
- Added new Bolt customer strategy;
- Extracted customer strategy with `executePaymentProviderCheckout`;
- Fixed tests of another payment providers customer strategies;

## Testing / Proof
All tests passed locally:
Screenshot 2021-08-11 at 12 27 23

## Sibling pull-requests
checkout-js: https://github.com/bigcommerce/checkout-js/pull/666


src/customer/strategies/googlepay/googlepay-customer-strategy.ts
```diff
@@ -66,6 +66,14 @@ export default class GooglePayCustomerStrategy implements CustomerStrategy {
         );
     }
 
+    executePaymentProviderCheckout(options?: ExecutePaymentProviderCheckoutOptions): Promise<InternalCheckoutSelectors> {
+        if (typeof options?.fallback === 'function') {
```
BC-ACherednichenko, line 70: per our discussion, may be better to to remove this condition `==='functtion'` and refer only to the presence of fallback

BC-ACherednichenko (approved):
Nicely done! just one minor suggestion

src/customer/strategies/googlepay/googlepay-customer-strategy.ts
Sireks: Thanks! 🙏 
Already updated

src/customer/strategies/amazon-pay-v2/amazon-pay-v2-customer-strategy.ts
```diff
@@ -63,6 +63,12 @@ export default class AmazonPayV2CustomerStrategy implements CustomerStrategy {
         );
     }
 
+    executePaymentProviderCheckout(options?: ExecutePaymentProviderCheckoutOptions): Promise<InternalCheckoutSelectors> {
```
animesh1987, line 66: Just want to understand if this is part of the scope? Since this task is for `Bolt`.

Sireks: Yes, it is a new generic method for all providers customer strategies.

Sireks: We can't build this kind of feature only for bolt, because we should have an ability to use the feature for another payment providers.

animesh1987: Thanks for the explanation. Nice work on doing that and can see the thinking of how we can use it for other providers too.

src/customer/customer-strategy-reducer.ts
```diff
@@ -174,6 +187,19 @@ function statusesReducer(
...
         });
 
+    case CustomerStrategyActionType.ExecutePaymentMethodCheckoutRequested:
+        return objectMerge(statuses, {
+            isExecutingPaymentMethodCheckout: true,
+            executePaymentMethodCheckoutMethodId: action.meta && action.meta.methodId,
+        });
+
+    case CustomerStrategyActionType.ExecutePaymentMethodCheckoutFailed:
+    case CustomerStrategyActionType.ExecutePaymentMethodCheckoutSucceeded:
```
animesh1987, line 197: Are there any errors that we need to catch and set in case of faliure?

src/customer/customer-strategy-selector.ts
```diff
@@ -7,10 +7,12 @@ import CustomerStrategyState, { DEFAULT_STATE } from './customer-strategy-state'
 export default interface CustomerStrategySelector {
     getSignInError(methodId?: string): Error | undefined;
     getSignOutError(methodId?: string): Error | undefined;
+    getExecutePaymentMethodCheckoutError(methodId?: string): Error | undefined;
```
animesh1987, line 10: getExecutePaymentMethodCheckoutError(methodId?: string): Error | undefined;
```suggestion
    getExecutePaymentMethodCheckoutError?(methodId?: string)?: Error | undefined;
```

Sireks: Do I need to fix all getError methods in the file? Or can I live it right how it is to match the stile of the file?

src/customer/customer-strategy-reducer.ts
Sireks: We catch all the errors in `_executePaymentMethodProvider` method in `BoltCustomerStrategy` and throw them to checkout page, where we handle them in those way that we need. 

I think that all of the errors in the '_executePaymentMethodProvider' method will be about failures on provider service side or about provided arguments errors from checkout side.

src/customer/customer-strategy-selector.ts
animesh1987: That will be nice you can 👍

src/customer/strategies/bolt/bolt-customer-strategy.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    signOut(options?: CustomerRequestOptions): Promise<InternalCheckoutSelectors> {
+        return this._store.dispatch(
+            this._customerActionCreator.signOutCustomer(options)
+        );
+    }
+
+    async executePaymentMethodCheckout(options?: ExecutePaymentMethodCheckoutOptions) {
+        const { methodId, fallback } = options || {};
+        const defaultFallback = () => {};
```
animesh1987, line 71: Why not use `noop` here?

src/customer/customer-request-options.ts
```diff
@@ -101,3 +108,17 @@ export interface CustomerInitializeOptions extends CustomerRequestOptions {
...
+ * current checkout flow.
+ *
+ * Some payment methods have specific suggestion for customer to pass
+ * the customer step. For example, Bolt suggests the customer to use
+ * their custom checkout with prefilled form values. As a result, you
+ * may need to provide additional information.
+ *
+ */
+export interface ExecutePaymentMethodCheckoutOptions extends CustomerRequestOptions {
+    fallback?(): void;
```
animesh1987, line 123: This method seems to be doing a few things, and fallback is a very generic name. On discussion this feels like something more like `continueWithCheckoutCallback` ?

src/customer/strategies/bolt/bolt-customer-strategy.ts
Sireks: We can use it here, thanks! 🙏

src/customer/customer-request-options.ts
Sireks: `continueWithCheckoutCallback` is more valuable and matching name for this case. Thanks, I will fix it today

src/customer/customer-strategy-selector.ts
Sireks: I can't do that :(
Because when I try to change the code `getExecutePaymentMethodCheckoutError?(methodId?: string)?: Error;` TSLint returns an error (see screenshot bellow)

Screenshot 2021-08-20 at 13 18 18

src/customer/strategies/bolt/bolt-customer-strategy.ts
Sireks: I added `noop` as default value for `fallback` and  removed checkoutFallback variable because it is unnecessary now.

src/customer/customer-strategy-selector.ts
animesh1987: Sorry was a typo, supposed to be 
`    getExecutePaymentMethodCheckoutError?(methodId?: string): Error;
`

Sireks: I think, that the reason why we should left it (`getExecutePaymentMethodCheckoutError(methodId?: string): Error | undefined;`) that we can have undefined as a result of the function call. Is it right?

animesh1987 (approved):

PR #1209, opened 2021-08-13
fix(payment): INT-4761 Improvements to the Afterpay strategy

josesonora:
## What? [INT-4761](https://jira.bigcommerce.com/browse/INT-4761)
Modify the strategy to load the payment method correctly and specify the store credit if necessary

## Why?
Afterpay is a multi-option provider but currently the strategy is loading the payment method as a unique method so it's causing a bad user experience

## Testing / Proof
**BEFORE**
image
**AFTER**
image

@bigcommerce/checkout @bigcommerce/payments

PR #1213, opened 2021-08-17
fix(payment): INT-4594 [Afterpay] display correct error message when amount is out of limits

orecio-bc:
## What? [INT-4594](https://jira.bigcommerce.com/browse/INT-4594)
Display the correct error message when trying to purchase with Afterpay and the amount is outside the limits provided by Afterpay (due to a Coupon code)

## Why?
When trying to purchase with Afterpay and the amount is outside the limits provided by Afterpay (due to a Coupon code), the initialize on BIGPAY will return an error response, which is going to be interpreted by BCAPP and then the payment strategy will show a more readable message to the shopper.

## Depends on
BCAPP
[#42340](https://github.com/bigcommerce/bigcommerce/pull/42340)

## Testing / Proof
https://drive.google.com/file/d/1kVV0eBRbDoQL9Ak66ciLrZdTonppKnd2/view?usp=sharing

### DIFF
--- a/src/payment/strategies/afterpay/afterpay-payment-strategy.ts
+++ b/src/payment/strategies/afterpay/afterpay-payment-strategy.ts
@@ -1,5 +1,6 @@
 import { CheckoutStore, CheckoutValidator, InternalCheckoutSelectors } from '../../../checkout';
-import { MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType } from '../../../common/error/errors';
+import { InvalidArgumentError, MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType, RequestError } from '../../../common/error/errors';
+import { RequestOptions } from '../../../common/http-request';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotCompletedError } from '../../../order/errors';
 import { RemoteCheckoutRequestSender } from '../../../remote-checkout';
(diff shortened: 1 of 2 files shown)


josesonora (approved):

src/payment/strategies/afterpay/afterpay-payment-strategy.ts
```diff
@@ -49,16 +50,20 @@ export default class AfterpayPaymentStrategy implements PaymentStrategy {
...
+        if (!gatewayId || !methodId) {
+            throw new PaymentArgumentInvalidError(['payment.gatewayId', 'payment.methodId']);
         }
 
         let state = this._store.getState();
         const currencyCode = state.cart.getCart()?.currency.code || '';
         const countryCode = this._mapCurrencyToISO2(currencyCode);
-        const { useStoreCredit } = payload;
+        let { useStoreCredit } = payload;
+        useStoreCredit = typeof useStoreCredit === 'undefined' ? this._store.getState().checkout.getCheckoutOrThrow().isStoreCreditApplied : useStoreCredit;
```
davidchin, line 66: I think this condition will cause `applyStoreCredit` call below to fire, even if there's no need to. You only need to call `applyStoreCredit` if `payload.useStoreCredit` is defined. What you have now is, if `payload.useStoreCredit` is undefined, call `applyStoreCredit` again with the same value that is currently set for the checkout session, which is redundant. As far as I can tell from reading the code, I don't think you need to make a change at all to achieve the goal of this PR?

orecio-bc: indeed, this change comes from this PR https://github.com/bigcommerce/checkout-sdk-js/pull/1209 which is needed because some of the changes that they are introducing there

orecio-bc:
@davidchin I removed the dependency with the PR that was introducing those changes, can you review it again? ♻️

davidchin (approved):
Thanks @orecio-bc 👍

JuanTorres077:
💚 Tested
QA Details: 
Working as Intended...
We tested max and min boundary,  and also we test the error message for them. 
Also we tested along with coupons. store credits were out of the scope...
for further details please take a look at below link:
[1213.pdf](https://github.com/bigcommerce/checkout-sdk-js/files/7022942/1213.pdf)

PR #1215, opened 2021-08-17
fix(payment): INT-4674 Using  forgetCheckout instead of signOut on Google Pay

BC-PabloRivera:
## What? [INT-4674](https://jira.bigcommerce.com/browse/INT-4674)
Using the forgetCheckout action instead of the signOut action when clicking sign out on Google Pay

## Why?
Using the signOut action was deleting all the checkout data

## Testing / Proof
image
After signing out
image

### DIFF
--- a/src/customer/strategies/googlepay/googlepay-customer-strategy.ts
+++ b/src/customer/strategies/googlepay/googlepay-customer-strategy.ts
@@ -61,9 +61,7 @@ export default class GooglePayCustomerStrategy implements CustomerStrategy {
             return Promise.resolve(this._store.getState());
         }
 
-        return this._store.dispatch(
-            this._remoteCheckoutActionCreator.signOut(payment.providerId, options)
-        );
+        return this._store.dispatch(this._remoteCheckoutActionCreator.forgetCheckout(payment.providerId, options));
     }
 
     executePaymentMethodCheckout(options?: ExecutePaymentMethodCheckoutOptions): Promise<InternalCheckoutSelectors> {
--- a/src/remote-checkout/remote-checkout-action-creator.ts
+++ b/src/remote-checkout/remote-checkout-action-creator.ts
@@ -1,6 +1,9 @@
 import { createAction, createErrorAction, Action } from '@bigcommerce/data-store';
-import { Observable, Observer } from 'rxjs';
+import { concat, defer, of, Observable, Observer } from 'rxjs';
+import { catchError } from 'rxjs/operators';
 
+import { CheckoutActionCreator } from '../checkout';
+import { throwErrorAction } from '../common/error';
 import { RequestOptions } from '../common/http-request';
 
 import { RemoteCheckoutActionType } from './remote-checkout-actions';
@@ -13,7 +16,8 @@ import { RemoteCheckoutStateData } from './remote-checkout-state';
  */
 export default class RemoteCheckoutActionCreator {
     constructor(
-        private _remoteCheckoutRequestSender: RemoteCheckoutRequestSender
+        private _remoteCheckoutRequestSender: RemoteCheckoutRequestSender,
+        private _checkoutActionCreator: CheckoutActionCreator
     ) {}
 
     initializeBilling(methodId: string, params?: { referenceId: string }, options?: RequestOptions): Observable<Action> {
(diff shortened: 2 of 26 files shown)


JoseAlvarezHdz (approved):
LGTM 👍🏾

leeBigCommerce (approved):

JelopezBC:
💚 Tested!

PR has been validated Google pay sign out is working as expected

Tested scenarios:

- Orders with Auth Only 
- Orders with Auth & Capture 
- Credit card, google pay.
- Payment providers:
- stripe
- authorize
- braintree
- adyenv2
- etc.

Please review attached document with validation screenshots:
[INT-4674 Google Pay - Signing out of Google Pay to choose another payment method restarts the checkout flow.docx](https://github.com/bigcommerce/checkout-sdk-js/files/7122517/INT-4674.Google.Pay.-.Signing.out.of.Google.Pay.to.choose.another.payment.method.restarts.the.checkout.flow.docx)

src/remote-checkout/remote-checkout-action-creator.ts
```diff
@@ -91,6 +91,21 @@ export default class RemoteCheckoutActionCreator {
         });
     }
 
+    forgetCheckout(methodId: string, options?: RequestOptions): Observable<Action> {
+        return Observable.create((observer: Observer<Action>) => {
+            observer.next(createAction(RemoteCheckoutActionType.SignOutRemoteCustomerRequested, undefined, { methodId }));
```
animesh1987, line 96: Can we please create new action types for this action and not reuse the ones from `SignOut`, since they are separate actions.

animesh1987: Also think you may need to reload the checkout as part of this, since there might be some information related to session that is not cleared for the shopper.

BC-PabloRivera: I don't think this needs to be reloaded because forgetCheckout deletes all the checkout information from the backend and in this case there is no information that needs to be loaded again in the frontend.

BC-PabloRivera:
♻️ @animesh1987

src/remote-checkout/remote-checkout-action-creator.ts
davidchin: @BC-PabloRivera `forgetCheckout` only removes certain session data. It doesn't wipe out all the checkout information. Most of the fields of checkout won't change after removing those session data, except for `payments` field. If you don't reload it, the browser client would be referencing stale data.

BC-PabloRivera: Ok. When you talk about reloading the checkout, is something like `checkoutActionCreator.loadCurrentCheckout()` what you expect or do you have something different in mind?

BC-PabloRivera:
♻️ @animesh1987 @davidchin

src/remote-checkout/remote-checkout-action-creator.ts
animesh1987: Sorry for the late reply, you should be able to call that in the action itself.

src/customer/strategies/googlepay/googlepay-customer-strategy.ts
```diff
@@ -61,9 +62,10 @@ export default class GooglePayCustomerStrategy implements CustomerStrategy {
             return Promise.resolve(this._store.getState());
         }
 
-        return this._store.dispatch(
-            this._remoteCheckoutActionCreator.signOut(payment.providerId, options)
-        );
+        const x = this._store.dispatch(this._remoteCheckoutActionCreator.forgetCheckout(payment.providerId, options));
+        this._store.dispatch(this._checkoutActionCreator.loadCurrentCheckout());
```
animesh1987, line 66: This should be called as part of the `forgetCheckout` action

BC-PabloRivera:
♻️ @animesh1987

src/remote-checkout/remote-checkout-action-creator.ts
```diff
@@ -91,6 +93,23 @@ export default class RemoteCheckoutActionCreator {
         });
     }
 
+    forgetCheckout(methodId: string, options?: RequestOptions): Observable<Action> {
+        return Observable.create((observer: Observer<Action>) => {
```
animesh1987, line 97: Sorry again but think we need to do a `concat` of calls. You can see example here https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/checkout/checkout-action-creator.ts#L37.

BC-PabloRivera:
♻️ @animesh1987

animesh1987 (approved):
Nice work, and thanks for the patience. LGTM 👍

BC-ISilva:
💚 Tested!
QE details:
Following scenarios and Payments providers have been validated:

**Providers:**
1. Authorize.net
2. StripeV3
3. AdyenV2.
4. Braintree.


**Scenarios:**
1. Place one order with Google pay.
2. Sign-Off Google pay. 
3. place order with credit card.
4. Void
5. Refund.

Please review attaches word document with validation details.
[INT-4674 - Google Pay - Signing out of Google Pay to choose another payment method restarts the checkout flow.docx](https://github.com/bigcommerce/checkout-sdk-js/files/7197148/INT-4674.-.Google.Pay.-.Signing.out.of.Google.Pay.to.choose.another.payment.method.restarts.the.checkout.flow.docx)

PR #1218, opened 2021-08-19
fix(payment): BOLT-47 fixed an issue of throwing payment method error…

Sireks:
… on initialization for Bolt Full Checkout flow

## What?
We should not throw an error on Bolt Payment Strategy initialisation if useBigcommerceCheckout is false and publishableKey is not provided.

## Why?
Because Bolt gets an error each time when they try to initialise Bolt Payment Strategy for Bolt Full Checkout flow

## Testing / Proof
Here is test result:
Screenshot 2021-08-19 at 12 40 45

@bigcommerce/checkout


BC-ACherednichenko (approved):
LGTM

vneutrino (approved):

animesh1987 (approved):

PR #866, opened 2020-05-08
feat(payment): INT-2119 adding payment and shipping strategies

BC-AxelMalagon:
Jira ticket: https://jira.bigcommerce.com/browse/INT-2119
## What?
Added a payment strategy in order to  handle the purchase process it was created a payment strategy using three-ds flow to redirect to amazon site.

Besides adding the payment strategy it was needed to create a shipping strategy for amazon pay where we bind an Edit button to a ChangeAction in amazon.

There's a scenario where in case the address is not valid is not possible to continue with checkout process and is needed to return to amazon and correct the shipping address on that end.

##### Related PRs
https://github.com/bigcommerce/checkout-js/pull/271

# Why?
Amazon pay v2 integration handle payment has a different flow process in order to retrieve checkoutSessionId from amazon and then place order, this is done so that in the end we redirect to amazon in order to finalize the purchase.

The shipping strategy was needed because at the shipping step it was not possible to have the required checkoutSessionId to bind the Edit button to amazon ChangeAction. At that step it was not possible use the customerStrategy nor the paymentStrategy to have the needed data.

@bigcommerce/checkout @bigcommerce/payments

PR #1219, opened 2021-08-19
feat(payment): INT-4231 deleting spinner functionality in amazonpay

deov31:
[INT-4231](https://jira.bigcommerce.com/browse/INT-4231)

## What?
I changed payment strategy to reject the promise that triggers the spinner in amazonpay payment step.

## Why?
Amazonpay team changed US region script to open a modal instead of performing a full redirect to their hosted page, this causes an infinite spinner running if shopper closes the modal.

The change covers US considering that is the only region using the modal, for the other regions (UK and EU) we will keep the same functionality as before.

## Testing / Proof
Demo issue to resolve:
https://drive.google.com/file/d/1nXrKE0YsvSFxsxjHED1CyFZSnE6vJC-q/view?usp=sharing

Demo US:
https://drive.google.com/file/d/1L8pXznmXBrfPZ6dTz91CWqSNij6O7HpY/view?usp=sharing

Demo UK:
https://drive.google.com/file/d/1Onur_lY6gkVD0oxFzWMbZ919c3XmnL9O/view?usp=sharing

Demo EU:
https://drive.google.com/file/d/1MEC2rPRouaxhJOsr4eyC3MmLo0SdBpXC/view?usp=sharing

### DIFF
--- a/src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-strategy.ts
+++ b/src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-strategy.ts
@@ -5,7 +5,7 @@ import { InvalidArgumentError, MissingDataError, MissingDataErrorType, NotInitia
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
 import { getShippableItemsCount } from '../../../shipping';
-import { PaymentArgumentInvalidError } from '../../errors';
+import { PaymentArgumentInvalidError, PaymentMethodCancelledError } from '../../errors';
 import PaymentActionCreator from '../../payment-action-creator';
 import PaymentMethod from '../../payment-method';
 import PaymentMethodActionCreator from '../../payment-method-action-creator';
@@ -40,11 +40,11 @@ export default class AmazonPayV2PaymentStrategy implements PaymentStrategy {
 
         await this._amazonPayV2PaymentProcessor.initialize(paymentMethod);
 
-        const { paymentToken } = paymentMethod.initializationData;
+        const { paymentToken, region } = paymentMethod.initializationData;
         const buttonId = amazonpay.editButtonId;
 
         if (paymentToken && buttonId) {
-            this._bindEditButton(buttonId, paymentToken, 'changePayment');
+            this._bindEditButton(buttonId, paymentToken, 'changePayment', this._isModalFlow(region));
         } else {
             this._walletButton = this._createSignInButton(paymentMethod);
         }
@@ -64,7 +64,7 @@ export default class AmazonPayV2PaymentStrategy implements PaymentStrategy {
         const state = await this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId));
         const paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(methodId);
 
-        const { paymentToken } = paymentMethod.initializationData;
+        const { paymentToken, region } = paymentMethod.initializationData;
 
         if (paymentToken) {
             const paymentPayload = {
@@ -93,6 +93,19 @@ export default class AmazonPayV2PaymentStrategy implements PaymentStrategy {
 
         this._walletButton.click();
 
+        // Focus of parent window used to try and detect the user cancelling the Amazon log in modal
+        // Should be refactored if/when Amazon add a modal close hook to their SDK
+        if (this._isModalFlow(region)) {
+            return new Promise((_, reject) => {
+                const onFocus = () => {
+                    window.removeEventListener('focus', onFocus);
+                    reject(new PaymentMethodCancelledError('Shopper needs to login to Amazonpay to continue'));
+                };
+
+                window.addEventListener('focus', onFocus);
+            });
+        }
+
         return new Promise<never>(noop);
     }
 
@@ -111,30 +124,32 @@ export default class AmazonPayV2PaymentStrategy implements PaymentStrategy {
         return Promise.resolve(this._store.getState());
     }
 
-    private _bindEditButton(id: string, sessionId: string, changeAction: AmazonPayV2ChangeActionType): void {
+    private _bindEditButton(id: string, sessionId: string, changeAction: AmazonPayV2ChangeActionType, isModalFlow: boolean): void {
         const button = document.getElementById(id);
 
         if (!button || !button.parentNode) {
             return;
         }
 
-        const clone = button.cloneNode(true);
-        button.parentNode.replaceChild(clone, button);
+        if (!isModalFlow) {
+            const clone = button.cloneNode(true);
+            button.parentNode.replaceChild(clone, button);
 
-        clone.addEventListener('click', () => this._showLoadingSpinner(() => new Promise(noop)));
+            clone.addEventListener('click', () => this._showLoadingSpinner());
+        }
 
         this._amazonPayV2PaymentProcessor.bindButton(id, sessionId, changeAction);
     }
 
-    private _showLoadingSpinner(callback?: () => Promise<void> | Promise<never>): Promise<InternalCheckoutSelectors> {
-        return this._store.dispatch(this._paymentStrategyActionCreator.widgetInteraction(() => {
+    private _isModalFlow(region: string) {
+        return region === 'us';
+    }
 
-            if (callback) {
-                return callback();
-            }
-
-            return Promise.reject();
-        }), { queueId: 'widgetInteraction' });
+    private _showLoadingSpinner(): Promise<InternalCheckoutSelectors> {
+        return this._store.dispatch(
+            this._paymentStrategyActionCreator.widgetInteraction(() => new Promise(noop)),
+            { queueId: 'widgetInteraction' }
+        );
     }
 
     private _createContainer(): HTMLDivElement {
--- a/src/shipping/strategies/amazon-pay-v2/amazon-pay-v2-shipping-strategy.ts
+++ b/src/shipping/strategies/amazon-pay-v2/amazon-pay-v2-shipping-strategy.ts
@@ -53,11 +53,12 @@ export default class AmazonPayV2ShippingStrategy implements ShippingStrategy {
 
         await this._amazonPayV2PaymentProcessor.initialize(paymentMethod);
 
-        const { paymentToken } = paymentMethod.initializationData;
+        const { paymentToken, region } = paymentMethod.initializationData;
         const buttonId = amazonpay.editAddressButtonId;
 
         if (paymentToken && buttonId) {
-            this._bindEditButton(buttonId, paymentToken, 'changeAddress');
+            const shouldShowLoadingSpinner = this._shouldShowLoadingSpinner(region);
+            this._bindEditButton(buttonId, paymentToken, 'changeAddress', shouldShowLoadingSpinner);
         }
 
         return this._store.getState();
@@ -69,29 +70,31 @@ export default class AmazonPayV2ShippingStrategy implements ShippingStrategy {
         return Promise.resolve(this._store.getState());
     }
 
-    private _bindEditButton(id: string, sessionId: string, changeAction: AmazonPayV2ChangeActionType): void {
+    private _bindEditButton(id: string, sessionId: string, changeAction: AmazonPayV2ChangeActionType, shouldShowLoadingSpinner: boolean ): void {
         const button = document.getElementById(id);
 
         if (!button || !button.parentNode) {
             return;
         }
 
-        const clone = button.cloneNode(true);
-        button.parentNode.replaceChild(clone, button);
+        if (shouldShowLoadingSpinner) {
+            const clone = button.cloneNode(true);
+            button.parentNode.replaceChild(clone, button);
 
-        clone.addEventListener('click', () => this._showLoadingSpinner(() => new Promise(noop)));
+            clone.addEventListener('click', () => this._showLoadingSpinner());
+        }
 
         this._amazonPayV2PaymentProcessor.bindButton(id, sessionId, changeAction);
     }
 
-    private _showLoadingSpinner(callback?: () => Promise<void> | Promise<never>): Promise<InternalCheckoutSelectors> {
-        return this._store.dispatch(this._shippingStrategyActionCreator.widgetInteraction(() => {
+    private _showLoadingSpinner(): Promise<InternalCheckoutSelectors> {
+        return this._store.dispatch(
+            this._shippingStrategyActionCreator.widgetInteraction(() => new Promise(noop)),
+            { queueId: 'widgetInteraction' }
+        );
+    }
 
-            if (callback) {
-                return callback();
-            }
-
-            return Promise.reject();
-        }), { queueId: 'widgetInteraction' });
+    private _shouldShowLoadingSpinner(region: string) {
+        return region !== 'us';
     }
 }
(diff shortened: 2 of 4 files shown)


mauricio-sg (approved):
LGTM 🤓

leeBigCommerce:
Seems a little fragile, what happens if the UK begins behaving like the US suddenly and also uses a modal? 

Also, what does Amazon suggest to do in this scenario? I've tried to dig through their documentation, but they don't seem to offer any way of knowing if the modal has been closed or any way of knowing _if_ they will use a modal.

Is this a problem that we've invented? I can't find any of the Amazon docs talking about how to handle overlays or loading states - so I think the done thing may be not to have them.

deov31:
Hi @leeBigCommerce, well in case any other region script start open a modal we will have to use same fix for it. I will make sure to let them know that if they decide to update any other script this will create a bug on our side.

I understand your concern but right now Amazon doesn't offer any callback or notification that we can use to handle the close event of the modal. We've talk to them and they are considering updating this on their side but unfortunately they don't have a date for when this will be an option. So, until then, this is what we can do to fix the bug, if they offer anything in the near future, we can revert this change and start managing the event instead of using this fix.

src/shipping/strategies/amazon-pay-v2/amazon-pay-v2-shipping-strategy.ts
```diff
@@ -69,17 +70,19 @@ export default class AmazonPayV2ShippingStrategy implements ShippingStrategy {
...
         const button = document.getElementById(id);
 
         if (!button || !button.parentNode) {
             return;
         }
 
-        const clone = button.cloneNode(true);
-        button.parentNode.replaceChild(clone, button);
+        if (shouldShowLoadingSpinner) {
+            const clone = button.cloneNode(true);
```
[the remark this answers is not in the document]
leeBigCommerce: Also applies to similar code in `src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-strategy.ts`

src/shipping/strategies/amazon-pay-v2/amazon-pay-v2-shipping-strategy.ts
```diff
@@ -69,17 +70,19 @@ export default class AmazonPayV2ShippingStrategy implements ShippingStrategy {
...
+            button.parentNode.replaceChild(clone, button);
 
-        clone.addEventListener('click', () => this._showLoadingSpinner(() => new Promise(noop)));
+            clone.addEventListener('click', () => this._showLoadingSpinner(() => new Promise(noop)));
```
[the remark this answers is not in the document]
leeBigCommerce: Also applies to similar code in `src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-strategy.ts`

src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-strategy.ts
```diff
@@ -93,7 +93,9 @@ export default class AmazonPayV2PaymentStrategy implements PaymentStrategy {
 
         this._walletButton.click();
 
-        return new Promise<never>(noop);
+        return this._isModalFlow(region)
+            ? Promise.reject(new PaymentMethodCancelledError('Shopper needs to login to Amazonpay to continue'))
```
leeBigCommerce, line 97: Why do we want reject in this scenario?

If I'm reading it right, we've just opened the modal for the customer and they are probably still in the process of logging in

leeBigCommerce:
Thanks for the explanation, @deov31 . Do Amazon recommend having a loading state our side? All the examples I saw in their documentation avoided having any overlays, spinners etc.

src/shipping/strategies/amazon-pay-v2/amazon-pay-v2-shipping-strategy.ts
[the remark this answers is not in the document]
deov31: This was defined since the creation of the amazonpayv2 strategy: https://github.com/bigcommerce/checkout-sdk-js/pull/866#discussion_r430956512

We need to clear the original button to attach the new behavior we want when a shopper logins to amazonpay, which is getting the behavior of the amazonpay button.

src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-strategy.ts
deov31: So, rejecting the promise is the way to not trigger the action for the spinner, rejecting will not cause any issue while shopper is in login process. Amazon scripts handles the login, shipping address selection and after that they will hit out backend with all that information, after that moment we will re render UCO with that data. So we are not actually expecting anything at that moment.

deov31:
> Thanks for the explanation, @deov31 . Do Amazon recommend having a loading state our side? All the examples I saw in their documentation avoided having any overlays, spinners etc.

Hey @leeBigCommerce, amazon have suggested not having spinners on our side, but this was after the current integration was implemented. Originally the US script, as the other regions, used to perform a full redirect to their hosted page, they updated the script without a warning and created the bug on our side. We asked for a callback, but unfortunately that is still in progress and not currently an option for us. We manage to fix the bug this way, but asking them in the future a callback or something we can use to manage properly this kind of behavior. Also in talks, we asked them to keep us in the loop if they consider updating the other regions scripts as well so we can handle it in time.

src/shipping/strategies/amazon-pay-v2/amazon-pay-v2-shipping-strategy.ts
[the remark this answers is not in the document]
deov31: Thanks for the suggestion!... looks good.

src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-strategy.ts
leeBigCommerce: I thought we were only using the loading spinner if `this._isModalFlow(region)` returned `false`? (line 123)

In which case, this reject is only happening when `this._isModalFlow(region)` returns `true`, which is when the loading spinner is not active?

src/shipping/strategies/amazon-pay-v2/amazon-pay-v2-shipping-strategy.ts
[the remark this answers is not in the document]
leeBigCommerce: What existing behaviour are we wanting to clear?

The amazon docs don't mention cloning and infer that using `bindChangeAction` is enough to update the button:

https://developer.amazon.com/docs/amazon-pay-checkout/amazon-pay-script.html

Given it predates your change we can leave it, I'm just not sure if/why we need it.

src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-strategy.ts
deov31: @leeBigCommerce I'm not sure if I understand, yes, we are using the loading spinner in regions that are not US, which is the validation of the function. In this return if the modal flow is true (we are using US region) we need to reject the promise to not trigger the loading spinner.

src/shipping/strategies/amazon-pay-v2/amazon-pay-v2-shipping-strategy.ts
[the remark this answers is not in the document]
mauricio-sg: @leeBigCommerce As `cloneNode` does not copy event listeners, I think it's used as a way to wipe out any preexisting `onclick` behavior of the passed element and just have the one bound by the Amazon script.

src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-strategy.ts
bc-mackxu: Does AmazonPay provide a callback method when user cancel the flow? As you cancel the flow before user click cancel button. It will be nice if we could hook on the callback method from AmazonPay and implement the related function.

deov31: Hey @bc-mackxu no, unfortunately they don't. Amazon team is considering adding a callback, but right now they don't offer it.

leeBigCommerce: Sorry @deov31 , I've finally grasped that the spinner is triggered outside of this strategy (`checkoutState.statuses.isPending()`) and not by the code in line 123 🤦 

I am a bit concerned that every US execution will result in a `PaymentMethodCancelledError` even when not cancelled, but actually continued.  Though I'm not sure if this will have any actual negative side-effects? (@bc-mackxu, @animesh1987  ?)

We're 'cancelling' not to _cancel_ the payment execution, but to remove a side effect of the pending execution (the spinner).

It's not perfect, but adding a `focus` event listener to the window and tying the promise rejection to that might help minimise these 'false cancelations' - if we need to. The re-focussing of this window doesn't necessarily mean the popup was closed, but it's the closest I think of, given Amazon's API does us no favours.

animesh1987: I have similar thoughts as @leeBigCommerce, if we do need to reject the promise then we probably need to create a new error type?

deov31: @animesh1987 I was wondering the same about @leeBigCommerce's comment, creating new error type seems like a good way to address concerns regarding `PaymentMethodCancelledError`. You guys would be ok with that? Something like `PaymentMethodModalClosed` would be ok?

animesh1987: @deov31 what I meant earlier was have you given a thought about what Lee mentioned about adding `focus` listener to the window and dealing with rejections when the focus changes? 

Also `PaymentMethodModalClosed` error doesn't seem right since we are actually opening the modal, if I understand this correctly.

ggarcia-bc: hello team @animesh1987 @leeBigCommerce is there any recommendation you could provide us to move forward? 
 or you think should we take this in another direction? from Amazon they suggested us to remove the spinner and we want to provide a solution with the best approach

animesh1987:
@deov31 any updates on this one? Thanks

deov31:
> @deov31 any updates on this one? Thanks

Hey @animesh1987, I've been considering your suggestion about the focus, but I don't think is a viable solution. The way the script creates the window after the button does not allow us to catch it or identify it (considering that it's concatenating a datetime) to add the event to the window object, also it would also bring us another scenarios that we would have to consider is shopper for examples start moving around tabs and windows. Can we consider our current solution to fix this bug?

Screen Shot 2021-09-27 at 20 45 50

leeBigCommerce:
@deov31 

> The way the script creates the window after the button does not allow us to catch it or identify it

I was suggesting putting the focus listener on our window (the parent), not the pop up window. Crudely, something like this:

```
    async execute(orderRequest: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {

        // etc etc...

        if (this._isModalFlow(region)) {
            return new Promise((_, reject) => {
                window.addEventListener(
                    'focus',
                    () => reject(new PaymentMethodCancelledError('Shopper needs to login to Amazonpay to continue')),
                    { once: true }
                );
            });
        }

        return new Promise<never>(noop);
    }
```

The idea being:

Given the modal is opened
When the user refocusses the main window (our best indication that they closed/abandoned the modal)
Then we reject out of the `execute` method with the `PaymentMethodCancelledError`

It's far from perfect, but I'm hoping it will cause fewer false Cancellations than otherwise _always_ rejecting in the modal flow.

deov31:
@leeBigCommerce thank you for the suggestion, I tested and console looks cleaner. Hopefully this looks better now.

src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-strategy.ts
```diff
@@ -93,6 +93,16 @@ export default class AmazonPayV2PaymentStrategy implements PaymentStrategy {
 
         this._walletButton.click();
 
+        if (this._isModalFlow(region)) {
+            return new Promise((_, reject) => {
+                window.addEventListener(
```
leeBigCommerce, line 98: Given it's not explicit form the code itself why we're doing the focus event,
I'd suggest adding a comment for future developers. Something like:

```
// Focus of parent window used to try and detect the user cancelling the Amazon log in modal
// Should be refactored if/when Amazon add a modal close hook to their SDK 
```

deov31: Agree.

src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-strategy.ts
```diff
@@ -40,11 +40,11 @@ export default class AmazonPayV2PaymentStrategy implements PaymentStrategy {
 
         await this._amazonPayV2PaymentProcessor.initialize(paymentMethod);
 
-        const { paymentToken } = paymentMethod.initializationData;
+        const { paymentToken, region } = paymentMethod.initializationData;
         const buttonId = amazonpay.editButtonId;
 
         if (paymentToken && buttonId) {
-            this._bindEditButton(buttonId, paymentToken, 'changePayment');
+            this._bindEditButton(buttonId, paymentToken, 'changePayment', this._isModalFlow(region));
```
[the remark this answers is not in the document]
deov31: We call function `_isModalFlow` within initialize and the other time is within execute. Constant will not share same context.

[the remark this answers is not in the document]
leeBigCommerce: My mistake, was tricked by GitHub's collapsed area 👍

src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-strategy.ts
```diff
@@ -93,6 +93,16 @@ export default class AmazonPayV2PaymentStrategy implements PaymentStrategy {
 
         this._walletButton.click();
 
+        if (this._isModalFlow(region)) {
+            return new Promise((_, reject) => {
+                window.addEventListener(
+                    'focus',
+                    () => reject(new PaymentMethodCancelledError('Shopper needs to login to Amazonpay to continue')),
+                    { once: true }
```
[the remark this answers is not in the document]
deov31: Thanks, i've checking browser supported versions and I don't think this could be a problem. Most browsers support it since few years ago. The only browser that not support this is IE9. You think a validation would be required for this scenario?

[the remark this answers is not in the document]
leeBigCommerce: IE11 doesn't support it either: https://caniuse.com/once-event-listener

and we still support IE11: https://github.com/bigcommerce/checkout-sdk-js#browser-support

we can play it safe by using a named function and removing it on invocation instead:

```
        if (this._isModalFlow(region)) {
            return new Promise((_, reject) => {
                const onFocus = () => {
                    window.removeEventListener('focus', onFocus);
                    reject(new PaymentMethodCancelledError('Shopper needs to login to Amazonpay to continue'));
                };

                window.addEventListener( 'focus', onFocus);
            });
        }
```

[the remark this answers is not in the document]
deov31: Made the change, thanks. @leeBigCommerce

leeBigCommerce (approved):
👍

leeBigCommerce:
@animesh1987 code owner review please 😄

deov31:
Thanks both, @leeBigCommerce @animesh1987 🎉 ! This PR is ready to merge now.

PR #1221, opened 2021-08-23
fix(payment): PAYMENTS-7221 fix the PPSDK BigPay Payments request headers

leeBigCommerce:
**N.B.** The PPSDK is behind a work in progress feature toggle

## What?

- For PPSDK requests to the BigPay `/payments` endpoint: 
  - Add `AUTHORIZATION` header populated by the order token
  - Null out `X-XSRF-TOKEN` header (as not to have its value set by default: https://github.com/bigcommerce/request-sender-js/pull/27)

## Why?

- BigPay need the `AUTHORIZATION` JWT value to work with the order
- The endpoint is not designed to use `X-XSRF-TOKEN` value, so its default inclusion is unnecessary and potentially misleading

## Testing / Proof

- Updated tests

### DIFF
--- a/src/payment/strategies/ppsdk/payment-processors/none-payment-processor.ts
+++ b/src/payment/strategies/ppsdk/payment-processors/none-payment-processor.ts
@@ -10,11 +10,19 @@ export class NonePaymentProcessor implements PaymentProcessor {
         private _stepHandler: StepHandler
     ) {}
 
-    process({ paymentMethod, bigpayBaseUrl }: ProcessorSettings) {
+    process({ paymentMethod, bigpayBaseUrl, token }: ProcessorSettings) {
         const paymentMethodId = `${paymentMethod.id}.${paymentMethod.method}`;
         const body = { payment_method_id: paymentMethodId };
+        const options = {
+            credentials: false,
+            body,
+            headers: {
+                authorization: token,
+                'X-XSRF-TOKEN': null,
+            },
+        };
 
-        return this._requestSender.post<PaymentsAPIResponse['body']>(`${bigpayBaseUrl}/payments`, { credentials: false, body })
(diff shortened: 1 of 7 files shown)


IamDavidovich:
@davidchin or @animesh1987 Any chance we can get some eyes on this one? It's kind of blocking everything at the moment.

animesh1987 (approved):

src/payment/strategies/ppsdk/payment-processors/none-payment-processor.ts
```diff
@@ -10,11 +10,19 @@ export class NonePaymentProcessor implements PaymentProcessor {
...
-    process({ paymentMethod, bigpayBaseUrl }: ProcessorSettings) {
+    process({ paymentMethod, bigpayBaseUrl, token }: ProcessorSettings) {
         const paymentMethodId = `${paymentMethod.id}.${paymentMethod.method}`;
         const body = { payment_method_id: paymentMethodId };
+        const options = {
+            credentials: false,
+            body,
+            headers: {
+                authorization: token,
+                'X-XSRF-TOKEN': null,
```
davidchin, line 21: I suggest introducing a new flag to request-sender library to turn off the default behaviour, e.g: `xsrf: false`. The reason is because with the current approach, the client needs to know that this specific header `X-XSRF-TOKEN` needs overriding. It'd be better if the option is declared explicitly as a part of the `RequestOption` interface instead.

leeBigCommerce: That would definitely be nicer, I was following the president here: https://github.com/bigcommerce/request-sender-js/pull/27 which established this `null`ing as a way to remove the `X-XSRF-TOKEN`.

Are we happy to move this improvement to a future ticket?

davidchin (approved):
I'm happy for you to do a fast-follow 👍

PR #1222, opened 2021-08-23
fix(payment): INT-4761 Improvements to the Afterpay strategy

orecio-bc:
## What?
Ensure `useStoreCredit` value is correct when Afterpay fails to process a payment (invalid card or similar) and the shopper wants to retry their payment using the correct data.

## Why?
Afterpay is having issues after it fails to process a payment (invalid card or similar) and the shopper wants to retry their payment, store credit was being applied even if the checkbox was not selected, which caused an error in Bigpay because the amounts were not matching.

## Testing / Proof
Before the changes https://drive.google.com/file/d/1y45TcAlyN5xqrJytDrWQ6hO0xSBQYQFe/view?usp=sharing
After the changes https://drive.google.com/file/d/166jI99wasflw8F14Ncx6ExISSRK9GH9F/view?usp=sharing


josesonora:
closing because it is not necessary

PR #1223, opened 2021-08-25
fix(checkout): INT-4798 Fix Moneris Hosted Field Validation

FranciscoOnt:
[INT-4798](https://jira.bigcommerce.com/browse/INT-4798)
## What?
Add a hosted verification fields validation to prevent the Hosted form flow to be executed when the hosted form fails to initialize due to the number of verification fields required being 0 and fallback to the regular flow.

## Why?
When using a vaulted instrument with Moneris thad does not require number and cvv verification, the hosted form factory will fail to initialize (because it requires at least one field to initialize), by applying this fix it will no longer fail to initialize when fields are not required and prevent failure when trying to submit a payment without hosted verification fields.

## Testing / Proof
Before:  [Demo Video](https://drive.google.com/file/d/1Dbar3Pk8jV8LXC3fgfjQjUr_ezYrNbFB/view?usp=sharing)

After: [Demo Video](https://drive.google.com/file/d/1WWDHRQOThUfcjpbni3k_SQcTjcXQinuK/view?usp=sharing)

### DIFF
--- a/src/payment/strategies/moneris/moneris-payment-strategy.ts
+++ b/src/payment/strategies/moneris/moneris-payment-strategy.ts
@@ -3,7 +3,7 @@ import { map } from 'lodash';
 import { isHostedInstrumentLike } from '../../';
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
 import { InvalidArgumentError , MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType } from '../../../common/error/errors';
-import { HostedForm, HostedFormFactory, HostedFormOptions } from '../../../hosted-form';
+import { HostedFieldValidationOptionsMap, HostedForm, HostedFormFactory, HostedFormOptions } from '../../../hosted-form';
 import { OrderActionCreator, OrderPaymentRequestBody, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
 import { StoreCreditActionCreator } from '../../../store-credit';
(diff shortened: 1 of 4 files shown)


cmedinilla:
♻️  @leeBigCommerce

leeBigCommerce (approved):

leeBigCommerce:
@animesh1987 code owner review please 😄

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -147,7 +147,7 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
     }
 
     private async _executeWithVaulted(payment: OrderPaymentRequestBody): Promise<InternalCheckoutSelectors> {
-        if (this._isHostedPaymentFormEnabled(payment.methodId)) {
+        if (this._isHostedPaymentFormEnabled(payment.methodId) && this._hostedForm) {
```
bc-akshatkaushik, line 150: Is there a scenario where hostedForm will not be available?

FranciscoOnt: @bc-akshatkaushik  What is validated before this step is if the payment is done with an instrument or with credit card, the hosted form Its not validated before.

cmedinilla: @bc-akshatkaushik

animesh1987: In this line https://github.com/bigcommerce/checkout-sdk-js/pull/1223/files#diff-d755dc9987b2df0ced566e63258108d7af6892bc592ed1cccd9f4ddbe119dfc9R167 we check for `this._isHostedPaymentFormEnabled(methodId)` which is what you are checking above as well. Hence checking why the double check?
```suggestion
        if (this._hostedForm) {
```

animesh1987 (approved):

PR #1228, opened 2021-08-27
fix(payment): INT-4802 Moneris Validate response before resolve

cmedinilla:
[INT-4802](https://jira.bigcommerce.com/browse/INT-4802)
## What?
Validate if the response is formatted as expected 

## Why?
The handler can be triggered before the response is send by Moneris 

## Testing / Proof
image

### DIFF
--- a/src/payment/strategies/moneris/moneris-payment-strategy.ts
+++ b/src/payment/strategies/moneris/moneris-payment-strategy.ts
@@ -126,6 +126,10 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
             frameref?.postMessage('tokenize', this._monerisURL(!!testMode));
 
             this._windowEventListener = (response: MessageEvent) => {
+                if (typeof response.data !== 'string') {
+                    return;
+                }
+
                 try {
                     resolve(this._handleMonerisResponse(response));
                 } catch (error) {
--- a/src/payment/strategies/moneris/moneris-payment-strategy.spec.ts
+++ b/src/payment/strategies/moneris/moneris-payment-strategy.spec.ts
@@ -277,6 +277,62 @@ describe('MonerisPaymentStrategy', () => {
             expect(paymentActionCreator.submitPayment).toHaveBeenCalledWith(expectedPayment);
         });
 
+        it('submits payment with vaulted card', async () => {
+            const expectedPayment = {
+                methodId: 'moneris',
+                paymentData: {
+                    instrumentId: '1234',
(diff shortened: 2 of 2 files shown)


FranciscoOnt (approved):
LGTM

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -239,7 +239,11 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
         return iframe;
     }
 
-    private _handleMonerisResponse(response: MessageEvent): string {
+    private _handleMonerisResponse(response: MessageEvent): string | undefined {
+        if (typeof(response.data) === 'object') {
```
leeBigCommerce, line 243: What causes this malformed message?

Do we need to handle the scenario where we get an incorrect message, followed by a correct message?

Currently I think we will reject and fail after the first malformed message.

leeBigCommerce: Should we filter out malformed messages instead?

Currently it looks like a malformed message will lead to us setting `nonce` as `undefined`, not submitting payment and exiting execution.

Instead, should we not filter them out and only resolve/reject based on messages received in the correct format?

cmedinilla: @leeBigCommerce i moved the validation at the moment of the event listener so the promise will not be resolved until we get a proper response

src/payment/strategies/moneris/moneris-payment-strategy.ts
```diff
@@ -126,6 +126,10 @@ export default class MonerisPaymentStrategy implements PaymentStrategy {
             frameref?.postMessage('tokenize', this._monerisURL(!!testMode));
 
             this._windowEventListener = (response: MessageEvent) => {
+                if (typeof(response.data) === 'object') {
```
leeBigCommerce, line 129: it's safer to test that values are of the type that we want, rather than if they are one of the type we do not.

e.g.
`if (typeof(response.data) !== 'string')`

cmedinilla: ♻️ @leeBigCommerce

leeBigCommerce (approved):

cmedinilla:
@bigcommerce/checkout code owners review please 🙏🏼

animesh1987 (approved):

PR #1231, opened 2021-09-01
feat(payment): PAYPAL-954 Update Braintree web SDK version

andriiVitvitskyi1990:
## What?
Update Braintree web SDK version

## Why?
According to task https://jira.bigcommerce.com/browse/PAYPAL-954

## Testing / Proof
...Tested on Dev and Int
Screenshot 2021-09-01 at 16 22 14

### DIFF
--- a/src/payment/strategies/braintree/braintree-script-loader.ts
+++ b/src/payment/strategies/braintree/braintree-script-loader.ts
@@ -5,6 +5,8 @@ import { GooglePayCreator } from '../googlepay';
 
 import { BraintreeClientCreator, BraintreeDataCollectorCreator, BraintreeHostedFieldsCreator, BraintreeHostWindow, BraintreePaypalCheckoutCreator, BraintreePaypalCreator, BraintreeThreeDSecureCreator, BraintreeVisaCheckoutCreator } from './braintree';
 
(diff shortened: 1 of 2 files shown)


animesh1987:
Can we get some screenshots/screencasts as testing proof please?

src/payment/strategies/braintree/braintree-script-loader.ts
```diff
@@ -13,7 +13,7 @@ export default class BraintreeScriptLoader {
 
     loadClient(): Promise<BraintreeClientCreator> {
         return this._scriptLoader
-            .loadScript('//js.braintreegateway.com/web/3.70.0/js/client.min.js')
+            .loadScript('//js.braintreegateway.com/web/3.81.0/js/client.min.js')
```
animesh1987, line 16: Just one think out of curiosity, does it make sense to make the verison  a constant? So we don't need to change in every place?

`const version = '3.81.0';`
and the can use it like
```
`//js.braintreegateway.com/web/${version}/js/client.min.js`
```

andriiVitvitskyi1990: fixed

animesh1987 (approved):

PR #1233, opened 2021-09-02
fix(payment): PAYMENTS-7252 pass methodIds on to PPSDK endpoint without transformation

leeBigCommerce:
**N.B. this work is protected by a WIP feature toggle**

## What?

- Transparently pass on `methodId` to PPSDK, rather than constructing it out of `${paymentMethod.id}.${paymentMethod.method}`

## Why?

- The BE is now sending the id value in the `this.that` format, so the client no longer needs to assemble it

## Testing / Proof

- Passing unit tests
- Manual test of WIP PPSDK functionality 

### DIFF
--- a/src/payment/strategies/ppsdk/ppsdk-payment-processor.ts
+++ b/src/payment/strategies/ppsdk/ppsdk-payment-processor.ts
@@ -1,9 +1,8 @@
 import { OrderPaymentRequestBody } from '../../../order';
-import { PPSDKPaymentMethod } from '../../ppsdk-payment-method';
 
 export interface ProcessorSettings {
     token: string;
-    paymentMethod: PPSDKPaymentMethod;
+    methodId: string;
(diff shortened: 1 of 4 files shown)


src/payment/strategies/ppsdk/ppsdk-payment-processor.ts
```diff
@@ -1,9 +1,8 @@
 import { OrderPaymentRequestBody } from '../../../order';
-import { PPSDKPaymentMethod } from '../../ppsdk-payment-method';
 
 export interface ProcessorSettings {
     token: string;
-    paymentMethod: PPSDKPaymentMethod;
+    methodId: string;
```
animesh1987, line 5: Do we need to check for any backwards compatibility?

animesh1987 (approved):

bc-chenli (approved):

PR #1235, opened 2021-09-02
feat(payment): BOLT-24 added account creation flag for Bolt Embedded One Click execution

Sireks:
## What?
Added account creation flag to Bolt Embedded One Click execution payment payload.

## Why?
Because of the task: https://jira.bigcommerce.com/browse/BOLT-24

## Testing / Proof
Unit tests
Manual test

Here is a screenshot of the checkbox:

## Sibling pull-request
checkout-js: https://github.com/bigcommerce/checkout-js/pull/690

@bigcommerce/checkout

### DIFF
--- a/src/order/order-request-body.ts
+++ b/src/order/order-request-body.ts
@@ -1,4 +1,4 @@
-import { CreditCardInstrument, HostedCreditCardInstrument, HostedInstrument, HostedVaultedInstrument, NonceInstrument, VaultedInstrument, WithCheckoutcomiDealInstrument, WithCheckoutcomFawryInstrument, WithCheckoutcomSEPAInstrument, WithDocumentInstrument, WithMollieIssuerInstrument } from '../payment';
+import { CreditCardInstrument, HostedCreditCardInstrument, HostedInstrument, HostedVaultedInstrument, NonceInstrument, VaultedInstrument, WithAccountCreation, WithCheckoutcomiDealInstrument, WithCheckoutcomFawryInstrument, WithCheckoutcomSEPAInstrument, WithDocumentInstrument, WithMollieIssuerInstrument } from '../payment';
 
 /**
  * An object that contains the information required for submitting an order.
@@ -39,5 +39,5 @@ export interface OrderPaymentRequestBody {
      * An object that contains the details of a credit card, vaulted payment
      * instrument or nonce instrument.
      */
-    paymentData?: CreditCardInstrument | HostedInstrument | HostedCreditCardInstrument | HostedVaultedInstrument | NonceInstrument | VaultedInstrument | CreditCardInstrument & WithDocumentInstrument | CreditCardInstrument & WithCheckoutcomFawryInstrument | CreditCardInstrument & WithCheckoutcomSEPAInstrument | CreditCardInstrument & WithCheckoutcomiDealInstrument | HostedInstrument & WithMollieIssuerInstrument;
+    paymentData?: CreditCardInstrument | HostedInstrument | HostedCreditCardInstrument | HostedVaultedInstrument | NonceInstrument | VaultedInstrument | CreditCardInstrument & WithDocumentInstrument | CreditCardInstrument & WithCheckoutcomFawryInstrument | CreditCardInstrument & WithCheckoutcomSEPAInstrument | CreditCardInstrument & WithCheckoutcomiDealInstrument | HostedInstrument & WithMollieIssuerInstrument | WithAccountCreation;
 }
--- a/src/payment/payment.ts
+++ b/src/payment/payment.ts
@@ -24,11 +24,16 @@ export type PaymentInstrument = (
     ThreeDSVaultedInstrument |
     VaultedInstrument |
     VaultedInstrument & WithHostedFormNonce |
-    VaultedInstrumentWithNonceVerification
+    VaultedInstrumentWithNonceVerification |
+    WithAccountCreation
 );
 
 export interface PaymentInstrumentMeta {
     deviceSessionId?: string;
+}
+
+export interface WithAccountCreation {
+    shouldCreateAccount?: boolean;
 }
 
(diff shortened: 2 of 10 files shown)


src/payment/strategies/bolt/bolt-payment-strategy.ts
```diff
@@ -183,6 +193,8 @@ export default class BoltPaymentStrategy implements PaymentStrategy {
 
         await this._store.dispatch(this._orderActionCreator.submitOrder(order, options));
 
+        const { shouldCreateAccount } = payment.paymentData as WithAccountCreation;
```
animesh1987, line 196: Can we not use `as` please and use type assertion or guard?

src/payment/strategies/bolt/bolt-payment-strategy.ts
```diff
@@ -267,6 +279,16 @@ export default class BoltPaymentStrategy implements PaymentStrategy {
         return boltEmbedded;
     }
 
+    private async _hasBoltAccount() {
```
animesh1987, line 282: This method as `async` in declaration, but no `await` in the function body?

src/order/order-request-body.ts
```diff
@@ -39,5 +39,5 @@ export interface OrderPaymentRequestBody {
      * An object that contains the details of a credit card, vaulted payment
      * instrument or nonce instrument.
      */
-    paymentData?: CreditCardInstrument | HostedInstrument | HostedCreditCardInstrument | HostedVaultedInstrument | NonceInstrument | VaultedInstrument | CreditCardInstrument & WithDocumentInstrument | CreditCardInstrument & WithCheckoutcomFawryInstrument | CreditCardInstrument & WithCheckoutcomSEPAInstrument | CreditCardInstrument & WithCheckoutcomiDealInstrument | HostedInstrument & WithMollieIssuerInstrument;
+    paymentData?: CreditCardInstrument | HostedInstrument | HostedCreditCardInstrument | HostedVaultedInstrument | NonceInstrument | VaultedInstrument | CreditCardInstrument & WithDocumentInstrument | CreditCardInstrument & WithCheckoutcomFawryInstrument | CreditCardInstrument & WithCheckoutcomSEPAInstrument | CreditCardInstrument & WithCheckoutcomiDealInstrument | HostedInstrument & WithMollieIssuerInstrument | HostedInstrument & WithAccountCreation;
```
animesh1987, line 42: Do we need to add `hostedInstrument` here?

Sireks: I don't think actually. But is it will be ok to live `WithAccountCreation` alone here?

animesh1987: Yes, I think so it's just a type being mapped and returned from checkout-js I assume?

Sireks: I checked this case. We should use `hostedInstrument` to prevent code duplication in `WithAccountCreation` interface. 

If we remove pairing of `hostedInstrument` and `WithAccountCreation` interfaces we will get a bunch of errors after project build.

Sireks: > Yes, I think so it's just a type being mapped and returned from checkout-js I assume?

Yes, you are right

src/order/order-request-body.ts
```diff
@@ -1,4 +1,4 @@
-import { CreditCardInstrument, HostedCreditCardInstrument, HostedInstrument, HostedVaultedInstrument, NonceInstrument, VaultedInstrument, WithCheckoutcomiDealInstrument, WithCheckoutcomFawryInstrument, WithCheckoutcomSEPAInstrument, WithDocumentInstrument, WithMollieIssuerInstrument } from '../payment';
+import { AccountCreationInstrument, CreditCardInstrument, HostedCreditCardInstrument, HostedInstrument, HostedVaultedInstrument, NonceInstrument, VaultedInstrument, WithCheckoutcomiDealInstrument, WithCheckoutcomFawryInstrument, WithCheckoutcomSEPAInstrument, WithDocumentInstrument, WithMollieIssuerInstrument } from '../payment';
```
animesh1987, line 1: Don't think instrument is the correct prefix here, sine instrument means `Instrument refers to a stored payment instrument, like stored credit card or paypal account`.

src/payment/payment.ts
```diff
@@ -31,6 +32,10 @@ export interface PaymentInstrumentMeta {
     deviceSessionId?: string;
 }
 
+export interface AccountCreationInstrument {
+    shouldCreateAccount?: boolean;
```
animesh1987, line 36: Since in the typeguard you check for it not being undefined, does it need to be undefined here?

src/order/order-request-body.ts
Sireks: The previous naming was better?

src/payment/payment.ts
Sireks: I will fix the typeguard

src/payment/strategies/bolt/bolt-payment-strategy.ts
```diff
@@ -267,6 +287,16 @@ export default class BoltPaymentStrategy implements PaymentStrategy {
...
     }
 
+    private _hasBoltAccount() {
+        const state = this._store.getState();
+        const customer = state.customer.getCustomer();
+        const billingAddress = state.billingAddress.getBillingAddress();
+        const email = customer?.email || billingAddress?.email || '';
+        const boltClient = this._getBoltClient();
+
+        return boltClient.hasBoltAccount(email);
```
animesh1987, line 297: Do we need to do some kind of check here, if this call fails?

Sireks: Sure. I think that we need to add some error handling here

animesh1987 (approved):
Nice work and thanks for the patience. LGTM 👍

PR #1242, opened 2021-09-16
feat(checkout): CHECKOUT-5777 Send SDK version as a header for all SDK API requests

bc-peng:
## What?
- Create a static value at `/src/common/http-request/internal-api-headers.ts` to serve as centralised management of SDK  version headers.
- Add it to all API requests.
- All `GET`, `POST`, `PUT`, ~~PATCH~~  or `DELETE`requests through `@bigcommerce/request-sender` and all occurences of `headers` have been checked to see whether they are fit to have SDK verision information in the header.
## Why? [Checkout-5777](https://jira.bigcommerce.com/browse/CHECKOUT-5777)
Now, we only send the checkout SDK version on a specific API request that creates orders. It seems like a good idea to simply send this header all the time, so we're able to track SDK version usage across many different endpoints.
## Testing / Proof
Fullscreen_17_9_21__1_40_pm

### DIFF
--- a/src/common/http-request/internal-api-headers.ts
+++ b/src/common/http-request/internal-api-headers.ts
@@ -1,1 +1,3 @@
 export const INTERNAL_USE_ONLY = 'This API endpoint is for internal use only and may change in the future';
+
+export const SDK_HEADERS = { 'X-Checkout-SDK-Version': LIBRARY_VERSION };
--- a/src/common/http-request/internal-api-headers.ts
+++ b/src/common/http-request/internal-api-headers.ts
@@ -1,3 +1,3 @@
 export const INTERNAL_USE_ONLY = 'This API endpoint is for internal use only and may change in the future';
 
-export const SDK_HEADERS = { 'X-Checkout-SDK-Version': LIBRARY_VERSION };
+export const SDK_VERSION_HEADERS = { 'X-Checkout-SDK-Version': LIBRARY_VERSION };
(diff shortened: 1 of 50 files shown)


src/common/http-request/internal-api-headers.ts
```diff
@@ -1 +1,3 @@
 export const INTERNAL_USE_ONLY = 'This API endpoint is for internal use only and may change in the future';
+
+export const SDK_HEADERS = { 'X-Checkout-SDK-Version': LIBRARY_VERSION };
```
davidchin, line 3:
```suggestion
export const SDK_VERSION_HEADERS = { 'X-Checkout-SDK-Version': LIBRARY_VERSION };
```

src/common/http-request/internal-api-headers.ts
```diff
@@ -1 +1,3 @@
 export const INTERNAL_USE_ONLY = 'This API endpoint is for internal use only and may change in the future';
+
+export const SDK_VERSION_HEADERS = { 'X-Checkout-SDK-Version': LIBRARY_VERSION };
```
davidchin, line 3: Could we move this into a separate file instead? This is because the header applies to public API as well, not just internal API.

bc-peng: Yes, of course. I was assuming this file is for internal use headers😅.

davidchin (approved):
LGTM! 👍

PR #1249, opened 2021-09-24
feat(payment): INT-4891 Add support to remount compliance section for Digital River

clopezh:
## What?
Refactor of clientToken since sellingEntity is needed to render the correct compliance section in the drop in element

## Why?
Per Digital River request: 

> The interim solution proposed to clients is to add our compliance DRJs Javascript element to the BC footer using the Script Manager. As a part of the JS, the selling entity is hardcoded.
> With the current beta clients (Warm Audio and Nvidia) for their immediate solution, we are able to set the selling entity to be DRGlobal Tech. However, once they begin to expand their reach to the EU, they will have additional selling entities to be supported. They will be OUT OF COMPLIANCE as this is a legal requirement for DR is MOR/SOR.
> 
> In order to solve this requirement, there are a couple of things to required in BC.
> 
> The selling entity from the DR checkout response needs to be sent to the front-end
> In the payment step, add javascript to replace/remount the compliance element created by the script manager.

## Testing / Proof
image
image


AlanMunozN (approved):
LGTM 👍

src/payment/strategies/digitalriver/digitalriver-payment-strategy.spec.ts
```diff
@@ -115,17 +115,28 @@ describe('DigitalRiverPaymentStrategy', () => {
...
         let onErrorCallback: (error: OnCancelOrErrorResponse) => {};
         let onSuccessCallback: (data?: OnSuccessResponse) => {};
+        let container: HTMLDivElement;
 
         beforeEach(() => {
             jest.spyOn(store.getState().billingAddress, 'getBillingAddressOrThrow').mockReturnValue(getBillingAddress());
             jest.spyOn(store.getState().customer, 'getCustomer').mockReturnValue(customer);
             jest.spyOn(digitalRiverScriptLoader, 'load').mockReturnValue(Promise.resolve(digitalRiverLoadResponse));
             jest.spyOn(digitalRiverLoadResponse, 'createDropin').mockReturnValue(digitalRiverComponent);
+            jest.spyOn(digitalRiverLoadResponse, 'createElement').mockReturnValue(digitalRiverElement);
```
bc-peng, line 130: Hi there, in `checkout-sdk-js`, it should be a blank line after `jest.spyOn()`([sample](https://github.com/bigcommerce/checkout-sdk-js/blob/49928fe40793871a7c99daa5bb1d29f798d760c5/src/payment/strategies/zip/zip-payment-strategy.spec.ts#L131)). Could we follow this style?

src/payment/strategies/digitalriver/digitalriver-payment-strategy.ts
```diff
@@ -303,4 +305,33 @@ export default class DigitalRiverPaymentStrategy implements PaymentStrategy {
...
+        if (complianceDiv) {
+            complianceDiv.innerHTML = '';
+
+            const complianceElement = this._getDigitalRiverJs().createElement('compliance', complianceOptions);
+            complianceElement.mount('compliance');
+        } else {
+            const drfooter = document.createElement('div');
+            drfooter.setAttribute('id', 'compliance');
+            drfooter.style.cssText = 'min-height: 45px;';
+            drfooter.classList.add('layout');
```
bc-peng, line 330: Could we provide more context for the style here? Why `min-height: 45px`, `layout`?

clopezh: Hi @bc-peng Actually this code was sent by the provider, due to their requirements.

src/payment/strategies/digitalriver/digitalriver-payment-strategy.spec.ts
clopezh: Correct, I'll do it

src/payment/strategies/digitalriver/digitalriver-payment-strategy.ts
```diff
@@ -303,4 +305,33 @@ export default class DigitalRiverPaymentStrategy implements PaymentStrategy {
 
         return this._store.dispatch(this._paymentActionCreator.submitPayment(paymentPayload));
     }
+
+    private mountComplianceSection(sellingEntity: string) {
```
animesh1987, line 309:
```suggestion
    private _mountComplianceSection(sellingEntity: string) {
```

src/payment/strategies/digitalriver/digitalriver-payment-strategy.ts
bc-peng: Understood. Thanks. 🙏🏻

PR #1250, opened 2021-09-27
feat(payment): PAYMENTS-7269 add method attribute to loaded order payload and selectors

leeBigCommerce:
## What?
...

## Why?
...

## Testing / Proof
...


leeBigCommerce:
Approach abandoned in favour of sending PPSDK method IDs in the `this.that` format

PR #1251, opened 2021-09-28
fix(checkout): ADYEN-260 fixed googlepay updates billing info, removed update of customer email

vneutrino:
## What?
Fixed googlepay updates billing info, removed update of customer email.
If customer is doing payment via googlepay cart button, system will pass googlepay email as customer email.
Changes ware made after this issue https://jira.bigcommerce.com/browse/INT-4900 and after revert of the previous PR related to the ADYEN-260

## Why?
Due to the https://jira.bigcommerce.com/browse/ADYEN-260

## Testing / Proof

https://user-images.githubusercontent.com/79574476/135050738-d1ce6282-bec2-4413-bb16-378cfd46bb8c.mov

https://user-images.githubusercontent.com/79574476/135050753-43a4a90e-e812-4a66-8d91-6f3dd54c1ec2.mov

### DIFF
--- a/src/payment/strategies/googlepay/googlepay-payment-processor.ts
+++ b/src/payment/strategies/googlepay/googlepay-payment-processor.ts
@@ -157,7 +157,7 @@ export default class GooglePayPaymentProcessor {
         return this._methodId;
     }
 
-    private _mapGooglePayAddressToBillingAddress(paymentData: GooglePaymentData, id: string): BillingAddressUpdateRequestBody {
+    private _mapGooglePayAddressToBillingAddress(paymentData: GooglePaymentData, id: string, customerEmail?: string): BillingAddressUpdateRequestBody {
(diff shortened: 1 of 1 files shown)


BC-ACherednichenko (commented):
I think it is worthy create a dedicated googlepay-payment-processor for adyen, this processor affects all provders at once, but, not necessarily we want to apply experience that necessarry only for adyen

animesh1987:
Do have that process for other strategies where we implement, gateway specific strategies for a provider?

vneutrino:
@BC-ACherednichenko @animesh1987 I think it’s better to make these changes for all providers that use GooglePay
Because they all have same problem with replacing customer email by GooglePay email

BC-ACherednichenko (approved):
per our discussion, it may be beneficial to apply this behavviour to all providers. if biling info exist we do not usually update after a redirect from 3rd-party

animesh1987 (approved):

PR #1266, opened 2021-10-08
feat(payment): INT-4897 Humm - Add Humm strategy as external payment

roberbnd:
## What? [INT-4897](https://jira.bigcommerce.com/browse/INT-4897)
Adding humm payment in the registry as external payment

## Why?
it's required to support humm

## Testing / Proof
image

### DIFF
--- a/src/payment/strategies/humm/humm-payment-strategy.spec.ts
+++ b/src/payment/strategies/humm/humm-payment-strategy.spec.ts
@@ -1,0 +1,133 @@
+import { createClient as createPaymentClient } from '@bigcommerce/bigpay-client';
+import { createAction, createErrorAction, Action } from '@bigcommerce/data-store';
+import { createFormPoster, FormPoster } from '@bigcommerce/form-poster';
+import { createRequestSender } from '@bigcommerce/request-sender';
+import { createScriptLoader } from '@bigcommerce/script-loader';
+import { merge } from 'lodash';
+import { of, Observable } from 'rxjs';
+
+import { createCheckoutStore, CheckoutRequestSender, CheckoutStore, CheckoutValidator } from '../../../checkout';
+import { getCheckoutStoreState } from '../../../checkout/checkouts.mock';
+import { RequestError } from '../../../common/error/errors';
+import { getResponse } from '../../../common/http-request/responses.mock';
+import { OrderActionCreator, OrderActionType, OrderRequestBody, OrderRequestSender } from '../../../order';
+import { getOrderRequestBody } from '../../../order/internal-orders.mock';
+import { getHumm } from '../../../payment/payment-methods.mock';
+import { createSpamProtection, PaymentHumanVerificationHandler } from '../../../spam-protection';
+import { PaymentArgumentInvalidError } from '../../errors';
+import PaymentActionCreator from '../../payment-action-creator';
+import { PaymentActionType } from '../../payment-actions';
+import PaymentMethod from '../../payment-method';
+import PaymentRequestSender from '../../payment-request-sender';
+import PaymentRequestTransformer from '../../payment-request-transformer';
+import { getErrorPaymentResponseBody } from '../../payments.mock';
+
+import HummPaymentStrategy from './humm-payment-strategy';
+
+describe('HummPaymentStrategy', () => {
+    let checkoutValidator: CheckoutValidator;
+    let checkoutRequestSender: CheckoutRequestSender;
+    let orderActionCreator: OrderActionCreator;
+    let orderRequestSender: OrderRequestSender;
+    let payload: OrderRequestBody;
+    let paymentActionCreator: PaymentActionCreator;
+    let paymentMethod: PaymentMethod;
+    let submitOrderAction: Observable<Action>;
+    let submitPaymentAction: Observable<Action>;
+    let store: CheckoutStore;
+    let strategy: HummPaymentStrategy;
+    let formPoster: FormPoster;
+
+    beforeEach(() => {
+        store = createCheckoutStore(getCheckoutStoreState());
+        orderRequestSender = new OrderRequestSender(createRequestSender());
+        checkoutRequestSender = new CheckoutRequestSender(createRequestSender());
+        checkoutValidator = new CheckoutValidator(checkoutRequestSender);
+        orderActionCreator = new OrderActionCreator(orderRequestSender, checkoutValidator);
+        paymentActionCreator = new PaymentActionCreator(
+            new PaymentRequestSender(createPaymentClient()),
+            orderActionCreator,
+            new PaymentRequestTransformer(),
+            new PaymentHumanVerificationHandler(createSpamProtection(createScriptLoader()))
+        );
+
+        paymentMethod = getHumm();
+        payload = merge({}, getOrderRequestBody(), {
+            payment: {
+                methodId: paymentMethod.id,
+                gatewayId: paymentMethod.gateway,
+            },
+        });
+
+        submitOrderAction = of(createAction(OrderActionType.SubmitOrderRequested));
+        submitPaymentAction = of(createAction(PaymentActionType.SubmitPaymentRequested));
+
+        formPoster = createFormPoster();
+
+        jest.spyOn(store, 'dispatch');
+
+        jest.spyOn(orderActionCreator, 'submitOrder')
+            .mockReturnValue(submitOrderAction);
+
+        jest.spyOn(paymentActionCreator, 'submitPayment')
+            .mockReturnValue(submitPaymentAction);
+
+        jest.spyOn(formPoster, 'postForm')
+            .mockReturnValue(Promise.resolve());
+
+        strategy = new HummPaymentStrategy(
+            store,
+            orderActionCreator,
+            paymentActionCreator,
+            formPoster
+        );
+
+    });
+
+    describe('#execute()', () => {
+        it('throws error when undefined payment is provided', async () => {
+            payload = {payment: undefined};
+
+            await expect(strategy.execute(payload)).rejects.toThrow(PaymentArgumentInvalidError);
+        });
+
+        it('returns checkout state', async () => {
+            const output = await strategy.execute(getOrderRequestBody());
+
+            expect(output).toEqual(store.getState());
+        });
+
+        it('redirect to Humm', async () => {
+            const data = JSON.stringify({data: 'data'});
+
+            const error = new RequestError(getResponse({
+                ...getErrorPaymentResponseBody(),
+                provider_data: data,
+                additional_action_required: {
+                    type: 'offsite_redirect',
+                    data : {
+                        redirect_url: 'https://sandbox-payment.humm.com',
+                    },
+                },
+                status:  'additional_action_required',
+            }));
+
+            jest.spyOn(paymentActionCreator, 'submitPayment')
+                .mockReturnValue(of(createErrorAction(PaymentActionType.SubmitPaymentFailed, error)));
+
+            strategy.execute(payload);
+
+            await new Promise(resolve => process.nextTick(resolve));
+            expect(formPoster.postForm).toHaveBeenCalledWith('https://sandbox-payment.humm.com', {data: 'data'});
+        });
+
+        it('reject payment when error is different to additional_action_required', async () => {
+            const error = new RequestError(getResponse(getErrorPaymentResponseBody()));
+
+            jest.spyOn(paymentActionCreator, 'submitPayment')
+                .mockReturnValue(of(createErrorAction(PaymentActionType.SubmitPaymentFailed, error)));
+
+            await expect(strategy.execute(getOrderRequestBody())).rejects.toThrowError(error);
+        });
+    });
+});
--- a/src/payment/strategies/humm/humm-payment-strategy.ts
+++ b/src/payment/strategies/humm/humm-payment-strategy.ts
@@ -1,0 +1,90 @@
+import { FormPoster } from '@bigcommerce/form-poster';
+
+import { PaymentStrategy } from '..';
+import { PaymentActionCreator } from '../..';
+import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
+import { OrderActionCreator, OrderRequestBody } from '../../../order';
+import { OrderFinalizationNotRequiredError } from '../../../order/errors';
+import { PaymentArgumentInvalidError } from '../../errors';
+import { PaymentRequestOptions } from '../../payment-request-options';
+
+export default class HummPaymentStrategy implements PaymentStrategy {
+    constructor(
+        private _store: CheckoutStore,
+        private _orderActionCreator: OrderActionCreator,
+        private _paymentActionCreator: PaymentActionCreator,
+        private _formPoster: FormPoster
+    ) { }
+
+    async execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        const { payment, ...order } = payload;
+
+        if (!payment?.methodId) {
+            throw new PaymentArgumentInvalidError(['payment.methodId']);
+        }
+
+        await this._store.dispatch(this._orderActionCreator.submitOrder(order, options));
+        try {
+            return await this._store.dispatch(this._paymentActionCreator.submitPayment({methodId: payment.methodId}));
+        } catch (error) {
+            if (this._isOffsiteRedirectResponse(error)) {
+                return this._handleOffsiteRedirectResponse(error);
+            }
+
+            return Promise.reject(error);
+        }
+    }
+
+    finalize(): Promise<InternalCheckoutSelectors> {
+        return Promise.reject(new OrderFinalizationNotRequiredError());
+    }
+
+    initialize(): Promise<InternalCheckoutSelectors> {
+        return Promise.resolve(this._store.getState());
+    }
+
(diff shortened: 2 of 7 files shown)


src/payment/strategies/humm/humm-payment-strategy.ts
```diff
@@ -0,0 +1,54 @@
...
+import { CheckoutStore,  InternalCheckoutSelectors } from '../../../checkout';
+import { RequestError } from '../../../common/error/errors';
+import { OrderActionCreator, OrderRequestBody } from '../../../order';
+import { OrderFinalizationNotRequiredError } from '../../../order/errors';
+import { PaymentArgumentInvalidError } from '../../errors';
+import PaymentActionCreator from '../../payment-action-creator';
+import { PaymentRequestOptions } from '../../payment-request-options';
+import PaymentStrategy from '../payment-strategy';
+
+export default class HummPaymentStrategy implements PaymentStrategy {
```
josesonora, line 13: change this to extend from `ExternalPaymentStrategy` and remove all methods that are not necessary to override

src/payment/strategies/humm/humm-payment-strategy.ts
```diff
@@ -0,0 +1,54 @@
...
+        return this._store.getState();
+    }
+
+    async deinitialize(): Promise<InternalCheckoutSelectors> {
+        return this._store.getState();
+    }
+
+    async execute(payload: OrderRequestBody, options?: PaymentRequestOptions): Promise<InternalCheckoutSelectors> {
+        const { payment, ...order } = payload;
+        await this._store.dispatch(this._orderActionCreator.submitOrder(order, options));
```
josesonora, line 32: create the order after validating all the required things

src/payment/strategies/humm/humm-payment-strategy.ts
```diff
@@ -0,0 +1,54 @@
...
+
+        if (!payment?.methodId) {
+            throw new PaymentArgumentInvalidError(['payment.methodId']);
+        }
+
+        try {
+            return await this._store.dispatch(this._paymentActionCreator.submitPayment({methodId: payment.methodId}));
+        } catch (error) {
+            return new Promise(() => {
+                if (!(error instanceof RequestError) || !some(error.body.errors, {code: 'additional_action_required'}) || !error.body.provider_data) {
```
josesonora, line 42: the `ExternalStrategy` already has a method to validate if the error is because of an additional action: `_isAdditionalActionRequired`

josesonora (changes requested):

roberbnd:
@josesonora  could your review?

src/payment/strategies/humm/humm-payment-strategy.ts
```diff
@@ -0,0 +1,59 @@
...
+import { OrderActionCreator, OrderRequestBody } from '../../../order';
+import { OrderFinalizationNotRequiredError } from '../../../order/errors';
+import { PaymentArgumentInvalidError } from '../../errors';
+import PaymentActionCreator from '../../payment-action-creator';
+import { PaymentRequestOptions } from '../../payment-request-options';
+import { ExternalPaymentStrategy } from '../external';
+
+export default class HummPaymentStrategy extends ExternalPaymentStrategy {
+
+    constructor(
```
josesonora, line 13: you do not need to do this if there is no custom logic in the implementation of the constructor

src/payment/strategies/humm/humm-payment-strategy.ts
```diff
@@ -0,0 +1,59 @@
...
+    ) {
+            super(
+            _store,
+            _orderActionCreator,
+            _paymentActionCreator,
+            _formPoster
+        );
+        }
+
+    async initialize(): Promise<InternalCheckoutSelectors> {
```
josesonora, line 27: this is not required, it's already implemented in the parent class and the logic here didn't change

src/payment/strategies/humm/humm-payment-strategy.ts
```diff
@@ -0,0 +1,59 @@
...
+        return this._store.getState();
+    }
+
+    async deinitialize(): Promise<InternalCheckoutSelectors> {
```
josesonora, line 31: similar here, this is not required, it's already implemented in the parent class and the logic here didn't change

src/payment/strategies/humm/humm-payment-strategy.ts
```diff
@@ -0,0 +1,59 @@
...
+                return Promise.reject(error);
+            }
+
+            return new Promise(() => {
+                this._formPoster.postForm(error.body.additional_action_required.data.redirect_url, JSON.parse(error.body.provider_data));
+            });
+        }
+    }
+
+    async finalize(): Promise<InternalCheckoutSelectors> {
```
josesonora, line 56: this is not required, it's already implemented in the parent class and the logic here didn't change

src/payment/strategies/humm/humm-payment-strategy.ts
```diff
@@ -0,0 +1,59 @@
...
+            throw new PaymentArgumentInvalidError(['payment.methodId']);
+        }
+
+        await this._store.dispatch(this._orderActionCreator.submitOrder(order, options));
+        try {
+            return await this._store.dispatch(this._paymentActionCreator.submitPayment({methodId: payment.methodId}));
+        } catch (error) {
+            if (!error.body.provider_data || this._isAdditionalActionRequired(error)) {
+                return Promise.reject(error);
+            }
```
josesonora, line 48: 🍹 One minor suggestion, not a blocker if you decide not to apply it:

If for some reason we don't receive the `provider_data`, the `additional action` will be thrown as a normal error and it won't have any sense to the shopper.

I'd recommend changing to this code:
```suggestion
            if (!this._isAdditionalActionRequired(error)) {
                return Promise.reject(error);
            }
            
            if (!error.body.provider_data) {
                throw new PaymentArgumentInvalidError(['payment. provider_data']);
            }
```

josesonora (changes requested):

roberbnd:
review please.. @josesonora

src/payment/strategies/humm/humm-payment-strategy.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+        });
+    });
+
+    describe('#execute()', () => {
+        beforeEach(() => {
+            const state = getCheckoutStoreStateWithOrder();
+            const internalSelectors = createInternalCheckoutSelectors(state);
+
+            jest.spyOn(internalSelectors.config, 'getContextConfig').mockReturnValue(undefined);
+        });
```
josesonora, line 106: this is not necessary, you can delete it and your tests will still passing

src/payment/strategies/humm/humm-payment-strategy.spec.ts
```diff
@@ -0,0 +1,165 @@
+import { createClient as createPaymentClient } from '@bigcommerce/bigpay-client';
+import { createAction, createErrorAction, Action } from '@bigcommerce/data-store';
+import { createFormPoster, FormPoster } from '@bigcommerce/form-poster';
+import { createRequestSender } from '@bigcommerce/request-sender';
+import { createScriptLoader } from '@bigcommerce/script-loader';
+import { merge } from 'lodash';
+import { of, Observable } from 'rxjs';
+
+import { createCheckoutStore, createInternalCheckoutSelectors, CheckoutRequestSender, CheckoutStore, CheckoutValidator } from '../../../checkout';
```
josesonora, line 9: after applying all the recommendations to the test cases please be aware of fixing any imports that need to be reorganized

josesonora (changes requested):

josesonora (approved):

src/payment/strategies/external/external-payment-strategy.ts
```diff
@@ -53,7 +53,7 @@ export default class ExternalPaymentStrategy implements PaymentStrategy {
         return Promise.resolve(this._store.getState());
     }
 
-    private _isAdditionalActionRequired(error: RequestError): boolean {
+    protected _isAdditionalActionRequired(error: RequestError): boolean {
```
leeBigCommerce, line 56: It's a bit odd that this method is named `_isAdditionalActionRequired` when it only detects offside redirects (a very specific additional action).

It also validates the shape of the request error, but then just returns a boolean. Meaning the code which calls it misses out on any type information (e.g. all the `Object is of type 'unknown'.` errors in `HummPaymentStrategy`)

This might be better changed into a type guard that turns the response from an `unknown` to a `OffsiteRedirectResponse` entity.

Something like:

```ts
    protected _isOffsiteRedirectResponse(response: unknown): response is OffsiteRedirectResponse {
        if (typeof response !== 'object' || response === null) {
            return false;
        }

        const partialResponse: Partial<OffsiteRedirectResponse> = response;

        if (!partialResponse.body) {
            return false;
        }

        const partialBody: Partial<OffsiteRedirectBody> = partialResponse.body;

        return partialBody.status === 'additional_action_required'
            && !!partialBody.additional_action_required
            && partialBody.additional_action_required.type === 'offsite_redirect';
    }
```

That way code which uses it benefits from the discerned type information (no more TypeScript errors) and we can write our code a bit more intuitively:

```ts
 } catch (error) {
            if (this._isOffsiteRedirectResponse(error)) {
                // do things safely knowing error is a OffsiteRedirectResponse
                // doing redirecting etc
            } esle {
                // handle it being an unknown error
            }
        }
```

src/payment/strategies/humm/humm-payment-strategy.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+                        redirect_url: 'https://sandbox-payment.humm.com',
+                    },
+                },
+                status:  'additional_action_required',
+            }));
+
+            jest.spyOn(paymentActionCreator, 'submitPayment')
+                .mockReturnValue(of(createErrorAction(PaymentActionType.SubmitPaymentFailed, error)));
+
+            jest.spyOn(JSON, 'parse').mockReturnValue({data: 'data'});
```
leeBigCommerce, line 118: Why is this needed?

src/payment/strategies/humm/humm-payment-strategy.ts
```diff
@@ -0,0 +1,32 @@
...
+            if (!this._isAdditionalActionRequired(error)) {
+                return Promise.reject(error);
+            }
+
+            if (!error.body.provider_data) {
+                throw new PaymentArgumentInvalidError(['payment.provider_data']);
+            }
+
+            return new Promise(() => {
+                this._formPoster.postForm(error.body.additional_action_required.data.redirect_url, JSON.parse(error.body.provider_data));
```
leeBigCommerce, line 28: why is `error.body.provider_data` a string that we need to parse as JSON

why not just have an actual JSON response?

roberbnd:
@leeBigCommerce  could you ♻️ ?

src/payment/strategies/humm/humm-payment-strategy.ts
```diff
@@ -0,0 +1,33 @@
...
+            throw new PaymentArgumentInvalidError(['payment.methodId']);
+        }
+
+        await this._store.dispatch(this._orderActionCreator.submitOrder(order, options));
+        try {
+            return await this._store.dispatch(this._paymentActionCreator.submitPayment({methodId: payment.methodId}));
+        } catch (error) {
+            if (isOffsiteRedirectResponse(error)) {
+                if (!error.body.provider_data) {
+                    throw new PaymentArgumentInvalidError(['payment.provider_data']);
```
leeBigCommerce, line 22: Is there a scenario where we could get a redirect response without an `error.body.provider_data`?

It seems odd to raise a `PaymentArgumentInvalidError` given we're in control of the BE that should be providing that specific argument.

src/payment/is-offsite-redirect-response.ts
```diff
@@ -0,0 +1,27 @@
...
+    additional_action_required: { type: string };
+    status: string;
+    provider_data?: string;
+}
+
+interface OffsiteRedirectResponse {
+    body: OffsiteRedirectBody;
+}
+
+export function isOffsiteRedirectResponse(response: unknown): response is OffsiteRedirectResponse {
```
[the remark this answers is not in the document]
roberbnd: yes, it's right

src/payment/strategies/humm/humm-payment-strategy.ts
roberbnd: no, we need it to continue, the values are required https://docs.shophumm.com.au/custom_integration/checkout_api.html

roberbnd:
@leeBigCommerce  could you ♻️  please?

src/payment/is-offsite-redirect-response.ts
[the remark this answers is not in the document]
leeBigCommerce: In which case, what value does `ExternalPaymentStrategy` provide?

A lot of our strategies implement their own `finalize`, `initialize` and `deinitialize` methods which are identical to those in `ExternalPaymentStrategy`. The benefit being that the overall strategy remains isolated with fewer levels of indirection (at the cost of a tiny bit of repeated code).

leeBigCommerce (commented):
See question here: https://github.com/bigcommerce/checkout-sdk-js/pull/1266#discussion_r748915262

src/payment/is-offsite-redirect-response.ts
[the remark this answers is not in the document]
roberbnd: well that is the idea we are trying to reduce repetitive code, humm strategy is the first class with inheritance of ExternalPaymentStrategy.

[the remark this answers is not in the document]
leeBigCommerce: I think we might be prematurely optimising here. 🤷 

Given that the only things inherited are the one-line `finalize`, `initialize` and `deinitialize` methods. 90% of `ExternalPaymentStrategy` is not built upon or even used, but completely replaced by `HummPaymentStrategy`.

The fact that some providers may need `provider_data` and others won't (https://github.com/bigcommerce/checkout-sdk-js/pull/1266#discussion_r747537890) also points to this potentially being the wrong abstraction. Especially if you take into account that the true payloads within `provider_data` might be different between providers (they are currently only superficially the same for being stringified JSON).

I'd recommend removing `ExternalPaymentStrategy` and instead making `HummPaymentStrategy` isolated and clearly tuned to its own specific use case. Something like: https://github.com/leeBigCommerce/checkout-sdk-js/blob/4158bfb3adc42c4830808c9c77536ea070696be1/src/payment/strategies/humm/humm-payment-strategy.ts

If future strategies show a big overlap, we can look at abstracting and sharing code at that point.

roberbnd:
@leeBigCommerce  could you ♻️ ?

leeBigCommerce (commented):
Were you able to look into: https://github.com/bigcommerce/checkout-sdk-js/pull/1266/files#r745194093 ?

The test seems to run fine without it

roberbnd:
@leeBigCommerce  one more time, please ♻️

src/payment/strategies/humm/humm-payment-strategy.ts
```diff
@@ -0,0 +1,90 @@
...
+import { PaymentActionCreator } from '../..';
+import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
+import { OrderActionCreator, OrderRequestBody } from '../../../order';
+import { OrderFinalizationNotRequiredError } from '../../../order/errors';
+import { PaymentArgumentInvalidError } from '../../errors';
+import { PaymentRequestOptions } from '../../payment-request-options';
+
+export default class HummPaymentStrategy implements PaymentStrategy {
+    constructor(
+        protected _store: CheckoutStore,
```
leeBigCommerce, line 13: These can be `private` now we're not expected this class to be extended 

(sorry, mistake in my example code!)

leeBigCommerce (approved):
👍

leeBigCommerce:
@bigcommerce/checkout Code owner review please 👋

src/payment/strategies/humm/humm-payment-strategy.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    initialize(): Promise<InternalCheckoutSelectors> {
+        return Promise.resolve(this._store.getState());
+    }
+
+    deinitialize(): Promise<InternalCheckoutSelectors> {
+        return Promise.resolve(this._store.getState());
+    }
+
+    private handleOffsiteRedirectResponse(response: OffsiteRedirectResponse): Promise<never> {
```
animesh1987, line 50:
```suggestion
    private _handleOffsiteRedirectResponse(response: OffsiteRedirectResponse): Promise<never> {
```

src/payment/strategies/humm/humm-payment-strategy.ts
```diff
@@ -0,0 +1,90 @@
...
+        const url = response.body.additional_action_required.data.redirect_url;
+        const data = response.body.provider_data;
+
+        return new Promise(() => {
+            this._formPoster.postForm(url, JSON.parse(data));
+        });
+    }
+
+    private isOffsiteRedirectResponse(response: unknown): response is OffsiteRedirectResponse {
```
animesh1987, line 59:
```suggestion
    private _isOffsiteRedirectResponse(response: unknown): response is OffsiteRedirectResponse {
```

roberbnd:
@animesh1987  Could you ♻️ ? please

animesh1987 (approved):

PR #1281, opened 2021-11-02
fix(payment): PAYMENTS-7408 suppress repeated thrown errors when resuming a PPSDK payment

leeBigCommerce:
## What?

- When trying to resume a PPSDK payment which has failed or errored: re-throw the original error once, then fallback to throwing only `OrderFinalizationNotRequiredError` if repeated
- The 'repeated' nature is deduced by storing payment IDs which have resulting in errors or failure in local storage (via the `BrowserStorage` class)

## Why?

- To consider the consumer 'notified' with the first error thrown
- In practice (with `checkout-js`) this leads to only one error modal being displayed (rather than the same one over and over with refreshes or repeated attempts to finalise the payment)

## Testing / Proof

- Updated tests

### DIFF
--- a/src/payment/strategies/ppsdk/ppsdk-completed-payments.ts
+++ b/src/payment/strategies/ppsdk/ppsdk-completed-payments.ts
@@ -1,0 +1,30 @@
+import { BrowserStorage } from '../../../common/storage';
+
+type CompletedPayments = string[];
+const COMPLETED_PAYMENTS_KEY = 'completed-payments';
+
+export class PPSDKCompletedPayments {
+    constructor(
+        private _browserStorage: BrowserStorage
+    ) { }
+
+    isCompleted(paymentId: string): boolean {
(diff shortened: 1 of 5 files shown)


src/payment/strategies/ppsdk/ppsdk-completed-payments.ts
```diff
@@ -0,0 +1,30 @@
+import { BrowserStorage } from '../../../common/storage';
+
+type CompletedPayments = string[];
+const COMPLETED_PAYMENTS_KEY = 'completed-payments';
+
+export class PPSDKCompletedPayments {
+    constructor(
+        private _browserStorage: BrowserStorage
+    ) { }
```
Tharaae, line 9:
```suggestion
    ) {}
```

Tharaae (approved):
LGTM

animesh1987 (approved):

PR #1282, opened 2021-11-03
feat(payment): CHECKOUT-6067 Filter applepay based on browser

animesh1987:
## What?
Filter apple pay based on the brower

## Why?
Since applepay only works in safari, therefore filter apple pay from the list of payment methods that are sent to checkout if the browser used is not safari.

## Future changes
- Update the interface of `ApplePaySession` on window object as per what is defined by apple.
- There is more filtering of payment methods that is done in checkout-js, move that logic slowly to SDK as well

## Testing / Proof
- Unit tests
- Screenshot

## Safari
Screen Shot 2021-11-03 at 3 34 40 pm

## Chrome
Screen Shot 2021-11-03 at 3 36 55 pm (2)

@bigcommerce/checkout

### DIFF
--- a/src/global.d.ts
+++ b/src/global.d.ts
@@ -1,3 +1,0 @@
-interface Window {
-    ApplePaySession(): void;
-}
--- a/src/global.d.ts
+++ b/src/global.d.ts
@@ -1,0 +1,3 @@
+interface Window {
+    ApplePaySession(): void;
+}
--- a/src/payment/payment-method-action-creator.ts
+++ b/src/payment/payment-method-action-creator.ts
@@ -4,8 +4,11 @@ import { Observable, Observer } from 'rxjs';
 import { cachableAction, ActionOptions } from '../common/data-store';
 import { RequestOptions } from '../common/http-request';
 
+import { PaymentMethod } from '.';
 import { LoadPaymentMethodsAction, LoadPaymentMethodAction, PaymentMethodActionType } from './payment-method-actions';
 import PaymentMethodRequestSender from './payment-method-request-sender';
+
+const APPLEPAYID = 'applepay';
 
 export default class PaymentMethodActionCreator {
     constructor(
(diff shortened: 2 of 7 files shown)


src/payment/payment-method-action-creator.ts
```diff
@@ -23,7 +26,7 @@ export default class PaymentMethodActionCreator {
                         sessionHash: response.headers['x-session-hash'],
                     };
 
-                    observer.next(createAction(PaymentMethodActionType.LoadPaymentMethodsSucceeded, response.body, meta));
+                    observer.next(createAction(PaymentMethodActionType.LoadPaymentMethodsSucceeded, this._filterApplePay(response.body), meta));
```
bc-peng, line 29: If we were to get more conditions to filter the payment methods, should we add lines like below one at a time
```
observer.next(createAction(PaymentMethodActionType.LoadPaymentMethodsSucceeded, this._filterSOMETHINGELSE(response.body), meta));
observer.next(createAction(PaymentMethodActionType.LoadPaymentMethodsSucceeded, this._filterSOMETHINGELSE(response.body), meta));
```
Or use one function to act as a filter?

animesh1987: This is the question I came up with while discussing the amazon pay response, so I am thinking we can move with this and possibly change the method and implementation based on the conditions when we look at the amazon ticket?

bc-peng (approved):

src/global.d.ts
```diff
@@ -0,0 +1,3 @@
+interface Window {
```
davidchin, line 1: Prefer not to define it in `global.d.ts` because `ApplePaySession` isn't defined for all browsers. You can add a question mark there but I think a better way is to just define an interface like `ApplePayWindow` and a type guard `isApplePayWindow`. You can then use the type guard to check if `window` is a regular `Window` or `ApplePayWindow` in your `filterApplePay` method.

davidchin (approved):

PR #1283, opened 2021-11-04
chore(payment): PAYMENTS-7409 remove unreachable, duplicate request error handling from PPSDK strategy

leeBigCommerce:
**N.B.** best reviewed with `Hide whitespace` enabled (the removal of a test's describe block has changed a lot of indenting)

## What?

- Remove a code branch which attempts to identify and 'handle' request errors from BigPay endpoints within the PPSDK strategy

## Why?

- This code branch will never be hit. It is called within the `then` method of our `RequestSender`'s `post` and `get` promises. Both of which will throw in a request error scenario, meaning our `handler` is not called.
- There is no difference between what the removed code was _trying_ to achieve and what the `RequestSender` throws on its own (which is why this orphan branch wasn't noticed in the first place)
- As a result we still get a generic `Something's gone wrong - An unexpected error has occurred.`

## Testing / Proof

- Passing tests


Tharaae (approved):
LGTM

animesh1987 (approved):

PR #1284, opened 2021-11-05
feat(checkout): CHECKOUT-5388 hide payment methods that cannot work with multiple consignments

bc-peng:
## What?
1. Add `providesShippingAddress: boolean` to interface `PaymentMethod` as it is already existing in API.

2. Pass `store:CheckoutStore` to class  `PaymentMethodActionCreator`, so that we can access consignments after loading payment methods.

3. Hide payment methods that cannot work with multiple consignments. A method with `providesShippingAddress:true` cannot handle multi-address shipping ([CHECKOUT-5388](https://jira.bigcommerce.com/browse/CHECKOUT-5388)).

4. Merge the filter for Apple Pay into one new generic method  `_filterPaymentMethods`.

5. Modify many tests to accommodate changes to the interface and class.

6. Add one new unit test.

## Why? 
We currently have hardcoded filter in `Payment.tsx`  for Amazon and AmazonPay, but instead the logic should come from the value returned by the API ([CHECKOUT-5388](https://jira.bigcommerce.com/browse/CHECKOUT-5388)).

## Testing / Proof
✅  Manual test:
If there are several consignments, AmazonPay will be removed.

https://user-images.githubusercontent.com/88361607/140459411-fa482a5b-25af-4e78-983a-8d193148955c.mov


src/payment/payment-method-action-creator.ts
```diff
@@ -52,12 +54,19 @@ export default class PaymentMethodActionCreator {
...
+    private _filterPaymentMethods(methods: PaymentMethod[]): PaymentMethod[] {
         return methods.filter(method => {
+            // remove ApplePay in a non-Apple environment
             if (method.id === APPLEPAYID && !isApplePayWindow(window)) {
                 return false;
             }
 
+            // remove methods that cannot work with multi-address shipping when multiple consignments detected
+            const consignments = this._store.getState().consignments.getConsignments();
+            if (consignments && consignments.length > 1 && method.providesShippingAddress) {
```
davidchin, line 66: `/api/storefront/payments` is aware of the cart so wouldn't it make more sense if the filtering is done in the backend, similar to how payment methods are currently filtered out if they are not compatible with the billing country and currency of the cart? 🤔 cc @bc-alfreid

davidchin: Ok I just realised this flag `providesShippingAddress` doesn't mean the shipping address of a cart is provided by a payment method. I think it just means that the payment method _can_ provide an address. So now I'm actually confused what we're trying to do here. If the flag really means that a payment _can_ provide an address, it doesn't actually mean that it's not compatible with multi-shipping? cc @rabinshr (sorry I just realised I pinged the wrong person in my last comment)

rabinshr: >I think it just means that the payment method can provide an address

Yes that is my understanding too 👍 

>If the flag really means that a payment can provide an address, it doesn't actually mean that it's not compatible with multi-shipping?

I believe the intention here is: from payments perspective, we are providing information in regards to whether or not a payment method supports capturing of shoppers address. The notion of, if it is compatible with multi-shipping, I guess is more of a checkout concern rather than payments. This is the reason why I don't think the filtering belongs in the payments.

Also to provide more context, this seems to be more specific to wallet payment methods where they take over the checkout and provide shipping related information.

Having said that, I think I have similar questions as to you David. It seems like we are filtering the payment method when cart has multiple consignment, meaning wallet didn't take over the checkout. There might have been some issues with wallet payment methods such as Amazon Pay that could have introduced this change. https://jira.bigcommerce.com/browse/CHECKOUT-5295?focusedCommentId=833948&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-833948

>If a wallet was selected in Cart page or Step 1 of checkout, we shall hide the multi-shipping option in step 2 (shipping section) of the checkout.
If someone chooses multi-shipping in step 2 of the checkout, we should hide Amazon (and other wallets) at are non multi-shipping compatible in step 4 (payments section) of the checkout.

This sounds like the logic we are trying to capture here? I don't have the full context into the original issue with Amazon. Maybe you remember the actual issue we tried to resolve something to do with wallet payments using their own shipping address instead of what is provided in the cart?

davidchin: Thanks @rabinshr. So I followed the chain of tickets yesterday and found out what we try to achieve. Those payment methods with `providesShippingAddress` flag as `true` are not compatible with multi-shipping because they can convert multi-shipping carts into a single-shipping cart. For example, when a shopper gets redirected to AmazonPay, all they see is a single shipping address, and when they are returned back to checkout, all the consignments get collapsed into one consignment with the address provided by Amazon (see the video attached to the ticket, starting from 8:33). Would you still say these payment methods are compatible with multi-shipping?

rabinshr: Thanks @davidchin.
Having had a look at the video and also having tried it locally on another wallet (Google Pay), it looks like this is an `AmazonPay` related feature. I don't think all wallets exhibit this behavior, certainly not GooglePay and ApplePay.

To answer your question around are wallets compatible with mult-shipping, I believe the answer is yes with the exception of `AmazonPay`. _Note:_ I haven't tested master pass specifically. They can collect address when started from cart or start of checkout (in which case, they will only allow single address selection), however if checkout is already started and then I don't believe they affect the payment section. That is, they should still appear in the payment list if multi shipping is selected in the checkout (except AmazonPay)

I guess the question is, are we trying to generalize this behavior across all wallets when this is only the case with AmazonPay?

bc-peng: Thank you for taking part in the discussions, Rabin and DC. 

I believe it would be more logical if the backend presented capable payment methods based on all accessible info. The frontend just needs to choose payment options depending on the browser environment.

animesh1987: Thanks for the feedback @rabinshr and @davidchin.
I think @rabinshr has a good point about if are trying to generalize this behaviour for all wallets, but it is only an issue for Amazon as of now. 
So does that mean that we should be able to filter Amazon as a payment method in the backend? Which makes this flag redundant completely? Or do we think that there will be more providers that will require this behaviour?

davidchin: Thank you so much @rabinshr for looking into this. I had the assumption that its the behaviour for all wallet providers that have the `providesShippingAddress` flag set as true. But you're correct, it is not the case. Only `paypalexpress`, `braintreepaypal`, `masterpass`, `chasepay` and `amazonpay` behave that way. And out of that list, only `masterpass` and `amazonpay` has `providesShippingAddress` set as true. 

> Mainly because this flag is only indicating whether a provider can supply shipping address

Unfortunately, that definition is also not entirely correct because there are other wallet providers that supply a shipping address but don't have the flag as true. So in other words, we have a flag that isn't consistently implemented. 🤦  And what makes the matter even more frustrating is that the flag isn't even in use. I believe we wanted to use it but never got around to do it until now.

> This is really a checkout concern and adding this into payments feels like the wrong side to put this responsibility on.

Yes I agree, I initially thought those methods simply can't be used to pay for multi-shipping cart because of some payment restriction, similar to how they are filtered out based on the billing country of a cart. But it turns out the flag has a different meaning, which is "methods that may erase consignment details if the shopper chooses to log into their wallet account at the end of the checkout flow rather than the start".

So I think the best next course of action is to make the flag consistent so that all providers that erase shipping details have the flag as true. Checkout doesn't have that information because payment is the one making the call to modify the cart directly, so we have to rely on payment side to return that information. Checkout can use the flag to either hide those payment methods if a cart has multi-shipping, or can warn the shopper that if they choose to pay with them their shipping details will be reset.

animesh1987: Think this goes back to product then, since as @davidchin said the flag is inconsistent and actually doesn't serve any purpose for now. So maybe we need to rethink as to what the flag means/does. Maybe we can change the flag to something like `supportMultiShipping` since that's what we are trying to do in its current state. 

And check with @bc-charlesho if we should either: 
- hide those payment methods if a cart has multi-shipping?
- can warn the shopper that if they choose to pay with them their shipping details will be reset.

rabinshr: >"methods that may erase consignment details if the shopper chooses to log into their wallet account at the end of the checkout flow rather than the start"

I see, this definition holds true for amazonpay.

>Only paypalexpress, braintreepaypal, masterpass, chasepay and amazonpay behave that way

I tested with `paypal` and it allows multi shipping without changing the consignment details. However, only shows a single address in Paypal but BC still keep the consignments intact
image
image
 
 Would this be why the flag was not set to true for Paypal?

animesh1987: Thanks @rabinshr for testing for paypal, think this just goes to prove that this flag was meant as a switch for providers that support multi shipping vs that don't in my opinion. 
Hence I still believe we need to rethink what the purpose of this flag is and shall we rename to something more suitable.

davidchin: Thanks @rabinshr for actually testing it as I only looked at the code. You're right, when PayPal is selected at the last step of checkout, consignment details don't get overridden. It only does it if you go back to the cart page and select PayPal there (as the smart button uses a different code path). Or if you try to check out using LCO. Anyway, unfortunately, it's quite messy at the moment.

> Maybe we can change the flag to something like supportMultiShipping since that's what we are trying to do in its current state.

@animesh1987 Based on [previous discussions](https://github.com/bigcommerce/bigcommerce/pull/38086#issuecomment-725175845), I think we concluded that payments will inform checkout that certain payment methods may provide an address through `providesShippingAddress`. We decided not to go with `supportMultiShipping` because it's rather misleading. From payments perspective, you can still pay for an order that has multiple addresses. So we ended up with `providesShippingAddress` instead.

The fact that it is inconsistent is because we haven't started using it. We just have to test ~`paypalexpress`~, `braintreepaypal`, `masterpass` and `chasepay` to verify if they behave like `amazonpay`. If they are, set them as true so it is consistent from that point onward and we can start using it. 🙂

bc-peng:
I'm closing this PR based on the outcomes of our discussion ([details](https://jira.bigcommerce.com/browse/CHECKOUT-5388?focusedCommentId=1062756&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1062756)).

PR #1285, opened 2021-11-08
Payments 7391 add ability to cancel abandoned PPSDK redirects

leeBigCommerce:
**N.B.** best read one commit at a time, with `Hide whitespace` enabled:

1. Moving redirection code to own folder
2. Adding redirection state code and use

## What?

- Throw a `PaymentMethodCancelledError` when the checkout page loads with an already pending redirect

## Why?

- This scenario occurs when a user has been redirected away, but then they use the browser back button to cancel out of that payment
- In future work we will add a way for the user to confirm that they wish to cancel, or if they want to be redirected again to try and continue

## Testing / Proof

- New and existing passing tests


src/payment/strategies/ppsdk/step-handler/continue-handler/redirect/RedirectionState.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export class RedirectionState {
+    private _isRedirecting: boolean;
+
+    constructor() {
+        this._isRedirecting = this.getCurrentUrl().searchParams.has(PENDING_REDIRECT_PARAM);
+    }
+
+    private getCurrentUrl(): URL {
+        return new URL(window.location.href);
```
davidchin, line 11: I believe we still support IE11 in storefront, so we can't use the native `URL` provided by browsers. You can however import `parseUrl` from `common/url` and use that instead if you prefer to deal with objects.

src/payment/strategies/ppsdk/step-handler/continue-handler/redirect/RedirectionState.ts
```diff
@@ -0,0 +1,43 @@
...
+        const url = this.getCurrentUrl();
+        url.searchParams.delete(PENDING_REDIRECT_PARAM);
+        this.replaceCurrentUrl(url);
+    }
+
+    get isRedirecting() {
+        return this._isRedirecting;
+    }
+
+    set isRedirecting(value: boolean) {
```
davidchin, line 34: We should probably make this a linter rule. But we follow this rule (https://github.com/airbnb/javascript#accessors--no-getters-setters) in our project. In this case, it's not clear that setting `true/false` on an object actually produces the side effect of changing the browser history. It'd be much clear if methods with intention-revealing names are called instead (e.g.: `setRedirectingParamOnUrl`, `removeRedirectingParamFromUrl`).

leeBigCommerce: Agreed on dropping the custom getter/setter - they are too 'magic': https://github.com/bigcommerce/checkout-sdk-js/pull/1285/commits/6954c7433dd2cd3f2cb15fd72a6e6d5df8eccfb1

I've stuck with `isRedirecting` and `setRedirecting` for the method names for now.

The idea was not to expose the url based implementation detail of the `RedirectionState` class to consumers. E.g. the consumer wants the `RedirectionState` to keep track of redirect activity, it is not concerned or opinionated about _how_ `RedirectionState` keeps that state.

src/payment/strategies/ppsdk/step-handler/continue-handler/redirect/RedirectionState.ts
leeBigCommerce: I've moved to using `parseUrl` from `common/url`: https://github.com/bigcommerce/checkout-sdk-js/pull/1285/commits/90b45fd57c67b7b5edc14081193e13606906917e

Though `parseUrl` doesn't have any real support for parsing search params, which has led to me doing a bit of regex and string manipulation within `RedirectionState` - more than I'd ideally like.

Depending on wider opinion, I'd be open to removing this ad hoc manipulation and moving to either:

- Using polyfill for `URL` and replacing all uses of `parseUrl` from `common/url`
- Swapping `common/url` for a more fully featured, open source package like https://github.com/unshiftio/url-parse

src/payment/strategies/ppsdk/step-handler/continue-handler/redirect/RedirectionState.ts
davidchin: Sure I'm good with that 👍

src/payment/strategies/ppsdk/step-handler/continue-handler/redirect/RedirectionState.ts
davidchin: I'm happy for you to go with either approach (use an OSS so it's more fully-featured). If we go with polyfill, it'd be good to pick one that doesn't modify the global `window`.

src/payment/strategies/ppsdk/step-handler/continue-handler/redirect/RedirectionState.ts
```diff
@@ -25,10 +24,10 @@ export class RedirectionState {
     }
 
     private currentUrlHasRedirectingParam(): boolean {
-        return paramRegex.test(this.getCurrentUrl().search);
+        return PENDING_REDIRECT_PARAM in this.getCurrentUrl().query;
```
leeBigCommerce, line 27: The result of `this.getCurrentUrl().query` is an object created via the `Object.create(null)` method, so does not have `hasOwnProperty` or any other prototypal methods - hence the `in`

src/payment/strategies/ppsdk/step-handler/continue-handler/redirect/RedirectionState.ts
leeBigCommerce: Polyfilling the `URL` was a non starter: `core-js` provides only a global altering version, `core-js-pure` provides an isolated version, doesn't provide types. All other deviate too far from the official spec to be worth while.

Instead we've gone for `query-string`, which just seeks to help manage url queries. It's not as full featured as `URL` but does move us to manipulating objects and not manipulating strings with regex and ad hoc concatenations: https://github.com/bigcommerce/checkout-sdk-js/pull/1285/commits/361e9211c3570a6d989da22f12d41f11763f7767

Which is then used in `RedirectionState`: https://github.com/bigcommerce/checkout-sdk-js/pull/1285/commits/60f7403fb21fe4efc91cdf88be1192d2bce90355

davidchin (approved):
LGTM! 👍

PR #1286, opened 2021-11-08
feat(payment): INT-5000 Throw error when payment method isn't available

clopezh:
[INT-5000](https://jira.bigcommerce.com/browse/INT-5000)
## What?
I added a validation when the payment method isn't available to explain the shopper what is happening, instead of do nothing

## Why?
Digital River requested it because the shopper doesn't know what happened if the drop in element isn't rendered
image

## Testing / Proof
image
image

## Dependency Of
https://github.com/bigcommerce/checkout-js/pull/748

### DIFF
--- a/src/payment/strategies/digitalriver/digitalriver-error.spec.ts
+++ b/src/payment/strategies/digitalriver/digitalriver-error.spec.ts
@@ -1,0 +1,11 @@
+import DigitalRiverError from './digitalriver-error';
+
+describe('DigitalRiverError', () => {
+    it('returns error name, type and message', () => {
+        const error = new DigitalRiverError('payment.digitalriver_checkout_error', 'digitalRiverCheckoutError');
+
+        expect(error.name).toEqual('digitalRiverCheckoutError');
+        expect(error.type).toEqual('payment.digitalriver_checkout_error');
+        expect(error.message).toEqual('There was an error while processing your payment. Please try again or contact us.');
+    });
+});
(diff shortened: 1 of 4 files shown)


AlanMunozN (approved):
lgtm

BC-ISilva:
💚 Tested!

QE details:
PR has been validated, issue is no longer available.

Following scenario has been validates:

- Use a Canadian shipping address, but introduce a US post code such as: 55343 **(positive and negative scenarios)**
- Use a US shipping address, but introduce a Canada post code such as: V6C 1B4 **(positive and negative scenarios)**

Please review attached document with validation details.

[INT-5000 - Post code validation does not occur causing checkouts call to error.docx](https://github.com/bigcommerce/checkout-sdk-js/files/7516809/INT-5000.-.Post.code.validation.does.not.occur.causing.checkouts.call.to.error.docx)

animesh1987:
Have we thought about translations for the above error message?

clopezh:
> Have we thought about translations for the above error message?

Hi @animesh1987 Actually in checkout-sdk there isn't a translator service or something similar we can use to translate that error message. No strategy has it.

Let me know if you have any other question,

Thanks

animesh1987:
Hey @clopezh for example we send errors from SDK to checkout. In checkout we read the `error.type` and translate it to a user facing error message. For example messages from https://github.com/bigcommerce/checkout-sdk-js/blob/67b1e0a818f16ba94b729683ea31362907980fad/src/common/error/errors/not-initialized-error.ts#L23 are translated here https://github.com/bigcommerce/checkout-js/pull/739/files#diff-ffda6ecdbe82592fc3308a6aea9fd06682b3484e9d019656c7d9bc2df35d365dR10.

animesh1987 (approved):

PR #1290, opened 2021-11-12
test(payment): INT-4929 increase code coverage for GooglePayProcessor

JoseAlvarezHdz:
## What? [INT-4929](https://jira.bigcommerce.com/browse/INT-4929)

Increase Code Coverage for googlepay-payment-processor.ts

## Why?

To increase code coverage

### Before 

Screen Shot 2021-11-12 at 11 03 42

### After

Screen Shot 2021-11-12 at 11 02 22


JoseAlvarezHdz:
The only case i was not able to test was this one. 
Because when `paymentDataRequest` is assign to `_paymentDataRequest`, it is used in another step as
`paymentDataRequest.allowedPaymentMethods[0].type,` so if we set as an empty objec
t/undefined/null it will fail as `position 0 is not found`

maybe we can refactor that code an instead of using it directly use the `_getPaymentDataRequest` function 

(but i think its worth it to create a ticket for the refactor because of QA testing)

Screen Shot 2021-11-12 at 11 08 48

vmparra (approved):
LGTM 👍🏻

TomA-R:
cc @bigcommerce/checkout

bc-peng (approved):

PR #1292, opened 2021-11-17
chore(payment): PAYMENTS-7402 rename ppsdk payment-processor to sub-strategy

leeBigCommerce:
**N.B.** Best read one commit at a time:

1. Renaming of files and folders
2. Renaming of the entities, functions etc within those files

## What?

- Rename all entities, functions and files that refer to 'Payment Processor's to 'Sub-Strategy's within the PPSDK

## Why?

- Our original 'Payment Processor' term made sense as our first use case (offsite redirects) only had the one method: `process` (similar to a payment strategy's `execute`)
- Our forthcoming 'Card' variant will need a fuller set of lifecycle methods (e.g. `deinitialize` - to be added in a future PR)
- As such, the shape and purpose of these classes will be similar to [PaymentStrategy](https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/payment/strategies/payment-strategy.ts), with the key differences of:
  - being used _within_ an existing strategy (the PPSDK Strategy)
  - using new PPSDK endpoints `/payments`
  - being orchestrated by the BE
- Hence the new term 'Sub-Strategy'

## Testing / Proof

- No functionality has been changed, purely naming
- Passing tests


Tharaae (approved):
LGTM

animesh1987 (approved):

PR #1297, opened 2021-11-25
feat(payment): CHECKOUT-6066 Add apple pay payment strategy

animesh1987:
## What?
Add a payment strategy for apple pay

## Why?
Add a new payment strategy for apple pay, so we can surface it as a payment method in checkout

## Testing / Proof
- Tests
- Screencast

https://user-images.githubusercontent.com/7134802/143333278-29a3b5be-20d4-4852-91bc-4100833e0cbb.mov

## Updated sheet
Screen Shot 2021-11-29 at 5 50 40 pm (2)

@bigcommerce/checkout


src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
```diff
@@ -0,0 +1,166 @@
...
+            paymentMethods: { getPaymentMethodOrThrow },
+        } = await this._store.dispatch(
+            this._paymentMethodActionCreator.loadPaymentMethod(methodId, options)
+        );
+
+        const paymentMethod = getPaymentMethodOrThrow(methodId);
+
+        await this._store.dispatch(this._orderActionCreator.submitOrder({}, options));
+​
+        assertApplePayWindow(window);
```
davidchin, line 62: I think this needs to be called before you create an `ApplePaySession`?

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
```diff
@@ -0,0 +1,166 @@
...
+        const { methodId } = payment;
+        const {
+            paymentMethods: { getPaymentMethodOrThrow },
+        } = await this._store.dispatch(
+            this._paymentMethodActionCreator.loadPaymentMethod(methodId, options)
+        );
+
+        const paymentMethod = getPaymentMethodOrThrow(methodId);
+
+        await this._store.dispatch(this._orderActionCreator.submitOrder({}, options));
```
davidchin, line 60: Need to forward `payload.useStoreCredit` if it's passed by the client.

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
```diff
@@ -0,0 +1,166 @@
...
+        return Promise.reject(new OrderFinalizationNotRequiredError());
+    }
+​
+    deinitialize(): Promise<InternalCheckoutSelectors> {
+        return Promise.resolve(this._store.getState());
+    }
+​
+    private _getBaseRequest(cart: Cart): ApplePayJS.ApplePayPaymentRequest {
+        return {
+            countryCode: 'US',
```
davidchin, line 82: Why is `US` hardcoded?

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
```diff
@@ -0,0 +1,166 @@
...
+            merchantCapabilities: ['supports3DS'],
+            supportedNetworks: [
+                'visa',
+                'masterCard',
+                'amex',
+                'discover',
+            ],
+            lineItems: [],
+            total: {
+                label: 'Total',
```
davidchin, line 93: Is this human-readable text? If so, you'll need to pass the translated string from CheckoutJS into here.

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
```diff
@@ -0,0 +1,166 @@
...
+            countryCode: 'US',
+            currencyCode: cart.currency.code,
+            merchantCapabilities: ['supports3DS'],
+            supportedNetworks: [
+                'visa',
+                'masterCard',
+                'amex',
+                'discover',
+            ],
+            lineItems: [],
```
davidchin, line 91: Do we need to display other line items such as tax, discount etc...?

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
```diff
@@ -0,0 +1,166 @@
...
+
+        const paymentMethod = getPaymentMethodOrThrow(methodId);
+
+        await this._store.dispatch(this._orderActionCreator.submitOrder({}, options));
+​
+        assertApplePayWindow(window);
+​
+        applePaySession.begin();
+​
+        // Applepay will handle the rest of the flow so return a promise that doesn't really resolve
```
davidchin, line 66: Is this comment still relevant? I can see that `_handleApplePayEvents` does call `resolve` in `onpaymentauthorized` handler?

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
```diff
@@ -0,0 +1,166 @@
...
+                        payment_data: token.paymentData,
+                        payment_method: token.paymentMethod,
+                        transaction_id: token.transactionIdentifier,
+                    },
+                },
+            },
+        };
+
+        return this._store.dispatch(this._paymentActionCreator.submitPayment(payment))
+            .then(() => {
```
davidchin, line 157: Prefer `async/await` when possible

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
```diff
@@ -0,0 +1,166 @@
...
+                applePaySession.completePayment(ApplePaySession.STATUS_SUCCESS);
+
+                return resolve(this._store.getState());
+            })
+            .catch(() => {
+                applePaySession.completePayment(ApplePaySession.STATUS_FAILURE);
```
davidchin, line 163: This needs to return a rejection otherwise the outer promise will resolve and trigger the success path.

src/payment/strategies/apple-pay/apple-pay-session-factory.ts
```diff
@@ -0,0 +1,5 @@
+export default class ApplePaySessionFactory {
+    create(request: ApplePayJS.ApplePayPaymentRequest): ApplePaySession {
+        return new ApplePaySession(1, request);
```
davidchin, line 3: Need to investigate if it's okay to upgrade to the latest version, which is believe is 12?

src/payment/payment.ts
```diff
@@ -162,6 +162,14 @@ interface BoltInstrument {
     };
 }
 
+interface AppleInstrument {
+    apple_pay_token: {
+        payment_data: ApplePayJS.ApplePayPaymentToken['paymentData'];
```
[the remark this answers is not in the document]
animesh1987: No reason but I did try that and there is an error about the member not being exported.

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
animesh1987: This is one of the things that we need to figure out how to, since I couldn't find the information on the cart object.

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
animesh1987: Yes it's the human readable text, which is displayed in apple sheet. Do we have an example of the translated text being passed?

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
animesh1987: I had to remove that since we didn't have this information as part of the cart. So feels more accurate to display the total price the shopper needs to pay.

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
animesh1987: Removing it now.

src/payment/strategies/apple-pay/apple-pay-session-factory.ts
animesh1987: I will look into it and see what the pros/cons are. Went with 1 since that's what we support in the cart page as of now.

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
```diff
@@ -0,0 +1,166 @@
...
+import PaymentMethod from '../../payment-method';
+import PaymentMethodActionCreator from '../../payment-method-action-creator';
+import { PaymentRequestOptions } from '../../payment-request-options';
+import PaymentStrategy from '../payment-strategy';
+
+import ApplePaySessionFactory from './apple-pay-session-factory';
+import { assertApplePayWindow } from './is-apple-pay-window';
+
+export const validationEndpoint = 'https://bigpay.service.bcdev/api/public/v1/payments/applepay/validate_merchant';
+const appleValidationUrl = 'https://apple-pay-gateway-cert.apple.com/paymentservices/startSession';
```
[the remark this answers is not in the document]
animesh1987: Realised the the `event` object for validatemerchant callback provides the URL. So updating to use that

src/payment/strategies/apple-pay/apple-pay-session-factory.ts
davidchin: It looks like they have different OS requirements: https://developer.apple.com/documentation/apple_pay_on_the_web/apple_pay_on_the_web_version_history. So v1 would be the most supported but has the least features?

src/payment/strategies/apple-pay/apple-pay-payment-initialize-options.ts
```diff
@@ -0,0 +1,26 @@
...
+ *         subtotalLabel: 'Sub total',
+ *     }
+ * });
+ * ```
+ */
+ export default interface ApplePayPaymentInitializeOptions {
+    /**
+     * Shipping label to be passed to apple sheet.
+     */
+     shippingLabel: string;
```
davidchin, line 20: I think this could be optional as we can provide a default value if not provided.

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
```diff
@@ -0,0 +1,189 @@
...
+
+const validationEndpoint = (bigPayEndpoint: string) => `${bigPayEndpoint}/api/public/v1/payments/applepay/validate_merchant`;
+
+interface ApplePayPromise {
+    resolve(value: InternalCheckoutSelectors | PromiseLike<InternalCheckoutSelectors>): void;
+    reject(reason?: Error): void;
+}
+​
+export default class ApplePayPaymentStrategy implements PaymentStrategy {
+    private _shippingLabel = '';
```
davidchin, line 28:
```suggestion
    private _shippingLabel: string = 'Shipping';
```

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
```diff
@@ -0,0 +1,189 @@
...
+    private _subTotalLabel = '';
```
davidchin, line 29:
```suggestion
    private _subTotalLabel: string = 'Subtotal';
```

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
```diff
@@ -0,0 +1,189 @@
...
+    private _getBaseRequest(cart: Cart, checkout: Checkout, config: StoreConfig): ApplePayJS.ApplePayPaymentRequest {
+        const { storeProfile: { storeCountryCode, storeName } } = config;
+        const lineItems: ApplePayJS.ApplePayLineItem[] = [
+            { label: this._subTotalLabel, amount: `${checkout.subtotal.toFixed(cart.currency.decimalPlaces)}`},
+        ];
+
+        checkout.taxes.forEach(tax =>
+            lineItems.push({ label: tax.name, amount: `${tax.amount}` }));
+
+        lineItems.push({ label: this._shippingLabel, amount: `${checkout.shippingCostTotal}`});
```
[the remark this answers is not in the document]
animesh1987: Good catch

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
animesh1987: The typescript compiler is fairly good at inferring types. So do we still need this?

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
```diff
@@ -0,0 +1,189 @@
...
+                    },
+                },
+            },
+        };
+
+        try {
+            await this._store.dispatch(this._paymentActionCreator.submitPayment(payment));
+            applePaySession.completePayment(ApplePaySession.STATUS_SUCCESS);
+
+            return promise.resolve(this._store.getState());
```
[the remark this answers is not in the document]
animesh1987: This is required here, since we return a promise as part of the `#execute` method, so when the payment fails/succeeds then we need to resolve/reject that promise.

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
```diff
@@ -0,0 +1,189 @@
...
+
+        checkout.taxes.forEach(tax =>
+            lineItems.push({ label: tax.name, amount: `${tax.amount}` }));
+
+        lineItems.push({ label: this._shippingLabel, amount: `${checkout.shippingCostTotal}`});
+
+        return {
+            countryCode: storeCountryCode,
+            currencyCode: cart.currency.code,
+            merchantCapabilities: ['supports3DS'],
```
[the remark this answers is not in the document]
animesh1987: Implemented a way to pass this config from checkout-js as part of initialize.

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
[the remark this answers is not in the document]
davidchin: Ah yes, you're right, I misread `promise.resolve` vs `Promise.resolve`. 👍

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
davidchin: They are different though, `shippingLabel: string` is different to `shippingLabel: 'Shipping'`

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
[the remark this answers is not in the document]
davidchin: I don't think they should be passed from `checkout-js`, as they are payment method configs. So perhaps `PaymentMethod#initializationData` would be a good place to propagate that data?

src/payment/strategies/apple-pay/apple-pay-session-factory.ts
animesh1987: Yes that's the correct assumption. Each new version is tailored for the features that come with that particular macOS. So may have more features but less support. I think we prefer to support more macOs's?

src/payment/strategies/apple-pay/apple-pay-payment-strategy.ts
```diff
@@ -0,0 +1,189 @@
...
+        private _sessionFactory: ApplePaySessionFactory
+    ) { }
+​
+    async initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
+        const { methodId } = options;
+        this._shippingLabel = options?.applepay?.shippingLabel || DefaultLabels.SHIPPING;
+        this._subTotalLabel = options?.applepay?.subtotalLabel || DefaultLabels.SUBTOTAL;
+        await this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(methodId));
+
+        return Promise.resolve(this._store.getState());
```
davidchin, line 50:
```suggestion
        return this._store.getState();
```

src/payment/strategies/apple-pay/apple-pay-payment-initialize-options.ts
```diff
@@ -0,0 +1,26 @@
...
+ export default interface ApplePayPaymentInitializeOptions {
+    /**
+     * Shipping label to be passed to apple sheet.
+     */
+     shippingLabel?: string;
+
+    /**
+     * Sub total label to be passed to apple sheet.
+     */
+     subtotalLabel: string;
```
davidchin, line 25:
```suggestion
     subtotalLabel?: string;
```

davidchin (approved):
Nice one @animesh1987, LGTM! 👍

PR #1307, opened 2021-12-21
feat(payments): CHECKOUT-6070 Add apple pay customer strategy

animesh1987:
## What?
Add apple pay customer strategy

## Why?
Add wallet button for apple pay, so shoppers can checkout using a wallet button in customer step.

## Testing / Proof
- Screencast

https://user-images.githubusercontent.com/7134802/146874158-8d39a584-d572-4f9f-b4e5-77117ee67673.mov

### DIFF
--- a/src/customer/create-customer-strategy-registry.ts
+++ b/src/customer/create-customer-strategy-registry.ts
@@ -226,7 +226,6 @@ export default function createCustomerStrategyRegistry(
         new ApplePayCustomerStrategy(
             store,
             requestSender,
-            remoteCheckoutActionCreator,
             paymentMethodActionCreator,
             new ConsignmentActionCreator(
                 new ConsignmentRequestSender(requestSender),
--- a/src/customer/create-customer-strategy-registry.ts
+++ b/src/customer/create-customer-strategy-registry.ts
@@ -9,6 +9,7 @@ import { FormFieldsActionCreator, FormFieldsRequestSender } from '../form';
 import { PaymentMethodActionCreator, PaymentMethodRequestSender } from '../payment';
 import { AmazonPayScriptLoader } from '../payment/strategies/amazon-pay';
 import { createAmazonPayV2PaymentProcessor } from '../payment/strategies/amazon-pay-v2';
+import { ApplePaySessionFactory } from '../payment/strategies/apple-pay';
 import { BoltScriptLoader } from '../payment/strategies/bolt';
 import { createBraintreeVisaCheckoutPaymentProcessor, BraintreeScriptLoader, BraintreeSDKCreator, VisaCheckoutScriptLoader } from '../payment/strategies/braintree';
 import { ChasePayScriptLoader } from '../payment/strategies/chasepay';
@@ -23,6 +24,7 @@ import CustomerStrategyActionCreator from './customer-strategy-action-creator';
 import { CustomerStrategy } from './strategies';
 import { AmazonPayCustomerStrategy } from './strategies/amazon';
 import { AmazonPayV2CustomerStrategy } from './strategies/amazon-pay-v2';
+import { ApplePayCustomerStrategy } from './strategies/apple-pay';
 import { BoltCustomerStrategy } from './strategies/bolt';
 import { BraintreeVisaCheckoutCustomerStrategy } from './strategies/braintree';
 import { ChasePayCustomerStrategy } from './strategies/chasepay';
@@ -212,6 +214,15 @@ export default function createCustomerStrategyRegistry(
                 new GooglePayStripeInitializer()
             ),
             formPoster
+        )
+    );
+
+    registry.register('applepay', () => 
+        new ApplePayCustomerStrategy(
+            store,
+            remoteCheckoutActionCreator,
+            paymentMethodActionCreator,
+            new ApplePaySessionFactory(),
         )
     );
 
--- a/src/customer/create-customer-strategy-registry.ts
+++ b/src/customer/create-customer-strategy-registry.ts
@@ -1,12 +1,15 @@
 import { createFormPoster } from '@bigcommerce/form-poster';
 import { RequestSender } from '@bigcommerce/request-sender';
-import { getScriptLoader } from '@bigcommerce/script-loader';
+import { createScriptLoader, getScriptLoader } from '@bigcommerce/script-loader';
+import { BillingAddressActionCreator, BillingAddressRequestSender } from '../billing';
 
-import { CheckoutActionCreator, CheckoutRequestSender, CheckoutStore } from '../checkout';
+import { CheckoutActionCreator, CheckoutRequestSender, CheckoutStore, CheckoutValidator } from '../checkout';
 import { Registry } from '../common/registry';
 import { ConfigActionCreator, ConfigRequestSender } from '../config';
 import { FormFieldsActionCreator, FormFieldsRequestSender } from '../form';
-import { PaymentMethodActionCreator, PaymentMethodRequestSender } from '../payment';
+import { OrderRequestSender } from '../order';
+import OrderActionCreator from '../order/order-action-creator';
+import { PaymentActionCreator, PaymentMethodActionCreator, PaymentMethodRequestSender, PaymentRequestSender, PaymentRequestTransformer } from '../payment';
 import { AmazonPayScriptLoader } from '../payment/strategies/amazon-pay';
 import { createAmazonPayV2PaymentProcessor } from '../payment/strategies/amazon-pay-v2';
 import { ApplePaySessionFactory } from '../payment/strategies/apple-pay';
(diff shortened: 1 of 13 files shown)


src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
```diff
@@ -0,0 +1,377 @@
...
+        paymentMethod: PaymentMethod
+    ): ApplePayJS.ApplePayPaymentRequest {
+        const { storeProfile: { storeCountryCode, storeName } } = config;
+        const { currency: { code } } = cart;
+
+        const { initializationData : { merchantCapabilities, supportedNetworks } } = paymentMethod;
+
+        return {
+            requiredBillingContactFields: ['postalAddress'],
+            requiredShippingContactFields: ['email', 'phone', 'postalAddress'],
```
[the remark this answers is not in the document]
davidchin: Also, phone number can be configured to be optional. So I believe we should respect the merchant's preference.

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
```diff
@@ -0,0 +1,377 @@
...
+            countryCode: storeCountryCode,
+            currencyCode: code,
+            merchantCapabilities,
+            supportedNetworks,
+            lineItems: [],
+            total: {
+                label: storeName,
+                amount: '0.01',
```
davidchin, line 155: Why is the amount hardcoded?

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
```diff
@@ -0,0 +1,377 @@
...
+        const { initializationData : { merchantCapabilities, supportedNetworks } } = paymentMethod;
+
+        return {
+            requiredBillingContactFields: ['postalAddress'],
+            requiredShippingContactFields: ['email', 'phone', 'postalAddress'],
+            countryCode: storeCountryCode,
+            currencyCode: code,
+            merchantCapabilities,
+            supportedNetworks,
+            lineItems: [],
```
davidchin, line 152: Have you tested the flow where a shopper has filled in everything, and then goes back to the customer step and click on the ApplePay button. Would they be forced to start from the beginning?

src/customer/create-customer-strategy-registry.ts
```diff
@@ -33,6 +39,7 @@ import { SquareCustomerStrategy } from './strategies/square';
 
 export default function createCustomerStrategyRegistry(
     store: CheckoutStore,
+    paymentClient: any,
```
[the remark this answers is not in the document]
animesh1987: I did try to find the type for this but can't find the type for this, it's been set to any in quite a few places. 

I can type it to use `ReturnType<typeof createPaymentClient>` which by inference resolves to `any` as well.

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
```diff
@@ -0,0 +1,377 @@
...
+        const container = document.getElementById(containerId);
+
+        if (!container) {
+            throw new InvalidArgumentError('Unable to create sign-in button without valid container ID.');
+        }
+
+        const button = document.createElement('input');
+        button.type = 'image';
+        button.src = this._paymentMethod?.logoUrl || '';
+        button.style.height = '40px';
```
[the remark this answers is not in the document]
animesh1987: Well probably need feedback from product, I did think the same but also personally prefer we can get feedback on this.

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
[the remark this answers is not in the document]
animesh1987: Sorry this is the information that we are asking apple to return to us. Require word here means what we want apple to return to us.
Wording from their demo 
```
If you require address or contact details from your customer, request them by passing values in the requiredShippingContactFields or requiredBillingContactFields.
```

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
animesh1987: This is for initial display of sheet, it's pretty much apple being weird to ask for an amount. This information is not rendered. But we can't pass an empty string as well.

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
animesh1987: This is one of the unresolved questions as to what we do when we close the apple pay modal. I did want to check with you that should we reset the checkout at this instance?

src/customer/create-customer-strategy-registry.ts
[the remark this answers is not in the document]
davidchin: Right, I guess it's a JS library so it doesn't have any type. I guess don't worry about it unless you're keen to write the TS definition for it.

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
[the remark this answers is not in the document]
davidchin: Sure I don't mind having the size hardcoded at this stage.

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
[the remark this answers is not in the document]
davidchin: Right I see. In that case `postalAddress` should still be omitted if the cart contains digital items only.

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
davidchin: What you described is a different flow to what I described. But yes, it's another flow we need to consider. I suppose it does make sense to reset the checkout details.

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
davidchin: And you can't just provide the grand total of the cart?

src/customer/create-customer-strategy-registry.ts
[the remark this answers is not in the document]
animesh1987: Probably will decouple this task, and do it as a separate task. Will be good to get an understanding of the that as well.

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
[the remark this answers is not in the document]
animesh1987: Cool will do that, I did think of that but wanted to check if it was a good place to do it here or do it as part of the shippingmethodselected callback.

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
animesh1987: Will probably sync on this once we are back from holidays.

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
animesh1987: Yes, can do that too 😄

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
bc-peng: In my dev store, when I goes back to the customer step by clicking `sign out`. `The shipping price you were quoted is no longer valid. Click OK to see the most up-to-date shipping prices` pops up. Then, everything reset.

https://user-images.githubusercontent.com/88361607/147194541-cbc0cc1f-48fe-4047-9c62-3e7ca5f6ac10.mov

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
```diff
@@ -0,0 +1,411 @@
...
+        };
+
+        applePaySession.onshippingcontactselected = async event =>
+            this._handleShippingContactSelected(applePaySession, config, event);
+
+        applePaySession.onshippingmethodselected = async event =>
+            this._handleShippingMethodSelected(applePaySession, config, event);
+
+        applePaySession.oncancel = async () => {
+            await this._store.dispatch(this._remoteCheckoutActionCreator.signOut(paymentMethod.id));
```
davidchin, line 209: What happens if these async calls fail in this callback? Should you trigger the `onError` handler, like what you're doing with `onvalidatemerchant`?

src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
```diff
@@ -206,9 +206,13 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
...
 
         applePaySession.oncancel = async () => {
-            await this._store.dispatch(this._remoteCheckoutActionCreator.signOut(paymentMethod.id));
+            try {
+                await this._store.dispatch(this._remoteCheckoutActionCreator.signOut(paymentMethod.id));
 
-            return this._store.dispatch(this._checkoutActionCreator.loadCurrentCheckout());
+                return this._store.dispatch(this._checkoutActionCreator.loadCurrentCheckout());
+            } catch (error) {
+                return this._onError(error);
```
davidchin, line 214: This `error` instance would be an internal error. You'll need to map it to an external error class, like `PaymentMethodCancelledError`

davidchin (approved):
Nice work @animesh1987! 👍

PR #1337, opened 2022-02-02
feat(shipping): CHECKOUT-6003 Add method to fetch available shipping options

animesh1987:
## What?
Add implementation to fetch available shipping options.

## Why?
We have an endpoint to fetch available shipping options as part of MLI project. Hence adding methods so partners can leverage that functionality via SDK

## Testing / Proof
- Tests
Screen Shot 2022-02-02 at 5 12 21 pm

@bigcommerce/checkout @bigcommerce/shipping

### DIFF
--- a/src/checkout/checkout-service.ts
+++ b/src/checkout/checkout-service.ts
@@ -444,7 +444,22 @@ export default class CheckoutService {
      * Loads a list of pickup options for a given criteria.
      *
      * ```js
-     * const state = await service.loadPickupOptions(query);
+     * const state = await service.loadPickupOptions({
+     *     search_area: {
+     *         radius: {
+     *             value: 1.4,
+     *             unit: 0
+     *         },
+     *         coordinates: {
+     *             latitude: 1.4,
+     *             longitude: 0
+     *         },
+     *     },
+     *     items: [{
+     *         variantId: 1,
+     *         quantity: 1
+     *     }]
+     * });
      *
      * console.log(state.data.getPickupOptions());
      * ```
--- a/src/checkout/checkout-service.ts
+++ b/src/checkout/checkout-service.ts
@@ -15,7 +15,7 @@ import { CountryActionCreator } from '../geography';
 import { OrderActionCreator, OrderRequestBody } from '../order';
 import { PaymentInitializeOptions, PaymentMethodActionCreator, PaymentRequestOptions, PaymentStrategyActionCreator } from '../payment';
 import { InstrumentActionCreator } from '../payment/instrument';
-import { ConsignmentsRequestBody, ConsignmentActionCreator, ConsignmentAssignmentRequestBody, ConsignmentUpdateRequestBody, ShippingCountryActionCreator, ShippingInitializeOptions, ShippingRequestOptions, ShippingStrategyActionCreator } from '../shipping';
+import { ConsignmentsRequestBody, ConsignmentActionCreator, ConsignmentAssignmentRequestBody, ConsignmentUpdateRequestBody, PickupOptionActionCreator, PickupOptionRequestBody, ShippingCountryActionCreator, ShippingInitializeOptions, ShippingRequestOptions, ShippingStrategyActionCreator } from '../shipping';
 import { SignInEmailActionCreator, SignInEmailRequestBody } from '../signin-email';
 import { SpamProtectionActionCreator, SpamProtectionOptions } from '../spam-protection';
 import { StoreCreditActionCreator } from '../store-credit';
@@ -61,6 +61,7 @@ export default class CheckoutService {
         private _orderActionCreator: OrderActionCreator,
         private _paymentMethodActionCreator: PaymentMethodActionCreator,
         private _paymentStrategyActionCreator: PaymentStrategyActionCreator,
+        private _pickupOptionActionCreator: PickupOptionActionCreator,
         private _shippingCountryActionCreator: ShippingCountryActionCreator,
         private _shippingStrategyActionCreator: ShippingStrategyActionCreator,
         private _signInEmailActionCreator: SignInEmailActionCreator,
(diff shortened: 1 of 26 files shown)


src/checkout/checkout-service.ts
```diff
@@ -438,6 +439,25 @@ export default class CheckoutService {
         return this._dispatch(action, { queueId: 'shippingCountries' });
     }
 
+    /**
+     * This is still in beta.
```
animesh1987, line 443: Not sure if we need to mention this here as part of checkout service?

theromulans: I'd say not? 😆

src/shipping/pickup-option.ts
```diff
@@ -1,3 +1,49 @@
...
+    search_area: SearchArea;
+    items: Item[];
+}
+
+export interface ConsignmentPickupOption {
     pickupMethodId: number;
 }
+
+export interface PickupOptionResponse {
+    results: PickupOption[];
```
bc-luke, line 48: I think each entry here could be called a `Result` and in that result are options for pickup.

animesh1987: Updated the interface name, can you please take another look @bc-luke

src/shipping/pickup-option.ts
```diff
@@ -1,3 +1,49 @@
...
+    item_quantities: Item;
+}
+
+export interface PickupOptionResult {
+    options: Option[];
+}
+
+export interface PickupOptionRequestBody {
+    search_area: SearchArea;
+    items: Item[];
```
davidchin, line 40: I think for the public interface, we should ask the client to pass in cart line item IDs instead, and then we internally map those IDs to product variant IDs.

animesh1987: @bc-luke can you please have a look at this

src/shipping/pickup-option.ts
```diff
@@ -1,3 +1,49 @@
...
+    pickup_method: PickupMethod;
+    item_quantities: Item;
+}
+
+export interface PickupOptionResult {
+    options: Option[];
+}
+
+export interface PickupOptionRequestBody {
+    search_area: SearchArea;
```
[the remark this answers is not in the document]
animesh1987: @bc-luke can you please have a look at this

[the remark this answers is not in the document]
bc-luke: No worries, we will take care of that :+1:

src/shipping/pickup-option.ts
bc-luke: I agree this would be friendlier to the developer. It does make our implementation more complex, adding the cart dependency, however. When we go on to make a version of this API with GraphQL, I expect we wouldn't have that dependency as it would go via storefront service direct to the shipping service.

bc-luke: Another aspect to consider is the response. We are planning to add the quantities of each item available via a given pickup method in the response, e.g.

```
  {
     "pickupOptions": [
       {
         "pickupMethod": {
           "id": 1,
           "locationId": 2,
           "displayName": "Pickup method 2",
           "collectionInstructions": "Instructions",
           "collectionTimeDescription": "Desc"
         },
         "itemQuantities": [
           {
             "variantId": 1,
             "quantity": 2
           }
         ]
       }
    ]
  }
```

I don't think we can map those back to line item IDs so easily. We may have to pass that extra context through to the service.
```

Comment (theromulans) [reply src/shipping/pickup-option.ts:40]: We can have multiple of the same variant in the cart but split across different line items, correct? Perhaps this would help guide toward the sense of using Cart ID. Thoughts?

Comment (bc-luke) [reply src/shipping/pickup-option.ts:40]: That's correct. We will allow (in the second phase) shoppers to pick up items that may share a variant ID from different locations. The current interface would support this, as you can supply multiple objects with the same variant ID.

An advantage of variant ID though is it does allow developers to query this API before an item is added to the cart if they choose to.

Comment (davidchin) [reply src/shipping/pickup-option.ts:40]: Hi @bc-luke, I think there's a bit of misunderstanding. I was asking @animesh1987 to do the mapping on the Checkout SDK level, since it already has a cart loaded in memory and is able to infer the variant ID from it. I wasn't asking to change the API.

I suggested this because clients of Checkout SDK use line item IDs to create consignments. So it'd be more natural to use line item IDs to find available pickup options for the purpose of creating a pickup consignment in checkout.

File: src/shipping/pickup-option-reducer.spec.ts [hunk 5]
```diff
@@ -25,14 +25,20 @@ describe('pickupOptionReducer()', () => {
         });
     });
 
-    it('returns a new state when pickup options are fetched', () => {
+    it.only('returns a new state when pickup options are fetched', () => {
```
davidchin, line 28: Can you remove `only`?

src/shipping/pickup-option-reducer.spec.ts
```diff
@@ -25,14 +25,20 @@ describe('pickupOptionReducer()', () => {
...
+        const query = getQueryForPickupOptions();
         const action: LoadPickupOptionsAction = {
             type: PickupOptionActionType.LoadPickupOptionsSucceeded,
+            meta: query,
             payload: [getPickupOptions()],
         };
 
+        const codedKey = btoa(`${query.consignmentId}-${JSON.stringify(query.searchArea)}`);
```
davidchin, line 36: `JSON.stringify` doesn't sort the object keys. Therefore if the keys are in a different order, it will produce a different hash key. So you either want to sort them first, or just manually flatten the values yourself.

animesh1987: May flatten it manually, since sorting can still be unreliable in this case.

src/shipping/pickup-option-reducer.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+function dataReducer(
+    data: PickupOptionQueryMap | undefined,
+    action: LoadPickupOptionsAction
+): PickupOptionQueryMap | undefined {
+    switch (action.type) {
+        case PickupOptionActionType.LoadPickupOptionsSucceeded:
+            if (action.meta) {
+                const flattenedMeta = objectFlatten(action.meta);
```
davidchin, line 30: Once you flatten the object, you need to sort the keys, otherwise the hash you generate won't be deterministic.

davidchin (approved):
LGTM! 👏

PR #1345, opened 2022-02-09
docs(checkout): CHECKOUT-6260 Update Order Interface Documentation

bc-peng:
## What?
Update `Order` Interface to document the new `consignments` property.

`consignments`'s back-end documentation is [here](https://github.com/bigcommerce/api-specs/blob/master/reference/orders.sf.yml).

 _* leave `pickup` consignments for future work._ 

## Why?
[ORDERS-4432](https://jira.bigcommerce.com/browse/ORDERS-4432) introduced a new `consignments` property to responses from the Storefront [Order](https://github.com/bigcommerce/checkout-sdk-js/blob/55edbb9e8f7ad895517809ed10bef6750a8914c2/docs/interfaces/order.md) API. 

The Order object returned by [getOrder](https://github.com/bigcommerce/checkout-sdk-js/blob/55edbb9e8f7ad895517809ed10bef6750a8914c2/docs/interfaces/checkoutstoreselector.md#getorder) already has a new `consignments` property, but it is not yet documented.

## Testing / Proof
Screen Shot 2022-02-09 at 2 56 25 pm

### DIFF
--- a/src/order/order.ts
+++ b/src/order/order.ts
@@ -11,6 +11,7 @@ export default interface Order {
     billingAddress: BillingAddress;
     cartId: string;
     coupons: Coupon[];
+    consignments: OrderConsignment[];
     currency: Currency;
     customerCanBeCreated: boolean;
     customerId: number;
@@ -59,4 +60,46 @@ export interface GiftCertificateOrderPayment extends OrderPayment {
         code: string;
         remaining: number;
     };
+}
+
+export interface OrderConsignment {
+    shipping: OrderShippingConsignment[];
+}
+
+export interface OrderShippingConsignment {
+    lineItems: Array<{
+        id: number;
+    }>;
+    shippingAddressId: number;
+    firstName: string;
+    lastName: string;
+    company: string;
+    address1: string;
+    address2: string;
+    city: string;
+    stateOrProvince: string;
+    postalCode: string;
+    country: string;
+    countryCode: string;
+    email: string;
+    phone: string;
+    itemsTotal: number;
+    itemsShipped: number;
+    shippingMethod: string;
+    baseCost: number;
+    costExTax: number;
+    costIncTax: number;
+    costTax: number;
+    costTaxClassId: number;
+    baseHandlingCost: number;
+    handlingCostExTax: number;
+    handlingCostIncTax: number;
+    handlingCostTax: number;
+    handlingCostTaxClassId: number;
+    shippingZoneId: number;
+    shippingZoneName: string;
+    customFields: Array<{
+        name: string;
+        value: string | null;
+    }>;
 }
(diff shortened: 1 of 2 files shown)


src/order/order.ts
```diff
@@ -60,3 +61,43 @@ export interface GiftCertificateOrderPayment extends OrderPayment {
...
+    company: string;
+    address1: string;
+    address2: string;
+    city: string;
+    stateOrProvince: string;
+    postalCode: number;
+    country: string;
+    countryCode: string;
+    email: string;
+    phone: number;
```
TomA-R, line 83: Is this actually a number?

src/order/order.ts
```diff
@@ -60,3 +61,43 @@ export interface GiftCertificateOrderPayment extends OrderPayment {
...
+    lineItems: LineItemMap;
+    shippingAddressId: number;
+    firstName: string;
+    lastName: string;
+    company: string;
+    address1: string;
+    address2: string;
+    city: string;
+    stateOrProvince: string;
+    postalCode: number;
```
TomA-R, line 79: Same question, is this actually a number?

bc-peng: Thanks for pointint out.
Just found out a [new doc](https://github.com/bigcommerce/api-specs/blob/master/reference/orders.sf.yml) about the API reponse.
WIll check these types and update this PR soon.

bc-peng: @TomA-R I have updated the code. Will add `pickup` consignments once MLI is live.
Would you mind taking another look at this PR?

src/order/orders.mock.ts
```diff
@@ -13,6 +13,7 @@ export function getOrder(): Order {
         baseAmount: 200,
         billingAddress: getBillingAddress(),
         cartId: 'b20deef40f9699e48671bbc3fef6ca44dc80e3c7',
+        consignments: [],
```
animesh1987, line 16: Can we add a better mock here please?

src/order/order.ts
```diff
@@ -60,3 +61,45 @@ export interface GiftCertificateOrderPayment extends OrderPayment {
...
+    costTax: number;
+    costTaxClassId: number;
+    baseHandlingCost: number;
+    handlingCostExTax: number;
+    handlingCostIncTax: number;
+    handlingCostTax: number;
+    handlingCostTaxClassId: number;
+    shippingZoneId: number;
+    shippingZoneName: string;
+    customFields: Array<{
```
animesh1987, line 101: Do the properties of this key be `{ name: string, value: string }`
Reference https://github.com/bigcommerce/api-specs/blob/master/reference/orders.sf.yml#L774

src/order/order.ts
TomA-R: Thanks Peng, lgtm

src/order/order.ts
bc-peng: Thanks for pointing this out. Could you please have another review?

animesh1987 (approved):
Nice work @bc-peng LGTM 👍

PR #1348, opened 2022-02-15
feat(payment): INT-5034 StripeUPE Add support for EPS

orecio-bc:
## What?
Add Support for EPS payment Method on StripeUpe

## Why?
As a customer, when the currency is EUR and the country is Austria, I want to be able to pay with EPS

## Testing / Proof
image

[https://drive.google.com/file/d/1yM7BMqOZHpBlmnnUQSfm3DPL75y_6aaY/view?usp=sharing](https://drive.google.com/file/d/1yM7BMqOZHpBlmnnUQSfm3DPL75y_6aaY/view?usp=sharing)

## Depends on
INT-5090
INT-5035

Ping @bigcommerce/apex-integrations @bigcommerce/kyiv-payments-team @bigcommerce/payments


mauricio-sg (approved):
LGTM but, can we increase the code coverage eventually? It's below compared to stripev3

animesh1987 (approved):

PR #1353, opened 2022-02-22
feat(payment): INT-5572 Stripe UPE Hide Fields

BC-PabloRivera:
## What? [INT-5572](https://jira.bigcommerce.com/browse/INT-5572)
Stripe Payment Element hides the following fields that are required earlier in the checkout flow: **email**, **address**, **city**, **postal code/zip code**, **country**, **state/province** for all supported Stripe UPE payment methods.
These fields are collected from the billing data and provided to stripe in the confirm step

## Why?
To avoid having my shopper re-enter any information in the payment step that was collected earlier in the checkout flow
So that I can allow them to checkout faster

## Testing / Proof
Before
image
After
image
Card element without Country/postal code fields

Before
image
After (name is not on the list of fields to hide)
image

### DIFF
--- a/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
+++ b/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
@@ -1,6 +1,7 @@
 import { includes, some } from 'lodash';
 
 import { isHostedInstrumentLike } from '../..';
+import { Address } from '../../../address';
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
 import { InvalidArgumentError, MissingDataError, MissingDataErrorType, NotInitializedError, NotInitializedErrorType, RequestError } from '../../../common/error/errors';
 import { OrderActionCreator, OrderRequestBody } from '../../../order';
@@ -13,7 +14,7 @@ import { PaymentInitializeOptions, PaymentRequestOptions } from '../../payment-r
 import PaymentStrategy from '../payment-strategy';
 
 import formatLocale from './format-locale';
-import { StripeElement, StripeElements, StripePaymentMethodType, StripeUPEClient } from './stripe-upe';
+import { AddressOptions, StripeConfirmPaymentData, StripeElement, StripeElements, StripePaymentMethodType, StripeUPEClient } from './stripe-upe';
 import StripeUPEScriptLoader from './stripe-upe-script-loader';
 
 const APM_REDIRECT = [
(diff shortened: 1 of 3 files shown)


src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
```diff
@@ -55,6 +56,17 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
 
         const stripeElement: StripeElement = this._stripeElements.getElement('payment') || this._stripeElements.create('payment',
         {
+            fields: {
+                billingDetails: {
+                    email: 'never',
```
rlopz, line 61: since you're using `never` several times maybe you should create a const for it

rlopz (approved):
LGTM 👍

bc-peng (commented):
LGTM! 👍🏻 
Could we add a before/after comparison picture in the description to show the changes?

src/payment/strategies/stripe-upe/stripe-upe.ts
```diff
@@ -18,6 +18,57 @@ export interface StripeError {
     payment_intent: PaymentIntent;
 }
 
+/**
+ * Object definition for part of the data sent to confirm the PaymentIntent.
+ */
+export interface AddressOptions {
+    city?: string;
```
animesh1987, line 25: Are these all optionals?

BC-PabloRivera: All of these can be optional depending on the method and the settings with which we create the stripe element.

BC-PabloRivera:
@bc-peng I added more images

bc-peng (approved):

src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
```diff
@@ -185,6 +194,47 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
...
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+        }
+
+        return {
+            elements: this._stripeElements,
+            redirect: StripeStringConstants.IF_REQUIRED,
+            confirmParams: {
+                payment_method_data: {
+                    billing_details: {
+                        email,
```
animesh1987, line 230: Here `email` if I understand correctly can be undefined? And address can be similar to `{ }`. Do we want to type check before assigning? Because if email is undefined and address is empty then this could be 
`billing_details: { email: undefined }`

BC-PabloRivera: I added type check

BC-PabloRivera: Although I don’t think these are needed since we always have information for these values in the payment step, and additionally, stripe throws an error if one of these fields is undefined.

animesh1987 (approved):

PR #1357, opened 2022-02-25
feat(payment): INT-5061 Stripe UPE localization

BC-PabloRivera:
## What? [INT-5061](https://jira.bigcommerce.com/browse/INT-5061)
Displays the Stripe UPE methods in the store profile language.
If automatic translation based on shopper’s browser language is enabled in store profile, displays the methods in the shoppers browser language.

## Why?
I want to have the payment integration elements in checkout use the language settings I've defined in my Store Profile Locale settings
So that I can provided a unified translated shopper experience

## Testing / Proof
image
image

image
image

image
(Browser language set to english)
image

## Depends on
https://github.com/bigcommerce/bigcommerce/pull/45192


src/payment/strategies/stripe-upe/format-locale.ts
```diff
@@ -0,0 +1,12 @@
+import { supportedLocales } from './stripe-upe-supported-locales';
+
+export default function formatLocale(localeLanguage: string): string {
+    const [language, country] = localeLanguage.replace(/_/g, '-').toLowerCase().split('-');
+    const countryLocales = supportedLocales[language];
+    if (!countryLocales) {
+        return 'auto';
+    }
+    const formattedLocale = `${language}-${country?.toUpperCase()}`;
```
animesh1987, line 9: In this case if language is `en` and country is undefined, then the formatted locale will be `en-undefined`. Maybe we address this in a better way?

BC-PabloRivera: In that case it would be `en-` because country has the optional chaining operator, and you cant call toUpperCase on undefined, and even then it doesn't matter because in the sentence below we check if this formattedLocale is in the supportedLocales array.
`en-` won't be in the array so it will return `countryLocales[0]` which in this case is `en`

BC-PabloRivera: Sorry, forgot the backtick will transform the undefined into a string, but the second part of my comment still holds

BC-PabloRivera:
♻️ @animesh1987 I added a change to address your comment

src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
```diff
@@ -43,7 +43,7 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
 
         const state = await this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(`${gatewayId}?method=${methodId}`));
         const paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(methodId);
-        const { initializationData: { stripePublishableKey, stripeConnectedAccount } } = paymentMethod;
+        const { initializationData: { stripePublishableKey, stripeConnectedAccount, shopperLanguage } } = paymentMethod;
```
animesh1987, line 46: Do we need to get locale as part of paymentMethod? Or do we need to respect the locale shopper has selected as part of checkout process?

BC-PabloRivera: This variable respects the language the merchant selected. If the merchant selects `Enable automatic translation based on shopper’s browser language` then this variable will have the shopper browser language.

animesh1987 (approved):

PR #1363, opened 2022-03-03
fix(common): CHECKOUT-6355 Remove token strings from test files

animesh1987:
## What?
Remove hardcoded tokens from test/mock files

## Why?
Snyk code scanner has identified that we have some hardcoded secrets in test files, this replaces them with something simpler.

## Testing / Proof
- Tests
Screen Shot 2022-03-03 at 3 08 59 pm

@bigcommerce/checkout


davidchin (approved):
Thanks for removing the dummy token.

PR #1366, opened 2022-03-08
fix(shipping): CHECKOUT-6003 Fix issues found in testing pickup options implementation

animesh1987:
## What?
Fix issues discovered in testing pickup option implementation

## Why?
While testing pickup options API figured out there are few implementation details that are missing. 
- Fix issue how key is created for encoding the option map
- Add getPickupOption to checkout selector that is used by checkout-js

## Testing / Proof
- Tests
- Screenshot

Screen Shot 2022-03-08 at 4 14 39 pm

@bigcommerce/checkout

### DIFF
--- a/src/checkout/checkout-store-selector.ts
+++ b/src/checkout/checkout-store-selector.ts
@@ -495,14 +495,7 @@ export function createCheckoutStoreSelectorFactory(): CheckoutStoreSelectorFacto
 
     const getPickupOptions = createSelector(
         ({ pickupOptions }: InternalCheckoutSelectors) => pickupOptions.getPickupOptions,
-        getPickupOptions => clone((consignmentId: string, searchArea: SearchArea) => {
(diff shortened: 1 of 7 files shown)


src/checkout/checkout-store-selector.ts
```diff
@@ -483,6 +493,18 @@ export function createCheckoutStoreSelectorFactory(): CheckoutStoreSelectorFacto
         getFlashMessages => clone(getFlashMessages)
     );
 
+    const getPickupOptions = createSelector(
+        ({ pickupOptions }: InternalCheckoutSelectors) => pickupOptions.getPickupOptions,
+        getPickupOptions => clone((consignmentId: string, searchArea: SearchArea) => {
```
davidchin, line 498: Just `getPickupOptions  => clone(getPickupOptions)` is sufficient?

davidchin (approved):

PR #1373, opened 2022-03-14
fix(checkout): CHECKOUT-6125 Safari and Firefox focus loss bugfix

bc-peng:
## What?
A bugfix for the widely recognised Safari and Firefox lost focus issue.

- [braintree-web@3.78.2: Fix issue where Desktop Safari required 2 tabs to tab forward](https://github.com/braintree/braintree-web/blob/df38a9b1031504472939a1e7315becd69b7fabed/src/hosted-fields/internal/components/base-input.js#L249) (no longer works in iOS v14.5/6)
- [Firefox, let's focus, please!](https://www.codegram.com/blog/firefox-lets-focus-please/)
- [Inconsistent behavior among browsers when clicking on buttons](https://zellwk.com/blog/inconsistent-button-behavior/)

When a user wants to move the focus to the next input field inside an iframe in Safari or Firefox, pressing 'tab' once will only land the focus onto the iframe's `body`. Therefore, a user needs to press `tab` twice to fill in a payment field.

This PR will force the focus of Firefox and Safari on the first text field inside the iframe, making the payment experience consistent across major browsers.

## Why?

Buggy experience in Firefox now.

https://user-images.githubusercontent.com/88361607/158277012-84fbbbde-4a5e-4a7a-868b-cceaf69d930d.mov

Smooth experience in Chrome now.

https://user-images.githubusercontent.com/88361607/158277580-9cf95a33-387c-41f0-90bc-bce10c68655b.mov

## Testing / Proof

Firefox:

https://user-images.githubusercontent.com/88361607/158277064-4437ad5f-057b-4944-b858-61b508ec78ce.mov

Safari:

https://user-images.githubusercontent.com/88361607/158277103-002e56f1-de28-495b-911d-a0be8885ce5d.mov

### DIFF
--- a/src/hosted-form/iframe-content/hosted-input-initializer.ts
+++ b/src/hosted-form/iframe-content/hosted-input-initializer.ts
@@ -31,9 +31,6 @@ export default class HostedInputInitializer {
         this._resetPageStyles(containerId);
         this._eventListener.listen();
 
-        // fixes the issue on Firefox/Safari where the input doesn't focus properly
-        window.addEventListener('focus', this.firefoxAndSafariBugfix);
-
         return fromEvent<HostedFieldAttachEvent>(
             this._eventListener as EventTargetLike<HostedFieldAttachEvent>,
             HostedFieldEventType.AttachRequested
@@ -55,24 +52,6 @@ export default class HostedInputInitializer {
                 take(1)
             )
             .toPromise();
-    }
-
-    firefoxAndSafariBugfix(): void {
-        if (document.activeElement === document.body) {
-            const textField = document.querySelectorAll('input')[0];
-            if (textField) {
-                const browserName = navigator.userAgent.toLowerCase();
-                if (browserName.indexOf('safari') > -1) {
-                    if (textField.value === '') {
-                        textField.setAttribute('value', ' ');
-                        textField.setSelectionRange(0, 1);
-                        textField.setAttribute('value', '');
-                    }
-                } else {
-                    textField.focus();
-                }
-            }
-        }
     }
 
     private _resetPageStyles(containerId: string) {
--- a/src/hosted-form/iframe-content/hosted-input-initializer.ts
+++ b/src/hosted-form/iframe-content/hosted-input-initializer.ts
@@ -31,10 +31,8 @@ export default class HostedInputInitializer {
         this._resetPageStyles(containerId);
         this._eventListener.listen();
 
-        const browserName = navigator.userAgent.toLowerCase();
-        if (browserName.indexOf('firefox') > -1 || browserName.indexOf('safari') > -1) {
-            window.addEventListener('focus', this.firefoxAndSafariBugfix);
-        }
+        // fixes the issue on Firefox/Safari where the input doesn't focus properly
+        window.addEventListener('focus', this.firefoxAndSafariBugfix);
 
         return fromEvent<HostedFieldAttachEvent>(
             this._eventListener as EventTargetLike<HostedFieldAttachEvent>,
@@ -64,12 +62,14 @@ export default class HostedInputInitializer {
             const textField = document.querySelectorAll('input')[0];
             if (textField) {
                 const browserName = navigator.userAgent.toLowerCase();
-                if (browserName.indexOf('firefox') > -1) {
+                if (browserName.indexOf('safari') > -1) {
+                    if (textField.value === '') {
+                        textField.setAttribute('value', ' ');
+                        textField.setSelectionRange(0, 1);
+                        textField.setAttribute('value', '');
+                    }
+                } else {
                     textField.focus();
-                } else if (browserName.indexOf('safari') > -1) {
-                    setTimeout(() => {
-                        textField.focus();
-                    }, 150);
                 }
             }
         }
--- a/src/hosted-form/iframe-content/hosted-input-initializer.ts
+++ b/src/hosted-form/iframe-content/hosted-input-initializer.ts
@@ -31,6 +31,11 @@ export default class HostedInputInitializer {
         this._resetPageStyles(containerId);
         this._eventListener.listen();
 
+        const browserName = navigator.userAgent.toLowerCase();
+        if (browserName.indexOf('firefox') > -1 || browserName.indexOf('safari') > -1) {
+            window.addEventListener('focus', this.firefoxAndSafariBugfix);
+        }
+
         return fromEvent<HostedFieldAttachEvent>(
             this._eventListener as EventTargetLike<HostedFieldAttachEvent>,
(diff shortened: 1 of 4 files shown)


src/hosted-form/iframe-content/hosted-input-initializer.ts
```diff
@@ -54,6 +59,22 @@ export default class HostedInputInitializer {
...
+        if (document.activeElement === document.body) {
+            const textField = document.querySelectorAll('input')[0];
+            if (textField) {
+                const browserName = navigator.userAgent.toLowerCase();
+                if (browserName.indexOf('firefox') > -1) {
+                    textField.focus();
+                } else if (browserName.indexOf('safari') > -1) {
+                    setTimeout(() => {
+                        textField.focus();
+                    }, 150);
```
bc-peng, line 72: Have tested `0`, `100`. It appears that `150` is an acceptable wait time for Safari to be visually ready.  There is no obvious waiting experience in actual use.

davidchin: Is there any harm in applying `focus` on all browsers so we don't have to do browser detection?

bc-peng: No, I don't think it's harmful as `document.activeElement === document.body` will detect a Firefox and Safari specific behaviour. It's possible that we only need to keep browser detection on line 69.

animesh1987: Interesting finding on the timeout 😄 

Just out of curiosity what does the `document.activeElement` is for chrome 🤔 

Is that by design we do browser detection in line #35 as well?

bc-peng: For Chrome, `document.activeElement` is always a text input field during the payment step. 
In this commit, I was thinking about applying a bugfix to Safari and Firefox first, then triggering different actions, hence I did two browser detections.

davidchin: Another question I have is, why `150` specifically? Does `0` also work?

bc-peng: `0` is not working. Safari still requires two tabs to jump between fields.
I read about `SAFARI_FOCUS_TIMEOUT` concept on the Braintree's hosted field [code](https://github.com/braintree/braintree-web/blob/75c1c2496daa529525d7b6bf852fc0f3264ab7e0/src/hosted-fields/external/hosted-fields.js#L1375). Not exactly the same case, but I think it tells me Safari needs some time to work.

bc-peng: Hi @davidchin @animesh1987 , I have updated the code.

1.  Apply the bugfix on all browsers.
2. Use Braintree's [method](https://github.com/braintree/braintree-web/blob/df38a9b1031504472939a1e7315becd69b7fabed/src/hosted-fields/internal/components/base-input.js#L249) to fix Safari's focus issue. The limitation is that this workaround is not working any more on the latest iOS 14.5.

Here is an updated demo for Safari.

https://user-images.githubusercontent.com/88361607/158500548-ee6e9a61-a4e0-4fbf-b37a-e6a7e42ad6dd.mov

animesh1987 (approved):
Nice work @bc-peng

src/hosted-form/iframe-content/hosted-input-initializer.ts
```diff
@@ -54,6 +57,24 @@ export default class HostedInputInitializer {
             .toPromise();
     }
 
+    firefoxAndSafariBugfix(): void {
+        if (document.activeElement === document.body) {
+            const textField = document.querySelectorAll('input')[0];
```
davidchin, line 62: Could we move the event handler to `HostedInput`? That way we don't have to query for `input`, as it holds a direct reference to the input element. Also, we can hook up to the `attach/detach` methods.

src/hosted-form/iframe-content/hosted-input.ts
```diff
@@ -246,4 +251,19 @@ export default class HostedInput {
             },
         });
     };
+
+    private firefoxAndSafariBugfix(targetField: HTMLInputElement): void {
```
mjhmatt, line 255: 🍹 Would it make sense to rename this function to something more description? `firefoxAndSafariBugfix` doesn't tell me much `forceFocusForToNextInput` or something along those lines?

src/hosted-form/iframe-content/hosted-input.ts
```diff
@@ -85,6 +88,8 @@ export default class HostedInput {
         this._form.removeEventListener('submit', this._handleSubmit);
         this._unloadFonts();
 
+        window.removeEventListener('focus', () => this.firefoxAndSafariBugfix(this._input));
```
davidchin, line 91: I don't think you're removing anything at the moment because you need to pass an actual reference to the callback function that's passed into `addEventListener`. So you'll need to do something like this:

```ts
private firefoxAndSafariBugfix = (targetField: HTMLInputElement) => {
```

And then 
```ts
window.addEventListener('focus', this.firefoxAndSafariBugfix);
```
```ts
window.removeEventListener('focus', this.firefoxAndSafariBugfix);
```

src/hosted-form/iframe-content/hosted-input.ts
bc-peng: Sure, how about `firefoxAndSafariForceFocusToInput`?

davidchin: You don't need the parameter because `this._input` is a property of the class. Also, please prefix the private method with an underscore.

mjhmatt: I think that's better, I am not sure browser specific is ideal as it could potentially impact other browsers as well right? But I think it is more descriptive 👍

bc-peng: @davidchin I did it this way at first. Then I found `this` refered to `window`, hence an error raised saying `this._input` is `undefined`.

davidchin: You won't get that problem if define the `firefoxAndSafariBugfix` function like this https://github.com/bigcommerce/checkout-sdk-js/pull/1373#discussion_r829698810

davidchin (approved):
Thank you @bc-peng!

src/hosted-form/iframe-content/hosted-input.ts
bc-peng: I learned the differences between function expression and declaration within a class. Thank you as alaways. 🙇🏻‍♂️

PR #1390, opened 2022-03-29
CHECKOUT-6264: Add payment integration service

davidchin:
## What?
Add `PaymentIntegrationService` to be used in payment strategies according to [this proposal](https://github.com/bigcommerce/documentation/pull/707).

## Why?
So that payment strategies, which will be extracted into their own packages, can depend on an established API contract implemented by the checkout team.

## Testing / Proof
Unit

@bigcommerce/checkout @bigcommerce/payments @bigcommerce/apex-integrations @bigcommerce/kyiv-payments-team

### DIFF
--- a/src/core/payment-integration/payment-integration-service.ts
+++ b/src/core/payment-integration/payment-integration-service.ts
@@ -1,0 +1,26 @@
+import { BillingAddressRequestBody } from '../../billing';
+import { OrderRequestBody } from '../../order';
+import { Payment } from '../../payment';
+import { ShippingAddressRequestBody } from '../../shipping';
+
+import PaymentIntegrationSelectors from './payment-integration-selectors';
+
+export default interface PaymentIntegrationService {
+    subscribe(subscriber: (state: PaymentIntegrationSelectors) => void, ...filters: Array<(state: PaymentIntegrationSelectors) => any>): () => void;
+
+    getState(): PaymentIntegrationSelectors;
+
+    loadCheckout(): Promise<PaymentIntegrationSelectors>;
+
+    loadPaymentMethod(methodId: string): Promise<PaymentIntegrationSelectors>;
+
+    submitOrder(payload?: OrderRequestBody): Promise<PaymentIntegrationSelectors>;
+
+    submitPayment(payment: Payment): Promise<PaymentIntegrationSelectors>;
+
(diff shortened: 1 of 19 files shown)


src/order/order-request-sender.ts
```diff
@@ -46,7 +46,7 @@ export default class OrderRequestSender {
         });
     }
 
-    submitOrder(body: InternalOrderRequestBody, { headers, timeout }: SubmitOrderRequestOptions = {}): Promise<Response<InternalOrderResponseBody>> {
+    submitOrder(body?: InternalOrderRequestBody, { headers, timeout }: SubmitOrderRequestOptions = {}): Promise<Response<InternalOrderResponseBody>> {
         const url = '/internalapi/v1/checkout/order';
 
         return this._requestSender.post<InternalOrderResponseBody>(url, {
```
animesh1987, line 52: Do we need to do a check if `body` is undefined here? Or `{ body: undefined }` is a valid payload for the API?

src/core/payment-integration/payment-integration-service.ts
```diff
@@ -0,0 +1,21 @@
+import { BillingAddressRequestBody } from '../../billing';
+import { OrderRequestBody } from '../../order';
+import { ShippingAddressRequestBody } from '../../shipping';
+
+import PaymentIntegrationSelectors from './payment-integration-selectors';
+
+export default interface PaymentIntegrationService {
```
animesh1987, line 7: What about methods related to payments, eg submitPayment, getPaymentMethod? Do we expose the `_paymentMethodActionCreator` and `_paymentActionCreator` to the strategy packages?

src/order/order-request-sender.ts
davidchin: `/internalapi/v1/checkout/order` endpoint doesn't actually require any payload. So yes, `{ body: undefined }` is valid.

src/core/payment-integration/payment-integration-service.ts
davidchin: Thanks, good spot, I missed them.

animesh1987 (approved):

animesh1987:
Thanks DC for leading the charge on this 🙇

serhii-tkachenko:
Great work DC! 👍

PR #1396, opened 2022-04-05
feat(payment): Payments 7707 Create PPSDK credit card substrategy

Tharaae:
## What?
Create a new payment strategy for Credit Card Payment methods by PPSDK payment providers, instead of using the existing generic credit card payment strategy.

## Why?
To allow handling of different credit card payment features that will be available via PPSDK, like human verification, 3D security check, vaulting credit cards, save as default payment method.

# Related PRs
checkout-js PR: https://github.com/bigcommerce/checkout-js/pull/841

## Testing / Proof
Unit tests
Screenshots
**Hosted form is rendered when PPSDK integration type experiment is turned on:**
Screen Shot 2022-04-05 at 11 06 49 am

**Method returns from bcapp as PAYMENT_TYPE_SDK:**
Screen Shot 2022-04-05 at 11 08 11 am

**The new payment endpoint is used:**
Screen Shot 2022-04-05 at 11 10 08 am

### DIFF
--- a/src/hosted-form/hosted-form.ts
+++ b/src/hosted-form/hosted-form.ts
@@ -2,6 +2,7 @@ import { noop, without } from 'lodash';
 
 import { IframeEventListener } from '../common/iframe';
 import { OrderPaymentRequestBody } from '../order';
+import { PaymentStrategyType } from '../payment';
 import { StepHandler } from '../payment/strategies/ppsdk/step-handler';
 import { PaymentHumanVerificationHandler } from '../spam-protection';
 
@@ -9,7 +10,7 @@ import { InvalidHostedFormConfigError } from './errors';
 import HostedField from './hosted-field';
 import HostedFormOptions from './hosted-form-options';
 import HostedFormOrderDataTransformer from './hosted-form-order-data-transformer';
-import { HostedInputEnterEvent, HostedInputEventMap, HostedInputEventType } from './iframe-content';
+import { HostedInputEnterEvent, HostedInputEventMap, HostedInputEventType, HostedInputSubmitSuccessEvent } from './iframe-content';
 
 type HostedFormEventCallbacks = Pick<HostedFormOptions, 'onBlur' | 'onCardTypeChange' | 'onFocus' | 'onEnter' | 'onValidate'>;
 
@@ -23,7 +24,7 @@ export default class HostedForm {
         private _payloadTransformer: HostedFormOrderDataTransformer,
         private _eventCallbacks: HostedFormEventCallbacks,
         private _paymentHumanVerificationHandler: PaymentHumanVerificationHandler,
-        ppsdkStepHandler?: StepHandler
+        private _ppsdkStepHandler: StepHandler
     ) {
         const { onBlur = noop, onCardTypeChange = noop, onFocus = noop, onValidate = noop } = this._eventCallbacks;
 
@@ -36,7 +37,7 @@ export default class HostedForm {
         this._eventListener.addListener(HostedInputEventType.CardTypeChanged, ({ payload }) => this._cardType = payload.cardType);
         this._eventListener.addListener(HostedInputEventType.BinChanged, ({ payload }) => this._bin = payload.bin);
 
-        this._eventListener.addListener(HostedInputEventType.SubmitSucceeded, ({ payload }) => ppsdkStepHandler?.handle(payload));
+        this._eventListener.addListener(HostedInputEventType.SubmitSucceeded, this._handleSubmitSuccess);
     }
 
     getBin(): string | undefined {
@@ -109,5 +110,15 @@ export default class HostedForm {
         const { onEnter = noop } = this._eventCallbacks;
 
         onEnter(payload);
+    };
+
+    private _handleSubmitSuccess: (event: HostedInputSubmitSuccessEvent) => Promise<void> = async ({ payload }) => {
+        const { paymentType, response } = payload;
+
+        if (paymentType === PaymentStrategyType.PPSDK) {
+            return this._ppsdkStepHandler.handle(response);
+        }
+
+        return Promise.resolve();
     };
 }
--- a/src/hosted-form/hosted-form.ts
+++ b/src/hosted-form/hosted-form.ts
@@ -2,6 +2,7 @@ import { noop, without } from 'lodash';
 
 import { IframeEventListener } from '../common/iframe';
 import { OrderPaymentRequestBody } from '../order';
+import { StepHandler } from '../payment/strategies/ppsdk/step-handler';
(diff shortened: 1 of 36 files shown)


src/hosted-form/hosted-form.ts
```diff
@@ -21,7 +22,8 @@ export default class HostedForm {
         private _eventListener: IframeEventListener<HostedInputEventMap>,
         private _payloadTransformer: HostedFormOrderDataTransformer,
         private _eventCallbacks: HostedFormEventCallbacks,
-        private _paymentHumanVerificationHandler: PaymentHumanVerificationHandler
+        private _paymentHumanVerificationHandler: PaymentHumanVerificationHandler,
+        ppsdkStepHandler?: StepHandler
```
davidchin, line 26: Could we make this a required dependency rather than optional? And instead of conditionally handling `HostedInputEventType.SubmitSucceeded` event depending on whether `StepHandler` is injected or not, what we can do is have the event carry a flag to indicate what kind of `payload` it is carrying (i.e.: if it's PPSDK response, handle it with `StepHandler`). That way it is more explicit and the same `HostedForm` instance is compatible with both non-PPSDK and PPSDK payment methods.

src/hosted-form/iframe-content/hosted-input-payment-handler.ts
```diff
@@ -43,15 +43,19 @@ export default class HostedInputPaymentHandler {
...
+            const response = await this._paymentRequestSender.submitPayment(
                 this._paymentRequestTransformer.transformWithHostedFormData(
                     values,
                     data,
                     this._inputStorage.getNonce() || ''
                 )
             );
 
-            this._eventPoster.post({ type: HostedInputEventType.SubmitSucceeded });
+            if (data.paymentMethod?.type === PaymentStrategyType.PPSDK) {
```
davidchin, line 54: This `if` statement could be lifted up to the event handler instead.

src/payment/strategies/ppsdk/sub-strategies/card-sub-strategy.ts
```diff
@@ -0,0 +1,58 @@
...
+        const { paymentSettings: { bigpayBaseUrl: host = '' } = {} } = config.getStoreConfig() || {};
+
+        if (!formOptions) {
+            throw new InvalidArgumentError();
+        }
+
+        const form = formOptions && this._hostedFormFactory.create(host, formOptions, this._stepHandler);
+
+        form.attach()
+            .then(() => {
```
davidchin, line 48: Prefer to use `async/await` if possible 🍹

src/payment/strategies/ppsdk/ppsdk-sub-strategy.ts
```diff
@@ -8,5 +9,9 @@ export interface SubStrategySettings {
 }
 
 export interface SubStrategy {
-    execute(settings: SubStrategySettings): Promise<void>;
+    execute(settings: SubStrategySettings): void;
```
davidchin, line 12: Are you sure you want to change from `Promise<void>` to `void`? I can see that the implementation actually treats this as an async method, but the interface doesn't seem to reflect that. e.g.: https://github.com/bigcommerce/checkout-sdk-js/blob/9123b21a61f2b9d69d38a6bbed99d6dd19c3f962/src/payment/strategies/ppsdk/sub-strategies/card-sub-strategy.ts#L31

src/payment/strategies/ppsdk/ppsdk-sub-strategy.ts
```diff
@@ -8,5 +9,9 @@ export interface SubStrategySettings {
...
+
+    initialize(options?: PaymentInitializeOptions): void;
```
davidchin, line 14: Likewise, you're waiting for this method to complete: https://github.com/bigcommerce/checkout-sdk-js/blob/9123b21a61f2b9d69d38a6bbed99d6dd19c3f962/src/payment/strategies/ppsdk/ppsdk-strategy.ts#L104

src/hosted-form/hosted-form.ts
Tharaae: This means that we will inject the `StepHandler` into other strategies that create a hosted form while we know that they don't need the `StepHandler`, like `StripeV3PaymentStrategy`, `MonerisPaymentStrategy` as well as the generic `CreditCardPaymentStrategy` and `CreditCardRedirectPaymentStrategy`. Why would we do this?

davidchin: You wouldn't though because the construction of `HostedForm` is handled by `HostedFormFactory`, so all these strategies don't need to supply the dependencies required to construct `HostedForm` or know about `StepHandler`. For example, you'd supply `StepHandler` in a similar way as `PaymentHumanVerificationHandler` inside the factory method.

Tharaae: https://github.com/bigcommerce/checkout-sdk-js/pull/1396/files#r848936962

Tharaae: I get what you mean, but please check [my comment below](/bigcommerce/checkout-sdk-js/pull/1396#discussion_r848936962) about the issue.

src/payment/create-payment-strategy-registry.ts
```diff
@@ -671,7 +671,7 @@ export default function createPaymentStrategyRegistry(
         new PPSDKStrategy(
             store,
             orderActionCreator,
-            createSubStrategyRegistry(requestSender, stepHandler),
+            createSubStrategyRegistry(store, orderActionCreator, requestSender, stepHandler),
             new PaymentResumer(requestSender, stepHandler),
             new BrowserStorage('PPSDK')
         )
```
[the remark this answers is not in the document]
davidchin: Spoke privately and explained how we can avoid `StepHandler` being an optional parameter and passing it conditionally in each payment strategy if we need it to be a stateful instance.

src/payment/strategies/ppsdk/sub-strategies/card-sub-strategy.ts
```diff
@@ -0,0 +1,57 @@
...
+export class CardSubStrategy implements SubStrategy {
+    protected _hostedForm?: HostedForm;
+
+    constructor(
+        private _store: CheckoutStore,
+        private _orderActionCreator: OrderActionCreator,
+        private _hostedFormFactory: HostedFormFactory
+    ) {}
+
+    async execute({ payment }: SubStrategySettings) {
```
davidchin, line 18: It's a good practice to write out the return type of a public method so it's clear to other engineers reading the code.

davidchin (approved):
A small comment, otherwise LGTM, nice work! 👍

PR #1400, opened 2022-04-11
fix(payment): CHECKOUT-6071 Add apple pay style container class

animesh1987:
## What?
Add apple pay container class to body so apple pay button can be rendered properly.

## Why?
In the legacy implementation, styles are setup in a way that there is a class added to document.body that sets apple pay buttons display to block. Since we will be ramping up the experiment slowly add the class in SDK so the implementation of the button doesn't break.

Once the experiment is rolled out to 100%, we can remove this line from SDK and the relevant style from bcapp.

## Testing / Proof
- Screenshot

Screen Shot 2022-04-11 at 10 57 59 am (2)
Screen Shot 2022-04-11 at 10 57 59 am

@bigcommerce/checkout

### DIFF
--- a/src/checkout-buttons/strategies/apple-pay/apple-pay-button-strategy.ts
+++ b/src/checkout-buttons/strategies/apple-pay/apple-pay-button-strategy.ts
@@ -84,6 +84,7 @@ export default class ApplePayButtonStrategy implements CheckoutButtonStrategy {
             throw new InvalidArgumentError('Unable to create wallet button without valid container ID.');
         }
 
+        document.body.classList.add('apple-pay-supported');
         const button = document.createElement('div');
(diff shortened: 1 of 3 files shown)


bc-peng (approved):
Thanks @animesh1987 for creating a quick fix. 🙇🏻‍♂️

For future reference, `cornerstone` automatically adds `apple-pay-supported` to two places regardless of whether Apple Pay is turned on or off. One for the [cart](https://github.com/bigcommerce/cornerstone/blob/master/assets/js/theme/cart.js#L30), the other for [the cart preview](https://github.com/bigcommerce/cornerstone/blob/f77c1f6c97c82d8c85e0d50e0094ddcb7a8202ff/assets/js/theme/global/cart-preview.js#L19).

PR #1404, opened 2022-04-13
chore(common): CHECKOUT-6591 Set node version to 14 for repo

animesh1987:
## What?
Update node version of sdk

## Why?
Before we start the next project, it's better to update to a newer node version

## Testing / Proof
- Tests

@bigcommerce/checkout


TomA-R:
Could you update .nvmrc too?

.nvmrc
```diff
@@ -1 +1 @@
-10
+14
```
davidchin, line 1:
```suggestion
14.18
```

To match with the version we're using in CircleCI

package.json
```diff
@@ -18,9 +18,13 @@
...
     "url": "https://github.com/bigcommerce/checkout-sdk-js/issues"
   },
+  "engines": {
+    "node": "14.18",
+    "npm": "6"
+  },
   "homepage": "https://github.com/bigcommerce/checkout-sdk-js",
   "scripts": {
-    "prepare": "check-node-version --node '>=6' --npm '>=6'",
+    "prepare": "check-node-version --node '>=14' --npm '>=6'",
```
davidchin, line 27:
```suggestion
    "prepare": "check-node-version --node '>=14.18' --npm '>=6'",
```

src/payment/strategies/moneris/moneris-payment-strategy.spec.ts
```diff
@@ -246,6 +246,24 @@ describe('MonerisPaymentStrategy', () => {
     });
 
     describe('#execute', () => {
+        it('throws error when moneris response is not successful', async () => {
```
davidchin, line 249: Do we need this change? It seems you just moved the `it` block to the top?

davidchin (approved):

src/payment/strategies/moneris/moneris-payment-strategy.spec.ts
animesh1987: Yes there is an issue how these tests are written, as postmessage payload was not leaking through tests and it was having intermittent behaviour. So I moved this one to top, and that fixed the issue. Once this is merged I will ping the integrations team to have a look as to what exactly was causing the issue with the test.

PR #1405, opened 2022-04-13
fix(shipping): CHECKOUT-6422 Use union type for consignment assignment interface

animesh1987:
## What?
Use union type for consignment assignment interface

## Why?
As part of interface update it causes a breaking change which is not out intention at this point. So implement a union interface with a type guard so custom checkouts can update sdk without having build issues for their checkout

## Testing / Proof
- Test

@bigcommerce/checkout


src/shipping/consignment-action-creator.ts
```diff
@@ -19,7 +19,7 @@ export default class ConsignmentActionCreator {
     ) {}
 
     unassignItemsByAddress(
-        consignment: ConsignmentAssignmentRequestBody,
+        consignment: ConsignmentAssignmentRequestBody ,
```
davidchin, line 22:
```suggestion
        consignment: ConsignmentAssignmentRequestBody,
```

davidchin (approved):

PR #1407, opened 2022-04-15
[POC] feat(payment): PAYPAL-1406 added initialisation and deinitialisation smart button (checkout button) methods to checkout service

serhii-tkachenko:
## What?
Added initialisation and deinitialisation smart button (checkout button) methods to checkout service to make an ability to initialise smart buttons on checkout page

## Why?
Due the initiative: https://jira.bigcommerce.com/browse/PAYPAL-1379

## Testing / Proof
Unit tests are not finished yet due the POC.

Here is a scheenshot of the braintree paypal smart button integration:
Screenshot 2022-04-15 at 12 13 42

### DIFF
--- a/src/checkout/checkout-service.ts
+++ b/src/checkout/checkout-service.ts
@@ -3,6 +3,7 @@ import { Observable } from 'rxjs';
 
 import { AddressRequestBody } from '../address';
 import { BillingAddressActionCreator, BillingAddressRequestBody } from '../billing';
+import { CheckoutButtonStrategyActionCreator } from '../checkout-buttons';
 import { createDataStoreProjection, DataStoreProjection } from '../common/data-store';
 import { ErrorActionCreator, ErrorMessageTransformer } from '../common/error';
 import { RequestOptions } from '../common/http-request';
@@ -68,7 +69,8 @@ export default class CheckoutService {
         private _spamProtectionActionCreator: SpamProtectionActionCreator,
         private _storeCreditActionCreator: StoreCreditActionCreator,
         private _subscriptionsActionCreator: SubscriptionsActionCreator,
-        private _formFieldsActionCreator: FormFieldsActionCreator
+        private _formFieldsActionCreator: FormFieldsActionCreator,
+        private _checkoutButtonStrategyActionCreator: CheckoutButtonStrategyActionCreator
     ) {
         this._errorTransformer = createCheckoutServiceErrorTransformer();
         this._selectorsFactory = createCheckoutSelectorsFactory();
@@ -344,6 +346,18 @@ export default class CheckoutService {
         const action = this._paymentMethodActionCreator.loadPaymentMethod(methodId, options);
 
         return this._dispatch(action, { queueId: 'paymentMethods' });
+    }
+
+    initializeSmartButton(options: any): Promise<CheckoutSelectors> {
+        const action = this._checkoutButtonStrategyActionCreator.initialize(options);
+
+        return this._dispatch(action, { queueId: 'checkoutButtonStrategy' });
+    }
+
+    deinitializeSmartButton(options: any): Promise<CheckoutSelectors> {
+        const action = this._checkoutButtonStrategyActionCreator.deinitialize(options);
+
+        return this._dispatch(action, { queueId: 'checkoutButtonStrategy' });
     }
 
     /**
--- a/src/checkout-buttons/checkout-button-options.ts
+++ b/src/checkout-buttons/checkout-button-options.ts
@@ -16,6 +16,13 @@ export interface CheckoutButtonOptions extends RequestOptions {
      * The identifier of the payment method.
      */
     methodId: CheckoutButtonMethodType;
+
+    /**
+     * The identifier of the payment provider providing the payment method. This
+     * option is only required if the provider offers multiple payment options.
+     * i.e.: PayPal Commerce, Adyen, Klarna, etc.
+     */
+    gatewayId?: string;
 }
 
 export interface CheckoutButtonInitializeOptions extends CheckoutButtonOptions {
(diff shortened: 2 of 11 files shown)


src/checkout/checkout-service.ts
```diff
@@ -346,6 +348,18 @@ export default class CheckoutService {
         return this._dispatch(action, { queueId: 'paymentMethods' });
     }
 
+    initializeSmartButton(options: any): Promise<CheckoutSelectors> {
```
davidchin, line 351: @BC-ACherednichenko @serhii-tkachenko As you might already know, the current method for initialising wallet buttons through `CheckoutService` is `initializeCustomer`. Depending on the payment method, it may initialise a wallet button, or it may initialise other things, such as replacing the email field with their own (e.g.: Bolt and Stripe Link). The reason for the design is because we anticipate that different payment methods might have different initialisation requirements for the customer step of checkout. And we don't want add a new method to the SDK every time a new payment method introduces something new.

There’s no plan to change the API design, or deprecate `initializeCustomer` method and replace it with more granular methods, such as `initializeSmartButton`, `initializeEmailField` etc... In other words, `initializeCustomer` needs to continue to function and fulfil its purpose like it is now. If we want to introduce `initializeSmartButton` for the purpose of providing SDK client the means to initialise wallet buttons only, both `initializeSmartButton` and `initializeCustomer` methods need to be able to initialise the same set of buttons. We want to avoid the situation where a wallet button for one payment method, e.g.: PayPal, can only be initialised by `initializeSmartButton` but not `initializeCustomer`, and vice versa. With that, I see `initializeSmartButton` as a nice-to-have rather than a must-have in order to complete the project.

Now, in terms of implementation, how can we achieve the above without duplicating code? Right now, wallet buttons for PayPal can already be initialised through `CheckoutButtonInitializer#initializeButton` and the implementation is done through `PayPalButtonStrategy` (a `CheckoutButtonStrategy`). On the other hand, `CheckoutService#initializeCustomer` delegates to `CustomerStrategy` and it does not utilise `CheckoutButtonStrategy`. For most payment methods that display wallet buttons in checkout, the implementation for `CheckoutButtonStrategy` and `CustomerStrategy` are exactly the same. This is obviously not ideal. So it'd be nice if we can have a single implementation so both `CheckoutButtonInitializer#initializeButton` and `CheckoutService#initializeCustomer` can reference.

One approach we can take is to extract the common code into a shared service, e.g.: `PayPalButtonInitializer` which both `PayPalButtonStrategy` and `PayPalCustomerStrategy` can reference. For example:

e.g.:
```ts
class PayPalCustomerStrategy {
    constructor(
        private store: ReadableCheckoutStore,
        private payPalButtoninitializer: PayPalButtonInitializer
    ) {}

    async initialize(options: CustomerInitializeOptions): Promise<InternalCheckoutSelectors> {
        await this.payPalButtoninitializer.initializeButton({ containerId: options.paypal.containerId });

    }
}
```

```ts
class PayPalButtonStrategy {
    constructor(
        private store: ReadableCheckoutStore,
        private payPalButtoninitializer: PayPalButtonInitializer
    ) {}

    async initialize(options: CheckoutButtonInitializeOptions): Promise<InternalCheckoutSelectors> {
        await this.payPalButtoninitializer.initializeButton({ containerId: options.paypal.containerId });
 
    }
}
```

Another approach we can take, is to use one of the strategies (e.g.: `PayPalButtonStrategy`) as the source of truth. So for the other type (e.g.: `CustomerStrategy`), we could have a generic strategy like `WalletButtonCustomerStrategy` that can delegate to the former. The strategy can then be registered against `paypal` and any other payment methods that don't require a specific `CustomerStrategy`. For example:

```ts
class WalletButtonCustomerStrategy {
    constructor(
        private store: ReadableCheckoutStore,
        private checkoutButtoninitializer: CheckoutButtonInitializer
    ) {}

    async initialize(options: CustomerInitializeOptions): Promise<InternalCheckoutSelectors> {
        await this.checkoutButtoninitializer.initializeButton({
            containerId: options[options.methodId].containerId,
            methodId: options.methodId,
        });

    }
}
```

There are pros and cons to both approaches. The first approach requires us to write boilerplate code but it offers flexibility as you can write code before and after `initializeButton` call inside each method-specific strategy. The second approach doesn't require us to write boilerplate code at the expense of flexibility. Both approaches allow us to achieve the aforementioned objective, which is to minimise code duplication while maintaining compatibility with the existing API.

Please let me know what you think about the above suggestions. 🙂

BC-ACherednichenko: Thank you for your input @davidchin!

> The reason for the design is because we anticipate that different payment methods might have different initialisation requirements for the customer step of checkout.

The key here is customer step of checkout. I concur, it has different capabilities as opposed to what we have on the cart page (the mentioned email, login functionality, etc). Nevertheless, the proposed change is not related to customer step, it just unlocks additional capabilities to the regular button strategy.

> There’s no plan to change the API design, or deprecate initializeCustomer method and replace it with more granular methods, such as initializeSmartButton, initializeEmailField etc...

Same here, customer strategy serves as multifunctional strategy for **customer** step, while we want to achive something generic and location agnostic. Moreover, the proposal is not to introduce something new, but, rather re-use existing code base with minimal or no changes to the existing API interfaces. For instance, if we have checkout button strategy that unlocks payments capabilities for all wallet methods, why do we need to consolidate customer strategy, can you break it down to me please?

> One approach we can take is to extract the common code into a shared service, e.g.: PayPalButtonInitializer which both PayPalButtonStrategy and PayPalCustomerStrategy can reference

I am more that positive about such solution, but, I am still not convinced it is the way to for us for top checkout buttons place?

As you may see [here](https://github.com/bigcommerce/checkout-js/pull/851/files#diff-510616288e5351c230e85700309bf9db005bc3caee9147d83b2f57cea0b75ad2R37)
either way we have to add a new component to the checkout-js layout to introduce top button place. And the key question here is which interface it consumes? (either checkout-button strategy interface or customer strategy one) 

> maintaining compatibility with the existing API

What if we stick to the existing intefaces in checkout-button strategy, in that case compatibility will not be an issue at all, right?

As I mentioned earlier, dupplication is a secondary issue which can be easilly resolved. The biggest concern is I am not sure we can benefit from the step / place specific strategy in the long. 

If we focus on making checkout-button strategy completely context agnostic, it will bring singificant value in the long term because there are other places where it can be leveraged, on contrary a customer strategy does not fit will with the idea of the agnostic strategy that can be initialized either on PDP, cart, or checkout pages.

PR #1410, opened 2022-04-18
feat(payment): PAYPAL-1177 added implementation to handle 3DS required error from backend

serhii-tkachenko:
## What?
- added implementation to handle 3DS required error for stored credit card;
- added extra check on 3DS regulation for credit card (not stored);
- added extra test coverage for Braintree strategy, payment processor and hosted form functionality;

## Why?
Some EU merchants are having the issue with payments made with stored credit cards. When the customer tried to pay with store credit card the Braintree had an error with "Authentication Required: The customer’s bank declined the transaction because a 3D Secure authentication was not performed during checkout" message.

The main point is in our implementation of 3DS verification for Braintree, or specifically in the fact that 3DS is skipped for stored credit card payments.

This behaviour seems to be by design, but the receiving banks are now rejecting these transactions due to this functionality. This is because of the new 3DS2 EU requirements stipulating that 3DS must be checked.

## Testing / Proof
Unit tests
Manual tests

Here is a screenshot of 3DS verification modal:
Screenshot 2022-04-19 at 11 08 50

### DIFF
--- a/src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
+++ b/src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
@@ -1,13 +1,14 @@
-import { pick } from 'lodash';
+import { some } from 'lodash';
 
 import { Address } from '../../../address';
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
-import { MissingDataError, MissingDataErrorType } from '../../../common/error/errors';
+import { MissingDataError, MissingDataErrorType, RequestError } from '../../../common/error/errors';
 import { OrderActionCreator, OrderPaymentRequestBody, OrderRequestBody } from '../../../order';
 import { OrderFinalizationNotRequiredError } from '../../../order/errors';
 import { PaymentArgumentInvalidError, PaymentMethodFailedError } from '../../errors';
+import { isHostedInstrumentLike, PaymentMethod } from '../../index';
 import isVaultedInstrument from '../../is-vaulted-instrument';
-import { CreditCardInstrument, NonceInstrument, PaymentInstrument, PaymentInstrumentMeta, VaultedInstrumentWithNonceVerification } from '../../payment';
+import { CreditCardInstrument, NonceInstrument, PaymentInstrument, PaymentInstrumentMeta } from '../../payment';
 import PaymentActionCreator from '../../payment-action-creator';
 import PaymentMethodActionCreator from '../../payment-method-action-creator';
 import { PaymentInitializeOptions, PaymentRequestOptions } from '../../payment-request-options';
@@ -19,6 +20,7 @@ export default class BraintreeCreditCardPaymentStrategy implements PaymentStrate
     private _is3dsEnabled?: boolean;
     private _isHostedFormInitialized?: boolean;
     private _deviceSessionId?: string;
+    private _paymentMethod?: PaymentMethod;
 
     constructor(
         private _store: CheckoutStore,
@@ -30,20 +32,21 @@ export default class BraintreeCreditCardPaymentStrategy implements PaymentStrate
 
     async initialize(options: PaymentInitializeOptions): Promise<InternalCheckoutSelectors> {
         const state = await this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(options.methodId));
-        const paymentMethod = state.paymentMethods.getPaymentMethod(options.methodId);
+        this._paymentMethod = state.paymentMethods.getPaymentMethod(options.methodId);
 
-        if (!paymentMethod || !paymentMethod.clientToken) {
+        if (!this._paymentMethod?.clientToken) {
             throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
         }
 
         try {
-            this._braintreePaymentProcessor.initialize(paymentMethod.clientToken, options.braintree);
+            this._braintreePaymentProcessor.initialize(this._paymentMethod.clientToken, options.braintree);
 
             if (this._isHostedPaymentFormEnabled(options.methodId, options.gatewayId) && options.braintree?.form) {
-                this._isHostedFormInitialized = await this._braintreePaymentProcessor.initializeHostedForm(options.braintree.form);
+                await this._braintreePaymentProcessor.initializeHostedForm(options.braintree.form);
+                this._isHostedFormInitialized = this._braintreePaymentProcessor.isInitializedHostedForm();
             }
 
-            this._is3dsEnabled = paymentMethod.config.is3dsEnabled;
+            this._is3dsEnabled = this._paymentMethod.config.is3dsEnabled;
             this._deviceSessionId = await this._braintreePaymentProcessor.getSessionId();
         } catch (error) {
             this._handleError(error);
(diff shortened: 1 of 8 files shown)


src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
```diff
@@ -177,6 +191,35 @@ export default class BraintreeCreditCardPaymentStrategy implements PaymentStrate
         };
     }
 
+    private async _processAdditionalAction(
+        error: Error,
+        payment: OrderPaymentRequestBody,
+        billingAddress: Address,
+        orderAmount: number
+    ): Promise<InternalCheckoutSelectors> {
+        // test: check handlerError function
```
animesh1987, line 200: Does this comment need to be here?

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
```diff
@@ -177,6 +191,35 @@ export default class BraintreeCreditCardPaymentStrategy implements PaymentStrate
...
+                    }),
+                },
+            }));
+        } catch (error) {
+            return this._processAdditionalAction(error, payment, billingAddress, orderAmount);
+        }
+    }
+
     private _mapToNonceInstrument(instrument: PaymentInstrument): NonceInstrument {
         return pick(instrument as NonceInstrument, 'nonce', 'shouldSaveInstrument', 'shouldSetAsDefaultInstrument');
```
animesh1987, line 224: Can we use a typeguard here for instrument rather than asserting it using `as`

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
```diff
@@ -177,6 +191,35 @@ export default class BraintreeCreditCardPaymentStrategy implements PaymentStrate
...
+                paymentData: {
+                    deviceSessionId: this._deviceSessionId,
+                    ...this._mapToNonceInstrument({
+                        ...payment.paymentData,
+                        ...await this._braintreePaymentProcessor.challenge3DSVerification(storedCreditCardNonce, orderAmount),
+                    }),
+                },
+            }));
+        } catch (error) {
+            return this._processAdditionalAction(error, payment, billingAddress, orderAmount);
```
animesh1987, line 219: This seems like some kind of recursion, where if the submitPayment fails then we call the method again. Does that seem right? Since it can run into an infinite loop.

src/payment/strategies/braintree/braintree-hosted-form.ts
```diff
@@ -21,12 +21,13 @@ export default class BraintreeHostedForm {
     private _cardNameField?: BraintreeRegularField;
     private _formOptions?: BraintreeFormOptions;
     private _type?: BraintreeHostedFormType;
+    private _isInitializedHostedForm?: boolean;
```
animesh1987, line 24: Out of curiosity, why don't we initialise this as the variable as being `false`?

src/payment/strategies/braintree/braintree-hosted-form.ts
```diff
@@ -133,6 +142,42 @@ export default class BraintreeHostedForm {
         }
     }
 
+    async tokenizeWith3DSRegulationCheck(billingAddress: Address, merchantAccountId: string): Promise<Pick<BraintreeHostedFieldsTokenizePayload, 'authenticationInsight'> & NonceInstrument> {
+        if (!this._cardFields) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+        }
+
+        try {
+            const { authenticationInsight, nonce } = await this._cardFields.tokenize(omitBy({
```
animesh1987, line 151: Can we please break this into multiple lines, it can be a bit harder to read in this current state.

src/payment/strategies/braintree/braintree-hosted-form.ts
```diff
@@ -133,6 +142,42 @@ export default class BraintreeHostedForm {
...
+        } catch (error) {
+            const errors = this._mapTokenizeError(error);
+
+            if (errors) {
+                this._formOptions?.onValidate?.({
+                    isValid: false,
+                    errors,
+                });
+
+                throw new PaymentInvalidFormError(errors as PaymentInvalidFormErrorDetails);
```
animesh1987, line 174: Type guard please

src/payment/strategies/braintree/braintree-hosted-form.ts
```diff
@@ -133,6 +142,42 @@ export default class BraintreeHostedForm {
         }
     }
 
+    async tokenizeWith3DSRegulationCheck(billingAddress: Address, merchantAccountId: string): Promise<Pick<BraintreeHostedFieldsTokenizePayload, 'authenticationInsight'> & NonceInstrument> {
```
DeveloperAlan, line 145: BraintreeHostedFieldsTokenizePayload has the type `details` but this method doesn't return any `details`, would the pick be necessary in this case?

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
serhii-tkachenko: Nope, I forgot to remove it

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
serhii-tkachenko: It can run into an infinite loop only in case if the backend will return the same "3DS required" error. This method are used to catch another errors that can be thrown in case of backend or braintree issues when we try to submit order with verified data.

In [this case](https://github.com/bigcommerce/checkout-sdk-js/pull/1410/files/b77e28e2eb93a851505b95c6b8935fb9b27d3680#diff-481762847de0d12fd9cbf38ac9e0555aca512bbdda28be8437d3320bfbdcca14R201) we will throw an error as a modal to the customer through checkout-js, otherwise we should trigger 3DS verification for stored credit card.

src/payment/strategies/braintree/braintree-hosted-form.ts
serhii-tkachenko: Oh..Nice catch up. We can do that. Thanks 👍

src/payment/strategies/braintree/braintree-hosted-form.ts
serhii-tkachenko: Ok, I will break this into multiple lines

src/payment/strategies/braintree/braintree-hosted-form.ts
serhii-tkachenko: I used Pick to get 'authenticationInsight' type from BraintreeHostedFieldsTokenizePayload interface. But, I think, we can create another interface to handle this case

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
serhii-tkachenko: I think we can

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
serhii-tkachenko: Perhaps, I can change it to handleError function to prevent infinite loop

src/payment/strategies/braintree/braintree-hosted-form.ts
serhii-tkachenko: We have the same row in two methods above. Is it ok to make it in another pr, like braintree hosted form class refactoring?

src/payment/strategies/braintree/braintree-hosted-form.ts
serhii-tkachenko: Fixed. Thanks you for your comment 🙏

src/payment/strategies/braintree/braintree-hosted-form.ts
serhii-tkachenko: Fixed

src/payment/strategies/braintree/braintree-hosted-form.ts
serhii-tkachenko: Fixed

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
serhii-tkachenko: Fixed

src/payment/strategies/braintree/braintree-credit-card-payment-strategy.ts
serhii-tkachenko: Fixed

animesh1987 (approved):
Nice work @serhii-tkachenko. Thanks for the patience.

PR #1415, opened 2022-04-26
fix(common): CHECKOUT-6268 Check in array while loading config

animesh1987:
## What?
Fix build process for cdn-deploy

## Why?
As a config-name args has a type of string[], due to which the build for cdn was failing. Apply the fix.

## Testing / Proof
Screen Shot 2022-04-26 at 2 07 19 pm

@bigcommerce/checkout

### DIFF
--- a/webpack-cdn.config.js
+++ b/webpack-cdn.config.js
@@ -85,7 +85,7 @@ async function getCdnLoaderConfig(options, argv) {
 
 // This configuration is for building distribution files for CDN deployment
 async function getConfigs(options, argv) {
-    if (argv.configName === 'umd-loader') {
+    if (argv.configName.includes('umd-loader')) {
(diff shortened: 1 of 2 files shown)


webpack-cdn.config.js
```diff
@@ -10,6 +10,7 @@ const { babelLoaderRules, getBaseConfig, libraryEntries, libraryName, srcPath }
 const baseOutputPath = path.join(__dirname, 'dist-cdn');
 
 async function getCdnConfig(options, argv) {
+    console.log('lolol runswwwwwww');
```
davidchin, line 13: 😄

animesh1987: 🤦‍♂️

davidchin (approved):

PR #1423, opened 2022-04-27
feat(payment): INT-3925 StripeV3: BOPIS

mauricio-sg:
## What? [INT-3925](https://bigcommercecloud.atlassian.net/browse/INT-3925)
Check if BOPIS is enabled and send `null` as the shipping address to Stripe if the shopper selects to pick up the entire order.

## Why?
When the shopper selects to pick up the entire order, nothing is passed in for shipping to Stripe.

## Testing / Proof
_**BC Control Panel:**_
Screen Shot 2022-04-27 at 2 40 11 PM
_**Stripe Control Panel:**_
Screen Shot 2022-04-27 at 2 30 33 PM

## Sibling PR
~👉 https://github.com/bigcommerce/bigcommerce/pull/46057 to get the `isBopisEnabled` flag.~
👉 https://github.com/bigcommerce/bigcommerce/pull/47101 to get the BOPIS data.

### DIFF
--- a/packages/core/src/payment/payment-request-transformer.ts
+++ b/packages/core/src/payment/payment-request-transformer.ts
@@ -1,6 +1,6 @@
 import { pick } from 'lodash';
 
-import { mapToInternalAddress } from '../address';
+import { mapToInternalAddress, Address } from '../address';
 import { mapToInternalCart } from '../cart';
 import { InternalCheckoutSelectors } from '../checkout';
 import { CheckoutButtonMethodType } from '../checkout-buttons/strategies';
@@ -26,7 +26,7 @@ export default class PaymentRequestTransformer {
         const customer = checkoutState.customer.getCustomer();
         const order = checkoutState.order.getOrder();
         const paymentMethod = checkoutState.paymentMethods.getPaymentMethod(payment.methodId, payment.gatewayId);
-        const shippingAddress = checkoutState.shippingAddress.getShippingAddress();
+        const shippingAddress = this._mapShippingAddress(checkoutState, paymentMethod);
         const consignments = checkoutState.consignments.getConsignments();
         const shippingOption = checkoutState.consignments.getShippingOption();
         const storeConfig = checkoutState.config.getStoreConfig();
@@ -143,5 +143,20 @@ export default class PaymentRequestTransformer {
                 ccNumber: this._cardNumberFormatter.unformat(values.cardNumber || ''),
                 hostedFormNonce: nonce,
             };
+    }
+
+    private _mapShippingAddress(checkoutState: InternalCheckoutSelectors, paymentMethod?: PaymentMethod): Address | undefined {
+        if (paymentMethod) {
+            const isBopisEnabled = paymentMethod.initializationData?.bopis?.enabled;
+            const requiredAddress = paymentMethod.initializationData?.bopis?.requiredAddress;
+            const consignments = checkoutState.consignments.getConsignments();
+            const isPickup = consignments?.every(consignment => consignment.selectedPickupOption);
+
+            if (isBopisEnabled && isPickup && requiredAddress === 'none') {
+                return undefined;
+            }
+        }
+
+        return checkoutState.shippingAddress.getShippingAddress();
     }
 }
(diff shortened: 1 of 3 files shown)


ggarcia-bc (approved):

src/payment/payment-request-transformer.ts
```diff
@@ -20,13 +20,13 @@ export default class PaymentRequestTransformer {
...
 
-    transform(payment: Payment, checkoutState: InternalCheckoutSelectors): PaymentRequestBody {
+    transform(payment: Payment, checkoutState: InternalCheckoutSelectors, address?: Address | null): PaymentRequestBody {
         const billingAddress = checkoutState.billingAddress.getBillingAddress();
         const checkout = checkoutState.checkout.getCheckout();
         const customer = checkoutState.customer.getCustomer();
         const order = checkoutState.order.getOrder();
         const paymentMethod = checkoutState.paymentMethods.getPaymentMethod(payment.methodId, payment.gatewayId);
-        const shippingAddress = checkoutState.shippingAddress.getShippingAddress();
+        const shippingAddress = address === undefined ? checkoutState.shippingAddress.getShippingAddress() : (address ?? undefined);
```
animesh1987, line 29: Do we need to do this, since `checkoutState.shippingAddress.getShippingAddress()` will return undefined if there is no address?

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -517,4 +517,13 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
             this._stripeElement = undefined;
         }
     }
+
+    private _getAddress(isBopisEnabled: boolean): Address | undefined | null {
```
animesh1987, line 521: This doesn't return an address at all. Does this seem correct?

src/payment/payment-request-transformer.ts
mauricio-sg: @animesh1987 What I'm doing in this PR is definitely not the best solution of all, but I want you to understand what we need to see if you can suggest something more robust.

If BOPIS is enabled, each provider will behave differently. Some of them require not to send an address at all, some others require to send the pickup address. So, I need a way to dynamically set this `shippingAddress` const in order to achieve what we want. And also I need a way to reuse code because the same logic is required for googlepay strategy and applepay js 😬

In this case, for StripeV3, we need to send nothing for `shippingAddress`. So, I made the address param optional to avoid breaking anything and if `address === undefined` it means that it should behave like before and use `checkoutState.shippingAddress.getShippingAddress()`. For this case, I'm sending a `null` to differentiate from `undefined`. If I send a `null`, `shippingAddress` will be explicitly set to `undefined` and no shipping address will be send to Stripe.

I was thinking of injecting some kind of strategy into this `PaymentRequestTransformer` that contains the required logic for each provider...

Please, let me know if something is unclear.

src/payment/strategies/stripev3/stripev3-payment-strategy.ts
mauricio-sg: Yes, I intentionally send `null`, please see my comment above.

animesh1987: Sorry think I didn't explain clearly, the function here doesn't return `Address` in any path, it either returns undefined or null?

src/payment/payment-request-transformer.ts
animesh1987: I think we can rethink the implementation in this case, since you can get paymentMethod and it's details from the payment passed. Also you have access to state. 

So I feel it maybe better to have a function here `_mapShippingAddress` or similar. And we can move the logic there. This avoids us to have the code in strategy, and then the need to pass an optional parameter down the method chain. Feel free to ping me on slack and we can work something out together.

packages/core/src/payment/payment-request-transformer.ts
```diff
@@ -144,4 +146,21 @@ export default class PaymentRequestTransformer {
                 hostedFormNonce: nonce,
             };
     }
+
+    private _mapShippingAddress(checkoutState: InternalCheckoutSelectors, paymentMethod?: PaymentMethod): Address | undefined {
+        if (
```
animesh1987, line 151: I think we discussed to use `initializationData` here for making the decision to resolve address or not. And we can also pass a property to decide if address is required or not.

mauricio-sg: I know, but we cannot determine that on the back-end because we rely on what the user selects, picking up the entire order, shipping, or a combination of them (in the future).

animesh1987 (approved):
Nice work @mauricio-sg

PR #1425, opened 2022-04-29
fix(checkout): CHECKOUT-6631 Sentry issue bugfix

bc-peng:
## What?
Fix Sentry Issue: [CHECKOUT-JS-4WG](https://sentry.io/organizations/bigcommerce/issues/2782844130/?referrer=jira_integration).

## Why?

Screen Shot 2022-04-29 at 11 49 52 am

When this happens, I'm guessing a shopper gets a message like this.

The reason for `t.filter is not a function` could be that `methods` within `methods.filter()` is an empty string. This bugfix provides a readable error message and instructs the shopper to try again.

## Testing / Proof
After applying the fix.
Screen Shot 2022-04-29 at 1 38 17 pm

### DIFF
--- a/src/payment/payment-method-action-creator.ts
+++ b/src/payment/payment-method-action-creator.ts
@@ -5,10 +5,10 @@ import { cachableAction, ActionOptions } from '../common/data-store';
 import { RequestOptions } from '../common/http-request';
 
 import { PaymentMethod } from '.';
+import { PaymentMethodInvalidError } from './errors';
 import { LoadPaymentMethodsAction, LoadPaymentMethodAction, PaymentMethodActionType } from './payment-method-actions';
 import PaymentMethodRequestSender from './payment-method-request-sender';
 import { isApplePayWindow } from './strategies/apple-pay';
-import {PaymentMethodInvalidError} from "./errors";
 
 const APPLEPAYID = 'applepay';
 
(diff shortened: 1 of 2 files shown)


src/payment/payment-method-action-creator.ts
```diff
@@ -8,6 +8,7 @@ import { PaymentMethod } from '.';
 import { LoadPaymentMethodsAction, LoadPaymentMethodAction, PaymentMethodActionType } from './payment-method-actions';
 import PaymentMethodRequestSender from './payment-method-request-sender';
 import { isApplePayWindow } from './strategies/apple-pay';
+import {PaymentMethodInvalidError} from "./errors";
```
animesh1987, line 11:
```suggestion
import { PaymentMethodInvalidError } from "./errors";
```

src/payment/payment-method-action-creator.ts
```diff
@@ -53,7 +56,8 @@ export default class PaymentMethodActionCreator {
     }
 
     private _filterApplePay(methods: PaymentMethod[]): PaymentMethod[] {
-        return methods.filter(method => {
+
+        return filter(methods, method => {
```
bc-peng, line 60: Note: We assume the back-end is good, so replace the vanilla `filter` with Lodash's `filter` method. Also, we have added an Array type check.

animesh1987 (approved):

PR #1428, opened 2022-04-29
fix(checkout): IINT-5126 [MPGS] Add delay between calls for 3ds

FranciscoOnt:
## What?
Add a delay of 2 seconds between the retries of the `authenticatePayer` calls for CBA_MPGS

## Why?
MPGS Documentation suggest waiting at around 10 seconds between the `initiateAuthentication` and `authenticatePayer`, this delay gives MPGS server enough time to run their processes and send a proper response for the authentication attempt.

## Testing / Proof
Unit tests

### DIFF
--- a/packages/core/src/payment/strategies/cba-mpgs/cba-mpgs-payment-strategy.spec.ts
+++ b/packages/core/src/payment/strategies/cba-mpgs/cba-mpgs-payment-strategy.spec.ts
@@ -99,6 +99,8 @@ describe('CBAMPGSPaymentStrategy', () => {
         submitOrderAction = of(createAction(OrderActionType.SubmitOrderRequested));
         submitPaymentAction = of(createAction(PaymentActionType.SubmitPaymentRequested));
 
+        jest.useFakeTimers();
+
         jest.spyOn(store.getState().config, 'getStoreConfigOrThrow')
             .mockReturnValue(getConfig().storeConfig);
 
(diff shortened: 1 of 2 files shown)


mauricio-sg (approved):
LGTM 🫥

src/payment/strategies/cba-mpgs/cba-mpgs-payment-strategy.ts
```diff
@@ -180,11 +180,14 @@ export default class CBAMPGSPaymentStrategy extends CreditCardPaymentStrategy {
                 return reject(new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized));
             }
 
-            this._threeDSjs.authenticatePayer(orderId, transactionId, data => {
+            this._threeDSjs.authenticatePayer(orderId, transactionId, async data => {
                 const error = data.error;
 
                 if (error) {
                     if (error.cause && error.cause === THREE_D_SECURE_BUSY && attempt < 5) {
+                        // Wait 2 seconds for MPGS server to process the `initiateAuthentication` call
```
animesh1987, line 188: Is there a document that this can be linked to? Or some other kind of reference.

animesh1987: If it says to wait for 10 seconds as in description, does it make sense to increase the wait time higher to 4 or so. So we don't hammer the API that much?

FranciscoOnt: @animesh1987 I've added a Link to the documentation.
Also I increased the delay time to 3 seconds, which gives a margin in the case the process finishes sooner (which happens on the test environment) or in case it finishes in time there's still 2 retries pending without waiting any longer.

animesh1987 (approved):

JuanTorres077:
💚 Tested
QA Details: 
Tested, working as expected with updates, unfortunately, there was not way to Test the retry scenarios. Since all transactions were passed at the first time, We also modified Chrome speed such as 3g slow and  getting passed at the first round... 
For further details, please take a look at below link:
[Evidence](https://drive.google.com/drive/folders/12rRturnVoqwW6SiRr_GflX0m-uiByjCu?usp=sharing)

PR #1432, opened 2022-05-05
chore(common): CHECKOUT-6268 Add nx package

animesh1987:
## What?
Wrap SDK core in Nx monorepo structure

## Why?
As part of extensions project, we are converting SDK into a monorepo. This is the first of many PRs in improving the overall developer experience.

## Testing / Proof
- Tests 
- Build

@bigcommerce/checkout


tmp/packages/core/tsconfig.generated.json
```diff
@@ -0,0 +1,10 @@
+{
```
davidchin, line 1: Is the `tmp` folder supposed to be committed?

packages/core/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+    "extends": "../../tsconfig.base.json",
+    "include": [
+        "src/common/types/card-validator.d.ts",
+        "src/**/*.ts",
+      ], 
+      "files": [
+        "src/common/types/card-validator.d.ts",
+        "src/common/types/webpack.d.ts"
```
davidchin, line 9: It seems to me that it's missing a closing `]` bracket.

.eslintrc.js
```diff
@@ -9,7 +9,8 @@
     },
     "parser": "@typescript-eslint/parser",
     "parserOptions": {
-        "project": "./tsconfig.json",
+        "project": "./packages/core/tsconfig.json",
```
davidchin, line 12: How come the eslintrc file at the root level is referencing a tsconfig file at the package level? Looking at what's done so far, it is the other way around where config files at the package level extend from the root level. Is this intentional?

animesh1987: Yes this is intentional since we use our implementation on eslint as of now. Which requires a tsconfig.json. In the follow up task once we migrate to eslint and use nx's lint package, we should be able to get rid of the above.

davidchin: We spoke offline and clarified what I was asking. Came up with a solution for the problem without having to reference package level config at the global level.

davidchin (approved):
LGTM, really nice work! 👍

PR #1446, opened 2022-05-23
feat(payment): BOLT-238 Card method swap

PavlenkoM:
## What?
Interfaces update for new `paymentMethodsWithCreditCardConflicts` in `storeConfig`. This need for implementation new logic in checkpot-js: [https://github.com/bigcommerce/checkout-js/pull/877](https://github.com/bigcommerce/checkout-js/pull/877)

## Why?
Because of task: [https://bigcommercecloud.atlassian.net/browse/BOLT-235](https://bigcommercecloud.atlassian.net/browse/BOLT-235)
For cases when merchant setup multiple payment providers and this providers has similar payment methods, we should hide this conflicts methods. 
For example:
Merchant setup Braintree and Bolt (with Braintree as payment system)
So customers will see two Credit card fields in payment section. And both of them will be processed by Braintree.
For this cases we need to hide similar fields and show only one Credit card field.
And we should keep workable all other payment methods for this providers, like embedded forms and other.
Video example of solution:

https://user-images.githubusercontent.com/9430298/169863639-332380e1-c125-45de-9d87-c101240ee023.mov

## Testing / Proof
...

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -112,6 +112,7 @@ export interface CheckoutSettings {
     orderTermsAndConditionsType: string;
     privacyPolicyUrl: string;
     providerWithCustomCheckout: string | null;
+    paymentMethodsWithCreditCardConflicts: PaymentMethodsWithConflicts[] | null;
     shippingQuoteFailedMessage: string;
     realtimeShippingProviders: string[];
     requiresMarketingConsent: boolean;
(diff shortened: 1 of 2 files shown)


bc-peng:
Hi there, is the back-end change for `checkout-settings` on Tier 3 now, can you please give me a hint about changes happened in the back-end?

animesh1987 (changes requested):
Still under discussion with the Bolt team, since we don't think that this API contract needs to be passed under checkout-settings API

PR #1448, opened 2022-05-24
feat(checkout): CHECKOUT-6703 post cartid for order creation API endpoint

bc-mackxu:
## What?
Added Cart Id as a required parameter for order creation API endpoint when creating an order.

## Why?
To support buy-now cart, we will need to pass cart id for order creation API endpoint when creating an order.

#### After place order is clicked, Order creation API endpoint is called
image
#### Cart id is included in the payload
image
#### Order created
image
#### Call GET order endpoint, order is created against the cart Id passed 
image

## Deployment
Will be merged after https://github.com/bigcommerce/bigcommerce/pull/46390 release to T3

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/order/internal-order-request-body.ts
+++ b/packages/core/src/order/internal-order-request-body.ts
@@ -1,6 +1,7 @@
 import { PaymentInstrument } from '../payment';
 
 export default interface InternalOrderRequestBody {
+    cartId: string;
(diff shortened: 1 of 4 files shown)


snaderiBC (approved):
👍

bc-peng (approved):

DeveloperAlan (approved):

PR #1450, opened 2022-05-25
feat(payment): PAYPAL-1466 fixed currency issue

andriiVitvitskyi1990:
## What?
Fixed default currency issue for Braintree SPB

## Why?
According to task https://bigcommercecloud.atlassian.net/browse/PAYPAL-1466

## Testing / Proof
...


packages/core/src/checkout-buttons/strategies/braintree/braintree-paypal-button-strategy.ts
```diff
@@ -41,7 +41,7 @@ export default class BraintreePaypalButtonStrategy implements CheckoutButtonStra
         const paymentMethod = this._paymentMethod = state.paymentMethods.getPaymentMethod(options.methodId);
         const isVenmoEnabled = paymentMethod?.initializationData?.isBraintreeVenmoEnabled;
         const storeState = await this._store.dispatch(this._checkoutActionCreator.loadDefaultCheckout());
-        const currency = storeState.config.getStoreConfig()?.shopperCurrency;
+        const currency = storeState.config.getStoreConfig()?.currency;
```
animesh1987, line 44: We should be using the currency of the cart here, which is the transacting currency. And not the currency from the storeconfig. 
`        const currency = storeState.getCartOrThrow()?.currency;
`

andriiVitvitskyi1990: fixed

PR #1460, opened 2022-06-02
feat(payment): PAYMENTS-7524 Human verification for PPSDK methods

Tharaae:
## What?
- Implement Human Verification feature for PPSDK payment methods on checkout as a payment step to be handled.
- Handle subsequent PPSDK payment steps.

## Why?
To support this feature when needed.

## Testing / Proof
Unit Tests
Screenshots
**First request payload**
Screen Shot 2022-06-05 at 4 15 26 pm

**First request response**
Screen Shot 2022-06-05 at 4 15 57 pm

**Second (last) request payload**
Screen Shot 2022-06-05 at 4 19 06 pm

**Second (Last) request response**
Screen Shot 2022-06-05 at 4 17 37 pm

### DIFF
--- a/packages/core/src/hosted-form/hosted-form.ts
+++ b/packages/core/src/hosted-form/hosted-form.ts
@@ -2,8 +2,7 @@ import { noop, without } from 'lodash';
 
 import { IframeEventListener } from '../common/iframe';
 import { OrderPaymentRequestBody } from '../order';
-import { PaymentStrategyType } from '../payment';
-import { StepHandler } from '../payment/strategies/ppsdk/step-handler';
+import { PaymentAdditionalAction } from '../payment';
 import { PaymentHumanVerificationHandler } from '../spam-protection';
 
 import { InvalidHostedFormConfigError } from './errors';
@@ -23,8 +22,7 @@ export default class HostedForm {
         private _eventListener: IframeEventListener<HostedInputEventMap>,
         private _payloadTransformer: HostedFormOrderDataTransformer,
         private _eventCallbacks: HostedFormEventCallbacks,
-        private _paymentHumanVerificationHandler: PaymentHumanVerificationHandler,
-        private _ppsdkStepHandler: StepHandler
+        private _paymentHumanVerificationHandler: PaymentHumanVerificationHandler
     ) {
         const { onBlur = noop, onCardTypeChange = noop, onFocus = noop, onValidate = noop } = this._eventCallbacks;
 
@@ -36,8 +34,6 @@ export default class HostedForm {
 
         this._eventListener.addListener(HostedInputEventType.CardTypeChanged, ({ payload }) => this._cardType = payload.cardType);
         this._eventListener.addListener(HostedInputEventType.BinChanged, ({ payload }) => this._bin = payload.bin);
-
-        this._eventListener.addListener(HostedInputEventType.SubmitSucceeded, this._handleSubmitSuccess);
     }
 
     getBin(): string | undefined {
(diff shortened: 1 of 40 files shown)


packages/core/src/spam-protection/payment-human-verification-handler.ts
```diff
@@ -12,13 +12,39 @@ export default class PaymentHumanVerificationHandler {
         private _googleRecaptcha: GoogleRecaptcha
     ) {}
 
-    async handle(error: Error): Promise<PaymentAdditionalAction> {
+    async handle(errorOrId: Error | string, key?: string): Promise<PaymentAdditionalAction> {
```
davidchin, line 15: Maybe you can overload the method instead so you get a better type definition?

```ts
async handle(error: Error): Promise<PaymentAdditionalAction>;
async handle(id: string, key: string): Promise<PaymentAdditionalAction>;
async handle(errorOrId: Error | string, key?: string): Promise<PaymentAdditionalAction> {
    // ...
}
```

packages/core/src/hosted-form/hosted-form.ts
```diff
@@ -68,14 +67,24 @@ export default class HostedForm {
...
                 this._fields.map(field => field.getType()),
                 this._payloadTransformer.transform(payload)
             );
+
+            const { response, data } = eventPayload;
+
+            if (data.paymentMethod?.type === PaymentStrategyType.PPSDK) {
+                await this._ppsdkStepHandler.handle(
+                    response,
+                    this._paymentHumanVerificationHandler,
```
davidchin, line 80: Avoid injecting this via `StepHandler#handle` method as you want its parameters to apply to all kinds of handler, not just human verification.

packages/core/src/payment/strategies/ppsdk/step-handler/step-handler.ts
```diff
@@ -11,15 +12,15 @@ export class StepHandler {
...
+    handle(response: PaymentsAPIResponse, callback?: (additionalAction: PaymentAdditionalAction) => Promise<void>): Promise<void> {
         const { body } = response;
 
         if (isSuccess(body)) {
             return handleSuccess();
         }
 
         if (isContinue(body)) {
-            return this._continueHandler.handle(body);
+            return this._continueHandler.handle(body, callback);
```
davidchin, line 23: `callback` needs to be triggered for all type of responses, not just `continue`. Otherwise, we should specify that this is a callback for a specific step, i.e.: `continueCallback`.

packages/core/src/payment/strategies/ppsdk/step-handler/continue-handler/continue-handler.ts
```diff
@@ -1,26 +1,31 @@
...
 
 export class ContinueHandler {
     constructor(
-        private _formPoster: FormPoster
+        private _formPoster: FormPoster,
+        private _humanVerificationHandler: PaymentHumanVerificationHandler
     ) {}
 
-    handle(body: Continue): Promise<void> {
+    handle(body: Continue, callback?: (additionalAction: PaymentAdditionalAction) => Promise<void>): Promise<void> {
```
davidchin, line 23: Again, this `callback` needs to apply to all types of `continue` responses, not just `resubmit_with_human_verification`. If not, we should make it a specific callback. Otherwise, the caller might wonder why the `callback` doesn't get trigger for other continue response types, such as `redirect`.

davidchin (approved):
Nice work @Tharaae, LGTM!

PR #1461, opened 2022-06-03
feat(payment): PAYPAL-0000 updated braintree paypal checkout button strategy to prepare strategies separation

serhii-tkachenko:
## What?
- updated braintree paypal checkout button strategy;
- added some updates to braintree checkout sdk;
- updated braintree types and interfaces;
- cleaned some code and fixed some tests;

## Why?
To prepare braintree checkout buttons strategies separation.

## Testing / Proof
Unit tests
Manual tests

US:
Screenshot 2022-06-06 at 12 18 42

Other countries:
Screenshot 2022-06-06 at 12 16 18


packages/core/src/checkout-buttons/strategies/braintree/get-valid-button-style.ts
```diff
@@ -0,0 +1,32 @@
...
+    };
+
+    return omitBy(validStyles, isNil);
+}
+
+function validateHeight(height?: number): number | undefined {
+    const minHeight = 25;
+    const maxHeight = 55;
+
+    if (typeof height !== 'number') {
```
animesh1987, line 25: Can we just return maxHeight over here?

packages/core/src/checkout-buttons/strategies/braintree/get-valid-button-style.ts
```diff
@@ -0,0 +1,32 @@
...
+        return undefined;
+    }
+
+    return (height < minHeight && minHeight)
```
animesh1987, line 29: Can we use if/else or switch here.

packages/core/src/checkout-buttons/strategies/braintree/map-to-legacy-shipping-address.ts
```diff
@@ -0,0 +1,22 @@
+import { LegacyAddress } from '../../../address';
+import { BraintreeDetails } from '../../../payment/strategies/braintree';
+
+export default function mapToLegacyShippingAddress(details: BraintreeDetails): Partial<LegacyAddress> {
+    const { email, phone, shippingAddress } = details;
+
+    const recipientName = shippingAddress?.recipientName || '';
```
animesh1987, line 7: Can we add a check and not use empty string as a fallback?

packages/core/src/checkout-buttons/strategies/braintree/get-valid-button-style.ts
serhii-tkachenko: Yes, it's a bit tricky to read

packages/core/src/checkout-buttons/strategies/braintree/get-valid-button-style.ts
serhii-tkachenko: I think we can

animesh1987 (approved):

PR #1463, opened 2022-06-06
feat(payment): PAYPAL-1381 added braintreevenmo checkout button strategy

serhii-tkachenko:
## What?
Added braintreevenmo checkout button strategy

## Why?
To separate braintree venmo functionality from braintreepaypal strategy to its own.

## Testing / Proof
Manual tests
Unit tests 

Screenshot 2022-05-12 at 12 55 08

Screenshot 2022-05-12 at 12 55 16

Screenshot 2022-05-12 at 12 55 32

### DIFF
--- a/packages/core/src/checkout-buttons/create-checkout-button-registry.ts
+++ b/packages/core/src/checkout-buttons/create-checkout-button-registry.ts
@@ -24,7 +24,7 @@ import { SubscriptionsActionCreator, SubscriptionsRequestSender } from '../subsc
 import { CheckoutButtonMethodType, CheckoutButtonStrategy } from './strategies';
 import { AmazonPayV2ButtonStrategy } from './strategies/amazon-pay-v2';
 import { ApplePayButtonStrategy } from './strategies/apple-pay';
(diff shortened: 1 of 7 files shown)


packages/core/src/checkout-buttons/create-checkout-button-registry.ts
```diff
@@ -117,6 +117,15 @@ export default function createCheckoutButtonRegistry(
         )
     );
 
+    registry.register(CheckoutButtonMethodType.BRAINTREE_VENMO, () =>
+        new BraintreeVenmoButtonStrategy(
+            store,
+            paymentMethodActionCreator,
+            new BraintreeSDKCreator(new BraintreeScriptLoader(scriptLoader)),
```
animesh1987, line 124: Can we move this to a variable, since I can see this being the same for other strategies.

serhii-tkachenko: Yes we can. It is a great idea

animesh1987 (approved):

PR #1466, opened 2022-06-07
chore(common): CHECKOUT-6377 Resolve circular dependencies and enable rule to enforce it

animesh1987:
## What?
- Add a common types package
- Remove payment-integration package and move types to types package, and implementation to core package

## Why?
Due to running into an issue with cyclic dependency. We want to move common-types to a separate package. So going forward we can have the following structure
- core -> payment-strategy (apple pay) -> common (to be named payment-integration)
- core -> common (to be named payment-integration)

We will eventually refactor to move all implementations into their own package and make the core package really light.

## Future Implementation
Remove imports for above types in core package, and start importing from common-types package.

## Testing / Proof
- Circle

@bigcommerce/checkout


davidchin:
@animesh1987 thanks for exploring this option! After thinking about this problem a bit more, although extracting TS interfaces, shared between `core` and `payment-integration`, into a separate package could work, perhaps a better option is to actually break the dependency arrow so that `payment-integration` no longer points to `core`?

For example:

```mermaid
flowchart LR
    core --> payment-integration
    core --> apple-pay --> payment-integration
```

To do that, we will need to duplicate the common types across to `payment-integration` package so they are no longer shared. It'd look something like the following:

In `payment-integration` package
```ts
import Checkout from './checkout';
import OrderRequestBody from './order-request-body';

export interface PaymentIntegrationSelectors {
    // ...
    getCheckout(): Checkout | undefined;
}

export interface PaymentIntegrationService {
    // ...
    submitOrder(payload?: OrderRequestBody): Promise<PaymentIntegrationSelectors>;
}
```

In `core` package
```ts
import { PaymentIntegrationService } from '@bigcommerce/checkout-sdk/payment-integration';
import { createApplePayPaymentStrategy } from '@bigcommerce/checkout-sdk/apple-pay';

import PaymentStrategyRegistry from './payment-strategy-registry';
import DefaultPaymentIntegrationService from './default-payment-integration-service';

export function createPaymentIntegrationService(): PaymentIntegrationService {
    return new DefaultPaymentIntegrationService(
        // ...
    );
}

export function createPaymentStrategyRegistry(): PaymentStrategyRegistry {
    const registry = new PaymentStrategyRegistry();
    const paymentIntegrationService = createPaymentIntegrationService();

    registry.register('applepay', () =>
        createApplePayPaymentStrategy(paymentIntegrationService);
    );

    return registry;
}
```

In `apple-pay` package

```ts
import { PaymentIntegrationService } from '@bigcommerce/checkout-sdk/payment-integration';

import ApplePayPaymentStrategy from './apple-pay-payment-strategy';

export function createApplePayPaymentStrategy(paymentIntegrationService: PaymentIntegrationService) {
    return new ApplePayPaymentStrategy(
        paymentIntegrationService
        // ...
    );
}
```

animesh1987:
Thanks for have a look @davidchin, what you have highlighted above is what the goal of this PR is as well. I think the only difference is the changing the name of the package. Is my understanding correct in that regards?

davidchin:
@animesh1987 practically speaking, yes, it would be the caae if we stop at its current form. 😄 I know the original plan is to copy the types over to `common-types`, and then refactor `core` to point to `common-types`, and lastly remove the duplicates from `core`. If we go with my latest suggestion, we'd not proceed with the last two steps.

animesh1987:
Yes I think we are aligned on what you outlined above, hence I updated the package name and references. 
Next step will be to add the apple pay package and see what we learn from that.

davidchin (approved):
LGTM

packages/core/src/payment-integration/create-payment-integration-selectors.spec.ts
```diff
@@ -14,9 +14,11 @@ describe('createPaymentIntegrationSelectors', () => {
         beforeEach(() => {
             internalSelectors = createInternalCheckoutSelectors(getCheckoutStoreStateWithOrder());
             subject = createPaymentIntegrationSelectors(internalSelectors);
+            console.log('This runs', subject.getBillingAddress);
         });
 
-        it('returns copy of billing address', () => {
+        it.only('returns copy of billing address', () => {
+            console.log('This runs second', subject);
```
bc-peng, line 21: Learned a lot from both of you, @animesh1987 and @davidchin , on this PR. 🙇🏻
This looks like some midway work, is it intended to be?

davidchin: No, I don't think it is intentional. Good catch! I'll update now.

bc-peng (approved):

packages/core/src/payment-integration/create-payment-integration-selectors.spec.ts
animesh1987: Nice catch @bc-peng

PR #1482, opened 2022-06-29
chore(common): CHECKOUT-6377 Setup apple pay package

animesh1987:
## What?
Setup apple pay package and move apple pay payment strategy

## Why?
As part of monorepo setup move apple pay to a separate package.

## Testing / Proof
- circle

@bigcommerce/checkout

### DIFF
--- a/packages/apple-pay/src/create-apple-pay-payment-strategy.spec.ts
+++ b/packages/apple-pay/src/create-apple-pay-payment-strategy.spec.ts
@@ -1,7 +1,7 @@
 import { PaymentIntegrationService, PaymentIntegrationServiceMock } from "@bigcommerce/checkout-sdk/payment-integration";
 import { RequestSender, createRequestSender } from "@bigcommerce/request-sender";
-import { createApplePayPaymentStrategy } from "./src";
-import ApplePayPaymentStrategy from "./src/apple-pay-payment-strategy";
+import { createApplePayPaymentStrategy } from ".";
+import ApplePayPaymentStrategy from "./apple-pay-payment-strategy";
 
 describe('createApplePayPaymentStrategy', () => {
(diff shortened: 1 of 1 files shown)


packages/apple-pay/create-apple-pay-payment-strategy.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { PaymentIntegrationService, PaymentIntegrationServiceMock } from "@bigcommerce/checkout-sdk/payment-integration";
```
davidchin, line 1: Should this file live inside `src` next to the file it is testing against?

packages/payment-integration/src/test-utils/line-items.mock.ts
```diff
@@ -0,0 +1,94 @@
+import { DigitalItem, GiftCertificateItem, PhysicalItem } from "../cart/line-item";
```
davidchin, line 1: We should be using single quotes, but I believe it'll be once we apply `@bigcommerce/eslint-config` to the project.

animesh1987: Yes agreed, I think this is my IDE doing things.

packages/apple-pay/create-apple-pay-payment-strategy.spec.ts
animesh1987: Good catch. It was created there by accident. Updated

packages/payment-integration/src/test-utils/line-items.mock.ts
animesh1987: This will move it as part of the `@bigcommerce/eslint-config` migration

davidchin (approved):

PR #1485, opened 2022-07-01
feat(payment): PAYPAL-1383 added PayPalCommerce Venmo button strategy

serhii-tkachenko:
## What?
Added PayPalCommerce Venmo button strategy

## Why?
It is a part of PayPalCommerce button strategy separation what we need to initialise and render PayPalCommerce buttons separately in different containers.

## Testing / Proof
Unit tests

### DIFF
--- a/packages/core/src/checkout-buttons/checkout-button-options.ts
+++ b/packages/core/src/checkout-buttons/checkout-button-options.ts
@@ -6,7 +6,7 @@ import { ApplePayButtonInitializeOptions } from './strategies/apple-pay';
 import { BraintreePaypalButtonInitializeOptions, BraintreePaypalCreditButtonInitializeOptions, BraintreeVenmoButtonInitializeOptions } from './strategies/braintree';
 import { GooglePayButtonInitializeOptions } from './strategies/googlepay';
 import { PaypalButtonInitializeOptions } from './strategies/paypal';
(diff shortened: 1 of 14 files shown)


packages/core/src/checkout-buttons/checkout-button-options.ts
```diff
@@ -119,4 +107,22 @@ export interface CheckoutButtonInitializeOptions extends CheckoutButtonOptions {
...
+     * The options that are required to facilitate PayPal. They can be omitted
+     * unless you need to support Paypal.
+     */
+    paypal?: PaypalButtonInitializeOptions;
+
+    /**
+     * The options that are required to facilitate PayPal Commerce. They can be omitted
+     * unless you need to support Paypal.
+     */
+    paypalCommerce?: PaypalCommerceButtonInitializeOptions;
```
animesh1987, line 121:
```suggestion
    paypalcommerce?: PaypalCommerceButtonInitializeOptions;
```

animesh1987 (approved):

packages/core/src/checkout-buttons/checkout-button-options.ts
serhii-tkachenko: It will be changed in one of the future pr

PR #1486, opened 2022-07-03
chore(common): CHECKOUT-6377 Move apple pay customer strategy

animesh1987:
## What?
As above

## Why?
Add customer wallet button strategy.

## Testing / Proof
- Circle

@bigcommerce/checkout


packages/apple-pay/src/apple-pay-session-factory.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+describe('apple pay session factory', () => {
+    it('throws an error if apple pay window object is not present', () => {
+        try {
+            assertApplePayWindow(window);
+        } catch (err) {
+            expect((err as Error).message).toBe('Apple pay is not supported');
+        }
+    });
+
+    it('', () => {
```
davidchin, line 13: Interesting test description 🤔

packages/core/src/payment-integration/default-payment-integration-service.ts
```diff
@@ -96,4 +100,20 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
 
         return this._storeProjection.getState();
     }
+
+    async signOut(methodId: string, options?: RequestOptions): Promise<PaymentIntegrationSelectors> {
```
davidchin, line 104: I would prefer not to expose any `/remote-checkout/*` as it is private and maintained by integration team. These endpoints can be called directly as they don't result in states that need to be managed via `CheckoutStore`.

packages/payment-integration/src/index.ts
```diff
@@ -1,17 +1,23 @@
...
 export { Order, OrderRequestBody } from './order';
-export { CardInstrument, Payment, PaymentInitializeOptions, PaymentMethod, PaymentRequestOptions, PaymentStrategyNew } from './payment';
+export { CardInstrument, Payment, PaymentInitializeOptions, PaymentMethod, PaymentRequestOptions } from './payment';
 export { default as PaymentIntegrationSelectors } from './payment-integration-selectors';
 export { default as PaymentIntegrationService } from './payment-integration-service';
-export { Consignment, ShippingAddress, ShippingAddressRequestBody } from './shipping';
-export { PaymentIntegrationServiceMock, getCart, getCheckout, getConfig, getOrderRequestBody } from './test-utils';
+export { Consignment, ShippingAddress, ShippingOption, ShippingAddressRequestBody } from './shipping';
+export { CustomerWalletButtonStrategy, PaymentStrategyNew } from './strategy';
+export { PaymentIntegrationServiceMock, getCart, getCheckout, getConfig, getConsignment, getOrderRequestBody, getShippingOption } from './test-utils';
```
davidchin, line 22: I just noticed this. I think mock files should be exported via different file. Not sure how smart Webpack is but ideally non-test code shouldn't be referring code written for tests, even though the reference is indirect.

packages/core/src/payment-integration/default-payment-integration-service.ts
animesh1987: So shall we remove this from the service and inject it directly when the strategy is constructed.

packages/payment-integration/src/index.ts
animesh1987: I am not sure if that is even possible, the only way I feel this can be possible is to move the mocks to a separate package?

davidchin (approved):

PR #1492, opened 2022-07-07
chore(common): CHECKOUT-000 Upgrade to typescript version 4.1

animesh1987:
## What?
- Upgrade to typescript 4.1
- Upgrade typedoc and typedoc-plugin-markdown to compatible versions and update docs command

## Why?
As part of working with Nx there are few features that require typescript version 4+. Plus we get better type checks plus the present version is fairly dated. 

## Testing / Proof
- Circle

@bigcommerce/checkout


davidchin (approved):
One small comment. Otherwise LGTM. Thanks for doing this, one step closer to the latest TS version!

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "typescript.tsdk": "node_modules/typescript/lib"
```
bc-peng, line 2: Think we'd better add `.vscode` and `.idea` to SDK's `.gitignore` as well. 🙇🏻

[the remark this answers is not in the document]
animesh1987: Updated as @bc-peng suggestion added them to `.gitignore`

animesh1987: Done

PR #1495, opened 2022-07-10
feat(checkout): CHECKOUT-6277 add eslint config to nx

DeveloperAlan:
## What?
Add `@bigcommerce/eslint-config` to the nx workspace

## Why?
So we can have the same eslint rules applied to all checkout sdk js and tools

## Testing / Proof

Screen Shot 2022-07-11 at 2 16 45 pm

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -24,10 +24,7 @@
                     }
                   ]
                 }
-              ],
-            "@typescript-eslint/naming-convention": "off",
-            "jsdoc/check-tag-names": "off",
-            "@typescript-eslint/default-param-last": "off"
+              ]
           }
         },
         {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,4 +1,7 @@
 {
+    "extends": [
+        "@bigcommerce/eslint-config"
+    ],
     "root": true,
     "ignorePatterns": ["**/*"],
     "plugins": ["@nrwl/nx"],
@@ -21,7 +24,10 @@
                     }
                   ]
                 }
-              ]
+              ],
+            "@typescript-eslint/naming-convention": "off",
+            "jsdoc/check-tag-names": "off",
+            "@typescript-eslint/default-param-last": "off"
           }
         },
         {
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "npm": "6"
   },
   "homepage": "https://github.com/bigcommerce/checkout-sdk-js",
+  "prettier": "@bigcommerce/eslint-config/prettier",
   "scripts": {
     "prepare": "check-node-version --node '>=14.18' --npm '>=6'",
     "build": "npm run bundle && npm run bundle-dts && npm run docs",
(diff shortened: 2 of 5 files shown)


packages/core/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-    "extends": "../../tsconfig.base.json",
+    "extends": "../../tsconfig.json",
```
davidchin, line 2: What's the reason for the rename?

package.json
```diff
@@ -77,10 +78,10 @@
   "devDependencies": {
     "@babel/core": "^7.6.2",
     "@babel/preset-env": "^7.6.2",
-    "@bigcommerce/tslint-config": "^4.0.0",
+    "@bigcommerce/eslint-config": "^1.0.1",
```
davidchin, line 81: Are you going to do the upgrade in a separate PR?

DeveloperAlan: Will do upgrade in a seperate PR since there's about 3000+ errors to fix up

packages/core/tsconfig.json
DeveloperAlan: Forgot to remove it from the v2.6.1 lint upgrade, will change it back

.eslintrc.json
```diff
@@ -21,7 +24,10 @@
                     }
                   ]
                 }
-              ]
+              ],
```
animesh1987, line 27: Can we move these configuration to package level? Or for the particular lines. Don't think we want to turn the rule off completely?

packages/payment-integration/tsconfig.json
```diff
@@ -1,4 +1,4 @@
 {
-    "extends": "../../tsconfig.base.json",
-  }
+  "extends": "../../tsconfig.base.json",
+}
```
animesh1987, line 3: EOL?

.eslintrc.json
```diff
@@ -21,7 +24,10 @@
...
+            "@typescript-eslint/naming-convention": "off",
+            "jsdoc/check-tag-names": "off",
+            "@typescript-eslint/default-param-last": "off"
```
bc-peng, line 30: Rules like `@typescript-eslint/default-param-last` are good for code readability. Could we consider setting it as `warn`?

package.json
```diff
@@ -23,6 +23,7 @@
     "npm": "6"
   },
   "homepage": "https://github.com/bigcommerce/checkout-sdk-js",
+  "prettier": "@bigcommerce/eslint-config/prettier",
```
bc-peng, line 26: Is this line working? I tried it in `checkout-js`, it was like it did not agree with `"@bigcommerce/eslint-config": "^1.0.1"`.

.eslintrc.json
animesh1987: Yes either that, but depends on how many instances there are. If not many better to turn off on specific lines.

package.json
animesh1987: Good catch, could be left over from 2.6.1 testing

DeveloperAlan: yep just leftover from 2.6.1

.eslintrc.json
DeveloperAlan: Sorry these rules were leftovers from the 2.6.1 upgrade, I'll remove these for now

davidchin (approved):

animesh1987 (approved):

PR #1499, opened 2022-07-11
feat(payment): INT-6276 Dispatch the thunk action to retrieve the payment method

clopezh:
## What?
Dispatch the loadPaymentMethod action to get the current state and generate the Klarna Session Token correctly

## Why?
Klarna strategy was affected when the session token is not generated and the session is not updated, so the transaction failed.

This issue was reported [INT-6276](https://bigcommercecloud.atlassian.net/browse/INT-6276)

## Testing / Proof
Manual test on Integrations


vmparra (approved):
LGTM 👍🏻

animesh1987 (approved):

bc-peng (approved):
LGTM, thanks.

PR #1501, opened 2022-07-14
CHECKOUT-6780: Automatically register strategies and surface initialisation options from payment integration packages

davidchin:
## What?
Automatically register strategies exported from payment integration packages and surface their initialisation options through the `CheckoutService` interface.

For example, if the entry point of the `applepay` package (i.e.: `index.ts`), exports a strategy factory function (i.e.: `createApplePayPaymentStrategy`), it will automatically be registered by an instance of `Registry<PaymentStrategy>` inside the `core` package. The factory will be registered against the `id`, `gateway` and/or `type` of a payment method as specified in the `applepay` package.

```ts
export default toResolvableModule(
    createApplePayPaymentStrategy,
    [{ id: 'applepay' }]
);
```

## Why?
So that payment integration packages can be used without manual intervention.

This PR depends on https://github.com/bigcommerce/checkout-sdk-js/pull/1503

## Testing / Proof
Unit


packages/payment-integration/coverage/base.css
```diff
@@ -0,0 +1,224 @@
+body, html {
```
animesh1987, line 1: Shall we add this in .gitignore? Feels like something we don't need to commit.

davidchin: Yes, I didn't realise it is committed. Thanks for spotting this!

packages/core/src/common/registry/resolve-id-registry.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    private _decodeToken(token: string): TToken {
+        return JSON.parse(atob(token));
+    }
+
+    private _resolveToken(token: string, registeredTokens: string[]): string | undefined {
+        const query = this._decodeToken(token);
+        const results: Array<{ token: string; matches: number }> = [];
+
+        for (const [, registeredToken] of Object.entries(registeredTokens)) {
```
animesh1987, line 34: Just out of curiosity is there a reason why we use Object.entries for `registeredTokens` instead of a simple `registeredTokens.forEach` ?

animesh1987 (approved):
Nice work @davidchin. LGTM, Just a minor question.

packages/core/src/common/registry/resolve-id-registry.ts
davidchin: Ah I didn't realise `registeredTokens` is `string[]` so a simple `forEach` should do.

PR #1503, opened 2022-07-15
CHECKOUT-6780: Add Nx plugin to generate source files required for registering strategies and surfacing their initialisation options

davidchin:
## What?
Add `workspace-tools` Nx plugin to generate source files required for registering strategies and surfacing their initialisation options. It has two generators - `auto-export` and `extend-interface` - that can executed using the following commands respectively:
* `npx nx generate @bigcommerce/checkout-sdk/workspace-tools:auto-export --projectName=core`
* `npx nx generate @bigcommerce/checkout-sdk/workspace-tools:extend-interface --projectName=core`

## Why?
So that it is possible to automatically locate and execute strategies exported from payment integration packages.

## Testing / Proof
Unit


animesh1987 (approved):
Nice work @davidchin

PR #1512, opened 2022-07-20
feat(payment): ADYEN-539 added vaulting card validation

vitalii-koshovyi:
## What?
Added custom Adyen validation for saved card validation, comparison of last 4 symbols entered and saved card.
checkout-js pr: https://github.com/bigcommerce/checkout-js/pull/935
## Why?
Due to the https://bigcommercecloud.atlassian.net/browse/ADYEN-539

## Testing / Proof

https://user-images.githubusercontent.com/79574476/180739964-1cdbb97a-4202-4b7c-a161-944cd86818bc.mov

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/payment/strategies/adyenv2/adyenv2.ts
+++ b/packages/core/src/payment/strategies/adyenv2/adyenv2.ts
@@ -182,9 +182,9 @@ export interface AdyenComponentEvents {
      * Called in case of an invalid card number, invalid expiry date, or
      *  incomplete field. Called again when errors are cleared.
      */
-    onError?(state: AdyenComponentState, component: AdyenComponent): void;
+    onError?(state: AdyenV2ValidationState, component: AdyenComponent): void;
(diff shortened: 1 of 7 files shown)


packages/core/src/payment/strategies/adyenv2/adyenv2.ts
```diff
@@ -833,6 +833,22 @@ export type AdyenComponentState = (
     CardState | WechatState
 );
 
+export interface AdyenV2ValidationState {
+    valid?: boolean;
```
animesh1987, line 837: Can we have these types as not all optional, since this means `{}` is same as `AdyenV2ValidationState` ?

vitalii-koshovyi: you're right! fixing

animesh1987 (approved):

PR #1522, opened 2022-07-27
feat(payment): INT-6342 fix klarna issue adding klarna v1 payment_method

jmontesbc:
## What?
I switched to use PaymentMethodRequestSender.loadPaymentMethod function instead paymentMethodActionCreator.loadPaymentMethod(gateway_id) to avoid the klarnav1 payment method adding.

## Why?
we noticed that when we use loadPaymentMethod(gateway_id) function, load Klarna v1 payment_methods

## Testing / Proof
image

image

### DIFF
--- a/packages/core/src/payment/strategies/klarnav2/klarnav2-payment-strategy.spec.ts
+++ b/packages/core/src/payment/strategies/klarnav2/klarnav2-payment-strategy.spec.ts
@@ -15,12 +15,12 @@ import { getOrderRequestBody } from '../../../order/internal-orders.mock';
 import { RemoteCheckoutActionCreator, RemoteCheckoutActionType, RemoteCheckoutRequestSender } from '../../../remote-checkout';
 import { PaymentMethodCancelledError, PaymentMethodInvalidError } from '../../errors';
 import PaymentMethod from '../../payment-method';
-import PaymentMethodRequestSender from '../../payment-method-request-sender';
 import { getKlarna } from '../../payment-methods.mock';
 
 import KlarnaPayments from './klarna-payments';
 import KlarnaV2PaymentStrategy from './klarnav2-payment-strategy';
 import KlarnaV2ScriptLoader from './klarnav2-script-loader';
+import KlarnaV2TokenUpdater from './klarnav2-token-updater';
 import { getEUBillingAddress, getEUBillingAddressWithNoPhone, getEUShippingAddress, getKlarnaV2UpdateSessionParams, getKlarnaV2UpdateSessionParamsForOC, getKlarnaV2UpdateSessionParamsPhone, getOCBillingAddress } from './klarnav2.mock';
 import { getResponse } from "../../../common/http-request/responses.mock";
 
@@ -38,7 +38,7 @@ describe('KlarnaV2PaymentStrategy', () => {
     let store: CheckoutStore;
     let strategy: KlarnaV2PaymentStrategy;
     let paymentMethodMock: PaymentMethod;
-    let paymentMethodRequestSender: PaymentMethodRequestSender
+    let klarnav2TokenUpdater: KlarnaV2TokenUpdater;
 
     beforeEach(() => {
         paymentMethodMock = { ...getKlarna(), id: 'pay_now', gateway: 'klarna'};
@@ -66,13 +66,13 @@ describe('KlarnaV2PaymentStrategy', () => {
             checkoutActionCreator
         );
         scriptLoader = new KlarnaV2ScriptLoader(createScriptLoader());
-        paymentMethodRequestSender = new PaymentMethodRequestSender(requestSender);
+        klarnav2TokenUpdater = new KlarnaV2TokenUpdater(requestSender);
         strategy = new KlarnaV2PaymentStrategy(
             store,
             orderActionCreator,
             remoteCheckoutActionCreator,
             scriptLoader,
-            paymentMethodRequestSender
+            klarnav2TokenUpdater
         );
 
         klarnaPayments = {
@@ -107,7 +107,7 @@ describe('KlarnaV2PaymentStrategy', () => {
         jest.spyOn(scriptLoader, 'load')
             .mockImplementation(() => Promise.resolve(klarnaPayments));
 
-        jest.spyOn(paymentMethodRequestSender, 'loadPaymentMethod')
+        jest.spyOn(klarnav2TokenUpdater, 'updateClientToken')
             .mockResolvedValue(getResponse(getKlarna()));
 
         jest.spyOn(store, 'subscribe');
@@ -156,7 +156,7 @@ describe('KlarnaV2PaymentStrategy', () => {
             strategy.execute(payload);
             expect(klarnaPayments.authorize)
                 .toHaveBeenCalledWith({ payment_method_category: paymentMethod.id }, getKlarnaV2UpdateSessionParamsPhone(), expect.any(Function));
-            expect(paymentMethodRequestSender.loadPaymentMethod)
+            expect(klarnav2TokenUpdater.updateClientToken)
                 .toHaveBeenCalledWith(paymentMethod.gateway, { params: { params: 'b20deef40f9699e48671bbc3fef6ca44dc80e3c7'} });
         });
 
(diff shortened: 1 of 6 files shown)


BC-ISilva:
💚 Tested:
QE Details:
PR has been validated, following scenarios have been tested.

1. Validate none extra Klarna Payment method are displaying on Klarna Drop in component.
2. Place order with Klarna Pay Later.
3. Place order with Klarna Pay Now.
4. Place order with Klarna Slice it.

Video from validation have been uploaded into following google drive folder:
https://drive.google.com/drive/folders/1B97C5Ma8VHrMNvXPCI1PoUNpaQeczddC

packages/core/src/payment/strategies/klarnav2/klarnav2-payment-strategy.spec.ts
```diff
@@ -95,7 +92,7 @@ describe('KlarnaV2PaymentStrategy', () => {
             useStoreCredit: true,
         });
 
-        loadPaymentMethodAction = of(createAction(PaymentMethodActionType.LoadPaymentMethodSucceeded, paymentMethod, { methodId: paymentMethod.id }));
+        // loadPaymentMethodAction = of(createAction(PaymentMethodActionType.LoadPaymentMethodSucceeded, paymentMethod, { methodId: paymentMethod.id }));
```
clopezh, line 95: Remove this line please

packages/core/src/payment/strategies/klarnav2/klarnav2-payment-strategy.spec.ts
```diff
@@ -69,12 +65,13 @@ describe('KlarnaV2PaymentStrategy', () => {
             checkoutActionCreator
         );
         scriptLoader = new KlarnaV2ScriptLoader(createScriptLoader());
+        paymentMethodRequestSender = new PaymentMethodRequestSender(requestSender);
```
clopezh, line 68: Blank line after this line 💅🏻

packages/core/src/payment/strategies/klarnav2/klarnav2-payment-strategy.spec.ts
```diff
@@ -159,7 +153,7 @@ describe('KlarnaV2PaymentStrategy', () => {
             strategy.execute(payload);
             expect(klarnaPayments.authorize)
                 .toHaveBeenCalledWith({ payment_method_category: paymentMethod.id }, getKlarnaV2UpdateSessionParamsPhone(), expect.any(Function));
-            expect(paymentMethodActionCreator.loadPaymentMethod).toHaveBeenCalledWith(paymentMethod.gateway);
+            // expect(paymentMethodActionCreator.loadPaymentMethod).toHaveBeenCalledWith(paymentMethod.gateway);
```
clopezh, line 156: Remove this line

packages/core/src/payment/strategies/klarnav2/klarnav2-payment-strategy.spec.ts
```diff
@@ -159,7 +153,7 @@ describe('KlarnaV2PaymentStrategy', () => {
             strategy.execute(payload);
             expect(klarnaPayments.authorize)
                 .toHaveBeenCalledWith({ payment_method_category: paymentMethod.id }, getKlarnaV2UpdateSessionParamsPhone(), expect.any(Function));
-            expect(paymentMethodActionCreator.loadPaymentMethod).toHaveBeenCalledWith(paymentMethod.gateway);
```
clopezh, line 162: Because you replace this method with loadPaymentMethod from PaymentMethodRequestSender, the expectation is to verify if that method have been called, you also need to create the mock.

packages/core/src/payment/strategies/klarnav2/klarnav2-payment-strategy.spec.ts
```diff
@@ -104,9 +101,6 @@ describe('KlarnaV2PaymentStrategy', () => {
         jest.spyOn(orderActionCreator, 'submitOrder')
             .mockReturnValue(submitOrderAction);
 
-        jest.spyOn(paymentMethodActionCreator, 'loadPaymentMethod')
-            .mockReturnValue(loadPaymentMethodAction);
-
```
clopezh, line 109: As I said in line: 162 you must create the mock for `PaymentMethodRequestSender.loadPaymentMethod`

clopezh (changes requested):
Hello @jmontesbc great work!! I have some comments for you, Could you please take a look and if you have any question please let me know.

jmontesbc:
hey @clopezh!! thanks for reviewing my pr!

I have already worked on your comments, sorry for the commented lines, everything it's ok now. 
Could you please review it again?

Please let me know if there's anything else that I can help with.

packages/core/src/payment/strategies/klarnav2/klarnav2-payment-strategy.ts
```diff
@@ -103,7 +103,14 @@ export default class KlarnaV2PaymentStrategy implements PaymentStrategy {
             throw new InvalidArgumentError('Unable to proceed because "payload.payment.gatewayId" argument is not provided.');
         }
 
-        const state = await this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(gatewayId));
+        const state = this._store.getState();
+        const cartId = state.cart.getCart()?.id;
```
clopezh, line 107: Instead of doing null check please validate with an if

clopezh (changes requested):
Hi @jmontesbc great job on the previous comments! I only have 1 more comment :D

clopezh (approved):
LGTM 👍🏻

animesh1987 (commented):
Hey @jmontesbc can you please explain why we need to inject `_paymentMethodrequestSender` into a payment strategy? Since we should not be injecting that in payment providers. 
Going forward the request sender won't this is the integration service that all payment strategies need to conform to https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/payment-integration/default-payment-integration-service.ts. Which will become harder for this change to align with. 

Also from the change it seems that we are pretty much only catching if the promise to `loadPaymentMethod` fails? Which should be the case in the current implementation as well.

clopezh:
Hello @animesh1987 
I understand your point, and this has an explanation why we're using `_paymentMethodRequestSender` instead of `_paymentMethodActionCreator`: 
Klarna has two "versions" inside the same module, Klarna V1 and Klarna V2:
The main differences are the way Klarna render the payment options:

- Klarna V1: 

Shows only 1 option and inside of it has all the options available.
The payload for this payment method is something like this: 
```
{
    "id": "klarna",
    "gateway": null,
    ...etc
}
```

- Klarna V2

Shows all the available options as a separate ones.
The payload for this options is something like this:
```
[
    {
        "id": "paynow",
        "gateway": "klarna",
        ...etc
    },
    {
        "id": "paylater",
        "gateway": "klarna",
        ...etc
    },
    {
        "id": "payovertime",
        "gateway": "klarna",
        ...etc
    },
]
```

So when I call `this._paymentMethodActionCreator.loadPaymentMethod("klarna")`
inside of the logic of this method, it validates if that payment method exists into the `state.paymentMethods`, and if we're using Klarna V2, the answer is NO, because the id for V1 is "klarna" and for V2 can be "paynow". That's why the array changed and added the new option on the fly.

To fix this and because we need to call `/api/storefront/payments/${methodId}` this endpoint to update the existing clientToken (and session on Klarna), we use directly `_paymentMethodRequestSender`, so we avoid to add the extra option and we refresh the token without impact to the user experience.

Animesh if you identify another way to fix this we're open to change the approach, but for now this is the path we found to fix the problem.

cc. @bc-erich @jmontesbc

animesh1987:
Hey @clopezh so to understand it correct the goal of this call is to update the existing client token?

Also the reason we are calling the requestSender directly is because methodId `klarna` fetches v1 payment method? 

Is there a way we can load the relevant method by passing in options same as 
`this._paymentMethodActionCreator.loadPaymentMethod(gatewayId, { ...options, params: { ...options?.params, method: methodId } })` ?

clopezh:
Hi @animesh1987 I tried your suggestion but this doesn't work: 
image

image

image

So I think this is not only checkout-sdk-js work but also bcapp and this impacts directly core code, and it implies more time and affectation in the deadline for this fix which is affecting several stores.

So my question is, Can we go with this solution to fix the stores and also add this topic for our sync calls and start the discussions to create an approach to make this available through `_paymentMethodActionCreator` and with a better analysis.

What do you think?

cc. @bc-erich

clopezh:
@animesh1987 now is ready for you review again :D 🚀

animesh1987 (approved):

PR #1528, opened 2022-07-28
CHECKOUT-6781: Generate new enum from existing enums

davidchin:
## What?
Add a build task that generates new enum by combining multiple enums into one.

## Why?
Currently we have `CheckoutButtonMethodType` exported as a public enum. This enum consists of various payment method IDs that support wallet buttons. As we're now moving towards the monorepo structure, we want each payment integration package to define what those method IDs are so the `core` package doesn't need to know about them. This build task can help achieve this goal. For example:

If `apple-pay` package exports the following enum
```ts
enum ApplePayButtonMethodType {
    APPLEPAY = "applepay",
}
```

And if `paypal-commerce` package exports the following enum
```ts
enum PayPalCommerceButtonMethodType {
    PAYPALCOMMERCE = "paypalcommerce",
    PAYPALCOMMERCE_APMS = "paypalcommercealternativemethods",
    PAYPALCOMMERCE_VENMO = "paypalcommercevenmo",
}
```

The task can combine the two into one, which can then be exported for public clients to consume.
```ts
enum CheckoutButtonMethodType {
    APPLEPAY = "applepay",
    PAYPALCOMMERCE = "paypalcommerce",
    PAYPALCOMMERCE_APMS = "paypalcommercealternativemethods",
    PAYPALCOMMERCE_VENMO = "paypalcommercevenmo",
}
```

## Testing / Proof
CircleCI


vitalii-koshovyi (approved):

packages/workspace-tools/src/generators/create-enum/templates/__outputName__
```diff
@@ -0,0 +1 @@
+<%- content %>
```
animesh1987, line 1: Do we need this?

davidchin: Yes, Nx needs an `ejs` template in order to generate files 🤷‍♂️

animesh1987 (approved):
Nice work @davidchin 👏

PR #1529, opened 2022-07-29
chore(common): CHECKOUT-6855 Load payment strategies from V2 as priority

animesh1987:
## What?
- Load payment methods from automatic v2 strategy registry as priority and use v1 as fallback
- Update webpack config to process packages with ts-loader 

## Why?
As we move towards payment strategy packages, we want to register strategies as part of build process. 
Also process packages with ts-loader, we need to have a rule per package for now [issue](https://github.com/TypeStrong/ts-loader/issues/647). 

## Testing / Proof
- tests

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/checkout/create-checkout-service.ts
+++ b/packages/core/src/checkout/create-checkout-service.ts
@@ -11,7 +11,8 @@ import { createCustomerStrategyRegistry, CustomerActionCreator, CustomerRequestS
 import { FormFieldsActionCreator, FormFieldsRequestSender } from '../form';
 import { CountryActionCreator, CountryRequestSender } from '../geography';
 import { OrderActionCreator, OrderRequestSender } from '../order';
-import { createPaymentClient, createPaymentStrategyRegistry, PaymentMethodActionCreator, PaymentMethodRequestSender, PaymentStrategyActionCreator } from '../payment';
+import { createPaymentClient, createPaymentStrategyRegistry, createPaymentStrategyRegistryV2, PaymentMethodActionCreator, PaymentMethodRequestSender, PaymentStrategyActionCreator } from '../payment';
+import { createPaymentIntegrationService } from '../payment-integration';
 import { InstrumentActionCreator, InstrumentRequestSender } from '../payment/instrument';
 import { createShippingStrategyRegistry, ConsignmentActionCreator, ConsignmentRequestSender, PickupOptionActionCreator, PickupOptionRequestSender, ShippingCountryActionCreator, ShippingCountryRequestSender, ShippingStrategyActionCreator } from '../shipping';
 import { SignInEmailActionCreator, SignInEmailRequestSender } from '../signin-email';
@@ -78,6 +79,7 @@ export default function createCheckoutService(options?: CheckoutServiceOptions):
     const subscriptionsActionCreator = new SubscriptionsActionCreator(new SubscriptionsRequestSender(requestSender));
     const formFieldsActionCreator = new FormFieldsActionCreator(new FormFieldsRequestSender(requestSender));
     const checkoutActionCreator = new CheckoutActionCreator(checkoutRequestSender, configActionCreator, formFieldsActionCreator);
+    const paymentIntegrationService = createPaymentIntegrationService(store);
 
     return new CheckoutService(
         store,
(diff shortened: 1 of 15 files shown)


packages/core/src/checkout/create-checkout-service.ts
```diff
@@ -78,6 +81,8 @@ export default function createCheckoutService(options?: CheckoutServiceOptions):
     const subscriptionsActionCreator = new SubscriptionsActionCreator(new SubscriptionsRequestSender(requestSender));
     const formFieldsActionCreator = new FormFieldsActionCreator(new FormFieldsRequestSender(requestSender));
     const checkoutActionCreator = new CheckoutActionCreator(checkoutRequestSender, configActionCreator, formFieldsActionCreator);
+    const paymentIntegrationService = createPaymentIntegrationService(store);
+    const registryV2: ResolveIdRegistry<PaymentStrategy, PaymentStrategyResolveId> = createPaymentStrategyRegistryV2(paymentIntegrationService);
```
davidchin, line 85:
```suggestion
    const registryV2 = createPaymentStrategyRegistryV2(paymentIntegrationService);
```

packages/core/src/payment/payment-strategy-action-creator.ts
```diff
@@ -19,6 +21,7 @@ import { PaymentStrategy } from './strategies';
 export default class PaymentStrategyActionCreator {
     constructor(
         private _strategyRegistry: PaymentStrategyRegistry,
+        private _strategyRegistryV2: ResolveIdRegistry<PaymentStrategyV2, PaymentStrategyResolveId>,
```
davidchin, line 24: You could import `PaymentStrategyRegistry` from `payment-strategy-registry-v2`, which is essentially an alias for `ResolveIdRegistry<PaymentStrategyV2, PaymentStrategyResolveId>`.

packages/core/src/payment/payment-strategy-action-creator.ts
```diff
@@ -46,14 +49,18 @@ export default class PaymentStrategyActionCreator {
...
+                            strategy = this._strategyRegistry.getByMethod(method)
+                        }
                     } else {
                         strategy = this._strategyRegistry.get(PaymentStrategyType.NO_PAYMENT_DATA_REQUIRED);
                     }
 
-                    return strategy
-                        .execute(payload, { ...options, methodId: payment.methodId, gatewayId: payment.gatewayId })
-                        .then(() => createAction(PaymentStrategyActionType.ExecuteSucceeded, undefined, meta));
+                    const promise: Promise<InternalCheckoutSelectors | void> = strategy.execute(payload, { ...options, methodId: payment.methodId, gatewayId: payment.gatewayId });
```
davidchin, line 61: You probably can keep what's there before instead of assigning to a temp variable.

packages/core/src/payment/payment-strategy-action-creator.ts
```diff
@@ -103,11 +118,19 @@ export default class PaymentStrategyActionCreator {
...
 
+            let strategy: PaymentStrategy | PaymentStrategyV2;
+
+            try {
+                strategy = this._strategyRegistryV2.get({ id: method.id });
+            } catch {
+                strategy = this._strategyRegistry.getByMethod(method)
+            }
+
+            const promise: Promise<InternalCheckoutSelectors | void> = strategy.initialize({ ...options, methodId, gatewayId });
```
davidchin, line 129: Likewise here.

webpack-common.config.js
```diff
@@ -44,7 +47,17 @@ async function getBaseConfig() {
                 },
                 {
                     test: /\.[tj]s$/,
-                    include: srcPath,
+                    include: coreSrcPath,
```
davidchin, line 50: Is there a way we can specify the `ts-loader` without having to do it individually for each package?

packages/core/src/payment/payment-strategy-action-creator.ts
animesh1987: I had to do this since typescript was confused about mixed return types since either of the strategy methods had their return types not compatible, so it's essentially setting a type.

webpack-common.config.js
animesh1987: I linked the ts-loader issue in the PR, we can try use tsconfig includes going forward but was not sure if we need to upgrade ts-loader for that. 
I am going to move this config to a new file so we can have all the config object in one place.

Another way is what we discussed about moving builds to each package, but that may require thinking around how to export types.

davidchin (approved):

PR #1540, opened 2022-08-03
chore(common): CHECKOUT-6855 Load customer strategies from V2 as priority

animesh1987:
## What?
Load customer methods from automatic v2 strategy registry as priority and use v1 as fallback

## Why?
As we move towards customer strategy packages, we want to register strategies as part of build process.

## Testing / Proof
- circle

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/customer/customer-request-options.ts
+++ b/packages/core/src/customer/customer-request-options.ts
@@ -32,6 +32,8 @@ export interface CustomerRequestOptions extends RequestOptions {
  * information in order to initialize the customer step of checkout.
  */
 export interface BaseCustomerInitializeOptions extends CustomerRequestOptions {
+    [key: string]: unknown;
+
     /**
      * The options that are required to initialize the customer step of checkout
      * when using Amazon Pay.
--- a/packages/core/src/customer/customer-strategy-action-creator.spec.ts
+++ b/packages/core/src/customer/customer-strategy-action-creator.spec.ts
@@ -11,18 +11,22 @@ import { MutationObserverFactory } from '../common/dom';
 import { Registry } from '../common/registry';
 import { ConfigActionCreator, ConfigRequestSender } from '../config';
 import { FormFieldsActionCreator, FormFieldsRequestSender } from '../form';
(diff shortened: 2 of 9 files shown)


packages/core/src/customer/customer-request-options.ts
```diff
@@ -31,6 +31,8 @@ export interface CustomerRequestOptions extends RequestOptions {
  * information in order to initialize the customer step of checkout.
  */
 export interface CustomerInitializeOptions extends CustomerRequestOptions {
+    [key: string]: unknown;
```
davidchin, line 34: I don't think you need this.

animesh1987: I rebased and this was still needed for the types to match :|

packages/core/src/customer/customer-request-options.ts
```diff
@@ -32,6 +32,8 @@ export interface CustomerRequestOptions extends RequestOptions {
  * information in order to initialize the customer step of checkout.
  */
 export interface BaseCustomerInitializeOptions extends CustomerRequestOptions {
+    [key: string]: unknown;
```
davidchin, line 35: I don't think this line is necessary

packages/core/src/customer/customer-strategy-action-creator.spec.ts
```diff
@@ -42,9 +46,11 @@ describe('CustomerStrategyActionCreator', () => {
         const googleRecaptchaScriptLoader = new GoogleRecaptchaScriptLoader(scriptLoader, mockWindow);
         const mutationObserverFactory = new MutationObserverFactory();
         const googleRecaptcha = new GoogleRecaptcha(googleRecaptchaScriptLoader, mutationObserverFactory);
-
+        
         state = getCheckoutStoreState();
         store = createCheckoutStore(state);
+        const paymentIntegrationService = createPaymentIntegrationService(store);
```
davidchin, line 52: No need for this temp variable? 🍹

davidchin (approved):

PR #1541, opened 2022-08-03
chore(common): CHECKOUT-6855 Load button strategies from V2 as priority

animesh1987:
## What?
Load checkout button methods from automatic v2 strategy registry as priority and use v1 as fallback

## Why?
As we move towards checkout button strategy packages, we want to register strategies as part of build process.

## Testing / Proof
- circle

@bigcommerce/checkout


packages/core/src/checkout-buttons/checkout-button-options.ts
```diff
@@ -19,6 +19,7 @@ export interface CheckoutButtonOptions extends RequestOptions {
 }
 
 export interface CheckoutButtonInitializeOptions extends CheckoutButtonOptions {
+    [ key: string ]: unknown;
```
davidchin, line 22: I don't think you need this after the rebase.

animesh1987: I rebased and this was still needed for the types to match :|

packages/core/src/checkout-buttons/checkout-button-options.ts
```diff
@@ -21,6 +21,8 @@ export interface CheckoutButtonOptions extends RequestOptions {
 }
 
 export interface BaseCheckoutButtonInitializeOptions extends CheckoutButtonOptions {
+    [key: string]: unknown;
```
davidchin, line 24: I don't think this line is necessary

davidchin (approved):

bc-peng (approved):

PR #1553, opened 2022-08-09
chore(common): CHECKOUT-000 Update ts loader for packages in webpack

animesh1987:
## What?
Move webpack package loader definition to a separate file.

## Why?
At the moment for every package we need to add a rule object with a path in webpack config. 

This PR eases that process by updating package list with package name.

## Testing / Proof
- build

### DIFF
--- a/package-loader-rule.js
+++ b/package-loader-rule.js
@@ -1,0 +1,19 @@
+const path = require('path');
+
+const packages = [
+    'core',
+    'apple-pay',
+    'payment-integration',
+];
+
+const getPackagesWithPath = packages.map(package => {
+    const packageSrcPath =  path.join(__dirname, `packages/${package}/src`);
+
+    return {
+        test: /\.[tj]s$/,
+        include: packageSrcPath,
+        loader: 'ts-loader',
+    };
+});
+
+module.exports = getPackagesWithPath;
--- a/package-loader-rule.js
+++ b/package-loader-rule.js
@@ -1,15 +1,10 @@
 const path = require('path');
 const { projects } = require('./workspace.json');
 
-const packages = [
-    'core',
-    'apple-pay',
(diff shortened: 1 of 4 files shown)


package-loader-rule.js
```diff
@@ -0,0 +1,19 @@
+const path = require('path');
+
+const packages = [
+    'core',
+    'apple-pay',
+    'payment-integration',
+];
```
DeveloperAlan, line 7: Would there be a more explicit way for a new package to be added? Seems a bit hidden for a new contributor to add new packages

DeveloperAlan (approved):

package-loader-rule.js
```diff
@@ -0,0 +1,19 @@
...
+
+const getPackagesWithPath = packages.map(package => {
+    const packageSrcPath =  path.join(__dirname, `packages/${package}/src`);
+
+    return {
+        test: /\.[tj]s$/,
+        include: packageSrcPath,
```
davidchin, line 14: I'm thinking, couldn't we just adjust the rule matcher so it matches all the input paths under `packages/*/src` (see https://webpack.js.org/configuration/module/#condition)? That way we don't have to maintain a static list?

animesh1987: Cool I will have a look at this later today.

animesh1987: I had a quick look and seems seems a bit tricky since we use `__dirname` as a prefix, not sure if that causes an issue. 

Also I had another thought where the list can be useful is to create the alias entry in webpack config, so maybe we can keep this change as is? 

This is a temporary change since when we have a package generator that should add the entries automatically.

davidchin (approved):
Nice work @animesh1987, this can be merged once the file is moved under `scripts/webpack` as discussed 🙂

PR #1555, opened 2022-08-11
feat(payment): INT-6407 AmazonPayV2: Add the new `estimatedOrderAmount` parameter

mauricio-sg:
## What? [INT-6407](https://bigcommercecloud.atlassian.net/browse/INT-6407)
1. Add the new `estimatedOrderAmount` parameter.
2. Add some missing tests.

## Why?
1. Because Amazon Pay will use this value to assess transaction risk and prevent buyers from selecting payment methods that can't be used to process the order.
2. To increase code coverage.

## Testing / Proof
Amazon Pay script logs the result:
estimatedOrderAmount

### DIFF
--- a/packages/core/src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-processor.ts
+++ b/packages/core/src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-processor.ts
@@ -64,25 +64,23 @@ export default class AmazonPayV2PaymentProcessor {
 
     private _getAmazonPayV2ButtonOptions(
         {
-            paymentMethods: { getPaymentMethodOrThrow },
-            config: { getStoreConfigOrThrow },
             cart: { getCart },
+            checkout: { getCheckout },
+            config: { getStoreConfigOrThrow },
+            paymentMethods: { getPaymentMethodOrThrow },
         }: InternalCheckoutSelectors,
         methodId: string,
         placement: AmazonPayV2Placement
     ): AmazonPayV2ButtonParameters {
         const {
-            config: {
-                merchantId,
-                testMode,
-            },
+            config: { merchantId, testMode },
             initializationData: {
                 checkoutLanguage,
-                ledgerCurrency,
                 checkoutSessionMethod,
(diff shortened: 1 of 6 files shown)


ggarcia-bc (approved):
LGTM 👍🏼

packages/core/src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-processor.ts
```diff
@@ -123,6 +123,16 @@ export default class AmazonPayV2PaymentProcessor {
                           publicKeyId,
                       },
                   };
+
+            return total
+                ? {
+                      ...buttonOptions,
+                      estimatedOrderAmount: {
+                          amount: total.toString(),
+                          currencyCode,
```
animesh1987, line 132: We we need the cart currency here? Since that's the transactional currency where as the storeConfig is for display purpose.

mauricio-sg: You're right, thanks!

packages/core/src/payment/strategies/amazon-pay-v2/amazon-pay-v2-payment-processor.ts
```diff
@@ -109,7 +107,9 @@ export default class AmazonPayV2PaymentProcessor {
         };
 
         if (features['PROJECT-3483.amazon_pay_ph4']) {
-            return /^(SANDBOX|LIVE)/.test(publicKeyId)
+            const amount = getCheckout()?.outstandingBalance.toString();
```
animesh1987, line 110:
```suggestion
            const amount = getCheckoutOrThrow()?.outstandingBalance.toString();
```

mauricio-sg: Since it's not a strictly required parameter for Amazon, I don't want it to be a blocker to checkout.

animesh1987 (approved):

animesh1987 (approved):

PR #1570, opened 2022-08-18
refactor(payment): PAYPAL-1538 switched PayPalCommerce common button strategy with V2 and removed old implementation

serhii-tkachenko:
## What?
- removed legacy PayPalCommerceButtonStrategy;
- switched PayPalCommerceV2ButtonStrategy to PayPalCommerceButtonStrategy;
- added extra todos to understand what should go next;

## Why?
To clean up deprecated PayPalCommerceCheckoutButtonStrategy.

## Testing / Proof
Unit tests
Manual tests

Screenshot 2022-09-06 at 15 51 57
Screenshot 2022-09-06 at 15 55 33

## Dependency
Should be merged when `PAYPAL-1389.paypal_commerce_button_strategies_update` will be turned on for all stores


packages/core/src/checkout-buttons/checkout-button-options.ts
```diff
@@ -122,13 +122,14 @@ export interface BaseCheckoutButtonInitializeOptions extends CheckoutButtonOptio
      * The options that are required to facilitate PayPal Commerce. They can be omitted
      * unless you need to support Paypal.
      */
+    // TODO: should be removed when PAYPAL-1539 hits Tier3
```
animesh1987, line 125: Is there a JIRA attached to this todo? Also has the change already hit T3?

packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -1,89 +1,121 @@
...
+        private _paypalScriptLoader: PaypalCommerceScriptLoader,
+        private _paypalCommerceRequestSender: PaypalCommerceRequestSender
     ) {}
 
     async initialize(options: CheckoutButtonInitializeOptions): Promise<void> {
-        let state = this._store.getState();
-        const paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(options.methodId);
-        const { initializationData } = paymentMethod;
+        const { paypalcommerce, containerId, methodId } = options;
+        const { initializesOnCheckoutPage, style } = paypalcommerce || {};
```
animesh1987, line 25: Do we need the empty object since https://github.com/bigcommerce/checkout-sdk-js/pull/1570/files#diff-a0e269a19247a2fe2de4543b3db3dc1cc7b26aeb62b056f0c6ecab82d449683cR35 we do validate?

packages/core/src/checkout-buttons/checkout-button-options.ts
serhii-tkachenko: Yes. Here it is:
https://bigcommercecloud.atlassian.net/browse/PAYPAL-1541

packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
serhii-tkachenko: We can avoid using empty object by moving this line under the `paypalcommerce` object check. But if we want to left this line on it's place, the typescript will have an error.

animesh1987 (approved):

PR #1576, opened 2022-08-19
fix(payment): STRIPE-106 Added a catch in case an exception occurs when confirming the payment using 3DS2

BC-SRodriguez:
## What? [STRIPE-106](https://bigcommercecloud.atlassian.net/browse/STRIPE-106) 

Added a catch in case an exception occurs when confirming the payment using 3DS2 and the client token is sent to try to confirm the payment in the backend.

## Why?

Some cards with 3DS2 skip the 3ds verification https://site-admin.stripe.com/guides/strong-customer-authentication#exemptions-to-strong-customer-authentication  which is causing an error to occur on the frontend but the charges are being completed on Stripe's side.

## Testing / Proof

https://user-images.githubusercontent.com/69487174/186690882-66cf367e-260c-4e02-87ba-72b7b43f439e.mov

## Dependency of
https://github.com/bigcommerce/bigpay/pull/5853

### DIFF
--- a/packages/core/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
+++ b/packages/core/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
@@ -489,7 +489,7 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
             const clientSecret = error.body.three_ds_result.token;
             const needsConfirm = false;
             let catchedConfirmError = false;
-            let result = null;
+            let result;
 
             try {
                 result = await this._getStripeJs().confirmCardPayment(clientSecret);
@@ -509,7 +509,7 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
                 throw new Error(result.error.message);
             }
 
-            const { id: token } = catchedConfirmError || result == null ? { id: clientSecret } : result?.paymentIntent || { id: '' };
+            const token = this._getPaymentToken(result?.paymentIntent, clientSecret, catchedConfirmError);
 
             const formattedPayload = {
                 credit_card_token: { token },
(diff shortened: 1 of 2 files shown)


vmparra (approved):
LGTM 👍🏻

packages/core/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -505,6 +509,8 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
                 throw new Error(result.error.message);
             }
 
+            const { id: token } = catchedConfirmError || result == null ? { id: clientSecret } : result?.paymentIntent || { id: '' };
```
animesh1987, line 512:
```suggestion
            const { id: token } = catchedConfirmError || result === null ? { id: clientSecret } : result?.paymentIntent || { id: '' };
```

animesh1987: Also as a result of this statement does it mean we will have a value of token based on conditionals? 

Is it better to move this logic to another function, feels like a bit too much happening in ternary conditions.

BC-SRodriguez:
Hi @animesh1987,
I addressed your comment, could you review it again please? ♻️

packages/core/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -516,13 +522,21 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
...
 
         throw error;
     }
 
+    private _getPaymentToken(paymentIntent: PaymentIntent | undefined, clientSecret: string, catchedConfirmError: boolean): string {
+        if (!paymentIntent || catchedConfirmError) {
+            return clientSecret;
+        }
+
+        return paymentIntent.id || '' ;
```
animesh1987, line 537: Again do we need to return an empty string as a token? Is that a valid value?

BC-SRodriguez: That's right. An empty string is not a valid value.

animesh1987 (approved):

BCHMercado:
💚  Testing done
https://bigcommercecloud.atlassian.net/jira/software/c/projects/STRIPE/boards/622?modal=detail&selectedIssue=STRIPE-106

PR #1577, opened 2022-08-22
fix(payment): PAYPAL-1639 fixed paypal commerce zoid issue

serhii-tkachenko:
## What?
Fixed PayPalCommerce zoid issue.
Zoid issue - is an implementation (or a bug) on PayPal side, what destroys all paypal components on the page if the script was loaded couple times in the same time.

It should help us to avoid a lot of Sentry issues with 'Zoid destroyed all components' message.

## Why?
To avoid an issue when PayPal removes its own components.

## Testing / Proof
Unit tests
Manual tests

### DIFF
--- a/packages/core/src/payment/strategies/paypal-commerce/paypal-commerce-script-loader.ts
+++ b/packages/core/src/payment/strategies/paypal-commerce/paypal-commerce-script-loader.ts
@@ -8,6 +8,7 @@ import { FundingType, PaypalCommerceHostWindow, PaypalCommerceInitializationData
 
 export default class PaypalCommerceScriptLoader {
     private _window: PaypalCommerceHostWindow;
+    private _paypalSdk?: Promise<PaypalCommerceSDK>;
 
     constructor(
         private _scriptLoader: ScriptLoader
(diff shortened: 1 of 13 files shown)


PavlenkoM (approved):

packages/core/src/payment/strategies/paypal-commerce/paypal-commerce-script-loader.ts
```diff
@@ -8,20 +8,29 @@ import { FundingType, PaypalCommerceHostWindow, PaypalCommerceInitializationData
...
     }
 
-    async loadPaypalCommerce(
+    async getPayPalSDK(
         paymentMethod: PaymentMethod<PaypalCommerceInitializationData>,
         currencyCode: string,
         initializesOnCheckoutPage?: boolean,
     ): Promise<PaypalCommerceSDK> {
-        const paypalSdkScriptConfig = this._getPayPalSdkScriptConfigOrThrow(paymentMethod, currencyCode, initializesOnCheckoutPage);
+        if (!this._paypalSdk) {
```
animesh1987, line 24: Just out of curiosity, is this an effort to cache the paypal sdk since it's being used by multiple clients?

serhii-tkachenko: Yes. In this case the method will return the same Promise for all initialized strategies what use PayPal SDK loader

animesh1987 (approved):

PR #1586, opened 2022-08-29
fix(payment): STRIPE-131 Added a catch in case an exception occurs when confirming the payment using 3DS2

BC-SRodriguez:
## What? [STRIPE-131](https://bigcommercecloud.atlassian.net/browse/STRIPE-131) 

Added a catch in case an exception occurs when confirming the payment using 3DS2 and the client token is sent to try to confirm the payment in the backend.

## Why?

Some cards with 3DS2 skip the 3ds verification https://site-admin.stripe.com/guides/strong-customer-authentication#exemptions-to-strong-customer-authentication  which is causing an error to occur on the frontend but the charges are being completed on Stripe's side.

## Testing / Proof

- [Video](https://drive.google.com/file/d/1YrI4J_GK7c6Dpuzt6dk8egn5RPCJzaN7/view?usp=sharing)

## Dependency of

https://github.com/bigcommerce/bigpay/pull/5899

### DIFF
--- a/packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
+++ b/packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
@@ -124,14 +124,23 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
             return await this._executeWithAPM(payment.methodId);
         }
 
-        const { paymentIntent, error } = await this._stripeUPEClient.confirmPayment(this._mapStripePaymentData());
+        let result;
+        let catchedConfirmError = false;
+        const paymentMethod = this._store.getState().paymentMethods.getPaymentMethodOrThrow(methodId);
+        const stripePaymentData = this._mapStripePaymentData();
 
-        if (error) {
-            this._throwDisplayableStripeError(error);
+        try {
+            result = await this._stripeUPEClient.confirmPayment(stripePaymentData);
+        } catch (error) {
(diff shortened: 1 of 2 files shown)


vmparra (approved):
LGTM 👍🏻

packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
```diff
@@ -124,14 +124,23 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
...
+            result = await this._stripeUPEClient.confirmPayment(stripePaymentData);
+        } catch (error) {
+            catchedConfirmError = true;
+        }
 
-        if (error) {
-            this._throwDisplayableStripeError(error);
+        if (result?.error) {
+            this._throwDisplayableStripeError(result.error);
             throw new PaymentMethodFailedError();
```
animesh1987, line 140: Do we still need this?

BC-SRodriguez: Yes. If we receive an error when calling `confirmPayment`, we must throw the exception. Only in case we catch an exception when calling `confirmPayment` would we continue with the flow.

animesh1987 (approved):

BCHMercado:
💚  Testing done !
https://bigcommercecloud.atlassian.net/browse/STRIPE-131

PR #1587, opened 2022-09-02
feat(storefront): STRF-9858 Bodl Service: Checkout Begin and Order Purchased

jairo-bc:
## What?

Added Bodl Events Service, that triggers 2 events: Checkout Begin and Order Purchased 

## Why?

1) Storefront will inject bodl events library (EventEmitter) into DOM, that will an interface to emit events and subscribe on events. This part is for emitting events.
2) had to update tslib, had an issue with VSCode interfaces 
more: https://stackoverflow.com/questions/58329178/the-syntax-requires-an-imported-helper-named-spreadarrays

## Testing / Proof

unit

@bigcommerce/checkout @bigcommerce/payments @bigcommerce/storefront-team

### DIFF
--- a/packages/core/src/bodl/bodl-emitter-service.ts
+++ b/packages/core/src/bodl/bodl-emitter-service.ts
@@ -1,0 +1,152 @@
+import { LineItemMap } from '../cart';
+import { CheckoutSelectors, CheckoutStoreSelector } from '../checkout';
+import { MissingDataError, MissingDataErrorType } from '../common/error/errors';
+
+import BodlService from "./bodl-service";
+import { BodlEventsCheckout, BODLProduct } from './bodl-window';
+
+export default class BodlEmitterService implements BodlService {
+    private _checkoutStarted = false;
+    private state?: CheckoutStoreSelector;
+
+    constructor(
+        private subscribe: (subscriber: (state: CheckoutSelectors) => void) => void,
+        private bodlEvents: BodlEventsCheckout
+    ) {
+        this.subscribe(state => {
+            this.setState(state.data);
+
+            const config = this.state?.getConfig();
+
+            if (!config) {
+                throw new MissingDataError(MissingDataErrorType.MissingCheckoutConfig);
+            }
+        });
+    }
+
+    setState(state: CheckoutStoreSelector) {
+        this.state = state;
+    }
+
+    checkoutBegin(): void {
+        if (this._checkoutStarted) {
+            return;
+        }
+
+        const checkout = this.state?.getCheckout();
+
+        if (!checkout) {
+            return;
+        }
+
+        const {
+            cart: {
+                cartAmount,
+                currency,
+                lineItems,
+                id,
+                coupons
+            },
+        } = checkout;
+
+        this.bodlEvents.emit('create_checkout_begin', {
+            id,
+            currency: currency.code,
+            cart_value: cartAmount,
+            coupon: coupons.map(coupon => coupon.code.toUpperCase()).join(','),
+            line_items: this.getProducts(lineItems, currency.code)
+        });
+
+        this._checkoutStarted = true;
+    }
+
+    orderPurchased(): void {
+        const order = this.state?.getOrder();
+
+        if (!order) {
+            return;
+        }
+
+        const {
+            currency,
+            isComplete,
+            orderId,
+            orderAmount,
+            shippingCostTotal,
+            lineItems,
+            cartId,
+            coupons
+        } = order;
+
+        if (!isComplete) {
+            return;
+        }
+
+        this.bodlEvents.emit('create_order_purchased', {
+            id: cartId,
+            currency: currency.code,
+            transaction_id: orderId,
+            cart_value: orderAmount,
+            coupon: coupons.map(coupon => coupon.code.toUpperCase()).join(','),
+            shipping_cost: shippingCostTotal,
+            line_items: this.getProducts(lineItems, currency.code),
+        });
+    }
+
+    private getProducts(lineItems: LineItemMap, currencyCode: string): BODLProduct[] {
+        const customItems: BODLProduct[] = (lineItems.customItems || []).map(item => ({
+            product_id: item.id,
+            product_sku: item.sku,
+            price: item.listPrice,
+            quantity: item.quantity,
+            product_name: item.name,
+            currency: currencyCode,
+        }));
+
+        const giftCertificateItems: BODLProduct[] = lineItems.giftCertificates.map(item => {
+            return {
+                product_id: item.id,
+                gift_certificate_id: item.id,
+                price: item.amount,
+                product_name: item.name,
+                gift_certificate_name: item.name,
+                gift_certificate_theme: item.theme,
+                quantity: 1,
+                currency: currencyCode,
+            };
+        });
+
+        const physicalAndDigitalItems: BODLProduct[] = [
+            ...lineItems.physicalItems,
+            ...lineItems.digitalItems,
+        ].map(item => {
+            let itemAttributes;
+
+            if (item.options && item.options.length) {
+                itemAttributes = item.options.map(option => `${option.name}:${option.value}`);
+                itemAttributes.sort();
+            }
+
+            return {
+                product_id: item.productId,
+                quantity: item.quantity,
+                product_name: item.name,
+                price: item.salePrice,
+                product_sku: item.sku,
+                variant_id: item.variantId,
+                discount: item.discountAmount,
+                brand_name: item.brand,
+                currency: currencyCode,
+                category_name: item.categoryNames ? item.categoryNames.join(', ') : '',
+            };
+        });
+
+        return [
+            ...customItems,
+            ...physicalAndDigitalItems,
+            ...giftCertificateItems,
+        ];
+    }
+
+}
+
--- a/packages/core/src/bodl/bodl-window.ts
+++ b/packages/core/src/bodl/bodl-window.ts
@@ -1,0 +1,53 @@
+export interface BODLProduct {
+    product_id: string | number;
+    product_name: string;
+    product_sku?: string;
+    variant_id?: number;
+    variant_sku?: string;
+    gift_certificate_id?: string | number;
+    gift_certificate_name?: string;
+    gift_certificate_theme?: string;
+    price: number;
+    sale_price?: number;
+    base_price?: number;
+    retail_price?: number;
+    quantity: number;
+    discount?: number;
+    index?: number;
+    brand_name?: string;
+    category_name?: string;
+    currency?: string;
+}
+
+
+export interface CheckoutBeginData {
+    id: string
+    currency: string,
+    cart_value: number,
+    coupon: string,
+    line_items: BODLProduct[]
+}
+
+
+export interface OrderPurchasedData {
+    id: string
+    currency: string,
+    transaction_id: number,
+    cart_value: number,
+    coupon: string,
+    shipping_cost: number,
+    line_items: BODLProduct[]
+
+}
+
+export interface BodlEventsCheckout { 
+    emit(eventName: string , data: CheckoutBeginData | OrderPurchasedData): boolean;
+}
+
+export interface BodlEvents {
+    checkout: BodlEventsCheckout;
+}
+
+export default interface BodlEventsWindow extends Window {
+    bodlEvents: BodlEvents;
+}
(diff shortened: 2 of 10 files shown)


jairo-bc:
Not sure if this is a flaky test, or is it bug in `master`. Here I tried locally on `master`

Screenshot 2022-09-02 at 18 14 40
Screenshot 2022-09-02 at 18 14 44
Screenshot 2022-09-02 at 18 14 48

animesh1987:
I was able to run all the tests locally and they seem to be passing ok, may I ask what node version are you using?

jairo-bc:
ok, I removed tslib, now locally are failing another steps, but not in CI 🥲

animesh1987:
> ok, I removed tslib, now locally are failing another steps, but not in CI 🥲

You might need to run `npm ci` again, if that doesn't work try clearing `node_modules` folder and run `npm ci` again. That should fix it 😄

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -0,0 +1,187 @@
...
+
+export default class BodlEmitterService implements BodlService {
+    private _checkoutStarted = false;
+
+    constructor(
+        private checkoutService: CheckoutService,
+        private bodlEvents: BodlEventsCheckout
+    ) {
+
+    }
```
animesh1987, line 17:
```suggestion
    constructor(
        private checkoutService: CheckoutService,
        private bodlEvents: BodlEventsCheckout
    ) { }
```

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    checkoutBegin(): void {
+        if (this._checkoutStarted) {
+            return;
+        }
+
+        const checkout = this.getCheckout();
```
animesh1987, line 24:
```suggestion
        const checkout = this.getCheckoutOrThrow();
```

animesh1987: That way you can avoid the if condition in the next line

animesh1987: Also if you just need cart data, does it make sense to just do `const cart = this.getCartOrThrow();`

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -0,0 +1,187 @@
...
+                id,
+                coupons
+            },
+        } = checkout;
+
+        this.bodlEvents.emit('create_checkout_begin', {
+            id,
+            currency: currency.code,
+            cart_value: cartAmount,
+            coupon: (coupons || []).map(coupon => coupon.code.toUpperCase()).join(','),
```
animesh1987, line 45: Do we need to do the `||` here? Since if there are no coupons applied it's value is `[]`

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -0,0 +1,187 @@
...
+
+        this._checkoutStarted = true;
+    }
+
+    orderPurchased(): void {
+        const order = this.getOrder();
+
+        if (!order) {
+            return;
+        }
```
animesh1987, line 57: `const order = this.getOrderOrThrow();`

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -0,0 +1,187 @@
...
+        if (!isComplete) {
+            return;
+        }
+
+        this.bodlEvents.emit('create_order_purchased', {
+            id: cartId,
+            currency: currency.code,
+            transaction_id: orderId,
+            cart_value: orderAmount,
+            coupon: (coupons || []).map(coupon => coupon.code.toUpperCase()).join(','),
```
animesh1987, line 80: Same as above

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -0,0 +1,187 @@
...
+            shipping_cost: shippingCostTotal,
+            line_items: this.getProducts(lineItems, currency.code),
+        });
+    }
+
+    private getOrder(): Order | undefined {
```
animesh1987, line 86: Dont think you need this method

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -0,0 +1,187 @@
...
+        const { data: { getOrder } } = this.checkoutService.getState();
+
+        return getOrder();
+    }
+
+    private getCheckout(): Checkout | undefined {
```
animesh1987, line 92: Same as this

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -0,0 +1,187 @@
...
+        const { data: { getCheckout } } = this.checkoutService.getState();
+
+        return getCheckout();
+    }
+
+    private getProducts(lineItems: LineItemMap, currency: string): BODLProduct[] {
+        const customItems: BODLProduct[] = (lineItems.customItems || []).map(item => ({
```
animesh1987, line 99: Same as above, default value is `[]`

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -0,0 +1,187 @@
...
+        ];
+    }
+
+    private toShopperCurrency(amount: number): number {
+        const { exchangeRate = 1 } = this.getShopperCurrency() || {};
+
+        return Math.round(amount * exchangeRate * 100) / 100;
+    }
+
+    private getShopperCurrency(): ShopperCurrency | undefined {
```
animesh1987, line 159: You should be able to get the shopper currency from cart object, the one in config is used for display purposes.

packages/core/src/bodl/noop-bodl-service.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export default class NoopBodlService implements BodlService {
+    checkoutBegin(): void {
+        return;
+    }
+
+    orderPurchased(): void {
+        return;
+    }
+}
```
animesh1987, line 11: EOL

animesh1987, line 11: EOL

packages/core/src/bodl/is-bodl-enabled.ts
```diff
@@ -0,0 +1,5 @@
+import BodlEventsWindow from './bodl-window';
+
+export function isBodlEnabled(window: Window): window is BodlEventsWindow {
+    return Boolean((window as BodlEventsWindow).bodlEvents);
```
animesh1987, line 4:
```suggestion
        return 'bodlEvents' in window;
```

packages/core/src/bodl/create-bodl-service.ts
```diff
@@ -0,0 +1,40 @@
...
+ * const bodlService = BodlService();
+ * bodlService.checkoutBegin();
+ * 
+ * ```
+ *
+ * @param {CheckoutService} checkoutService - An instance of CheckoutService
+ * @returns an instance of `BodlService`.
+ */
+export default function createBodlService(
+    checkoutService: CheckoutService,
```
animesh1987, line 23: Instead of checkoutService can we pass the store here?

animesh1987: Which injects all this in the background https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/checkout/create-checkout-service.ts

packages/core/src/bodl/bodl-emitter-service.ts
jairo-bc: What this method will do? I don't want to break  the application (throw an error) if there is no available data.

packages/core/src/bodl/bodl-emitter-service.ts
jairo-bc: makes sense

packages/core/src/bodl/bodl-emitter-service.ts
jairo-bc: It can be undefined

packages/core/src/bodl/bodl-emitter-service.ts
jairo-bc: I don't have this method in the class, what kind of implementation are you suggesting here?

packages/core/src/bodl/create-bodl-service.ts
jairo-bc: I'm going to use here in the same way https://github.com/bigcommerce/checkout-js/blob/42ee27d679f81eaabc8b43771b08575525962d4a/packages/core/src/app/checkout/CheckoutApp.tsx#L67
So, checkoutService will be available for me

animesh1987: Ahh let me come back on this tomorrow and see if there is a better way to do this.

packages/core/src/bodl/bodl-emitter-service.ts
jairo-bc: I was using as an example https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/analytics/analytics-step-tracker.ts#L361

There are all prices converted through `shopperCurrency`. Are yo saying I should use `cart.currency `instead?

packages/core/src/bodl/bodl-emitter-service.ts
animesh1987: I was going by the assumption of injecting store here as compared to checkoutService

packages/core/src/bodl/create-bodl-service.ts
animesh1987: Hey @jairo-bc , was able to check with David today and we came with a solution where should be able to pass the state https://github.com/animesh1987/checkout-sdk-js/commit/ba48907aa9448a56a6fa4e41b8915acb83dff32a. This should simplify a lot of code in bodl-emitter-service. 

Please feel free to reach out if you have any questions.

cc @davidchin @bc-peng

packages/core/src/bodl/bodl-emitter-service.ts
animesh1987: Yes, `cart.currency` is better here

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    orderPurchased(): void {
+        const order = this.state?.getOrder();
+
+        if (!order) {
+            return;
+        }
+
+
```
animesh1987, line 71: Extra line?

packages/core/src/bodl/is-bodl-enabled.ts
```diff
@@ -0,0 +1,5 @@
+import BodlEventsWindow from './bodl-window';
+
+export function isBodlEnabled(window: Window): window is BodlEventsWindow {
+        return 'bodlEvents' in window;
```
animesh1987, line 4: indentation seems off, should be 4 spaces

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -0,0 +1,174 @@
...
+    sale_price?: number;
+    base_price?: number;
+    retail_price?: number;
+    quantity: number;
+    discount?: number;
+    index?: number;
+    brand_name?: string;
+    category_name?: string;
+    currency?: string;
+}
```
animesh1987, line 174: EOL

packages/core/src/bodl/bodl-window.ts
```diff
@@ -0,0 +1,14 @@
+export interface BodlEventsCheckout { 
+    emit(eventName: string | symbol, ...args: any[]): boolean;
```
animesh1987, line 2: Does this need to be `any[]` here, given the two events we have above they seem to be an object? Unless I am missing something

jairo-bc: No, it's basically coming from here. https://www.npmjs.com/package/events
But yeah, I'm not using it in other case than object

packages/core/src/bodl/bodl-window.ts
```diff
@@ -0,0 +1,14 @@
...
+    on(eventName: string | symbol, listener: (...args: any[]) => void): this;
```
animesh1987, line 3: Same for this one as well

PR #1594, opened 2022-09-12
chore(common): CHECKOUT-000 Remove apple pay code in core repo

animesh1987:
## What?
- Remove apple pay code from core repo
- Move apple pay window interface and type guard to shared package
- Update gitignore

## Why?
We have moved apple pay code to monorepo for a month now, the code in core is not executed anymore.

## Testing / Proof
- circle


packages/apple-pay-integration/src/apple-pay-session-factory.ts
```diff
@@ -1,10 +1,4 @@
-interface ApplePayWindow extends Window {
-    ApplePaySession: ApplePaySession;
-}
-
-function isApplePayWindow(window: Window): window is ApplePayWindow {
-    return "ApplePaySession" in window;
-}
+import { isApplePayWindow, ApplePayWindow } from '@bigcommerce/checkout-sdk/payment-integration-api';
```
bc-peng, line 1: All look good to be. If it won't be reused for payment methods other than ApplePay, maybe we can leave it in Apple Pay pkg?

animesh1987: It's used here https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/payment/payment-method-action-creator.ts#L69

bc-peng (approved):

packages/payment-integration-api/src/util-types/index.ts
```diff
@@ -1,3 +1,4 @@
-export { default as Omit } from "./omit";
-export { default as RequestOptions } from "./request-options";
-export { default as RequireAtLeastOne } from "./require-at-least-one";
+export { default as isApplePayWindow, ApplePayWindow } from './is-apple-pay-window'
```
davidchin, line 1: `isApplePayWindow` and `ApplePayWindow` should belong to the `apple-pay-integration` package instead. Also, I just noticed that `PaymentMethodActionCreator` in `core` references `isApplePayWindow`. That ideally should be refactored as `core` shouldn't be aware of individual payment methods.

animesh1987: Ideally it should be in `apple-pay-integration`, but since we filter payment methods on whether they are supported or not hence we reference it there. Based on that I decided to move this to a shared type so can be used. 

We can look into the uses, and have another look to see if we can remove that later.

snaderiBC:
Hey @animesh1987 
Instead of `000`, would it make sense to like this PR to ticket that we have? [CHECKOUT-6830](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-6830)

PR #1595, opened 2022-09-13
feat(checkout): CHECKOUT-6860 Use numeric keypad

bc-peng:
## What?
Using a numeric keypad to enter numbers.

## Why?
Speed up the checkout.

## Testing / Proof
Before:

After:
Copied the newly generated html code, then made a static page for testing on mobile. Page source code:
```
Card number:<br>
<input id="card-number" placeholder="" autocomplete="cc-number" aria-label="Credit Card Number" inputmode="numeric" pattern="[0-9]*" value=""><br>
Name:<br>
<input id="card-name" placeholder="" autocomplete="cc-name" aria-label="Name on Card" inputmode="text" value="">
```

https://user-images.githubusercontent.com/88361607/189790937-d5ba8968-97f7-4385-a121-55955c1993e7.mov

### DIFF
--- a/packages/core/src/hosted-form/iframe-content/hosted-input.spec.ts
+++ b/packages/core/src/hosted-form/iframe-content/hosted-input.spec.ts
@@ -122,7 +122,7 @@ describe('HostedInput', () => {
     it('configures input with expected attributes', () => {
         input.attach();
 
-        // tslint:disable-next-line:no-non-null-assertion
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         const element = container.querySelector('input')!;
 
         expect(element.id)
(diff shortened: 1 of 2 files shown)


animesh1987:
Just want to check right now when we add the cc numbers, the formatting is done automatically. Does that still work as expected?

bc-peng:
> Just want to check right now when we add the cc numbers, the formatting is done automatically. Does that still work as expected?

It is working as usual on desktop. I have replaced the SDK in `checkout-js` e2e local testing env.
You can see the formatter is working on Bigpay.

https://user-images.githubusercontent.com/88361607/189805571-3ce365e8-ea4e-4a09-a655-02cf4ba262c2.mov

animesh1987 (approved):

PR #1597, opened 2022-09-13
feat(payment): PAYPAL-1625 added buy now for PayPalCommerceButtonStrategy

serhii-tkachenko:
## What?
...

## Why?
...

## Testing / Proof
...

### DIFF
--- a/packages/core/src/cart/cart-request-body.ts
+++ b/packages/core/src/cart/cart-request-body.ts
@@ -1,0 +1,15 @@
+/**
+ * An object that contains the information required for creating cart.
+ */
+export default interface CartRequestBody {
+    locale?: string; // TODO: do I really need this one?
+    source?: string; // TODO: Update with data type
+    lineItems: Array<{
+        productId: number;
+        quantity: number;
+        optionSelections?: {
+            optionId: number;
+            optionValue: number | string;
+        };
+    }>;
+}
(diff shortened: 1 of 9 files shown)


packages/core/src/cart/cart-request-body.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * An object that contains the information required for creating cart.
+ */
+export default interface CartRequestBody {
```
animesh1987, line 4: Can we call this `BuyNowCartRequestBody`?

packages/core/src/cart/cart-request-body.ts
```diff
@@ -0,0 +1,15 @@
...
+    locale?: string; // TODO: do I really need this one?
+    source?: string; // TODO: Update with data type
```
animesh1987, line 6:
```suggestion
    source: 'BUY_NOW';
```

packages/core/src/cart/cart-request-body.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * An object that contains the information required for creating cart.
+ */
+export default interface CartRequestBody {
+    locale?: string; // TODO: do I really need this one?
```
animesh1987, line 5: maybe not :D

packages/core/src/cart/cart-request-body.ts
serhii-tkachenko: Yes 👍

PR #1603, opened 2022-09-19
feat(payment): BOLT-386 Add BODL analytics tracking events

PavlenkoM:
## What?
Add additional BODL analytics events emitors to checkout.
Checkout-js PR: [https://github.com/bigcommerce/checkout-js/pull/1024](https://github.com/bigcommerce/checkout-js/pull/1024)

## Why?
Because of task: [https://bigcommercecloud.atlassian.net/browse/BOLT-386](https://bigcommercecloud.atlassian.net/browse/BOLT-386)

## Testing / Proof
Manual tested
Screenshot 2022-10-22 at 13 20 39

### DIFF
--- a/packages/core/src/bodl/bodl-emitter-service.ts
+++ b/packages/core/src/bodl/bodl-emitter-service.ts
@@ -1,12 +1,16 @@
+import { AnalyticStepOrder, AnalyticStepType } from './analytics-steps';
 import { LineItemMap } from '../cart';
 import { CheckoutSelectors, CheckoutStoreSelector } from '../checkout';
 import { MissingDataError, MissingDataErrorType } from '../common/error/errors';
 
 import BodlService from "./bodl-service";
-import { BodlEventsCheckout, BODLProduct } from './bodl-window';
+import { BodlEventsCheckout, BodlEventsPayload, BODLProduct } from './bodl-window';
 
 export default class BodlEmitterService implements BodlService {
     private _checkoutStarted = false;
+    private _emailEntryBegan = false;
+    private _shippingOptionsShown = false;
+    private _completedSteps: { [key: string]: boolean } = {};
     private state?: CheckoutStoreSelector;
 
     constructor(
@@ -55,8 +59,8 @@ export default class BodlEmitterService implements BodlService {
             currency: currency.code,
             cart_value: cartAmount,
             coupon: coupons.map(coupon => coupon.code.toUpperCase()).join(','),
-            line_items: this.getProducts(lineItems, currency.code),
-            channel_id: channelId,
+            line_items: this._getProducts(lineItems, currency.code),
+            channel_id: channelId
         });
 
         this._checkoutStarted = true;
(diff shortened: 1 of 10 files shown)


solofeed (approved):
LGTM! 👍 
Just to clarify, you decided not to integrate the bold and use checkout service directly, right?

PavlenkoM:
> 

We decide not to create additional services per each providers (currently for Bolt). And use BODL as single interface for all required events. And payments providers from their side will subscribe to BODL events `window.bodlEvents.checkout.on()` for tracking.

serhii-tkachenko (approved):
Awesome! Great job 👍

packages/core/src/customer/customer-request-options.ts
```diff
@@ -150,5 +150,10 @@ export interface BaseCustomerInitializeOptions extends CustomerRequestOptions {
  *
  */
 export interface ExecutePaymentMethodCheckoutOptions extends CustomerRequestOptions {
+    checkoutPaymentMethodExecuted?(data?: CheckoutPaymentMethodExecutedOptions): void;
```
animesh1987, line 153: Is this a bolt specific propery?

packages/core/src/customer/strategies/bolt/bolt-customer-strategy.spec.ts
```diff
@@ -170,10 +170,14 @@ describe('BoltCustomerStrategy', () => {
             jest.spyOn(boltCheckout, 'hasBoltAccount').mockReturnValue(true);
 
             await strategy.initialize({ methodId: 'bolt' });
-            await strategy.executePaymentMethodCheckout(options);
 
-            expect(boltCheckout.hasBoltAccount).toHaveBeenCalledWith('test@bigcommerce.com');
-            expect(boltCheckout.openCheckout).toHaveBeenCalled();
+            try {
+                /* eslint-disable @typescript-eslint/ban-ts-comment */
```
animesh1987, line 175: why the eslint disable and ignores?

packages/core/src/bodl/bodl-service.ts
```diff
@@ -1,4 +1,16 @@
+import { BodlEventsPayload } from './bodl-window';
+
 export default interface BodlService {
     checkoutBegin(): void;
     orderPurchased(): void;
+    stepCompleted(step?: string): void;
```
animesh1987, line 6: All the methods here seem to be taking an optional parameter? Is that by design?

animesh1987:
Can we please add tests for the new methods in bold-emitter service.

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -148,5 +207,25 @@ export default class BodlEmitterService implements BodlService {
         ];
     }
 
+    private trackCompletedStep(step: AnalyticStepType) {
```
animesh1987, line 210:
```suggestion
    private _trackCompletedStep(step: AnalyticStepType) {
```

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -148,5 +207,25 @@ export default class BodlEmitterService implements BodlService {
...
+        this._completedSteps[step] = true;
+        this.bodlEvents.emit('bodl_checkout_step_completed', { step });
+    }
+
+    private hasStepCompleted(step: AnalyticStepType): boolean {
```
animesh1987, line 215:
```suggestion
    private _hasStepCompleted(step: AnalyticStepType): boolean {
```

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -148,5 +207,25 @@ export default class BodlEmitterService implements BodlService {
...
+        return this._completedSteps[step];
+    }
+
+    private backfillCompletedSteps(currentStep: AnalyticStepType): void {
```
animesh1987, line 219:
```suggestion
    private _backfillCompletedSteps(currentStep: AnalyticStepType): void {
```

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -148,5 +207,25 @@ export default class BodlEmitterService implements BodlService {
...
+        for (const step of AnalyticStepOrder) {
```
animesh1987, line 220: Do we need to use `for...of` here?

packages/core/src/customer/customer-request-options.ts
PavlenkoM: Currently it uses only for Bolt, because only Bolt has such functionality, and request such analytics event. 
If in feature some other payment method will need such callback it can be reused as well.

packages/core/src/customer/strategies/bolt/bolt-customer-strategy.spec.ts
PavlenkoM: you are right in this case no need😅

packages/core/src/bodl/bodl-service.ts
PavlenkoM: Yes, it is by design, because for some cases this options can be not set and received as undefined. We handle this undefined cases in code if it needed.

packages/core/src/bodl/bodl-emitter-service.ts
PavlenkoM: done

packages/core/src/bodl/bodl-emitter-service.ts
PavlenkoM: done

packages/core/src/customer/strategies/bolt/bolt-customer-strategy.ts
```diff
@@ -1,4 +1,4 @@
-import { noop } from 'rxjs';
+// import { noop } from 'rxjs';
```
animesh1987, line 1: Comment?

packages/core/src/bodl/bodl-emitter-service.ts
PavlenkoM: After discussion - will change to `forEach`

packages/core/src/customer/strategies/bolt/bolt-customer-strategy.ts
PavlenkoM: thanks, done

animesh1987 (approved):

PR #1613, opened 2022-09-22
feat(payment): BOLT-282 fixed analytics feature for Bolt Checkout

bc-nick:
## What?
GA doesn't work with Bolt Checkout if we navigate to the confirmation page from any page other than the checkout page.
Added `_setExtraItemsForAnalytics()` method for set up extra data to LocalStorage for GA

## Why?
According to task [BOLT-282](https://bigcommercecloud.atlassian.net/browse/BOLT-282)

## Testing / Proof
Validation is not available in the test environment
Need to check in the production environment

### DIFF
--- a/packages/core/src/analytics/analytics-extra-items-manager.ts
+++ b/packages/core/src/analytics/analytics-extra-items-manager.ts
@@ -1,0 +1,54 @@
+import { LineItemMap } from '../cart';
+import { ExtraItemsData } from './analytics-step-tracker';
+
+const ORDER_ITEMS_STORAGE_KEY = 'ORDER_ITEMS';
+
+export default class AnalyticsExtraItemsManager {
+    constructor(
+        private storage: StorageFallback,
+    ) {}
+
+    saveExtraItemsData(id: string, lineItems: LineItemMap): ExtraItemsData {
+        const data = [
+            ...lineItems.physicalItems,
+            ...lineItems.digitalItems,
+        ].reduce<ExtraItemsData>((result, item) => {
+            result[item.productId] = {
+                brand: item.brand ? item.brand : '',
+                category: item.categoryNames ? item.categoryNames.join(', ') : '',
+            };
+
+            return result;
+        }, {});
+
+        try {
+            this.storage.setItem(this.getStorageKey(id), JSON.stringify(data));
+
+            return data;
+        } catch (err) {
+            return {};
+        }
+    }
+
+    readExtraItemsData(id: string): ExtraItemsData | null {
+        try {
+            const item = this.storage.getItem(this.getStorageKey(id));
+
+            return item ? JSON.parse(item) : null;
+        } catch (err) {
+            return null;
+        }
+    }
+
+    clearExtraItemData(id: string): void {
+        try {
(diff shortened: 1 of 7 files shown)


packages/core/src/payment/strategies/bolt/bolt-payment-strategy.ts
```diff
@@ -351,4 +356,23 @@ export default class BoltPaymentStrategy implements PaymentStrategy {
             throw new PaymentArgumentInvalidError();
         }
     }
+
+    private _setExtraItemsForAnalytics() {
```
solofeed, line 360: Since we already have an access modifier, we can omit underscore

```suggestion
    private setExtraItemsForAnalytics() {
```

solofeed (approved):
LGTM! 🚀

PavlenkoM (approved):

packages/core/src/payment/strategies/bolt/bolt-payment-strategy.ts
```diff
@@ -351,4 +356,23 @@ export default class BoltPaymentStrategy implements PaymentStrategy {
             throw new PaymentArgumentInvalidError();
         }
     }
+
+    private setExtraItemsForAnalytics() {
```
animesh1987, line 360:
```suggestion
    private _setExtraItemsForAnalytics() {
```

solofeed: https://github.com/bigcommerce/checkout-sdk-js/pull/1613#discussion_r980117760 😅

packages/core/src/analytics/analytics-extra-items-manager.ts
```diff
@@ -0,0 +1,54 @@
...
+            ...lineItems.physicalItems,
+            ...lineItems.digitalItems,
+        ].reduce((result, item) => {
+            result[item.productId] = {
+                brand: item.brand ? item.brand : '',
+                category: item.categoryNames ? item.categoryNames.join(', ') : '',
+            };
+
+            return result;
+        }, {} as ExtraItemsData);
```
animesh1987, line 22: Can we not use type casting here? And use typeguards if necessary?

bc-nick: @animesh1987 
We can use another syntax for this
Something like that `reduce<ExtraItemsData>`

I think typeguards are unnecessary here

bc-nick: @animesh1987 changed to `reduce<ExtraItemsData>`

packages/core/src/payment/strategies/bolt/bolt-payment-strategy.ts
animesh1987: @solofeed I get the private modifier above, but in TS/JS it's still not natively supported so we try and keep it consistent by adding the `_`.

animesh1987: There is a new proposal to add `#` to mark properties as private and that will have better checking, but till then we wait

animesh1987 (approved):

PR #1616, opened 2022-09-26
feat(payment): STRIPE-25 Stripe Link State

FranciscoOnt:
## What? [STRIPE-25](https://bigcommercecloud.atlassian.net/browse/STRIPE-25)
- Add `stripeLinkAuthenticated` to the customer's data state and its corresponding actions to update it.
- Also add `| undefined ` to StripeUPE `getStyles` function.

## Why?
- It was suggested to move this state to the `checkout-sdk` from `checkout-js` to avoid adding more states handled by the checkout component.
- `getStyles` function may also  have to return `undefined` on newer implementations of the function due to additional validations.

## Dependency Of
https://github.com/bigcommerce/checkout-js/pull/1017

### Without Link Authenticated
image (9)

image (10)

### With Link Authenticated
image (11)
image (12)

### Without Link Enabled
image (13)


clopezh (approved):
LGTM 👍🏻

animesh1987 (approved):

PR #1618, opened 2022-09-27
fix(payment): INT-6478 SquareV2 - Fail Gracefully when payment provider unavailable

polojimenez:
## What?
Added a try/catch with a reject msg and refactored it to return an error if the provider scripts path fails.

## Why?
Square is having degraded performance on their API and it is causing all checkouts to fail if a store has square enabled.

## Testing / Proof
https://drive.google.com/file/d/1rbZXSOWFunQXrnk07fGcP06YOFA1KV9w/view?usp=sharing

## Linked to this [PR](https://github.com/bigcommerce/checkout-js/pull/1047) of Checkout JS

### DIFF
--- a/packages/core/src/payment/strategies/square/square-script-loader.ts
+++ b/packages/core/src/payment/strategies/square/square-script-loader.ts
@@ -2,7 +2,7 @@ import { ScriptLoader } from '@bigcommerce/script-loader';
 
 import { PaymentMethodClientUnavailableError } from '../../errors';
 
-import { SquareFormOptions, SquareScriptCallBack } from './square-form';
+import SquarePaymentForm, { SquareFormOptions, SquareScriptCallBack } from './square-form';
 import SquareWindow from './square-window';
 
 export default class SquareScriptLoader {
@@ -23,7 +23,7 @@ export default class SquareScriptLoader {
                 '//js.squareupsandbox.com/v2/paymentform' :
                 '//js.squareup.com/v2/paymentform');
 
-        return (options: SquareFormOptions) => {
+        return (options: SquareFormOptions): SquarePaymentForm => {
             if (!this._isSquareWindow(this._window)) {
                 throw new PaymentMethodClientUnavailableError();
             }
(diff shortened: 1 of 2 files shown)


abrahamMartinezBc (approved):
LGMT 👍🏻

rlopz (approved):

BC-AAcevedo:
:green_heart: tested
QA Details: When the payment provider is unavailable, the purchase fails prompting an error message, you cannot try to finish with order using Square again, but you can complete it using another provider, Tested with only Square as a payment provider, with Zip to have an alternative, and with both Hosted forms enabled and disabled

[INT-6478 SquareV2 - Fail Gracefully when payment provider unavailable oct 5.docx](https://github.com/bigcommerce/checkout-sdk-js/files/9728614/INT-6478.SquareV2.-.Fail.Gracefully.when.payment.provider.unavailable.oct.5.docx)

packages/core/src/payment/strategies/square/square-script-loader.ts
```diff
@@ -18,17 +18,20 @@ export default class SquareScriptLoader {
     }
 
     async load(testmode?: boolean): Promise<SquareScriptCallBack> {
-        await this._scriptLoader
+        return this._scriptLoader
```
animesh1987, line 21: Can we try and use async/await here?

polojimenez: Done, I moved the code as it was and threw the error from the strategy.

polojimenez:
Hey @animesh1987 I have addressed your comments do you have chance to take a look again? ♻️

animesh1987 (approved):

BC-AAcevedo:
💚 tested
QA Details: When the payment provider is unavailable, the purchase fails prompting an error message, you cannot try to finish the order using Square again, but you can complete it by using another provider, Tested with only Square as a payment provider, with Zip to have an alternative, and with both Hosted forms enabled and disabled


[INT-6478 SquareV2 - Fail Gracefully when payment provider unavailable oct 14.docx](https://github.com/bigcommerce/checkout-sdk-js/files/9789956/INT-6478.SquareV2.-.Fail.Gracefully.when.payment.provider.unavailable.oct.14.docx)

PR #1620, opened 2022-09-28
feat(payment): PAYPAL-1626 added Buy Now functionality to PayPalCommerceCredit button strategy

serhii-tkachenko:
## What?
Added Buy Now functionality to PayPalCommerceCredit button strategy

## Why?
Due the task: https://bigcommercecloud.atlassian.net/browse/PAYPAL-1626

## Testing / Proof
Unit tests
Manual tests

### DIFF
--- a/packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-credit-button-strategy.ts
+++ b/packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-credit-button-strategy.ts
@@ -1,5 +1,7 @@
 import { FormPoster } from '@bigcommerce/form-poster';
 
+import { CartRequestSender } from '../../../cart';
+import { BuyNowCartCreationError } from '../../../cart/errors';
 import { CheckoutActionCreator, CheckoutStore } from '../../../checkout';
 import { InvalidArgumentError, MissingDataError, MissingDataErrorType, } from '../../../common/error/errors';
 import { PaymentMethodClientUnavailableError } from '../../../payment/errors';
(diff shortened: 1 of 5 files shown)


packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-credit-button-strategy.ts
```diff
@@ -36,20 +40,35 @@ export default class PaypalCommerceCreditButtonStrategy implements CheckoutButto
...
         }
 
-        const state = await this._store.dispatch(this._checkoutActionCreator.loadDefaultCheckout());
-        const currencyCode = state.cart.getCartOrThrow().currency.code;
-        const paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(methodId);
-        this._paypalCommerceSdk = await this._paypalScriptLoader.getPayPalSDK(paymentMethod, currencyCode, initializesOnCheckoutPage);
+        const { buyNowInitializeOptions, currencyCode, initializesOnCheckoutPage, messagingContainerId } = paypalcommercecredit;
 
-        this._renderButton(containerId, methodId, initializesOnCheckoutPage, style);
+        if (buyNowInitializeOptions) {
```
animesh1987, line 45: One question do we need a whole different set of code block? The only difference I see is how currencyCode is passed? Maybe we can have something like

```
const currencyCode = currencyCode ||  cart.currency.code;

 if (!currencyCode) {
                throw new InvalidArgumentError(`Unable to initialize payment because "options.paypalcommercecredit.currencyCode" argument is not provided.`);
 }
 
 // rest          
```

serhii-tkachenko: We should have another block of code due to the state variable. 
In buy now case we should not load default checkout, because we don't have access to checkout-settings, but for default flow we should

serhii-tkachenko: You can check the difference on line 46 and 55

animesh1987 (approved):

PR #1641, opened 2022-10-13
feat(payment): STRIPE-165 V3 Initialize payment for vaulted cards

HadesDX:
## What? [STRIPE-165](https://bigcommercecloud.atlassian.net/browse/STRIPE-165)
Do initialization step for vault cards

## Why?
So level3 data (line items an others) can be added on payment intent creation

## Depends on
https://github.com/bigcommerce/bigpay/pull/6158


packages/core/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -233,9 +234,11 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
     }
 
     private async _executeWithVaulted(payment: OrderPaymentRequestBody, token: string, shouldSetAsDefaultInstrument: boolean | undefined): Promise<InternalCheckoutSelectors> {
+        const paymentMethod = this._store.getState().paymentMethods.getPaymentMethodOrThrow(payment.methodId);
```
vmparra, line 237: I think we could just send this value as argument from calling function.

packages/core/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -95,14 +95,15 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
                 await this._store.dispatch(this._orderActionCreator.submitOrder(order, options));
             }
 
+            const state = await this._store.dispatch(this._paymentMethodActionCreator.loadPaymentMethod(`${gatewayId}`, { params: { method: methodId } }));
```
vmparra, line 98: I think we're already loading the payment method in line #107. Wouldn't we removing one?

vmparra (commented):
Hi @HadesDX ,

There're a couple of comments on regards of this change.

packages/core/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
HadesDX: sure!, pulled up

packages/core/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
HadesDX: we do for non vaulted, by moving it to this line, we get it initialized for vaulted too, so we can have the quote available for the level3 data

packages/core/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
```diff
@@ -95,14 +95,16 @@ export default class StripeV3PaymentStrategy implements PaymentStrategy {
...
+
             if (isVaultedInstrument(paymentData)) {
                 await this._store.dispatch(this._orderActionCreator.submitOrder(order, options));
                 const { instrumentId } = paymentData;
+                const paymentMethod = this._store.getState().paymentMethods.getPaymentMethodOrThrow(payment.methodId);
 
-                return await this._executeWithVaulted(payment, instrumentId, shouldSetAsDefaultInstrument);
+                return await this._executeWithVaulted(payment, instrumentId, shouldSetAsDefaultInstrument, paymentMethod.clientToken);
```
bc-peng, line 105: Do we need to check `paymentMethod.clientToken` is defined since `clientToken` is optional by the type definition?

HadesDX:
@vmparra ♻️

packages/core/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
FranciscoOnt: can't we just pass `gatewayId ` as is?

FranciscoOnt (approved):
LGTM

packages/core/src/payment/strategies/stripev3/stripev3-payment-strategy.ts
HadesDX: @FranciscoOnt  passing it directly would not work because its an optional

bc-peng (approved):
LGTM, cheers.

BCHMercado:
Testing done ! 💚 
https://bigcommercecloud.atlassian.net/jira/software/c/projects/STRIPE/boards/622?modal=detail&selectedIssue=STRIPE-165

PR #1648, opened 2022-10-19
fix(payment): PAYPAL-1725 fixed the issue with passing fake data as customer address from last session in PayPalCommerceInlineCheckoutButtonStrategy

serhii-tkachenko:
## What?
Fixed the issue with passing fake data as customer address from last session in PayPalCommerceInlineCheckoutButtonStrategy by clearing billing and consignment address before creating PayPal order

## Why?
Sometimes user can skip accelerated checkout in an unfinished state. In this case, fake data, that used to operate with shipping options before getting valid data from paypal, saves in checkout state. The user gets prev saved fake data from prev session in current one.

This code fixes this issue.

## Testing / Proof
Manual tests
Unit tests

Before:
Fake data can be found in shipping address section
Screenshot 2022-10-19 at 15 20 48

After:
The user sees prefilled shipping address from user's paypal account
Screenshot 2022-10-19 at 15 21 43


bc-peng (approved):
LGTM, nice work.

PR #1653, opened 2022-10-24
refactor(payment): PAYPAL-1742 removed Fake data implementation due to BE changes

serhii-tkachenko:
## What?
We used some 'Fake' data to fill all required empty fields to make an ability to create paypal order. We removed 'Fake' data implementation in this PR.

## Why?
We found a way how to avoid using 'Fake' data in Checkout SDK.

## Testing / Proof
Unit tests
Manual tests

### DIFF
--- a/packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-inline-checkout-button-strategy.ts
+++ b/packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-inline-checkout-button-strategy.ts
@@ -5,7 +5,7 @@ import { OrderActionCreator } from '../../../order';
 import { PaymentActionCreator } from '../../../payment';
 import { PaymentMethodClientUnavailableError } from '../../../payment/errors';
 import { ConsignmentActionCreator, ShippingOption } from '../../../shipping';
-import { ApproveCallbackActions, ApproveCallbackPayload, CompleteCallbackDataPayload, PaypalCheckoutButtonOptions, PaypalCommerceRequestSender, PaypalCommerceScriptLoader, PaypalCommerceSDK, PayPalOrderAddress, PayPalOrderDetails, ShippingAddressChangeCallbackPayload, ShippingOptionChangeCallbackPayload } from '../../../payment/strategies/paypal-commerce';
+import { ApproveCallbackActions, ApproveCallbackPayload, CompleteCallbackDataPayload, PaypalCheckoutButtonOptions, PaypalCommerceRequestSender, PaypalCommerceScriptLoader, PaypalCommerceSDK, ShippingAddressChangeCallbackPayload, ShippingOptionChangeCallbackPayload } from '../../../payment/strategies/paypal-commerce';
 import { CheckoutButtonInitializeOptions } from '../../checkout-button-options';
 
 import CheckoutButtonStrategy from '../checkout-button-strategy';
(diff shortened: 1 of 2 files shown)


packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-inline-checkout-button-strategy.ts
```diff
@@ -153,21 +151,36 @@ export default class PaypalCommerceInlineCheckoutButtonStrategy implements Check
...
             await this._updateOrder();
         } else {
-            const address = this._getValidAddress(
-                orderDetails.payer.name,
-                orderDetails.payer.email_address,
-                orderDetails.payer.address
-            );
+            const { payer } = orderDetails;
+
+            const address = this._getAddress({
```
animesh1987, line 174: Do we need to care about address_line_2?

serhii-tkachenko: Nope, because PayPal didn't pass the data to fill `address_line_2` on our side

packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-inline-checkout-button-strategy.ts
```diff
@@ -245,25 +258,14 @@ export default class PaypalCommerceInlineCheckoutButtonStrategy implements Check
...
-            lastName: address?.lastName || 'Fake',
-            email: address?.email || 'fake@fake.fake',
+            firstName: address?.firstName || '',
+            lastName: address?.lastName || '',
+            email: address?.email || '',
             phone: '',
             company: '',
-            address1: address?.address1 || 'Fake street',
+            address1: address?.address1 || '',
             address2: '',
```
animesh1987, line 269: Same can we set `address2` value if present?

serhii-tkachenko: Is it's ok to use `address?.address1` for `address2`?

serhii-tkachenko: Unfortunately, PayPal didn't pass the data for `address2`

animesh1987: No think should use `address2` if provided? Is there a flag that we can pass to get that data 🤔 https://developer.paypal.com/docs/api/orders/v2/#definition-address_portable looking at this we should be able to get `address_line_2`, or is that the wrong doc ?

serhii-tkachenko: Here is another doc for Accelerated Checkout feature, but there is no information about an order interface that we can get from `actions.order.get()`

https://developer.paypal.com/beta/accelerated-checkout/

animesh1987: Discussed offline and seems the accelerated beta is only for US only. Should be ok in that case.

animesh1987 (approved):

serhii-tkachenko:
Thanks @animesh1987 🙏

PR #1664, opened 2022-10-28
fix(payment): PAYPAL-1758 added condition to submit payment in onComplete method for auth&capture transaction only for PayPalInlineCheckoutButton

serhii-tkachenko:
## What?
Added condition to submit payment in `onComplete` method for auth&capture transactions for PayPalInlineCheckoutButtonStrategy

## Why?
We should not make secondary submit payment request for 'Authentication only' transaction types because it causes an error on BigPay side. Order data will be updated after 'Capture Funds' for transaction in CP by using Webhooks. 

## Sibling PRs:
bcapp: https://github.com/bigcommerce/bigcommerce/pull/49407

## Testing / Proof
Unit tests
Manual tests
QA tests

### DIFF
--- a/packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-inline-checkout-button-strategy.ts
+++ b/packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-inline-checkout-button-strategy.ts
@@ -196,7 +196,13 @@ export default class PaypalCommerceInlineCheckoutButtonStrategy implements Check
         methodId: string,
         callback?: () => void
     ): Promise<void> {
-        await this._submitPayment(methodId, data.orderID);
+        const state = this._store.getState();
+        const paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(methodId);
+        const { intent } = paymentMethod.initializationData;
+
+        if (intent === 'capture') {
+            await this._submitPayment(methodId, data.orderID);
+        }
 
         if (callback) {
             callback();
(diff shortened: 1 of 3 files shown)


packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-inline-checkout-button-strategy.ts
```diff
@@ -196,7 +196,13 @@ export default class PaypalCommerceInlineCheckoutButtonStrategy implements Check
         methodId: string,
         callback?: () => void
     ): Promise<void> {
-        await this._submitPayment(methodId, data.orderID);
+        const state = this._store.getState();
+        const paymentMethod = state.paymentMethods.getPaymentMethodOrThrow(methodId);
+        const { intent } = paymentMethod.initializationData;
+
+        if (intent === 'capture') {
```
animesh1987, line 203: Are there many `intent` types? I am asking since if we know that there can be a few then maybe we can create an intent Enum?

serhii-tkachenko: We have 2 possible values for intent: 'capture' and 'authorize'. I think we can make an Enum for that

serhii-tkachenko: Done 👍

animesh1987 (approved):

PR #1667, opened 2022-10-28
feat(payment): PAYPAL-1722 added skip checkout functionality to PDP

andriiVitvitskyi1990:
## What?
Added ability for skip checkout to work on PDP (paypalcommerce).

## Why?
According to task https://bigcommercecloud.atlassian.net/browse/PAYPAL-1722
smart-payment-buttons PR: https://github.com/bigcommerce/smart-payment-buttons/pull/12
bcapp PR: https://github.com/bigcommerce/bigcommerce/pull/49657
## Testing / Proof
Tested on dev and int

### DIFF
--- a/packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
+++ b/packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
@@ -123,6 +123,7 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
             try {
                 const { body: cart } = await this._cartRequestSender.createBuyNowCart(cartRequestBody);
                 this._buyNowCartId = cart.id;
+                await this._store.dispatch(this._checkoutActionCreator.loadCheckout(cart.id));
             } catch (error) {
                 throw new BuyNowCartCreationError();
             }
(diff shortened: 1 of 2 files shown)


packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -123,6 +123,7 @@ export default class PaypalCommerceButtonStrategy implements CheckoutButtonStrat
             try {
                 const { body: cart } = await this._cartRequestSender.createBuyNowCart(cartRequestBody);
                 this._buyNowCartId = cart.id;
+                await this._store.dispatch(this._checkoutActionCreator.loadCheckout(cart.id));
```
animesh1987, line 126: Can we please update tests to reflect this.

serhii-tkachenko:
The main point of this PR is make PayPal Shipping Options compatible with Buy Now cart. In this particular case we should to have checkout available to make `updateAddress` (billingAddress and consignment method), works correctly for our needs. 

We don't use `loadDefaultCheckout` method from `checkoutActionCreator`, due to the issue with primary cart. We should get specific checkout by using buy now cart id, to make sure, that we have all data that we need for this flow.

packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts
andriiVitvitskyi1990: added tests

animesh1987 (approved):

PR #1670, opened 2022-11-02
feat(checkout): STRF-10153 Data Layer Updates - Field Additions and Changes

jairo-bc:
## What?

- We are changing name for base price field, adding sale price and purchase price
- Channel id for both events
- Tax total for order purchased
- Rename sku

## Why?

Updated BODL payload to match desired structure: [link](https://docs.google.com/document/d/13qlYa9gtuVcjs2iBZ-jWP74W1GPRQw9J496t7QYbbJg/edit#heading=h.1m5ajudz3zss)

## Testing / Proof
Should be merged after [bcapp PR](https://github.com/bigcommerce/bigcommerce/pull/49561) is merged

### DIFF
--- a/packages/core/src/bodl/bodl-emitter-service.ts
+++ b/packages/core/src/bodl/bodl-emitter-service.ts
@@ -45,8 +45,9 @@ export default class BodlEmitterService implements BodlService {
                 currency,
                 lineItems,
                 id,
-                coupons
+                coupons,
             },
+            channelId,
         } = checkout;
 
         this.bodlEvents.emitCheckoutBeginEvent({
@@ -54,7 +55,8 @@ export default class BodlEmitterService implements BodlService {
             currency: currency.code,
             cart_value: cartAmount,
             coupon: coupons.map(coupon => coupon.code.toUpperCase()).join(','),
-            line_items: this.getProducts(lineItems, currency.code)
+            line_items: this.getProducts(lineItems, currency.code),
+            channel_id: channelId,
         });
 
         this._checkoutStarted = true;
@@ -75,7 +77,9 @@ export default class BodlEmitterService implements BodlService {
             shippingCostTotal,
             lineItems,
             cartId,
-            coupons
+            coupons,
+            channelId,
+            taxTotal,
         } = order;
 
         if (!isComplete) {
@@ -86,6 +90,8 @@ export default class BodlEmitterService implements BodlService {
             id: cartId,
             currency: currency.code,
             transaction_id: orderId,
+            tax: taxTotal,
+            channel_id: channelId,
             cart_value: orderAmount,
             coupon: coupons.map(coupon => coupon.code.toUpperCase()).join(','),
             shipping_cost: shippingCostTotal,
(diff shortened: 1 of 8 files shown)


packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -96,8 +102,10 @@ export default class BodlEmitterService implements BodlService {
     private getProducts(lineItems: LineItemMap, currencyCode: string): BODLProduct[] {
         const customItems: BODLProduct[] = (lineItems.customItems || []).map(item => ({
             product_id: item.id,
-            product_sku: item.sku,
-            price: item.listPrice,
+            sku: item.sku,
+            base_price: item.listPrice,
+            sale_price: item.listPrice,
```
animesh1987, line 107: Item also has a `salePrice` property, should we be using that?

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -75,7 +77,9 @@ export default class BodlEmitterService implements BodlService {
             shippingCostTotal,
             lineItems,
             cartId,
-            coupons
+            coupons,
+            channelId,
```
animesh1987, line 81: From what I can see we don't have `channelId` here, can you please add it here as well
https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/order/order.ts#L9

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -45,16 +45,18 @@ export default class BodlEmitterService implements BodlService {
                 currency,
                 lineItems,
                 id,
-                coupons
+                coupons,
             },
+            channelId,
```
animesh1987, line 50: Same as https://github.com/bigcommerce/checkout-sdk-js/pull/1670/files#r1013808067 comment, but for checkout interface.

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -96,8 +102,10 @@ export default class BodlEmitterService implements BodlService {
     private getProducts(lineItems: LineItemMap, currencyCode: string): BODLProduct[] {
         const customItems: BODLProduct[] = (lineItems.customItems || []).map(item => ({
             product_id: item.id,
-            product_sku: item.sku,
-            price: item.listPrice,
+            sku: item.sku,
+            base_price: item.listPrice,
+            sale_price: item.listPrice,
+            purchase_price: item.listPrice,
```
animesh1987, line 108: May need to check with other members if `listPrice` is the best field for this 🤔

animesh1987: I see you have a differnt logic on line 146

packages/core/src/bodl/bodl-emitter-service.ts
animesh1987: See you have already done this.

packages/core/src/bodl/bodl-emitter-service.ts
animesh1987: Same as above

packages/core/src/bodl/bodl-emitter-service.ts
jairo-bc: @animesh1987 per doc custom item has only listPrice available, I'm quite not sure if other fields make impact or used for this case

packages/core/src/bodl/bodl-emitter-service.ts
jairo-bc: yep, I added

packages/core/src/bodl/bodl-emitter-service.ts
jairo-bc: it's pretty much the same as below

animesh1987 (approved):

PR #1697, opened 2022-11-21
refactor(checkout): CHECKOUT-6397 Refactor external payment method

bc-peng:
## What?
Refactor the existing `external` payment strategy to use the new payment integration JS API.

## Why?
Lay the foundation work for restructuring Checkout SDK into a monorepo.

## Testing / Proof
- Circle.


packages/external-integration/.eslintrc.json
```diff
@@ -0,0 +1,4 @@
+{
+    "extends": ["../../.eslintrc.json"],
+    "ignorePatterns": ["!**/*"]
```
bc-peng, line 3: I have to override `ignorePatterns` so that the lint command can pass.

PR #1707, opened 2022-11-24
refactor(checkout): CHECKOUT-7107 Refactor `hosted-form`

bc-peng:
## What?
Add `HostedForm` to the payment integration JS API.

## Why?
Unblock refactoring subsequent payment methods, such as `credit-card-payment-method`, which depends on `HostedForm`, `HostedFormOptions`, `HostedFieldType`, and `HostedFormFactory`.

## Testing / Proof
- Added one unit test.
- CI checks.

### DIFF
--- a/packages/core/src/payment-integration/default-payment-integration-service.ts
+++ b/packages/core/src/payment-integration/default-payment-integration-service.ts
@@ -17,6 +17,7 @@ import { OrderActionCreator } from "../order";
 import PaymentActionCreator from "../payment/payment-action-creator";
 import PaymentMethodActionCreator from "../payment/payment-method-action-creator";
 import { ConsignmentActionCreator } from "../shipping";
+
 import PaymentIntegrationStoreProjectionFactory from "./payment-integration-store-projection-factory";
 
 export default class DefaultPaymentIntegrationService
--- a/packages/core/src/payment-integration/default-payment-integration-service.ts
+++ b/packages/core/src/payment-integration/default-payment-integration-service.ts
@@ -1,4 +1,6 @@
 import {
+    HostedForm,
+    HostedFormOptions,
     PaymentIntegrationService,
     PaymentIntegrationSelectors,
     BillingAddressRequestBody,
@@ -10,11 +12,11 @@ import {
 import { BillingAddressActionCreator } from "../billing";
 import { CheckoutStore, CheckoutActionCreator } from "../checkout";
 import { DataStoreProjection } from "../common/data-store";
+import { HostedFormFactory } from "../hosted-form";
 import { OrderActionCreator } from "../order";
 import PaymentActionCreator from "../payment/payment-action-creator";
 import PaymentMethodActionCreator from "../payment/payment-method-action-creator";
 import { ConsignmentActionCreator } from "../shipping";
-
 import PaymentIntegrationStoreProjectionFactory from "./payment-integration-store-projection-factory";
 
 export default class DefaultPaymentIntegrationService
@@ -35,6 +37,12 @@ export default class DefaultPaymentIntegrationService
         this._storeProjection = this._storeProjectionFactory.create(
             this._store
         );
+    }
+
+    createHostedForm(host: string, options: HostedFormOptions): HostedForm {
+        const hostedFormFactory = new HostedFormFactory(this._store);
+
+        return hostedFormFactory.create(host,options);
     }
 
     /* eslint-disable  @typescript-eslint/no-explicit-any */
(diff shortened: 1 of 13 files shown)


packages/core/src/payment-integration/default-payment-integration-service.ts
```diff
@@ -10,11 +12,11 @@ import {
 import { BillingAddressActionCreator } from "../billing";
 import { CheckoutStore, CheckoutActionCreator } from "../checkout";
 import { DataStoreProjection } from "../common/data-store";
+import { HostedFormFactory } from "../hosted-form";
 import { OrderActionCreator } from "../order";
 import PaymentActionCreator from "../payment/payment-action-creator";
 import PaymentMethodActionCreator from "../payment/payment-method-action-creator";
 import { ConsignmentActionCreator } from "../shipping";
-
```
bc-peng, line 17: Will add this back later.

packages/core/src/payment-integration/default-payment-integration-service.ts
```diff
@@ -37,6 +40,12 @@ export default class DefaultPaymentIntegrationService
         );
     }
 
+    createHostedForm(host: string, options: HostedFormOptions): HostedForm {
+        const hostedFormFactory = new HostedFormFactory(this._store);
```
animesh1987, line 44: Does it make sense to instantiate this in the constructor? Or pass an instance to this class and instantiate that in `create-payment-integration-service.ts` files?

packages/payment-integration-api/src/hosted-form/hosted-form-options.ts
```diff
@@ -0,0 +1,52 @@
+//copied from packages/core/src/hosted-form/hosted-form-options.ts
```
animesh1987, line 1: Do we need to have this comment? Also shall we remove the interfaces in core package in this case?  Since hosted form is a payments thing so we should be able to remove that.

cc @davidchin

animesh1987 (commented):
Nice work @bc-peng, just wondering if we need to add and expose any utils related to hosted form from test-utils package.

packages/payment-integration-api/src/hosted-form/hosted-form-options.ts
bc-peng: We can remove it in another PR as a quick follow-up to keep this PR simple.

packages/core/src/payment-integration/default-payment-integration-service.ts
bc-peng: I didn't include it in the constructor because 'hostedFormFactory' does not appear to be required all the time and we only need the factory method to create a `HostedForm` for a method.

bc-peng:
> Nice work @bc-peng, just wondering if we need to add and expose any utils related to hosted form from test-utils package.

`HostedForm` looks almost untouched as it stays in the core.
So far I think credit card payment method does not require exposing any utils.

packages/core/src/payment-integration/default-payment-integration-service.ts
animesh1987: I think it should be ok, since that's how it was passed to the providers before as well. And other dependencies are also passed in the same way to the service. 

Also better not to `new` every time the `createHostedForm` is called. 

cc @davidchin

packages/payment-integration-api/src/hosted-form/iframe-content.ts
```diff
@@ -0,0 +1,65 @@
+//copied from packages/core/src/hosted-form/iframe-content
```
bc-peng, line 1: Will remove this in the next PR, too.

packages/payment-integration-api/src/hosted-form/hosted-form.ts
```diff
@@ -0,0 +1,13 @@
+import { OrderPaymentRequestBody } from "../order";
+import { PaymentAdditionalAction } from "../payment";
+
+import { HostedInputSubmitSuccessEvent } from "./hosted-input-events";
+
+export interface HostedForm {
+    attach(): Promise<void>;
+    detach(): void;
+    getBin(): string | undefined;
```
animesh1987, line 9:
```suggestion
    getBin?(): string;
```

animesh1987 (approved):
Nice work @bc-peng , left a minor comment 👍

packages/payment-integration-api/src/payment-integration-service.ts
```diff
@@ -1,11 +1,18 @@
 import { BillingAddressRequestBody } from "./billing";
+import { HostedForm } from "./hosted-form";
+import HostedFormOptions from "./hosted-form/hosted-form-options";
```
davidchin, line 3: Can you combine line 2 and 3? 🍹

davidchin (approved):
LGTM, nice work!

packages/payment-integration-api/src/hosted-form/hosted-form.ts
bc-peng: ewww, the previous implementation was this. Hence it cannot only return `string` if I am not mistaken.
```
getBin(): string | undefined {
        return this._bin;
    }
```

PR #1710, opened 2022-11-29
refactor(checkout): CHECKOUT-6394 Refactor credit card payment method

bc-peng:
## What?
Refactor the existing `credit-card` payment strategy to use the new checkout payment integration JS API.

## Why?
Restructuring `checkout-sdk` into a monorepo.

## Testing / Proof
- CI checks.
- ~Manual testing: `checkout-js` e2e credit card payment method test with changes from this PR.~

### DIFF
--- a/packages/core/src/common/registry/resolve-id-registry.ts
+++ b/packages/core/src/common/registry/resolve-id-registry.ts
@@ -4,10 +4,10 @@ import Registry from './registry';
 export default class ResolveIdRegistry<TType, TToken extends { [key: string]: unknown }> {
     private _registry: Registry<TType>;
 
-    constructor() {
+    constructor(private _useFallback = false) {
         this._registry = new Registry({
             tokenResolver: this._resolveToken.bind(this),
-            useFallback: false,
+            useFallback: this._useFallback,
         });
     }
 
@@ -29,15 +29,19 @@ export default class ResolveIdRegistry<TType, TToken extends { [key: string]: un
 
     private _resolveToken(token: string, registeredTokens: string[]): string | undefined {
         const query = this._decodeToken(token);
-        const results: Array<{ token: string; matches: number }> = [];
+        const results: Array<{ token: string; matches: number; default: boolean }> = [];
 
         registeredTokens.forEach((registeredToken) => {
             const resolverId = this._decodeToken(registeredToken);
-            const result = { token: registeredToken, matches: 0 };
+            const result = { token: registeredToken, matches: 0, default: false };
 
             for (const [key, value] of Object.entries(resolverId)) {
                 if (key in query && query[key] === value) {
                     result.matches++;
+                }
+
+                if (key === 'default' && value === true) {
+                    result.default = true;
                 }
             }
 
@@ -48,8 +52,16 @@ export default class ResolveIdRegistry<TType, TToken extends { [key: string]: un
             .sort((a, b) => b.matches - a.matches)
             .filter((result) => result.matches > 0)[0];
 
-        if (matched.token) {
+        if (matched && matched.token) {
             return matched.token;
+        }
+
+        if (this._useFallback) {
+            const defaultToken = results.find((result) => result.default)?.token;
+
+            if (defaultToken) {
+                return defaultToken;
+            }
         }
 
         throw new Error('Unable to resolve to a registered token with the provided token.');
--- a/packages/core/src/payment/payment-strategy-registry.spec.ts
+++ b/packages/core/src/payment/payment-strategy-registry.spec.ts
@@ -1,4 +1,5 @@
 import { CheckoutStore, createCheckoutStore, InternalCheckoutSelectors } from '../checkout';
+import { InvalidArgumentError } from '../common/error/errors';
 import { getConfigState } from '../config/configs.mock';
 import { getFormFieldsState } from '../form/form.mock';
 import { OrderFinalizationNotRequiredError } from '../order/errors';
@@ -8,7 +9,6 @@ import {
     getAmazonPay,
     getBankDeposit,
     getBraintree,
-    getBraintreePaypal,
     getCybersource,
     getPPSDK,
 } from './payment-methods.mock';
@@ -91,18 +91,8 @@ describe('PaymentStrategyRegistry', () => {
             expect(registry.getByMethod(getAmazonPay())).toBeInstanceOf(AmazonPayPaymentStrategy);
         });
 
-        it('returns credit card strategy if none is registered with method name', () => {
-            expect(registry.getByMethod(getBraintree())).toBeInstanceOf(CreditCardPaymentStrategy);
-        });
-
-        it('returns new strategy instance if multiple methods belong to same type', () => {
-            expect(registry.getByMethod(getBraintree())).toBeInstanceOf(CreditCardPaymentStrategy);
-            expect(registry.getByMethod(getBraintreePaypal())).toBeInstanceOf(
-                CreditCardPaymentStrategy,
-            );
-            expect(registry.getByMethod(getBraintree())).not.toBe(
-                registry.getByMethod(getBraintreePaypal()),
-            );
+        it('throws error if none is registered with method name (expected V1 behavior)', () => {
+            expect(() => registry.getByMethod(getBraintree())).toThrow(InvalidArgumentError);
         });
 
         it('returns offsite strategy if none is registered with method name and method is hosted', () => {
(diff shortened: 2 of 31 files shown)


packages/credit-card-integration/src/lib/create-credit-card-payment-strategy.ts
```diff
@@ -0,0 +1,19 @@
+import {
+    PaymentStrategyFactory,
+    toResolvableModule,
+} from "@bigcommerce/checkout-sdk/payment-integration-api";
+
+import CreditCardPaymentStrategy from "./credit-card-payment-strategy";
+
+const createCreditCardPaymentStrategy: PaymentStrategyFactory<
+    CreditCardPaymentStrategy
+    > = (paymentIntegrationService) => {
```
davidchin, line 10:
```suggestion
> = (paymentIntegrationService) => {
```

packages/credit-card-integration/src/lib/create-credit-card-payment-strategy.ts
```diff
@@ -0,0 +1,19 @@
...
+
+    return new CreditCardPaymentStrategy(
+        paymentIntegrationService,
+    );
+};
+
+export default toResolvableModule(createCreditCardPaymentStrategy, [
+    { id: 'creditcard' },
```
davidchin, line 18: Not sure if this would work because credit card methods have ids such as `authorizenet` instead of `creditcard` 🤔 Have you tested it end-to-end in your VM with actual payment methods?

packages/credit-card-integration/src/index.ts
```diff
@@ -0,0 +1,3 @@
+// export { default as createCreditCardPaymentStrategy} from "./lib/create-credit-card-payment-strategy";
```
vitalii-koshovyi, line 1: do we need this comment?

packages/credit-card-integration/src/index.ts
```diff
@@ -0,0 +1,3 @@
...
+export { CreditCardPaymentInitializeOptions, WithCreditCardPaymentInitializeOptions } from "./lib/credit-card-payment-initialize-options";
```
vitalii-koshovyi, line 2: Why do we need `lib` sub-directory?

packages/credit-card-integration/src/index.ts
bc-peng: I removed `create-credit-card-payment-strategy` from this PR. 
Once we've decided whether or not to include it, I'll remove or uncomment this line.

packages/credit-card-integration/src/index.ts
bc-peng: `lib` comes from `nx` generator. I have removed the folder.

packages/credit-card-integration/src/lib/create-credit-card-payment-strategy.ts
bc-peng: I have removed `create-credit-card-payment-strategy.ts` from this PR.

As you mentioned, `CreditCardPaymentStrategy` is used as a base class right now ([sample](https://github.com/bigcommerce/checkout-sdk-js/search?q=%22extends+CreditCardPaymentStrategy%22)),  therefore creating a credit card payment strategy makes no sense.

packages/credit-card-integration/src/lib/credit-card-payment-strategy.ts
```diff
@@ -0,0 +1,123 @@
...
+        const { paymentSettings: { bigpayBaseUrl: host = '' } = {} } = state.getStoreConfigOrThrow();
+
+        if (!formOptions) {
+            throw new InvalidArgumentError();
+        }
+
+        const form = this._paymentIntegrationService.createHostedForm(host, formOptions);
+
+        return form.attach()
+            .then(() => {
```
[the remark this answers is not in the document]
bc-peng: This is simply a copy from the current code.
https://github.com/bigcommerce/checkout-sdk-js/blob/2b4e0d6b49d83029771d00de2d86bfab5068ff7f/packages/core/src/payment/strategies/credit-card/credit-card-payment-strategy.ts#L51

packages/credit-card-integration/src/lib/create-credit-card-payment-strategy.ts
davidchin: FYI `CreditCardPaymentStrategy` is also used directly: https://github.com/bigcommerce/checkout-sdk-js/blob/aaac6a157ec6783782460ca1851859d245aedff2/packages/core/src/payment/create-payment-strategy-registry.ts#L354 Right now, it is the default strategy - meaning if nothing matches, it'd fall back to this. We could follow the same pattern by passing `{ default: true }` to `toResolvableModule`. However, I think moving forward, we should explicit out the method IDs.

vitalii-koshovyi (approved):

packages/core/src/common/registry/resolve-id-registry.ts
```diff
@@ -51,6 +55,14 @@ export default class ResolveIdRegistry<TType, TToken extends { [key: string]: un
             return matched.token;
         }
 
+        if (this._useFallback) {
```
bc-peng, line 58: Hi @davidchin , I have added `defaultToken` in `_resolveToken`.

However, as we still keep the old `CreditCardPaymentStrategy` in `core` as the default payment strategy and base class for many strategies, I used a `useFallback` flag to control whether we have a default token.

All changes have been tested on my dev store. (To identify the modified SDK code, I added a custom console log.)

https://user-images.githubusercontent.com/88361607/204983770-4d400a0e-62af-4a01-8b87-4de301f7dba9.mov

packages/credit-card-integration/src/create-credit-card-payment-strategy.ts
```diff
@@ -0,0 +1,19 @@
...
+    CreditCardPaymentStrategy
+> = (paymentIntegrationService) => {
+
+    return new CreditCardPaymentStrategy(
+        paymentIntegrationService,
+    );
+};
+
+export default toResolvableModule(createCreditCardPaymentStrategy, [
+    { default: true },
```
davidchin, line 18: I think a bit more work needs to be done if we are using this approach. Right now, we use `PaymentStrategyRegistryV2` to resolve first, and then fall back to `PaymentStrategyRegistry`, e.g.: https://github.com/bigcommerce/checkout-sdk-js/blob/aaac6a157ec6783782460ca1851859d245aedff2/packages/core/src/payment/payment-strategy-action-creator.ts#L92-L103

So if there's a default strategy registered with the V2 registry, we'll not able to use V1 as a fallback. Perhaps we can flip it around so we resolve with V1 registry first and then V2 registry, as V1 registry no longer has a default strategy.

PS: I still prefer to list out the methods explicitly but I understand it's a lot more work to figure out which one to list out so I guess `default` is a good compromise for now.

packages/core/src/payment/payment-strategy-registry.spec.ts
```diff
@@ -79,14 +80,8 @@ describe('PaymentStrategyRegistry', () => {
...
         });
 
-        it('returns credit card strategy if none is registered with method name', () => {
-            expect(registry.getByMethod(getBraintree())).toBeInstanceOf(CreditCardPaymentStrategy);
-        });
-
-        it('returns new strategy instance if multiple methods belong to same type', () => {
-            expect(registry.getByMethod(getBraintree())).toBeInstanceOf(CreditCardPaymentStrategy);
-            expect(registry.getByMethod(getBraintreePaypal())).toBeInstanceOf(CreditCardPaymentStrategy);
-            expect(registry.getByMethod(getBraintree())).not.toBe(registry.getByMethod(getBraintreePaypal()));
```
bc-peng, line 89: Hi , @davidchin . I have:

- Fliped the order of V1 and V2 registrys, modified unit tests accordingly.
- Removed the old credit-card payment method from registry as it is no longer the default payment method.
- Tested all changes on a store.

https://user-images.githubusercontent.com/88361607/205209741-0c85feef-4559-4f63-b356-1a8bb3a117f9.mov

Just want to point out, because V1 won't return a default method now, this unit test has been moved. 
https://github.com/bigcommerce/checkout-sdk-js/blob/92bd1e66ec20d6cd4bbace5eb1af789ebd5c9b72/packages/core/src/common/registry/resolve-id-registry.spec.ts#L68.

animesh1987 (approved):

davidchin (approved):
Nice work @bc-peng!

PR #1711, opened 2022-11-29
feat(payment): BOLT-408 Add BODL event for customer suggestion init

PavlenkoM:
## What?
Add new Bodl analytics event to Bolt customer suggestion initialization

## Why?
Because of task: [https://bigcommercecloud.atlassian.net/browse/BOLT-408](https://bigcommercecloud.atlassian.net/browse/BOLT-408)

## Testing / Proof
Screenshot 2022-11-29 at 10 45 32

### DIFF
--- a/packages/core/src/bodl/bodl-emitter-service.ts
+++ b/packages/core/src/bodl/bodl-emitter-service.ts
@@ -120,6 +120,10 @@ export default class BodlEmitterService implements BodlService {
         this.bodlEvents.emit('bodl_checkout_email_entry_began');
     }
 
+    customerSuggestionInit(payload?: BodlEventsPayload) {
+        this.bodlEvents.emit('bodl_checkout_customer_suggestion_initialization', payload);
+    }
+
     customerSuggestionExecute() {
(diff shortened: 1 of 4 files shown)


serhii-tkachenko (approved):
LGTM 👍

vitalii-koshovyi (approved):

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -120,6 +120,10 @@ export default class BodlEmitterService implements BodlService {
         this.bodlEvents.emit('bodl_checkout_email_entry_began');
     }
 
+    customerSuggestionInit(payload?: BodlEventsPayload) {
+        this.bodlEvents.emit('bodl_checkout_customer_suggestion_initialization', payload);
```
animesh1987, line 124: I think we have an opportunity as a follow up to have proper typing for these events as well. Maybe we can tackle it in a future iteration?

animesh1987 (approved):

PR #1714, opened 2022-11-30
chore(common): CHECKOUT-6276 Update eslint for @bigcommerce/eslint

animesh1987:
## What?
As above

## Why?
As we have moved to monorepo structure, its a good idea to have a defined set of code rules and guidelines as we get more teams owning and contributing code.

## Testing / Proof
- Circle


vitalii-koshovyi (approved):
From Adyen perspective seems fine

packages/external-integration/.eslintrc.json
```diff
@@ -1,4 +1,12 @@
 {
     "extends": ["../../.eslintrc.json"],
-    "ignorePatterns": ["!**/*"]
+    "ignorePatterns": ["!**/*"],
```
bc-peng, line 3: Since we don't have `"ignorePatterns": ["**/*"]` in base, do we still need this line?

animesh1987: It's mostly copy paste, should be able to remove this.

packages/core/src/common/error/errors/custom-error.ts
```diff
@@ -4,11 +4,6 @@ export default interface CustomError extends Error {
...
 
-export function isCustomError(error: any): error is CustomError {
-    return (
-        typeof error.message === 'string' &&
-        typeof error.type === 'string' &&
-        (typeof error.subtype === 'string' || !error.subtype) &&
-        error instanceof Error
-    );
+export function isCustomError(error: unknown): error is CustomError {
+    return typeof error === 'object' && error !== null && 'message' in error && 'type' in error;
```
bc-peng, line 8: I see `subtype` checking is removed, is this intentional?

animesh1987: Yes, intentional since not all custom errors have a subtype from what I found.  We will need to type this properly later.

bc-peng (approved):
LGTM. Congrats on conquering another hardcore task.

davidchin (approved):
Assuming most changes are automatically applied by the linter.

PR #1719, opened 2022-12-06
fix(analytics): DATA-10159 Fix prices for Universal Analytics when non default currency used

bc-rmalyavc:
## What? [DATA-10159](https://bigcommercecloud.atlassian.net/browse/DATA-10159)
Convert prices for Universal Analytics payload when non-default currency used

## Why?
When currency, different from default one is used, the prices in the payload were not converted accordingly to exchange rate. At the same time, current currency code was sent, which caused issues in Google Analytics console. **Example below (before):**
Screenshot 2022-12-06 at 13 48 28
In this case:
- the price for the product is set up in JPY (15 JPY).
- the currency for current country is EUR, so the price is converted (0.11 EUR)
- the price sent to the analytic is 15JPY as set up by default, but the currency is EUR

## Testing / Proof
Manually tested on dev store
**Example below (after):**
Screenshot 2022-12-06 at 13 58 09
In this case:
- prices are set up in JPY
- the currency selected is USD
- prices are converted to USD when sent to the analytic. The currency sent is also USD


packages/core/src/analytics/analytics-step-tracker.ts
```diff
@@ -419,7 +419,7 @@ export default class AnalyticsStepTracker implements StepTracker {
             return {
                 product_id: item.productId,
                 sku: item.sku,
-                price: item.salePrice,
+                price: this.toShopperCurrency(item.salePrice),
```
solofeed, line 422: From Jira

```
Expected Result

Product price will remain in JPY
```

PR #1723, opened 2022-12-08
chore(common): CHECKOUT-6407 Move offsite strategy to separate package

animesh1987:
## What?
As above

## Why?
As part of decomposing payment strategies, we need to move offsite payment strategy to its own package.

## Testing / Proof
- circle


packages/core/src/payment/payment-strategy-registry.spec.ts
```diff
@@ -105,10 +101,6 @@ describe('PaymentStrategyRegistry', () => {
             );
         });
 
-        it('returns offsite strategy if none is registered with method name and method is hosted', () => {
-            expect(registry.getByMethod(getAdyenAmex())).toBeInstanceOf(OffsitePaymentStrategy);
-        });
-
```
bc-peng, line 111: I think it still should be tested or we may add an offsite e2e test.
The rest are all good.

davidchin (approved):
Nice work, LGTM! 👍

packages/payment-integration-api/src/payment-integration-service.ts
```diff
@@ -29,7 +33,7 @@ export default interface PaymentIntegrationService {
 
     submitPayment(payment: Payment): Promise<PaymentIntegrationSelectors>;
 
-    finalizeOrder(): Promise<PaymentIntegrationSelectors>;
+    finalizeOrder(options?: RequestOptions): Promise<PaymentIntegrationSelectors>;
```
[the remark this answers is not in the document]
animesh1987: Yes, since some methods require so. And the underlying methods accept it as well.

packages/core/src/payment/payment-strategy-registry.spec.ts
animesh1987: This was pretty much a dummy test from what I understood. But you are right we should write an offsite test and we do have a task to do so as well.

PR #1724, opened 2022-12-08
refactor(checkout): CHECKOUT-6406 Refactor offline payment method

bc-peng:
## What?
Refactor the existing `offline` payment strategy to use the new checkout payment integration JS API.

## Why?
Lay the foundation work for restructuring SDK into a monorepo.

## Testing / Proof
- CI checks.


packages/core/src/checkout/checkout-service.spec.ts
```diff
@@ -200,19 +199,19 @@ describe('CheckoutService', () => {
             getResponse(getCountriesResponseBody()),
         );
 
-        paymentStrategy = new OfflinePaymentStrategy(store, orderActionCreator);
+        paymentStrategyRegistry = new PaymentStrategyRegistry(store);
```
animesh1987, line 202: Do we need to change this?

packages/offline-integration/.eslintrc.json
```diff
@@ -0,0 +1,24 @@
+{
+    "extends": ["../../.eslintrc.json"],
+    "ignorePatterns": ["!**/*"],
+    "overrides": [
+        {
+            "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
+            "rules": {
```
animesh1987, line 7: Do we need to ignore these rules? The strategy files seem pretty simple.

packages/core/src/checkout/checkout-service.spec.ts
bc-peng: This looks bit confusing, it is the same line as the line213 before.
I put `const paymentIntegrationService` up first, then `paymentStrategy` and `jest.spyOn`. Somehow, this line got moving up.

packages/offline-integration/.eslintrc.json
bc-peng: Yes, make sense. I removed all overrides here.

packages/core/src/checkout/checkout-service.spec.ts
bc-peng: I moved it next to the `paymentStrategyRegistryV2` now.

packages/offline-integration/src/create-offline-payment-strategy.spec.ts
```diff
@@ -0,0 +1,20 @@
+import { PaymentIntegrationService } from '@bigcommerce/checkout-sdk/payment-integration-api';
+import { PaymentIntegrationServiceMock } from '@bigcommerce/checkout-sdk/payment-integrations-test-utils';
+
+import createExternalPaymentStrategy from './create-offline-payment-strategy';
```
animesh1987, line 4: copy paste?

packages/offline-integration/.eslintrc.json
animesh1987: Do we need the override for the spec files? Since there is an issue with jest types that imply that `expect(something)` returns type of `any`.

packages/offline-integration/src/offline-payment-strategy.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+        });
+    });
+
+    describe('#finalize()', () => {
+        it('throws error to inform that order finalization is not required', async () => {
+            try {
+                await strategy.finalize();
+            } catch (error) {
+                expect(error).toBeInstanceOf(OrderFinalizationNotRequiredError);
+            }
```
animesh1987, line 51:
```suggestion
await expect(strategy.finalize()).rejects.toThrow(OrderFinalizationNotRequiredError);
```

packages/offline-integration/src/create-offline-payment-strategy.spec.ts
bc-peng: 🤦🏻‍♂️ Updated.

packages/offline-integration/.eslintrc.json
bc-peng: I restored three rules.
How could CI check pass without turning off any rules in the previous commit😂

animesh1987: From what I know they are lint issues and we don't run lint for build, but only on release 🤦

animesh1987 (approved):
Nice work @bc-peng 👏

packages/offline-integration/.eslintrc.json
davidchin: > From what I know they are lint issues and we don't run lint for build, but only on release 🤦

Are you sure that's the case? I think we run the linter on every PR? https://github.com/bigcommerce/checkout-sdk-js/blob/f0cb31f27d7d4487bf713de9acd638b95734dcce/.circleci/config.yml#L33

davidchin (approved):

packages/offline-integration/.eslintrc.json
animesh1987: Yes we do on another look, which is more confusing as to how come the build passed? Might test this in a next PR to validate.

PR #1727, opened 2022-12-09
chore(common): CHECKOUT-6401 Move legacy payment strategy to separate package

animesh1987:
## What?
As above

## Why?
As part of decomposing payment strategies, we need to move legacy payment strategy to its own package.

## Testing / Proof
- circle


bc-peng (approved):
LGTM 👍

PR #1730, opened 2022-12-13
feat(payment): PAYPAL-1836 added PayPalCommerceCustomer strategy

serhii-tkachenko:
## What?
Added PayPalCommerceCustomer strategy

## Why?
To add PayPal Commerce button to the customer step

## Testing / Proof
Here is a screenshot of the PayPal button in customer step and a video of the result:
Screenshot 2022-12-13 at 14 08 06

https://user-images.githubusercontent.com/25133454/207329128-0a3e051d-a49a-4d0c-adc6-e850b0c23051.mov

## Sibling PRs
bigcommerce: https://github.com/bigcommerce/bigcommerce/pull/50236
checkout-js: https://github.com/bigcommerce/checkout-js/pull/1123

### DIFF
--- a/packages/core/src/customer/strategies/paypalcommerce/paypalcommerce-customer-strategy.ts
+++ b/packages/core/src/customer/strategies/paypalcommerce/paypalcommerce-customer-strategy.ts
@@ -1,0 +1,411 @@
+import { FormPoster } from '@bigcommerce/form-poster';
+import { noop } from 'lodash';
+
+import { BillingAddressActionCreator, BillingAddressRequestBody } from '../../../billing';
+import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
+import {
+    InvalidArgumentError,
+    MissingDataError,
+    MissingDataErrorType,
+    RequestError,
+} from '../../../common/error/errors';
+import { OrderActionCreator } from '../../../order';
+import { PaymentActionCreator, PaymentMethodActionCreator } from '../../../payment';
+import { PaymentMethodClientUnavailableError } from '../../../payment/errors';
+import {
+    ApproveCallbackActions,
+    ApproveCallbackPayload,
+    ButtonsOptions,
+    PaypalCommerceRequestSender,
+    PaypalCommerceScriptLoader,
+    PaypalCommerceSDK,
+    PayPalOrderDetails,
+    ShippingAddressChangeCallbackPayload,
+    ShippingOptionChangeCallbackPayload,
+} from '../../../payment/strategies/paypal-commerce';
+import { ConsignmentActionCreator, ShippingOption } from '../../../shipping';
+import CustomerActionCreator from '../../customer-action-creator';
+import CustomerCredentials from '../../customer-credentials';
+import {
+    CustomerInitializeOptions,
+    CustomerRequestOptions,
+    ExecutePaymentMethodCheckoutOptions,
+} from '../../customer-request-options';
+import CustomerStrategy from '../customer-strategy';
+import PaypalCommerceCustomerInitializeOptions from './paypalcommerce-customer-initialize-options';
+
+export default class PaypalCommerceCustomerStrategy implements CustomerStrategy {
+    private _paypalCommerceSdk?: PaypalCommerceSDK;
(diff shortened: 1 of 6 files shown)


packages/core/src/customer/strategies/paypalcommerce/paypalcommerce-customer-strategy.ts
```diff
@@ -0,0 +1,411 @@
...
+
+    deinitialize(): Promise<InternalCheckoutSelectors> {
+        return Promise.resolve(this._store.getState());
+    }
+
+    signIn(
+        credentials: CustomerCredentials,
+        options?: CustomerRequestOptions,
+    ): Promise<InternalCheckoutSelectors> {
+        return this._store.dispatch(
```
animesh1987, line 103: Just out of curiosity how will this be triggered? I ask since when we have paypal in checkout page wallet buttons will be at the top level? So we won't have the email address from the shopper 🤔

packages/core/src/customer/strategies/paypalcommerce/paypalcommerce-customer-strategy.ts
```diff
@@ -0,0 +1,411 @@
...
+        const state = this._store.getState();
+        const consignment = state.consignments.getConsignmentsOrThrow()[0];
+
+        const availableShippingOptions = consignment.availableShippingOptions || [];
+
+        const recommendedShippingOption = availableShippingOptions.find(
+            (option) => option.isRecommended,
+        );
+        const selectedShippingOption = selectedShippingOptionId
+            ? availableShippingOptions.find((option) => option.id === selectedShippingOptionId)
```
animesh1987, line 350: Will this return the same `selectedShippingOption`?

packages/core/src/customer/strategies/paypalcommerce/paypalcommerce-customer-strategy.ts
```diff
@@ -0,0 +1,411 @@
...
+        const address = this._getAddress({
+            city: data.shippingAddress.city,
+            countryCode: data.shippingAddress.country_code,
+            postalCode: data.shippingAddress.postal_code,
+            stateOrProvinceCode: data.shippingAddress.state,
+        });
+
+        try {
+            // Info: we use the same address to fill billing and consignment addresses to have valid quota on BE for order updating process
+            // on this stage we don't have access to valid customer's address accept shipping data
```
animesh1987, line 258:
```suggestion
            // on this stage we don't have access to valid customer's address except shipping data
```
?

packages/core/src/customer/strategies/paypalcommerce/paypalcommerce-customer-strategy.ts
serhii-tkachenko: As far as I remember, this method is repeating the default sigh in method without any effort for PayPal. The wallet buttons will be placed in customer step but in the top of the block, right?

I don't think that we should throw any error here. You can double check default customer strategy, also it has the same implementation.

packages/core/src/customer/strategies/paypalcommerce/paypalcommerce-customer-strategy.ts
serhii-tkachenko: My bad, thanks!

packages/core/src/customer/strategies/paypalcommerce/paypalcommerce-customer-strategy.ts
serhii-tkachenko: `selectedShippingOptionId` - is an id of the shipping option provided by PayPal.

In this case it will return an available shipping option data with id provided by PayPal if it exist. In another case, when PayPal didn't provides a `selectedShippingOptionId` we should use previous selected shipping options from consignment object.

animesh1987: Thanks for the explanation, didn't realise that `selectedShippingOptionId` was an id returned from paypal.

animesh1987 (approved):

PR #1733, opened 2022-12-14
fix(payment): BOLT-362 fix Bolt track.js error in console

PavlenkoM:
## What?
Fix error on loading Bolt track.js script

## Why?
Because of tack: [https://bigcommercecloud.atlassian.net/browse/BOLT-362](https://bigcommercecloud.atlassian.net/browse/BOLT-362)

## Testing / Proof
Before:
Screenshot 2022-12-14 at 14 41 07

After:
Screenshot 2022-12-14 at 14 46 20

Test:
Screenshot 2022-12-14 at 17 02 11

### DIFF
--- a/packages/core/src/payment/strategies/bolt/bolt-script-loader.spec.ts
+++ b/packages/core/src/payment/strategies/bolt/bolt-script-loader.spec.ts
@@ -11,6 +11,15 @@ describe('BoltScriptLoader', () => {
     let scriptLoader: ScriptLoader;
     let mockWindow: BoltHostWindow;
 
+    const expectedScriptOptions = (id: string, publishableKey?: string) => {
+        return expect.objectContaining({
+            attributes: expect.objectContaining({
+                id,
+                'data-publishable-key': publishableKey,
+            }),
+        });
+    };
+
     beforeEach(() => {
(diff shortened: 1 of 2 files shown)


serhii-tkachenko (approved):
LGTM! 👍

vitalii-koshovyi (approved):

animesh1987 (approved):

packages/core/src/payment/strategies/bolt/bolt-script-loader.spec.ts
```diff
@@ -11,6 +11,15 @@ describe('BoltScriptLoader', () => {
     let scriptLoader: ScriptLoader;
     let mockWindow: BoltHostWindow;
 
+    const expectedScriptOptions = (id: string, publishableKey?: string) => {
```
animesh1987, line 14: Do we need to move this to a function? Seems like can be part of the test itself.

PavlenkoM: I move it to function to prevent multiple code duplication.
No logic inside this function, only constructor for data object with specific parameters, which different for different scripts but the same for all cases.

PR #1744, opened 2022-12-21
chore(common): CHECKOUT-6401 Move legacy payment strategy to separate package

animesh1987:
## What?
As above

## Why?
As part of decomposing payment strategies, we need to move legacy payment strategy to its own package.

## Testing / Proof
- circle


animesh1987:
There are only two payment methods that are classified as legacy `testgateway` and `paypalexpress`. Paypal Express has it's own strategy and hence doesn't use legacy strategy which leaves only testgateway.

davidchin (approved):

PR #1745, opened 2022-12-21
feat(payment): PAYMENTS-8378 Send multiple addresses to trusted_shipping_address endpoint in case of multi-shipping

Tharaae:
## What?
In case of multi-shipping, we want to send an array of addresses to the `trusted_shipping_address` endpoint instead of a single address.

## Why?
This is required to allow vaulting cards with multi-shipping.

## Testing / Proof
Screenshots
**Before**
Screen Shot 2022-12-22 at 10 36 42 am

**After (with multiple addresses)**
Screen Shot 2022-12-22 at 4 33 43 pm

**After (with single address)**
Screen Shot 2022-12-22 at 4 44 11 pm

### DIFF
--- a/packages/core/src/shipping/shipping-address-selector.ts
+++ b/packages/core/src/shipping/shipping-address-selector.ts
@@ -9,8 +9,9 @@ import ConsignmentState, { DEFAULT_STATE } from './consignment-state';
 
 export default interface ShippingAddressSelector {
     getShippingAddress(): Address | undefined;
-    getShippingAddresses(): Address[];
     getShippingAddressOrThrow(): Address;
+    getShippingAddresses(): Address[];
+    getShippingAddressesOrThrow(): Address[];
 }
 
 export type ShippingAddressSelectorFactory = (state: ConsignmentState) => ShippingAddressSelector;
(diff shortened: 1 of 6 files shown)


packages/payment-integration-api/src/payment-integration-selectors.ts
```diff
@@ -52,6 +52,7 @@ export default interface PaymentIntegrationSelectors {
     getPaymentMethodOrThrow(methodId: string, gatewayId?: string): PaymentMethod;
 
     getShippingAddress(): ShippingAddress | undefined;
+    getShippingAddresses(): ShippingAddress[];
```
bc-peng, line 55: All look good. Does it make sense to create `getShippingAddressesOrThrow` also?

Tharaae: We can, but do we need it? I wanted to just update the core interface as needed only.

bc-peng (approved):
Nice work 👍

packages/core/src/shipping/shipping-address-selector.ts
```diff
@@ -36,10 +38,35 @@ export function createShippingAddressSelectorFactory(): ShippingAddressSelectorF
         },
     );
 
+    const getShippingAddresses = createSelector(
```
animesh1987, line 41: Hey @Tharaae sorry just noticed this, can we add tests to cover this as well please. Thanks.

PR #1746, opened 2022-12-23
CHECKOUT-7148: Throw `CartConsistencyError` when server detects inconsistent projection data

davidchin:
## What?
Throw `CartInconsistencyError` when server detects inconsistent projection data.

## Why?
To notify browser clients about the error so they can handle it accordingly when it is thrown.

## Testing / Proof
Unit + Manual

https://user-images.githubusercontent.com/667603/209246970-95ba9508-7912-4243-ad3e-189c8fb787d6.mov

### DIFF
--- a/packages/core/src/cart/errors/cart-consistency-error.ts
+++ b/packages/core/src/cart/errors/cart-consistency-error.ts
@@ -1,0 +1,17 @@
+import { StandardError } from '../../common/error/errors';
+
+/**
+ * This error is thrown when the server detects inconsistency in cart data since it is last requested,
+ * for example, product prices or eligible discounts have changed.
+ */
+export default class CartConsistencyError extends StandardError {
+    constructor(message?: string) {
+        super(
+            message ||
+                'Your checkout could not be processed because some details have changed. Please review your order and try again.',
+        );
+
+        this.name = 'CartConsistencyError';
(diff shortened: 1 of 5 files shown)


packages/core/src/cart/errors/cart-consistency-error.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * This error is thrown when the server detects inconsistency in cart data since it is last requested,
+ * for example, product prices or eligible discounts have changed.
+ */
+export default class CartConsistencyError extends StandardError {
+    constructor(message?: string) {
+        super(
+            message ||
+                'Your checkout could not be processed because some details have changed. Please review your order and try again.',
```
animesh1987, line 11: I forgot if we translate these messages or not? Or if there was a mechanism to do the translation since this is a user facing message.

davidchin: @animesh1987 It gets translated in `checkout-js`. See  https://github.com/bigcommerce/checkout-js/pull/1138/files#diff-00368bf808596bf313d17f20d0b174a28cd10bfb3e6651656cec53e395016f6aR29

animesh1987 (approved):

PR #1747, opened 2022-12-24
fix(payment): STRIPE-281 Fixed error when making purchases with only digital products

BC-SRodriguez:
## What? [STRIPE-281](https://bigcommercecloud.atlassian.net/browse/STRIPE-281) 

No exception is thrown when retrieving shipping address when there are only digital products in the order.

## Why?

Trying to purchase digital products results in an error "Unable to proceed because shipping address data is unavailable" on payment step

## Testing / Proof

https://user-images.githubusercontent.com/69487174/209418303-9723d6d4-81ea-4552-8e51-92855db6a2f7.mov

### DIFF
--- a/packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
+++ b/packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
@@ -390,7 +390,11 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
             appearance,
         });
 
-        const { postalCode } = state.shippingAddress.getShippingAddressOrThrow();
+        const {
+            billingAddress: { getBillingAddress },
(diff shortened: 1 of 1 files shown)


animesh1987 (commented):
Hey @BC-SRodriguez have you tested the regular product, also maybe we can have a follow up task to cleanup the flow. Since for digital products we don't need to consider shipping address so we can ignore that flow.

BCHMercado:
💚 Testing done !
https://bigcommercecloud.atlassian.net/browse/STRIPE-281

BC-SRodriguez:
> Hey @BC-SRodriguez have you tested the regular product, also maybe we can have a follow up task to cleanup the flow. Since for digital products we don't need to consider shipping address so we can ignore that flow.

Hi @animesh1987 We have also tested with regular products.

animesh1987 (approved):

PR #1749, opened 2022-12-28
fix(payment): CHECKOUT-000 Fix default setup for offsite payment strategy

animesh1987:
## What?
Fix v1 payment registry automatically registering offsite/hosted strategies.

## Why?
Since offsite payment registry has been moved to a separate package we need to remove the fallback in V1 registry. Also set the type for strategy to setup correctly, since as of now the fallback loads credit-card strategy.

## Testing / Proof
- circle


bc-peng (approved):
Tested on my employe store. The offsite strategy is working as expected.
https://user-images.githubusercontent.com/88361607/209892666-dc64b253-b959-4137-a768-47e85680dff0.mov

PR #1751, opened 2022-12-30
feat(payment): PAYPAL-1864 created paypal commerce package

serhii-tkachenko:
## What?
Created PayPalCommerce own package:
- created initial package structure;
- added paypal commerce request sender;
- added paypal commerce script loader;
- added paypal commerce customer strategy;

## Why?
To reduce amount of PayPalCommerce code in core package in the future

## Testing / Proof
Unit tests
Manual tests

Here how paypal commerce customer strategy works on Checkout page:
Screenshot 2022-12-29 at 11 57 24


PavlenkoM (approved):

.github/CODEOWNERS
```diff
@@ -5,3 +5,4 @@
 /packages/core @bigcommerce/checkout
 /packages/payment-integration @bigcommerce/checkout
 /packages/payment-integration-test-utils @bigcommerce/checkout
+/packages/paypal-commerce @bigcommerce/kyiv-payments-team
```
animesh1987, line 8: Thanks for doing this, since you are here can you please remove line 2. Thanks 🙇

packages/core/src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
```diff
@@ -9,6 +9,7 @@ export interface ClickActions {
     resolve(): Promise<void>;
 }
 
+// Moved to package
```
animesh1987, line 12: What does this comment mean?

packages/offsite-integration/jest.config.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
     globals: {
         'ts-jest': {
             tsconfig: '<rootDir>/tsconfig.spec.json',
+            diagnostics: false,
```
animesh1987, line 7: What does this change?

packages/payment-integration-api/src/common/types/webpack.d.ts
```diff
@@ -0,0 +1 @@
+declare const LIBRARY_VERSION: string;
```
animesh1987, line 1: Do we need to move this declaration here?

packages/payment-integration-api/src/payment/payment-method.ts
```diff
@@ -1,7 +1,7 @@
 import PaymentMethodConfig from './payment-method-config';
 import InitializationStrategy from './payment-method-initialization-strategy';
 
-export default interface PaymentMethod {
+export default interface PaymentMethod<T = any> {
```
animesh1987, line 4: Can we not use `any`? And come up with better generic typing? Or is that too big a task?

packages/paypal-commerce-integration/.eslintrc.json
```diff
@@ -0,0 +1,42 @@
+{
+    "extends": ["../../.eslintrc.json"],
+    "ignorePatterns": ["!**/*"],
+    "overrides": [
+        {
+            "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
+            "rules": {
+                "@typescript-eslint/naming-convention": "off",
```
animesh1987, line 8: This seems ok as of now. But we should try and fix whatever we can for now, and maybe create items to fix these items as well.

.github/CODEOWNERS
serhii-tkachenko: Sure

packages/core/src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
serhii-tkachenko: This comment means that the interface already exist in paypal commerce package and can be removed in the nearest feature. The comments were left to understand which interfaces didn't exist in paypal commerce package

packages/payment-integration-api/src/common/types/webpack.d.ts
serhii-tkachenko: This declaration needed to remove IDE errors in places where LIBRARY_VERSION in use

serhii-tkachenko: We should have this declaration here because LIBRARY_VERSION was duplicated into payment-integration-api project to avoid using same variable in other packages from core package

packages/payment-integration-api/src/payment/payment-method.ts
serhii-tkachenko: I just copy and paste the same implementation as we have in core package. I don't think that we should searching about better type in this pr, because the pr has another goal

packages/offsite-integration/jest.config.js
serhii-tkachenko: This change removes jest errors what occurs due to global LIBRARY_VERSION variable in payment-integration-api package

packages/core/src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
serhii-tkachenko: I will remove this comments, because the file should be clear and understandable for each developer, not only for me

packages/paypal-commerce-integration/.eslintrc.json
serhii-tkachenko: I removed almost all ts-lint naming-convention errors, but we should keep this rule off due to the bigpay or paypal sdk specific variable names. Just don't want to use ts-ignore all over the place

packages/core/src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
animesh1987: Sounds good, maybe good to add a more meaningful commit. Also we have an issue right now if we move types that need to be exported as part of SDK they need to be duplicated in `core` or else they are not exported correctly as part of SDK build.

.github/CODEOWNERS
```diff
@@ -1,5 +1,4 @@
 * @bigcommerce/checkout
-* @bigcommerce/kyiv-payments-team
```
animesh1987, line 2: Thanks 🙇

animesh1987 (approved):
Nice work @serhii-tkachenko

packages/core/src/payment/strategies/paypal-commerce/paypal-commerce-sdk.ts
serhii-tkachenko: As far as I understand we can remove that types with all other code that we have in core package after moving all other code to its own package. Right?

PR #1760, opened 2023-01-06
feat(payment): STRIPE-194 Clear Prefilled Shipping Address Information If Shopper Logs Out Of Link

AlanMunozN:
## What? [STRIPE-194](https://bigcommercecloud.atlassian.net/browse/STRIPE-194)
Destroy the StripeShippingElement and create another when the element exists

## Why?
StripeShippingElement is created in blank the first time since there is no previous shipping info and when the StripeLink user logs out and returns to the step, the last StripeShippingElement is called (because we have one element already created), and the Stripe get method doesn't accept parameters by default, that's why the fields will render in the blank(because last StripeShippingElement was created with no parameters filled by default), To fix this we are deleting the previous element created and creating it again with new default values, with this approach we can prevent it from leaving blank

## Testing / Proof
 https://drive.google.com/file/d/1YZSS76WX_nUta-Qqk8D15q4gg-GsEz6c/view?usp=sharing
 
[STRIPE-194]: https://bigcommercecloud.atlassian.net/browse/STRIPE-194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/shipping/strategies/stripe-upe/stripe-upe-shipping-strategy.spec.ts
+++ b/packages/core/src/shipping/strategies/stripe-upe/stripe-upe-shipping-strategy.spec.ts
@@ -21,6 +21,7 @@ import {
 } from '../../../payment/strategies/stripe-upe';
 import {
     getShippingStripeUPEJsMock,
+    getShippingStripeUPEJsMockWithAnElementCreated,
     getStripeUPEShippingInitializeOptionsMock,
 } from '../../../shipping/strategies/stripe-upe/stripe-upe-shipping.mock';
 import ConsignmentActionCreator from '../../consignment-action-creator';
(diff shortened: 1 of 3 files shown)


JavvmBC (approved):
LGTM

BC-ISilva:
💚 Tested!

QE Details.
PR has been validated, shipping address details from shopper still populated/available after sigh out/log out from stripe link.

Following scenarios were validated.

- Shopper logout from Stripe Link and then try to edit Shipping address.
- Shopper login to Stripe Link and then try to edit Shipping address.
- Shopper is not login to Stripe Link and edit shipping address.

Please review attached word file with validation details.
[STRIPE-194 - Keep Shipping Address Information If Shopper Logs Out Of Link.docx](https://github.com/bigcommerce/checkout-sdk-js/files/10430311/STRIPE-194.-.Keep.Shipping.Address.Information.If.Shopper.Logs.Out.Of.Link.docx)

bc-peng (approved):
LGTM, 👍

animesh1987:
Hey @AlanMunozN as per the video when a shopper logs out of link, and goes to shipping step there is a pre-filled shipping form? Is that the correct behaviour? Or we need to reset the shipping address as well 🤔

BC-AOrtiz:
> Hey @AlanMunozN as per the video when a shopper logs out of link, and goes to shipping step there is a pre-filled shipping form? Is that the correct behaviour? Or we need to reset the shipping address as well 🤔

Hi @animesh1987 the pre-filled shipping is the correct behavior, we did in that way to be similar like googlepay works

PR #1765, opened 2023-01-10
fix(payment): INT-7103 SquareV2: Assume that `postalCode` is valid unless it changes

mauricio-sg:
## What? [INT-7103](https://bigcommercecloud.atlassian.net/browse/INT-7103)
Exempt the postal code input field from validation if it hasn't been rendered yet.

## Why?
There are regions that do not prompt for a postal code, so we need to allow buyers from those regions to continue as long as the other fields are valid. For those regions that prompt for a postal code, as soon as the input is displayed, events will be fired and the field will start to be validated.

## Testing / Proof
https://user-images.githubusercontent.com/4843328/211678871-ab7f4e6a-76ed-4225-9da2-adca82ecd733.mov

[INT-7103]: https://bigcommercecloud.atlassian.net/browse/INT-7103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,3 +5,4 @@
 /packages/payment-integration @bigcommerce/checkout
 /packages/payment-integration-test-utils @bigcommerce/checkout
 /packages/paypal-commerce @bigcommerce/kyiv-payments-team
+/packages/squarev2-integration @bigcommerce/apex-leads
--- a/packages/squarev2-integration/src/squarev2-payment-processor.spec.ts
+++ b/packages/squarev2-integration/src/squarev2-payment-processor.spec.ts
@@ -10,7 +10,7 @@ import { PaymentIntegrationServiceMock } from '@bigcommerce/checkout-sdk/payment
 import { getSquareV2MockFunctions } from './mocks/squarev2-web-payments-sdk.mock';
 import SquareV2PaymentProcessor from './squarev2-payment-processor';
 import SquareV2ScriptLoader from './squarev2-script-loader';
-import { CardFieldNames, CardInputEventTypes, Square } from './types';
+import { Square } from './types';
 
 describe('SquareV2PaymentProcessor', () => {
     let squareV2ScriptLoader: SquareV2ScriptLoader;
@@ -106,11 +106,7 @@ describe('SquareV2PaymentProcessor', () => {
                 containerId: 'card-container',
                 onValidationChange,
             });
-            squareV2MockFunctions.simulateEvent(
-                'focusClassAdded' as CardInputEventTypes,
-                'cardNumber' as CardFieldNames,
-                true,
-            );
+            squareV2MockFunctions.simulateEvent('focusClassAdded', 'cardNumber', true);
 
             expect(onValidationChange).toHaveBeenCalledTimes(1);
             expect(onValidationChange).toHaveBeenCalledWith(false);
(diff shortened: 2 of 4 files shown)


AlanMunozN (approved):

packages/squarev2-integration/src/squarev2-payment-processor.ts
```diff
@@ -133,7 +133,7 @@ export default class SquareV2PaymentProcessor {
             'cardNumber',
             'expirationDate',
             'cvv',
-            'postalCode',
+            /* Let's assume that `postalCode` is valid unless it changes. */
```
animesh1987, line 136: Do we need to have this comment? It feels pretty redundant since we are not checking for that anymore.

packages/squarev2-integration/src/squarev2-payment-processor.spec.ts
```diff
@@ -187,6 +187,34 @@ describe('SquareV2PaymentProcessor', () => {
...
+
+        it("should exempt postalCode from validation if it hasn't been rendered yet", async () => {
+            const onValidationChange = jest.fn();
+
+            await processor.initializeCard({
+                containerId: 'card-container',
+                onValidationChange,
+            });
+            squareV2MockFunctions.simulateEvent(
+                'focusClassAdded' as CardInputEventTypes,
```
animesh1987, line 199: Can we avoid type assertion here?

packages/squarev2-integration/src/squarev2-payment-processor.spec.ts
```diff
@@ -187,6 +187,34 @@ describe('SquareV2PaymentProcessor', () => {
...
+                'expirationDate' as CardFieldNames,
+                true,
+            );
+            squareV2MockFunctions.simulateEvent(
+                'errorClassRemoved' as CardInputEventTypes,
+                'cvv' as CardFieldNames,
+                true,
+            );
+
+            expect(onValidationChange).toHaveBeenCalledTimes(2);
```
animesh1987, line 214: It seems like we have 3 events being fired, and only two times `onValidationChange` is triggered. Can you please explain that?

animesh1987:
Hey @mauricio-sg also this package should be owned, by integrations team in Github codeowners file. Can you please add that to.

packages/squarev2-integration/src/squarev2-payment-processor.ts
mauricio-sg: I left that comment so we don't lose visibility that this field should go there. In theory, the Square team will fix the issue at some point. I can remove it if it makes noise instead.

packages/squarev2-integration/src/squarev2-payment-processor.spec.ts
mauricio-sg: Done.

packages/squarev2-integration/src/squarev2-payment-processor.spec.ts
mauricio-sg: It's because of the [distinctuntilchanged](https://www.learnrxjs.io/learn-rxjs/operators/filtering/distinctuntilchanged) operator.
> Only emit when the current value is different than the last.

The observer in this case `onValidationChange` will only receive a value different than the last. So intead of having something like this:
onValidationChange(false);

BC-BMedina:
💚 Tested! 
QE details: 
JCB and China UnionPay cards are working now with experiment **PROJECT-4113.squarev2_web_payments_sdk** enable, here is the evidence of the regression made for this fix:
[INT-7103 Folder Evidence](https://drive.google.com/drive/folders/1OGIKoEnco47wYM2_Gwzdc_1T3j6dVG6t?usp=sharing)

mauricio-sg:
@animesh1987, your comments were addressed ♻️

packages/squarev2-integration/src/squarev2-payment-processor.spec.ts
animesh1987: Makes sense, so it will fire once with false regardless. Thanks for explanation.

.github/CODEOWNERS
```diff
@@ -5,3 +5,4 @@
 /packages/payment-integration @bigcommerce/checkout
 /packages/payment-integration-test-utils @bigcommerce/checkout
 /packages/paypal-commerce @bigcommerce/kyiv-payments-team
+/packages/squarev2-integration @bigcommerce/apex-integrations
```
animesh1987, line 8:
```suggestion
/packages/squarev2-integration @bigcommerce/apex-leads
```

mauricio-sg: Done.

.github/CODEOWNERS
```diff
@@ -5,3 +5,4 @@
 /packages/payment-integration @bigcommerce/checkout
 /packages/payment-integration-test-utils @bigcommerce/checkout
 /packages/paypal-commerce @bigcommerce/kyiv-payments-team
+/packages/squarev2-integration @bigcommerce/apex-leads
```
vmparra, line 8: Please set this to @bigcommerce/apex-integrations

mauricio-sg: Done.

animesh1987 (approved):

PR #1774, opened 2023-01-17
fix(billing): CHECKOUT-7214 Merge billing address as compared to replace

animesh1987:
## What?
Merge billing address returned from load order endpoint with existing address in state.

## Why?
Since load order call returns a billing address without an id, it causes side effects since we replace the existing object. In order to maintain reference to the existing id we need to merge the two objects.

## Testing / Proof
- Circle

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/billing/billing-address-reducer.ts
+++ b/packages/core/src/billing/billing-address-reducer.ts
@@ -2,8 +2,8 @@ import { Action, combineReducers, composeReducers } from '@bigcommerce/data-stor
 
 import { CheckoutAction, CheckoutActionType } from '../checkout';
 import { clearErrorReducer } from '../common/error';
-import { objectMerge, objectSet, replace } from '../common/utility';
-import { OrderAction, OrderActionType } from '../order';
+import { objectSet, replace } from '../common/utility';
+import { OrderAction } from '../order';
 import { SubscriptionsActionType, UpdateSubscriptionsAction } from '../subscription';
 
 import BillingAddress from './billing-address';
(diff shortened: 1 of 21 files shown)


packages/core/src/billing/billing-address-reducer.ts
```diff
@@ -35,8 +35,9 @@ function dataReducer(
         case BillingAddressActionType.UpdateBillingAddressSucceeded:
         case BillingAddressActionType.ContinueAsGuestSucceeded:
         case CheckoutActionType.LoadCheckoutSucceeded:
-        case OrderActionType.LoadOrderSucceeded:
             return replace(data, action.payload && action.payload.billingAddress);
+        case OrderActionType.LoadOrderSucceeded:
+            return objectMerge(data, action.payload && action.payload.billingAddress);
```
snaderiBC, line 40: This is in depth of the framework/library. Time bomb! ⏲️ 💣

bc-peng (approved):

packages/core/src/billing/billing-address-reducer.ts
animesh1987: Yes, and the funny part was the unit test assume that the mock for getOrder object will have a billing address with valid id.

davidchin: It seems odd that we're replacing or merging order billing address with checkout billing address, as they are essentially different entities, even though their values might be the same. Git history shows that this line was added in #298 to fix a problem. But I think a better fix for that prior problem is to either undo this commit https://github.com/bigcommerce/checkout-sdk-js/pull/283/commits/aba8bd216e82f13006f1889958473df0419e8bf8 or have a separate reducer for order billing address and coupon.

animesh1987: I think it's due to us using the same billing address in order-confirmation page as well? In that case shall we have a separate reference to billing-address for order confirmation page? Therefore billing address has a separate reducer for order confirmation page with its own selector?

davidchin (approved):

PR #1782, opened 2023-01-19
feat(payment): PAYPAL-1882 sending PPC APM V2 gateway instead of PPC APM

serhii-tkachenko:
## What?
Added temporary PayPalCommerceAlternativeMethods transformer to PayPalCommerceAlternativeMethodsV2

## Why?
PayPalCommerceAlternativeMethods gateway should be sent to BigPay without transforming to PayPalCommerce gateway. We have only one possible way how to make it as smooth as possible, transform temporary gateway to what we need. 

This code will be removed in the future.

## Testing / Proof
Unit tests
Manual tests

Here is a screenshot of payment method gateway sent to the BigPay:
Screenshot 2023-01-19 at 13 57 19

## Sibling PR:
bigpay-client-js: https://github.com/bigcommerce/bigpay-client-js/pull/155
bigcommerce: https://github.com/bigcommerce/bigcommerce/pull/50776

### DIFF
--- a/packages/core/src/payment/payment-request-transformer.ts
+++ b/packages/core/src/payment/payment-request-transformer.ts
@@ -164,6 +164,7 @@ export default class PaymentRequestTransformer {
             return { ...paymentMethod, id: CheckoutButtonMethodType.BRAINTREE_PAYPAL };
         }
 
+        // TODO: this block of code should be removed in PAYPAL-1921
         if (
             paymentMethod.gateway === CheckoutButtonMethodType.PAYPALCOMMERCE_APMS &&
             storeConfig?.checkoutSettings.features['PAYPAL-1883.paypal-commerce-split-gateway']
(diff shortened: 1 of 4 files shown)


packages/core/src/payment/payment-request-transformer.ts
```diff
@@ -159,6 +164,16 @@ export default class PaymentRequestTransformer {
             return { ...paymentMethod, id: CheckoutButtonMethodType.BRAINTREE_PAYPAL };
         }
 
+        if (
```
animesh1987, line 167: Do we need to add this logic here? Can we do this where we build the payment request itself?

serhii-tkachenko: I think, it is the best place for this logic due to the code above.

We can't move this logic to places like strategy because payment method loads with provided gateway and method on submitPayment method call. 

This code is temporary due to BigPay refactoring process. It will be removed right after paypalcommercealternativemethods will be separated from paypalcommerce gateway

animesh1987: Is there a ticket for removing this temporary code here?

serhii-tkachenko: Yes. Here it is: https://bigcommercecloud.atlassian.net/browse/PAYPAL-1921

animesh1987 (approved):

PR #1784, opened 2023-01-19
refactor(payment): PAYPAL-1918 moved LoadingIndicator from core to ui package

serhii-tkachenko:
## What?
Moved LoadingIndicator from core to ui package

## Why?
We should be able to use LoadingIndicator in any checkout-sdk package

## Testing / Proof
Unit tests
Manual tests


PavlenkoM (approved):

vitalii-koshovyi (approved):

bc-nick (approved):

packages/ui/src/loading-indicator/loading-indicator.ts
```diff
@@ -9,19 +9,19 @@ const DEFAULT_STYLES: LoadingIndicatorStyles = {
 const ROTATION_ANIMATION = 'embedded-checkout-loading-indicator-rotation';
 
 export default class LoadingIndicator {
-    private _container: HTMLElement;
-    private _indicator: HTMLElement;
-    private _styles: LoadingIndicatorStyles;
+    private container: HTMLElement;
+    private indicator: HTMLElement;
+    private styles: LoadingIndicatorStyles;
```
bc-peng, line 14: LGTM. Is this name change intentional?

serhii-tkachenko: Yes. This changes are related to `naming-convention` eslint rule what should be removed in the future.

Eslint rule in ui package: https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/ui/.eslintrc.json#L8

packages/ui/src/loading-indicator/loading-indicator.spec.ts
```diff
@@ -25,6 +25,22 @@ describe('LoadingIndicator', () => {
...
+    it('renders loading indicator with provided styles prop', () => {
+        indicator = new LoadingIndicator({
+            styles: {
+                size: 10,
+            },
+        });
+
+        indicator.show(parentId);
+
+        const child = parent.firstChild as HTMLElement;
```
animesh1987, line 37: Can we avoid using `as` here? Seems like we are adding the eslint rule just to bypass this condition check?

serhii-tkachenko: Yes, I added extra rule only for spec.ts files. We have the same rule added in several other packages.

serhii-tkachenko: I don't think that we should spend extra time to avoid using `as` here

animesh1987: We have that rule in other packages since we added eslint last month, and to fix all the issues is a major task. 
But the goal going forward is to avoid adding rule overrides.

serhii-tkachenko: Is it possible to skip it for now? I will update this tests and remove extra eslint rules right after removing paypal commerce implementation from code package

animesh1987:
```
function isHtmlElement(value?: HTMLElement): value is HTMLElement {
    return value !== undefined && value instanceof HTMLElement;
}
```

should be similar to this.

serhii-tkachenko: I found a way how to make it without typeguard. Thanks!

animesh1987 (approved):
Nice work @serhii-tkachenko.

PR #1791, opened 2023-01-24
feat(payment): PAYPAL-1898 added CartRequestSender to payment integration api package

serhii-tkachenko:
## What?
Added CartRequestSender to payment integration api package

## Why?
We don't have an ability to migrate PayPalCommerce strategies to their own package because payment-integration-api package doesn't have buy now cart creation method what core package has.

## Testing / Proof
Unit tests

### DIFF
--- a/packages/core/src/payment-integration/default-payment-integration-service.ts
+++ b/packages/core/src/payment-integration/default-payment-integration-service.ts
@@ -159,9 +159,13 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
         return this._storeProjection.getState();
     }
 
-    async createBuyNowCart(buyNowCartRequestBody: BuyNowCartRequestBody): Promise<Cart> {
+    async createBuyNowCart(
+        buyNowCartRequestBody: BuyNowCartRequestBody,
+        options?: RequestOptions,
+    ): Promise<Cart> {
         const { body: buyNowCart } = await this._cartRequestSender.createBuyNowCart(
             buyNowCartRequestBody,
+            options,
         );
 
         return buyNowCart;
--- a/packages/core/src/payment-integration/default-payment-integration-service.ts
+++ b/packages/core/src/payment-integration/default-payment-integration-service.ts
@@ -1,5 +1,7 @@
 import {
     BillingAddressRequestBody,
+    BuyNowCartRequestBody,
+    Cart,
     HostedForm,
     HostedFormOptions,
     InitializeOffsitePaymentConfig,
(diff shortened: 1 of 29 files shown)


vitalii-koshovyi (approved):

PavlenkoM (approved):

packages/payment-integration-api/src/cart/buy-now-cart-request-body.ts
```diff
@@ -0,0 +1,18 @@
+import { CartSource } from './cart-source';
+
+interface LineItem {
+    productId: number;
+    quantity: number;
```
animesh1987, line 5: Do we need to have variant ID as well here?

packages/core/src/payment-integration/default-payment-integration-service.ts
```diff
@@ -154,4 +158,12 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
 
         return this._storeProjection.getState();
     }
+
+    async createBuyNowCart(buyNowCartRequestBody: BuyNowCartRequestBody): Promise<Cart> {
```
animesh1987, line 162: https://github.com/bigcommerce/checkout-sdk-js/pull/1791/files#diff-a9aaa3275847637b88e61e83799c3ca27bb47afe157ff50fe27345ea9f6e4042R58 we have options defined in the interface but not here? Is that intentional?

packages/payment-integration-api/src/cart/buy-now-cart-request-body.ts
serhii-tkachenko: As far as I know `variantId === optionId`

serhii-tkachenko: For this case we have `optionSelections`

packages/core/src/payment-integration/default-payment-integration-service.ts
serhii-tkachenko: I forgot to add options for this method. Yeah, the `options?: RequestOptions` should be added here

serhii-tkachenko: Fixed

packages/payment-integration-api/src/cart/buy-now-cart-request-body.ts
animesh1987: There are various types of combinations as you can see here https://github.com/bigcommerce/bigcommerce/blob/master/app/Http/Storefront/Api/Cart/CartValidator.php#L88 it's an option.

serhii-tkachenko: We don't need it here

animesh1987: Maybe just add it as an optional now. Since partners might want to use if they can and storefront API allows for that.

serhii-tkachenko: Sure

animesh1987 (approved):
Thanks for the patience and discussions @serhii-tkachenko. LGTM 👍

PR #1793, opened 2023-01-25
fix(payment): PAYMENTS-8239 puts recommended shipping option first in the applepay payment modal

bc-amit:
## What?
Puts recommended shipping option before non-recommended ones in the applepay shipping method dropdown.

## Why?
Before this change, when Merchant has selected the `Least expensive, excluding pickup in store` option in the checkout settings in the Control Panel. While shopper is paying via ApplePay, `Pickup in store` is the default selected option in the ApplePay payment modal.

The expectation is for `Pickup in store` to not be the default selected option. 

The issues occurs because triggering the selection of the recommended option here: https://github.com/bigcommerce/checkout-sdk-js/pull/1793/files#diff-1e21ac0b56385d124ac395b7ea974cf5c4b5bfe4975ad290180a53c6e6cfb4abR317
Doesn't select the appropriate select option in applepay modal. It continues to show the shipping option list received from the API as is, making it seem like `Pickup in store` is selected.  

In this PR we have put the shipping option that is `isRecommended` before the other options, making the UI consistent with the selection.

## Testing / Proof
Additional tests. 

https://user-images.githubusercontent.com/107909200/214488794-93c95d47-6021-4046-9ce2-a1f5605d8e0c.mov 

### DIFF
--- a/packages/core/src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
+++ b/packages/core/src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
@@ -323,14 +323,27 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
               ]
             : [];
 
-        unselectedOptions?.forEach((option) =>
-            shippingOptions.push({
-                label: option.description,
-                amount: `${option.cost.toFixed(decimalPlaces)}`,
-                detail: option.additionalDescription,
-                identifier: option.id,
-            }),
-        );
+        unselectedOptions
+            ?.filter((option) => option.isRecommended)
+            ?.forEach((option) =>
+                shippingOptions.push({
+                    label: option.description,
+                    amount: `${option.cost.toFixed(decimalPlaces)}`,
+                    detail: option.additionalDescription,
+                    identifier: option.id,
+                }),
+            );
+
+        unselectedOptions
+            ?.filter((option) => !option.isRecommended)
+            ?.forEach((option) =>
+                shippingOptions.push({
+                    label: option.description,
+                    amount: `${option.cost.toFixed(decimalPlaces)}`,
+                    detail: option.additionalDescription,
+                    identifier: option.id,
+                }),
+            );
 
         if (!isShippingOptions(availableOptions)) {
             throw new Error('Shipping options not available.');
--- a/packages/core/src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
+++ b/packages/core/src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
@@ -323,9 +323,11 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
               ]
             : [];
 
-        unselectedOptions
-            ?.filter((option) => option.isRecommended)
-            ?.forEach((option) =>
+        if (unselectedOptions) {
+            [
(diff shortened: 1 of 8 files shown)


packages/core/src/customer/strategies/apple-pay/apple-pay-customer-strategy.ts
```diff
@@ -323,14 +323,27 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
...
+        unselectedOptions
+            ?.filter((option) => !option.isRecommended)
+            ?.forEach((option) =>
+                shippingOptions.push({
+                    label: option.description,
+                    amount: `${option.cost.toFixed(decimalPlaces)}`,
+                    detail: option.additionalDescription,
+                    identifier: option.id,
+                }),
+            );
```
bc-peng, line 346: Can we use `sortBy()` from `lodash` to avoid writing a similar block?

https://lodash.com/docs/4.17.15#sortBy
Here is a sample usage:
```
var users = [
  { 'user': 'fred',   'age': 48, flag:false },
  { 'user': 'barney', 'age': 36, flag:true },
  { 'user': 'fred',   'age': 40, flag:false },
  { 'user': 'barney', 'age': 34, flag:false }
];
 
_.sortBy(users, ['flag']).reverse();

// output: Array (4 items)
// 0: Object {user: "barney", age: 36, flag: true}
// 1: Object {user: "barney", age: 34, flag: false}
// 2: Object {user: "fred", age: 40, flag: false}
// 3: Object {user: "fred", age: 48, flag: false}
```

bc-amit: @bc-peng I think there can be an ordering issue while sorting this way. Take following example:

```js

// Users from API
var users = [
  { 'user': 'fred',   'age': 30, flag: false },
  { 'user': 'barney', 'age': 32, flag: true },
  { 'user': 'fred',   'age': 36, flag: false },
  { 'user': 'barney', 'age': 40, flag: false }
];
 
_.sortBy(users, ['flag']).reverse();

// After sorting
// 0: Object {user: "barney", age: 32, flag: true}
// 1: Object {user: "barney", age: 40, flag: false}
// 2: Object {user: "fred", age: 36, flag: false}
// 3: Object {user: "fred", age: 30, flag: false}

// What we actually want
// 0: Object {user: "barney", age: 32, flag: true}
// 1: Object {user: "fred", age: 30, flag: false}
// 2: Object {user: "fred", age: 36, flag: false}
// 3: Object {user: "barney", age: 40, flag: false}
```

Options available in Control Panel for checkout settings:
Screen Shot 2023-01-30 at 2 50 21 pm

I think, we would want to keep the ordering from the API response intact, apart from bumping the `isRecommended` options. 🤔

bc-peng: In that case, maybe
```
shippingOptions = [
     unselectedOptions?.filter((option) => option.isRecommended),
     ...unselectedOptions?.filter((option) => option.!isRecommended),
]
```

bc-amit: @bc-peng updated! 🙏

bc-peng (approved):
LGTM, thanks.

PR #1796, opened 2023-01-26
fix(payment): PAYPAL-1930 add intent to braintree config

bc-nick:
## What?
Add intent to braintree config

## Why?
To make it possible to switch intent in Braintree SDK configuration due to the changes in CP Braintree Settings form

## Testing / Proof
Unit tests
Manual tests

## Depends on
https://github.com/bigcommerce/bigcommerce/pull/50910

### DIFF
--- a/packages/core/src/payment/strategies/braintree/braintree-sdk-creator.ts
+++ b/packages/core/src/payment/strategies/braintree/braintree-sdk-creator.ts
@@ -11,6 +11,7 @@ import {
     BraintreeError,
     BraintreeHostedFields,
     BraintreeHostedFieldsCreatorConfig,
+    BraintreeIntent,
     BraintreeModule,
     BraintreePaypal,
     BraintreePaypalCheckout,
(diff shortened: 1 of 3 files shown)


serhii-tkachenko (approved):
LGTM! 👍

packages/core/src/payment/strategies/braintree/braintree-sdk-creator.ts
```diff
@@ -71,7 +72,7 @@ export default class BraintreeSDKCreator {
     }
 
     async getPaypalCheckout(
-        config: { currency: string },
+        config: { currency: string; indent?: BraintreeIntent },
```
animesh1987, line 75: Why are we calling it `indent` here? Is that a typo?

serhii-tkachenko: Yeah, its a typo :(

serhii-tkachenko:
Closed the PR, because we should make some extra work to make it working correctly. Will create another PR after test

PR #1798, opened 2023-01-27
chore(common): CHECKOUT-6271 Make changes to circle config

animesh1987:
## What?
- Move lint to a separate process
- Remove test artifact uploads
- Split test runs into two jobs one for core package and one for the rest of the packages
- Split test runs of core package into two parallel streams

## Why?
At the moment circlet CI runs take around 10 mins approximately, we can optimise this and have a quicker turn around in our CI steps.

## Testing / Proof
- Circle

Updated CI times for PRs

Screen Shot 2023-01-30 at 10 03 14 pm

@bigcommerce/checkout @icatalina

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,7 +145,8 @@ workflows:
             branches:
               only: master
           requires:
-            - test
+            - test-core
+            - test-packages
             - lint
             - build
             - ci/validate-commits
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,16 +26,19 @@ jobs:
     steps:
       - ci/pre-setup
       - node/npm-install
+      - run: mkdir ~/junit
       - run:
           name: "Run unit tests for core package"
           command: |
             TEST=$(circleci tests glob "packages/core/**/*.spec.js" | circleci tests split --split-by=timings)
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -119,7 +145,9 @@ workflows:
             branches:
               only: master
           requires:
-            - test
+            - test-core
+            - test-packages
+            - lint
             - build
             - ci/validate-commits
```
icatalina, line 152: I guess we could replace this with the tasks creating the packages? 🤔 

eg.
```yaml
requires:
  - ci/notify-success
```

icatalina (commented):
For eslint we have to add the `maxWarnings: 0` option in nx:

https://nx.dev/packages/linter/executors/eslint#maxwarnings

https://github.com/bigcommerce/eslint-config/tree/master/packages/eslint-config#usage

image

.circleci/config.yml
animesh1987: Not sure, because we don't build packages separately. Core package still is the orchestrator or the builder if you call it that imports everything and builds everything.

icatalina (approved):

bc-peng (approved):

PR #1808, opened 2023-01-30
feat(payment): PAYPAL-1906 Create braintree checkout-sdk package

PavlenkoM:
## What?
...

## Why?
...

## Testing / Proof
...


packages/braintree-integration/src/braintree-paypal/braintree-paypal-customer-strategy.ts
```diff
@@ -1,78 +1,78 @@
...
+        private braintreeHostWindow: BraintreeHostWindow,
     ) {}
 
-    async initialize(options: CustomerInitializeOptions): Promise<InternalCheckoutSelectors> {
+    async initialize(
+        options: CustomerInitializeOptions & WithBraintreePaypalCustomerInitializeOptions,
+    ): Promise<void> {
         const { braintreepaypal, methodId } = options;
-        const { container, onError, buttonHeight = 40 } = braintreepaypal || {};
+        const { container, buttonHeight, onError } = braintreepaypal || {};
```
vitalii-koshovyi, line 45: Is default height is needed?

vitalii-koshovyi: oh see, it below

vitalii-koshovyi (approved):
Thumbs Up for braintree-integration-service.ts

packages/braintree-integration/src/braintree-paypal/braintree-paypal-customer-strategy.ts
PavlenkoM: This height was added by checkout team, I think it is default option for all buttons on customer step
But in future styles for braintree buttons will get from initialisation data

serhii-tkachenko (approved):
Nice work! 👍

packages/payment-integration-api/src/payment-integration-selectors.ts
```diff
@@ -49,7 +49,7 @@ export default interface PaymentIntegrationSelectors {
     getPaymentRedirectUrlOrThrow(): string;
 
     getPaymentMethod(methodId: string, gatewayId?: string): PaymentMethod | undefined;
-    getPaymentMethodOrThrow(methodId: string, gatewayId?: string): PaymentMethod;
+    getPaymentMethodOrThrow<T = any>(methodId: string, gatewayId?: string): PaymentMethod<T>;
```
bc-peng, line 52: Can I know why adding `<T = any>` here?

PavlenkoM: Because this method returns instance of `PaymentMethod` and this interface has generic type for `initializationData`
[https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/payment-integration-api/src/payment/payment-method.ts](https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/payment-integration-api/src/payment/payment-method.ts)

And if we need to [describe correct type](https://github.com/bigcommerce/checkout-sdk-js/pull/1808/files#diff-152a1fd0ab36037c134c9d74cdfdadc7780a8b2f5b053a95a0ebf3f0d11dfcfcR68) for `PaymentMethod.initializationData` we need to describe it as generic
any by default because it is using by default in [original](https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/payment-integration-api/src/payment/payment-method.ts) `PaymentMethod` interface

animesh1987: Thanks @PavlenkoM can you try `T = unknown` instead of `any` here.

PavlenkoM: Yes, you are right, with `unknown` looks better

PavlenkoM: changing `any` to `unknown` effects other providers
lets leave `any` and change to `unknown` in separate PR?
Screenshot 2023-03-15 at 16 31 34

PavlenkoM: PR with generic type fix
[https://github.com/bigcommerce/checkout-sdk-js/pull/1907](https://github.com/bigcommerce/checkout-sdk-js/pull/1907)

animesh1987 (approved):

PR #1813, opened 2023-01-31
refactor(payment): PAYPAL-1891 added PayPalCommerceVenmoButtonStrategy to paypal commerce integration package

serhii-tkachenko:
## What?
Added PayPalCommerceVenmoButtonStrategy to paypal commerce integration package

## Why?
As part of migration paypal commerce code from core to paypal commerce integration package

## Testing / Proof
Unit tests
Manual tests

### DIFF
--- a/packages/paypal-commerce-integration/src/index.ts
+++ b/packages/paypal-commerce-integration/src/index.ts
@@ -27,3 +27,11 @@ export { WithPayPalCommerceCreditButtonInitializeOptions } from './paypal-commer
  */
 export { default as createPayPalCommerceInlineButtonStrategy } from './paypal-commerce-inline/create-paypal-commerce-inline-button-strategy';
 export { WithPayPalCommerceInlineButtonInitializeOptions } from './paypal-commerce-inline/paypal-commerce-inline-button-initialize-options';
+
+/**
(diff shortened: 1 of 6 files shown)


packages/paypal-commerce-integration/src/paypal-commerce-types.ts
```diff
@@ -173,14 +173,24 @@ export interface PayPalCommerceButtonsOptions {
     createOrder(): Promise<string>;
     onApprove(
         data: ApproveCallbackPayload,
-        actions: ApproveCallbackActions,
+        actions?: ApproveCallbackActions,
```
vitalii-koshovyi, line 176: should this field be optional?

vitalii-koshovyi (approved):

packages/paypal-commerce-integration/src/paypal-commerce-types.ts
serhii-tkachenko: Actually I don't think so. As far as I remember, there are several providers what don't use actions in onApprove callback. However it does not mean that the field is not provided by PayPal... I will make it required. 

Thanks for your comment!

PR #1825, opened 2023-02-08
feat(checkout): CHECKOUT-7231 Add checkoutUXSettings to interface

bc-peng:
## What?
Add `checkoutUXSettings` to the `CheckoutSettings` interface.

## Why?
We will introduce a few UX options to merchants in the following weeks.

## Related PR
- https://github.com/bigcommerce/bigcommerce/pull/51091

## Testing / Proof
CI checks.

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -92,6 +92,7 @@ export interface StoreCurrency {
 export interface CheckoutSettings {
     features: { [featureName: string]: boolean };
     checkoutBillingSameAsShippingEnabled: boolean;
+    checkoutUXSettings: { [settingName: string]: boolean };
     enableOrderComments: boolean;
     enableTermsAndConditions: boolean;
     googleMapsApiKey: string;
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -94,7 +94,7 @@ export type UXSettingNames = 'walletButtonsOnTop';
 export interface CheckoutSettings {
     features: { [featureName: string]: boolean };
     checkoutBillingSameAsShippingEnabled: boolean;
-    checkoutUXSettings: { [key in UXSettingNames]: boolean };
+    CheckoutUserExperienceSettings: { [key in UXSettingNames]: boolean };
     enableOrderComments: boolean;
     enableTermsAndConditions: boolean;
     googleMapsApiKey: string;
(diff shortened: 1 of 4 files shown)


packages/payment-integration-api/src/config/config.ts
```diff
@@ -91,6 +91,7 @@ export interface StoreCurrency {
 export interface CheckoutSettings {
     features: { [featureName: string]: boolean };
     checkoutBillingSameAsShippingEnabled: boolean;
+    checkoutUXSettings: { [settingName: string]: boolean };
```
animesh1987, line 94: Does it make sense to have a list of `settingNames`? Since this can mean 
```
{
     checkoutUXSettings: {
          'test': true
     }
}
```
is valid?

bc-peng: I think we treat it the same way as the feature flags. Your sample seems OK.
Otherwise, we need to update SDK every time we create a new option.

animesh1987: Yes I think that should be ok, because we are adding other settings. Just need to be mindful. 
For feature flags its a list of strings so easier.

PascalZajac:
🍹 Totally bike-shedding but I dislike initialisms in names, could we use `checkoutUserExperienceSettings`?

animesh1987:
> 🍹 Totally bike-shedding but I dislike initialisms in names, could we use `checkoutUserExperienceSettings`?

Thanks for the timely feedback, since I hadn't merged the BCApp PR where initialisms were created. Have updated the [BCApp PR](https://github.com/bigcommerce/bigcommerce/pull/51091/files#diff-415f063a7b1d80d094fac51d8a6c217e6a4d4b661175eb3adf656a68c55b8837R85) so we can update accordingly.

packages/core/src/config/config.ts
```diff
@@ -89,9 +89,12 @@ export interface StoreCurrency {
     thousandsSeparator: string;
 }
 
+export type UserExperienceSettingNames = 'walletButtonsOnTop';
+
 export interface CheckoutSettings {
     features: { [featureName: string]: boolean };
     checkoutBillingSameAsShippingEnabled: boolean;
+    CheckoutUserExperienceSettings: { [key in UserExperienceSettingNames]: boolean };
```
bc-peng, line 97: I have updated the code based on ther server side PR. 🙇🏻‍♂️
cc @PascalZajac @animesh1987

animesh1987 (approved):

PR #1828, opened 2023-02-08
chore(common): CHECKOUT-7279 Update to node 16

animesh1987:
## What?
Update node version of sdk

## Why?
Node v14 which is the version currently in use is coming to end of life, hence we should be making this upgrade to maintain the support.

## Testing / Proof
- Tests

@bigcommerce/checkout


animesh1987:
Closing this PR, since realised that we might need to upgrade nx as a prerequisite.

PR #1831, opened 2023-02-09
feat(checkout): CHECKOUT-7231 Update checkoutUserExperienceSettings

bc-peng:
## What?
The name should be `checkoutUserExperienceSettings`, not `CheckoutUserExperienceSettings`.

## Why?
So it is consistent with the other names of the interface.

## Testing / Proof
CI checks.

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -94,7 +94,7 @@ export type UserExperienceSettingNames = 'walletButtonsOnTop';
 export interface CheckoutSettings {
     features: { [featureName: string]: boolean };
     checkoutBillingSameAsShippingEnabled: boolean;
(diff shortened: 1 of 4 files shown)


packages/core/src/config/config.ts
```diff
@@ -94,7 +94,7 @@ export type UserExperienceSettingNames = 'walletButtonsOnTop';
 export interface CheckoutSettings {
     features: { [featureName: string]: boolean };
     checkoutBillingSameAsShippingEnabled: boolean;
-    CheckoutUserExperienceSettings: { [key in UserExperienceSettingNames]: boolean };
+    checkoutUserExperienceSettings: { [key in UserExperienceSettingNames]: boolean };
```
bc-peng, line 97: I think `UserExperienceSettingNames` is OK.

animesh1987 (approved):

PR #1835, opened 2023-02-13
feat(payment): PAYMENTS-8493 add untrusted shipping address card veri…

bc-chenli:
## What?
We are mapping `untrusted_shipping_address_card_verification_mode` to cardIntruemt mode.

This PR is a dependency of https://github.com/bigcommerce/checkout-js/pull/1195

## Why?
At BigPay, we have updated the trusted shipping endpoint to return instruments that include `untrusted_shipping_address_card_verification_mode`. We are going to use this attribute to determine which field should be shown on the payment form when untrusted shipping is detected.

For example, if a mode value `cvv` returns, the payment form should only show cvv input box, which means the shopper can only use cvv to finish the payment. If a mode value `number` returns, the number filed will be required.

For the logic how we gonna use this mode value, please refer to PR 
https://github.com/bigcommerce/checkout-js/pull/1195

## Testing / Proof
Local testing
untrusted shipping + cvv only
image
untrusted shipping + pan
image

Unit tests

## Depends on
none

## Dependency of
https://github.com/bigcommerce/checkout-js/pull/1195

### DIFF
--- a/packages/core/src/payment/instrument/instrument-response-body.ts
+++ b/packages/core/src/payment/instrument/instrument-response-body.ts
@@ -1,4 +1,4 @@
-import PaymentInstrument from './instrument';
+import PaymentInstrument, { UntrustedShippingCardVerificationType } from './instrument';
 
 export interface InstrumentError {
     code: number;
@@ -35,6 +35,7 @@ export interface CardInternalInstrument extends BaseInternalInstrument {
     iin: string;
     last_4: string;
     method_type: 'card';
+    untrusted_shipping_address_card_verification_mode: UntrustedShippingCardVerificationType;
 }
 
 export interface PayPalInternalInstrument extends BaseInternalAccountInstrument {
--- a/packages/core/src/payment/instrument/instrument.mock.ts
+++ b/packages/core/src/payment/instrument/instrument.mock.ts
@@ -1,6 +1,7 @@
 import PaymentInstrument, {
     CardInstrument,
     InstrumentRequestContext,
+    UntrustedShippingCardVerificationType,
     VaultAccessToken,
 } from './instrument';
 import {
@@ -35,6 +36,7 @@ export function getInstruments(): PaymentInstrument[] {
             trustedShippingAddress: true,
             defaultInstrument: true,
             method: 'credit_card',
+            untrustedShippingCardVerificationMode: UntrustedShippingCardVerificationType.PAN,
             type: 'card',
         },
         {
@@ -48,6 +50,7 @@ export function getInstruments(): PaymentInstrument[] {
             trustedShippingAddress: false,
             defaultInstrument: false,
             method: 'credit_card',
+            untrustedShippingCardVerificationMode: UntrustedShippingCardVerificationType.PAN,
             type: 'card',
         },
         {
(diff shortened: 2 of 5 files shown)


packages/core/src/payment/instrument/instrument-response-body.ts
```diff
@@ -35,6 +35,7 @@ export interface CardInternalInstrument extends BaseInternalInstrument {
     iin: string;
     last_4: string;
     method_type: 'card';
+    untrusted_shipping_address_card_verification_mode: string;
```
Tharaae, line 38: Why not using the type enum here?
```suggestion
    untrusted_shipping_address_card_verification_mode: UntrustedShippingCardVerificationType;
```

packages/core/src/payment/instrument/instrument.mock.ts
```diff
@@ -162,6 +167,7 @@ export function getInternalInstrumentsResponseBody(): InternalInstrumentsRespons
                 default_instrument: true,
                 method: 'credit_card',
                 method_type: 'card',
+                untrusted_shipping_address_card_verification_mode: 'pan',
```
Tharaae, line 170: Same here, why not use the enum value?
```suggestion
                untrusted_shipping_address_card_verification_mode: UntrustedShippingCardVerificationType.PAN,
```

packages/core/src/payment/instrument/map-to-card-instrument.spec.ts
```diff
@@ -13,6 +14,7 @@ describe('mapToCardInstrument', () => {
             expiry_year: '2020',
             expiry_month: '12',
             default_instrument: false,
+            untrusted_shipping_address_card_verification_mode: 'pan',
```
Tharaae, line 17: Same here.

packages/core/src/payment/instrument/instrument.mock.ts
bc-chenli: I think this is the mock data for the request which is a string acutally

packages/core/src/payment/instrument/map-to-card-instrument.spec.ts
bc-chenli: Same reason as above.

packages/core/src/payment/instrument/instrument-response-body.ts
bc-chenli: It is a response body object, we should use string to reflect what we received.

bc-chenli: Updated, since it is simpler to use enum here

bc-peng (commented):
LGTM, 👏.
It looks like we need to recreate the same enum in `checkout-js` to use its value, is that a right pratice? @animesh1987

animesh1987:
> It looks like we need to recreate the same enum in checkout-js to use its value, is that a right practice?

Thanks for looking into this, I had a chat with Chen and there is an issue with exporting enums. So we need to duplicate enums in the client. For reference https://stackoverflow.com/questions/63701326/exporting-enum-in-d-ts-leads-to-cant-resolve-error

animesh1987 (approved):

PR #1848, opened 2023-02-16
[SPIKE] feat(payment): hack braintree options

animesh1987:
## What?
Expose braintree options

## Why?
In order to provide more information to SDK clients for error states

## Testing / Proof
N/A


animesh1987:
@serhii-tkachenko can you please take a look and see if this works as part of extending the interface.

serhii-tkachenko:
Hey @animesh1987. Thanks for your effort!
As far as I see, it should work without any issues 👍

PR #1854, opened 2023-02-16
feat(payment): INT-7503 add floating labels and change little css details for stripe link

AlanMunozN:
## What? [INT-7503](https://bigcommercecloud.atlassian.net/browse/INT-7503)
Add floating labels behind an experiment

## Why?
for StripeLink, we are looking to match the same styles used by BC

## Testing / Proof
**Customer**
**BC  / vs / LINK**
image
**Shipping**
**BC  / vs / LINK**
image

## Merge at the same time
https://github.com/bigcommerce/checkout-js/pull/1202

[INT-7503]: https://bigcommercecloud.atlassian.net/browse/INT-7503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BC-ISilva:
💚 Tested!

QE Details:

PR implementation has been validated, following scenarios were tested:

1. Experiment “CHECKOUT-6879.enable_floating_labels“ OFF
2. Experiment “CHECKOUT-6879.enable_floating_labels“ ON
3. All Stripe elements, labels are rendered inside the corresponding text field
4. All Stripe elements, when the shopper starts typing, then labels. shall remain and float to the top of the field allowing for the entered text to be prominent.
5. Stripe Elements will inherit all style rules defined in Page Builder (Page Builder > Theme Styles > Checkout Page > Form Input Fields) so that the Stripe Elements match the style of the native BigCommerce elements.
6. Works across desktop and mobile browsers (Safari and FireFox browsers were used for validation).

Please review attached word file with validation details.
[INT-7503  - [Link] Support Floating Labels for Stripe Address Element.docx](https://github.com/bigcommerce/checkout-sdk-js/files/10851504/INT-7503.-.Link.Support.Floating.Labels.for.Stripe.Address.Element.docx)

JavvmBC (approved):

packages/core/src/customer/strategies/stripe-upe/stripe-upe-customer-strategy.ts
```diff
@@ -98,8 +104,24 @@ export default class StripeUPECustomerStrategy implements CustomerStrategy {
...
+                    labels: 'floating',
+                    variables: {
+                        ...appearance?.variables,
+                        fontSizeBase: '14px',
+                    },
+                    rules: {
+                        ...appearance?.rules,
+                        '.Input': {
+                            ...appearance?.rules?.['.Input'],
+                            padding: '7px 13px 5px 13px',
```
animesh1987, line 121: Where are we getting this numbers? Also can see the same repeated in shipping strategy...Can we make this a constant and use it that way?

AlanMunozN: @animesh1987 I got it from the elements of BC (left and right), and for the top and bottom I had to adjust manually in order to emulate the BC styles since the iframe does not look similar using the same pixels of BC,

by the way, I moved it into a function to use in customer, shipping, and payment steps, let me know your thoughts on this, tomorrow I'll try to check the approach (including the JS approach), and QA will check it again

packages/core/src/payment/strategies/stripe-upe/stripe-upe-custom-styles.ts
```diff
@@ -0,0 +1,65 @@
+import { StripeFormMode, StripeUPEAppearanceOptions } from './stripe-upe';
+
+interface Styles {
+    [key: string]: string;
+}
+
+export const getStripeCustomStyles = (
```
bc-peng, line 7: We generally avoid placing CSS styles in the SDK, and using inline styles limits other developers' ability to change the UI via CSS override. Is it possible to put them into `checkout-js`?

packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
```diff
@@ -356,35 +356,16 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
...
-                        boxShadow: styles.fieldInnerShadow,
-                    },
-                },
-            };
-        }
+        const {
+            config: { getStoreConfig },
+        } = this._store.getState();
+        const features = getStoreConfig()?.checkoutSettings.features;
+        const experiment = features && features['CHECKOUT-6879.enable_floating_labels'];
```
bc-peng, line 363: We'll shortly alter the condition to read it from the checkout configuration.
Therefore, the best way of moving forward is to use this helper function in `checkout-js`.
https://github.com/bigcommerce/checkout-js/blob/master/packages/core/src/app/common/utility/isFloatingLabelEnabled.ts
I think you may pass all of these UI properties from `checkout-js` while keeping the logic here.

packages/core/src/payment/strategies/stripe-upe/stripe-upe-custom-styles.ts
animesh1987: I think I agree with @bc-peng here, we should not put styles in SDK. Since it's supposed to be a wrapper for API that powers checkout. Unless we have a good reason for we should have styles in checkout-js.

packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
animesh1987: Same as above, I see the experiment is used to get styles from `getStripeCustomStyles` function. We should have this function in checkout-js if needed.

AlanMunozN: Hey @animesh1987 and @bc-peng, I made changes, and it's ready for another review ♻️

bc-peng (approved):
LGTM, thanks for your patience.

animesh1987 (approved):

vitalii-koshovyi (approved):

PR #1857, opened 2023-02-17
feat(payment): BOLT-465 move AnalyticsExtraItemsManager to payment-integration-api package

PavlenkoM:
## What?
Move `AnalyticsExtraItemsManager` interface to paymant-integration-api package

## Why?
to get ability to use `AnalyticsExtraItemsManager` interfaces in other payment integration packages

## Testing / Proof
Unit tests:
Screenshot 2023-02-17 at 18 48 13

### DIFF
--- a/packages/analytics/.eslintrc.json
+++ b/packages/analytics/.eslintrc.json
@@ -1,0 +1,18 @@
+{
+    "extends": ["../../.eslintrc.json"],
+    "ignorePatterns": ["!**/*"],
+    "overrides": [
+        {
+            "files": ["*.ts", "*.tsx"],
+            "rules": {
+                "@typescript-eslint/no-unsafe-return": "off"
+            }
+        },
+        {
+            "files": ["*.spec.ts", "*.spec.tsx"],
+            "rules": {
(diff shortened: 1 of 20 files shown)


animesh1987:
Thanks for raising this, but I don't feel this is the right approach. We are bloating the `payment-integration-api` in this case 🤔 and creating a ticket coupling with the API at the same time. Which will make evolution of both harder going forward. 

My suggestion will be to move analytics piece of code to a similar package similar to what we did here https://github.com/bigcommerce/checkout-js/tree/master/packages/analytics.

vitalii-koshovyi (approved):

PavlenkoM:
> Thanks for raising this, but I don't feel this is the right approach. We are bloating the `payment-integration-api` in this case 🤔 and creating a ticket coupling with the API at the same time. Which will make evolution of both harder going forward.
> 
> My suggestion will be to move analytics piece of code to a similar package similar to what we did here https://github.com/bigcommerce/checkout-js/tree/master/packages/analytics.

Hi @animesh1987
I agree with you, and have moved this changes to new analytics package.

animesh1987 (approved):

PR #1877, opened 2023-02-22
feat(payment): INT-7217 Adding channel name to send in stripe metadata

JavvmBC:
## What? [INT-7217](https://bigcommercecloud.atlassian.net/browse/INT-7217)
Adding channel name to payment payload formated data to send it in the metadata to stripe dashboard

## Why?
More Information to the store owners

## Testing / Proof
Screenshot 2023-02-23 at 12 48 09

## Dependency of
https://github.com/bigcommerce/bigpay/pull/6724

[INT-7217]: https://bigcommercecloud.atlassian.net/browse/INT-7217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/payment/payment.ts
+++ b/packages/core/src/payment/payment.ts
@@ -242,4 +242,5 @@ export interface StripeV3FormattedPayload {
     bigpay_token?: {
         token: string;
     };
+    channel_name?: string;
 }
--- a/packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
+++ b/packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
@@ -238,6 +238,11 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
         }
     }
 
+    private _getChannelName() {
+        const channel_link = this._store.getState().config.getStoreConfigOrThrow().links.siteLink;
+        return channel_link.split('://')[1];
+    }
+
     private async _executeWithAPM(methodId: string): Promise<InternalCheckoutSelectors> {
         const paymentMethod = this._store
             .getState()
@@ -250,6 +255,7 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
                     vault_payment_instrument: false,
                     confirm: false,
                     set_as_default_stored_instrument: false,
+                    channel_name: this._getChannelName(),
                 },
             },
         };
@@ -279,6 +285,7 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
                     vault_payment_instrument: shouldSaveInstrument,
                     confirm: false,
                     set_as_default_stored_instrument: shouldSetAsDefaultInstrument,
+                    channel_name: this._getChannelName(),
                 },
             },
         };
@@ -315,6 +322,7 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
                         confirm: false,
                         client_token: paymentMethod.clientToken,
                         set_as_default_stored_instrument: shouldSetAsDefaultInstrument,
+                        channel_name: this._getChannelName(),
                     },
                 },
             };
@@ -499,6 +507,7 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
                             confirm: false,
                             vault_payment_instrument: shouldSaveInstrument,
                             set_as_default_stored_instrument: shouldSetAsDefaultInstrument,
+                            channel_name: this._getChannelName(),
                         },
                     },
                 };
@@ -559,6 +568,7 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
                         },
                         confirm: false,
                         set_as_default_stored_instrument: shouldSetAsDefaultInstrument,
+                        channel_name: this._getChannelName(),
                     },
                 },
             };
--- a/packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.spec.ts
+++ b/packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.spec.ts
@@ -388,6 +388,7 @@ describe('StripeUPEPaymentStrategy', () => {
                                 confirm: false,
                                 client_token: 'myToken',
                                 set_as_default_stored_instrument: true,
+                                channel_name: 'store-k1drp8k8.bcapp.dev/',
                             },
                         },
                     });


mauricio-sg (approved):
LGTM

packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
```diff
@@ -238,6 +238,14 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
         }
     }
 
+    private _getChannelName() {
```
rabinshr, line 241: should this be channel specific url instead of channel name?

JavvmBC: I think it would be more understandable if we left it as channle_url instead of name, assigned channel_name due to the ticket description. What are your thoughts?

bc-peng: Since it is not an actual channel name, and we have code like this `channel_name.split('://')[1];`, could you please rename it?

packages/core/src/payment/payment.ts
```diff
@@ -244,4 +244,5 @@ export interface StripeV3FormattedPayload {
     bigpay_token?: {
         token: string;
     };
+    channel_name?: string;
```
animesh1987, line 247: Does this need to be optional?

packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
```diff
@@ -238,6 +238,14 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
...
+        const channel_name =
+            this._store.getState().config.getConfig()?.storeConfig?.links.siteLink ||
+            'problem getting channel name';
```
animesh1987, line 244: This seems a bit odd that if we don't have a `siteLink` then we return `problem getting channel name` as a channel name 🤔

packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
```diff
@@ -238,6 +238,14 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
         }
     }
 
+    private _getChannelName() {
+        const channel_name =
+            this._store.getState().config.getConfig()?.storeConfig?.links.siteLink ||
```
animesh1987, line 243:
```suggestion
            this._store.getState().config.getStoreConfigOrThrow().links.siteLink
```

animesh1987:
Also have we verified if `siteLink` is the field we need for channel information?

BC-ISilva:
💚 Tested!!

QE Details.

Implementation has been validated. Required metadata is now available for merchants in Stripe SandBox.

Following scenarios have been validated:

1. Pass the channel name to Stripe as part of [the metadata on the payment intent object](https://stripe.com/docs/api/payment_intents/object#payment_intent_object-metadata) with the following key-value pair: Storefront: {{CHANNEL NAME}}
2. Pass the order management URL to Stripe as part of the of [the metadata on the payment intent object](https://stripe.com/docs/api/payment_intents/object#payment_intent_object-metadata) with the following key-value pair: Order_View: {{CHANNEL URL}}/manage/orders/{{ORDER.ID}}
3. The key-value pairs for Storefront and Order View passed to the payment-intent object can be seen in the Stripe dashboard when looking at the page for an individual payment
4. If the merchant has multiple storefronts, the channel name and channel URL of the storefront passed to Stripe, will change based on which storefront the order was placed through. 
5. Full redirect.
6. Order types (Auth Only and Auth and Capture)
7. Review Metadata from Stripe Sandbox.

 **Above scenarios were tested for StripUPE, StripeLINK and StripeV3.**

Please review attached world file with validation details.
[INT-7217 - Pass Channel Name and Order Management URL To Stripe As Metadata.docx](https://github.com/bigcommerce/checkout-sdk-js/files/10994251/INT-7217.-.Pass.Channel.Name.and.Order.Management.URL.To.Stripe.As.Metadata.docx)

packages/core/src/payment/payment.ts
JavvmBC: This implementation is for stripe upe, until we make it for stripe V3 we can remove the optional because they share the interface

packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
JavvmBC: Thank you for the suggestion it is a way better approach

JavvmBC:
> Also have we verified if `siteLink` is the field we need for channel information?

From all the fields available from storeConfig this is the best way to get the channel name

packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
JavvmBC: What do you think of channel_link?

bc-peng: Perhaps `channel_url`?

TomA-R (commented):
Apologies for the late review but is there a reason this is sent by the browser rather than synced on the backend during the order sync?

JavvmBC:
> Apologies for the late review but is there a reason this is sent by the browser rather than synced on the backend during the order sync?

We decided to send it on the pyament_request from checkout to bigpay because we thought it was the best place to send the data due to the fact that we are already sending the store information in that request

TomA-R:
> > Apologies for the late review but is there a reason this is sent by the browser rather than synced on the backend during the order sync?
> 
> We decided to send it on the pyament_request from checkout to bigpay because we thought it was the best place to send the data due to the fact that we are already sending the store information in that request

Thank you for the explanation! There are a few issues that come to mind with this:

1. This can be tampered with: anyone can change the channel name to an arbitrary value. Depending on merchant's business rules that can either be harmless, or problematic. Considering the channel name will likely surface in various downstream systems, there's also more potential for sending malicious input. While it's probably not the biggest risk, it doesn't seem like something the client _should_ be able to change/tamper with.
2. From what I understand, this is actually just incorrect. Channels have a channel name stored in our DB. The channel name isn't just the domain name but is configurable. In your dev environment, the channel name will be something like `My Dev Store 176796912` rather than `store-k1drp8k8.bcapp.dev/`
3. This makes the data available on a per-integration basis. Stores with a custom checkout forked before this change won't send it at all, and if we want the channel name to be available in other integrations, we'll need to add it to the client-side payload for the integration.

My suggestion would be to add it to the order sync data. I'd be happy to catch up about this on Slack/on a call too so we don't discuss internal architecture details on an open-source PR

cc @IamDavidovich you may also have some views on this and/or can validate my suggestions

rabinshr:
Yep, I agree with what @TomA-R said. 
Adding it to order sync data (payment session) makes sense as we would also be doing this for any other integration that might need it. 
I thought that was one of the original requests (in original bigpay PR) https://github.com/bigcommerce/bigpay/pull/6331#issuecomment-1341787433.

We currently do similar thing for `checkout_url` which is used on bigpay side to build return_url. This may be a different url to channel's url itself but perhaps we could do similar thing. https://github.com/bigcommerce/bigcommerce/blob/d2d8b574e9fd035240e65fece2cc36e28a1278e5/lib/Services/Bigpay/Payment/Builders/AuthTokenRequestBuilder.php#L50

IamDavidovich:
Just to provide some more context, yes, this info shouldn't be coming from the browser for the reasons above. 

We're actually no longer sending a lot of the data sent from the storefront through to the provider for exactly this reason - all the order data, store data, etc. is all gathered from the synced payment session. (Parts of the request are still used in other ways, but I'm hoping to remove those soon too). 

Going forward, the only things that should be used from the front-end request are the shopper-provided details and the relevant identifiers. I.e. the actual card/token data, hosted form IDs, etc.

BC-AOrtiz:
This PR is not need it any more, moved the functionality to the back end

PR #1878, opened 2023-02-22
refactor(checkout): CHECKOUT-7304 Change PayPal wallet buttons height

bc-peng:
## What?
Set the height of the PayPal buttons at the top of the checkout page to `36px` rather than `40px`.

## Why?
So we can provide a standardised wallet button UI to shoppers.

## Testing / Proof

### DIFF
--- a/packages/core/src/customer/strategies/braintree/braintree-paypal-credit-customer-strategy.ts
+++ b/packages/core/src/customer/strategies/braintree/braintree-paypal-credit-customer-strategy.ts
@@ -141,9 +141,11 @@ export default class BraintreePaypalCreditCustomerStrategy implements CustomerSt
         let hasRenderedSmartButton = false;
 
         if (paypal) {
+            const checkoutUserExperienceSettings = this._store.getState().config.getConfig()
+                ?.storeConfig.checkoutSettings.checkoutUserExperienceSettings;
             const fundingSources = [paypal.FUNDING.PAYLATER, paypal.FUNDING.CREDIT];
             const commonButtonStyle = {
-                height: 40,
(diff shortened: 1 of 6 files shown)


packages/core/src/customer/strategies/braintree/braintree-paypal-credit-customer-strategy.ts
```diff
@@ -141,9 +141,11 @@ export default class BraintreePaypalCreditCustomerStrategy implements CustomerSt
         let hasRenderedSmartButton = false;
 
         if (paypal) {
+            const checkoutUserExperienceSettings = this._store.getState().config.getConfig()
+                ?.storeConfig.checkoutSettings.checkoutUserExperienceSettings;
             const fundingSources = [paypal.FUNDING.PAYLATER, paypal.FUNDING.CREDIT];
             const commonButtonStyle = {
-                height: 40,
+                height: checkoutUserExperienceSettings?.walletButtonsOnTop ? 36 : 40,
```
davidchin, line 148: Could we instead pass the value (36) from Checkout JS into here? I imagine the height depends on the value we set in Checkout JS. So custom checkouts might have a different value?

davidchin (approved):

packages/core/src/customer/strategies/braintree/braintree-paypal-credit-customer-strategy.ts
bc-peng: Thank you for your approval, DC.

animesh1987:
@serhii-tkachenko tagging you for visibility around this change.

PR #1883, opened 2023-02-27
feat(payment): PAYMENTS-8527 expose interface for Braintree real time host form filed validation

bc-chenli:
## What?
This is the following PR for the POC to expose the interface for Braintree real-time host form filed validation
https://github.com/bigcommerce/checkout-sdk-js/pull/1848

## Why?
RealReal would like the Braintree host form filed to have the real-time validation interface. This PR expose it.

## Testing / Proof
Unit tests
Error status
image

### DIFF
--- a/packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
+++ b/packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
@@ -7,6 +7,8 @@ import { NonceInstrument } from '../../payment';
 
 import {
     BraintreeBillingAddressRequestData,
+    BraintreeFormErrorDataKeys,
+    BraintreeFormErrorsData,
     BraintreeHostedFields,
     BraintreeHostedFieldsCreatorConfig,
     BraintreeHostedFieldsState,
@@ -254,20 +256,20 @@ export default class BraintreeHostedForm {
         }
     }
 
-    private _mapErrors(fields: any): any {
-        const errors = {};
+    private _mapErrors(fields: BraintreeHostedFieldsState['fields']): BraintreeFormErrorsData {
+        const errors: BraintreeFormErrorsData = {};
 
         if (fields) {
             for (const [key, value] of Object.entries(fields)) {
-                const { isValid, isEmpty, isPotentialyValid } = value as any;
+                if (value && this._isValidParam(key)) {
+                    const { isValid, isEmpty, isPotentiallyValid } = value;
 
-                (errors as any)[key] = {
-                    [key]: {
+                    errors[key] = {
                         isValid,
                         isEmpty,
-                        isPotentialyValid,
-                    },
-                };
+                        isPotentiallyValid,
+                    };
+                }
             }
         }
 
@@ -449,7 +451,7 @@ export default class BraintreeHostedForm {
         this._formOptions?.onCardTypeChange?.({
             cardType:
                 event.cards.length === 1
-                ? event.cards[0].type.replace(/^master\-card$/, 'mastercard') /* eslint-disable-line */
+                    ? event.cards[0].type.replace(/^master\-card$/, 'mastercard',) /* eslint-disable-line */
                     : undefined,
         });
     };
@@ -468,4 +470,20 @@ export default class BraintreeHostedForm {
             errors: this._mapValidationErrors(event.fields),
         });
     };
+
+    private _isValidParam(
+        formErrorDataKey: string,
+    ): formErrorDataKey is BraintreeFormErrorDataKeys {
+        switch (formErrorDataKey) {
+            case 'number':
+            case 'cvv':
+            case 'expirationDate':
+            case 'postalCode':
+            case 'cardholderName':
+            case 'cardType':
+                return true;
+            default:
+                return false;
+        }
+    }
 }
--- a/packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
+++ b/packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
@@ -254,6 +254,26 @@ export default class BraintreeHostedForm {
         }
     }
 
+    private _mapErrors(fields: any): any {
+        const errors = {};
+
+        if (fields) {
+            for (const [key, value] of Object.entries(fields)) {
+                const { isValid, isEmpty, isPotentialyValid } = value as any;
+
+                (errors as any)[key] = {
+                    [key]: {
+                        isValid,
+                        isEmpty,
+                        isPotentialyValid,
+                    },
+                };
+            }
+        }
+
+        return errors;
+    }
+
     private _mapValidationErrors(
         fields: BraintreeHostedFieldsState['fields'],
     ): BraintreeFormFieldValidateEventData['errors'] {
@@ -403,6 +423,7 @@ export default class BraintreeHostedForm {
     private _handleBlur: (event: BraintreeHostedFieldsState) => void = (event) => {
         this._formOptions?.onBlur?.({
             fieldType: this._mapFieldType(event.emittedBy),
+            errors: this._mapErrors(event.fields),
         });
     };
 
--- a/packages/core/src/payment/strategies/braintree/braintree-payment-options.ts
+++ b/packages/core/src/payment/strategies/braintree/braintree-payment-options.ts
@@ -1,4 +1,4 @@
-import { BraintreeVerifyPayload } from './braintree';
+import { BraintreeFormErrorsData, BraintreeVerifyPayload } from './braintree';
 
 /**
  * A set of options that are required to initialize the Braintree payment
@@ -173,7 +173,7 @@ export type BraintreeFormFieldStyles = Partial<
 
 export interface BraintreeFormFieldKeyboardEventData {
     fieldType: string;
-    errors?: any;
+    errors?: BraintreeFormErrorsData;
 }
 
 export type BraintreeFormFieldBlurEventData = BraintreeFormFieldKeyboardEventData;
--- a/packages/core/src/payment/strategies/braintree/braintree-payment-options.ts
+++ b/packages/core/src/payment/strategies/braintree/braintree-payment-options.ts
@@ -173,6 +173,7 @@ export type BraintreeFormFieldStyles = Partial<
 
 export interface BraintreeFormFieldKeyboardEventData {
     fieldType: string;
+    errors?: any;
 }
 
 export type BraintreeFormFieldBlurEventData = BraintreeFormFieldKeyboardEventData;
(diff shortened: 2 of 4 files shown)


packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
```diff
@@ -254,6 +254,26 @@ export default class BraintreeHostedForm {
...
 
+    private _mapErrors(fields: any): any {
+        const errors = {};
+
+        if (fields) {
+            for (const [key, value] of Object.entries(fields)) {
+                const { isValid, isEmpty, isPotentialyValid } = value as any;
+
+                (errors as any)[key] = {
+                    [key]: {
```
bc-chenli, line 265: I am thinking maybe we don't require an additional `[key]` layer here. Since the current format is 
```typescript
errors: {
        cvv: {
            cvv: {
                isEmpty: true,
                isPotentialyValid: true,
                isValid: false,
            },
        },
    },
```
Without it, it will be looks like 
```typescript
errors: {
      cvv: {
          isEmpty: true,
          isPotentialyValid: true,
          isValid: false,
      },
 },
```
Which is less repetitive.

animesh1987: Good thinking 😄

packages/core/src/payment/strategies/braintree/braintree-payment-options.ts
```diff
@@ -173,6 +173,7 @@ export type BraintreeFormFieldStyles = Partial<
 
 export interface BraintreeFormFieldKeyboardEventData {
     fieldType: string;
+    errors?: any;
```
animesh1987, line 176: I suppose we should be able to type this out better as well.

packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
serhii-tkachenko: Let's try to use the second option, because it looks much better than prev one

packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
```diff
@@ -254,6 +254,26 @@ export default class BraintreeHostedForm {
         }
     }
 
+    private _mapErrors(fields: any): any {
```
serhii-tkachenko, line 257: The fields interface can be updated with `BraintreeHostedFieldsState['fields']`:

```suggestion
    private _mapErrors(fields: BraintreeHostedFieldsState['fields']): any {
```

packages/core/src/payment/strategies/braintree/braintree-payment-options.ts
serhii-tkachenko, line 176: Is it possible to update the `errors` interface?
As far as I see, `errors` should be an object with optional keys and values inside. We know the basic shape of this param and can create another interface for that.

Base on your comment we can create something like this:
```
export interface BraintreeFormError {
    isEmpty: boolean;
    isPotentialyValid: boolean;
    isValid: boolean;
}

export interface BraintreeFormErrorsData {
    number?: BraintreeFormError;
    expirationDate?: BraintreeFormError;
    expirationMonth?: BraintreeFormError;
    expirationYear?: BraintreeFormError;
    cvv?: BraintreeFormError;
    postalCode?: BraintreeFormError;
}
```

packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
```diff
@@ -254,6 +254,26 @@ export default class BraintreeHostedForm {
...
+        const errors = {};
+
+        if (fields) {
+            for (const [key, value] of Object.entries(fields)) {
+                const { isValid, isEmpty, isPotentialyValid } = value as any;
```
serhii-tkachenko, line 262: If you make all [this](https://github.com/bigcommerce/checkout-sdk-js/pull/1883/files#r1121657345) changes value any type can be updated with BraintreeFormErrorData

```suggestion
                const { isValid, isEmpty, isPotentialyValid } = value as BraintreeFormErrorData;
```

packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
bc-chenli: Let me update this. 👍

packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
animesh1987: Also will try to avoid using `as` keyword and use a type guard, for reference https://blog.logrocket.com/how-to-use-type-guards-typescript

packages/core/src/payment/strategies/braintree/braintree-payment-options.ts
bc-chenli: Thanks for the advice!

bc-chenli: Good call 👍

packages/core/src/payment/strategies/braintree/braintree.ts
```diff
@@ -6,7 +6,7 @@ import {
     TokenizePayload,
 } from '../googlepay';
 import { PaypalAuthorizeData, PaypalButtonOptions, PaypalButtonRender, PaypalSDK } from '../paypal';
-
+import { Omit } from '../../../common/types';
```
serhii-tkachenko, line 9: Omit is a default method for typescript. This line can be removed.

packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
```diff
@@ -254,6 +255,26 @@ export default class BraintreeHostedForm {
...
 
+    private _mapErrors(fields: BraintreeHostedFieldsState['fields']): BraintreeFormErrorsData {
+        const errors = {};
+
+        if (fields) {
+            for (const [key, value] of Object.entries(fields)) {
+                if (value) {
+                    const { isValid, isEmpty, isPotentiallyValid } = value;
+
+                    (errors as any)[key] = {
```
serhii-tkachenko, line 266: Is there any chance to remove `any` type here?

packages/core/src/payment/strategies/braintree/braintree.ts
bc-chenli: I see, but I think in our repo this is required, otherwise, a build error will be thrown. I can see quite a few places are introducing it.

packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
```diff
@@ -254,6 +255,26 @@ export default class BraintreeHostedForm {
...
 
+    private _mapErrors(fields: BraintreeHostedFieldsState['fields']): BraintreeFormErrorsData {
+        const errors: BraintreeFormErrorsData = {};
+
+        if (fields) {
+            for (const [key, value] of Object.entries(fields)) {
+                if (value) {
+                    const { isValid, isEmpty, isPotentiallyValid } = value;
+
+                    errors[key as keyof BraintreeFormErrorsData] = {
```
bc-chenli, line 266: I tried to eliminate this `as` keyword but can't think of a neat way to do so. The way I can think of is either add [key: string] accessor to the BraintreeFormErrorsData or manual mapping the content with if/else, would like to see what will the advice here.

packages/core/src/payment/strategies/braintree/braintree.ts
animesh1987: This was added a long time ago before our version of TS had the method, can you please try the native `Omit` and see if it works fine.

bc-chenli: I tried, but it is not working. 😢 That is the reason I added it here

packages/core/src/payment/strategies/braintree/braintree.ts
```diff
@@ -300,14 +300,28 @@ export interface BraintreeHostedFieldsCard {
...
 }
 
+export interface BraintreeHostedFieldsFieldData extends BraintreeFormFieldState {
+    container: HTMLElement;
+}
+
+export type BraintreeFormErrorData = Omit<BraintreeFormFieldState, 'isFocused'>;
+
+export interface BraintreeFormErrorsData {
+    number?: BraintreeFormErrorData;
```
animesh1987, line 317: Do all of them need to be optional?

bc-chenli: _resolved offline_

packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
```diff
@@ -447,4 +471,18 @@ export default class BraintreeHostedForm {
             errors: this._mapValidationErrors(event.fields),
         });
     };
+
+    private _isValidParam(k: string): k is BraintreeFormErrorDataKeys {
```
animesh1987, line 475: Nice work, just one comment can we use a more meaningful name than `k`?

packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
```diff
@@ -254,6 +256,27 @@ export default class BraintreeHostedForm {
         }
     }
 
+    private _mapErrors(fields: BraintreeHostedFieldsState['fields']): BraintreeFormErrorsData {
+        const errors: BraintreeFormErrorsData = {};
```
animesh1987, line 260: Do we need to assign the type here?

bc-chenli: I don't think so, what is the drawback to assign type here?

packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
bc-chenli: Good call, let me update this 👍

packages/core/src/payment/strategies/braintree/braintree-hosted-form.ts
```diff
@@ -254,6 +256,27 @@ export default class BraintreeHostedForm {
...
+                    if (this._isValidParam(key)) {
+                        errors[key] = {
+                            isValid,
+                            isEmpty,
+                            isPotentiallyValid,
+                        };
+                    }
+                }
+            }
+        }
```
serhii-tkachenko, line 275: Can we simplify the code in this way?

```suggestion
        if (fields) {
            for (const [key, value] of Object.entries(fields)) {
                if (value && this._isValidParam(key)) {
                    const { isValid, isEmpty, isPotentiallyValid } = value;

                    errors[key] = {
                        isValid,
                        isEmpty,
                        isPotentiallyValid,
                    };
                }
            }
        }
```

serhii-tkachenko (approved):
LGTM! 👍

animesh1987 (approved):
Nice work 👍

PR #1886, opened 2023-03-01
feat(checkout): CHECKOUT-7313 Introduce wallet button support selectors

bc-peng:
## What?
Create a new selector to better support the new wallet button container implementation.

- `isInitializedCustomer(methodId?: string): boolean;`

Update the existing selector to retrieve the error object of any unsuccessful customer method (wallet button).

- `getInitializeCustomerError(methodId?: string): Error | undefined;`

## Why?
As a result, the new wallet button container can:
- Show skeletons to represent the actual loading process.
- Hide any faulty wallet buttons.

## Testing / Proof
- CI checks.
- Manual testing.
   There should be three buttons, however only Paypal is successfully initialised in the end.

   https://user-images.githubusercontent.com/88361607/222036304-e70df991-7597-4d9a-a7d0-ae6709e1835d.mov

### DIFF
--- a/packages/core/src/checkout/checkout-store-selector.ts
+++ b/packages/core/src/checkout/checkout-store-selector.ts
@@ -125,6 +125,14 @@ export default interface CheckoutStoreSelector {
     getPaymentMethods(): PaymentMethod[] | undefined;
 
     /**
+     * Gets a list of successfully initialized wallet button methodIds.
+     *
+     * @returns The list of successfully initialized wallet button methodIds if
+     * there is any, otherwise undefined.
+     */
+    getInitializedWalletButtons(methodIds: string[]): string[] | undefined;
+
+    /**
      * Gets a payment method by an id.
      *
      * The method returns undefined if unable to find a payment method with the
@@ -544,6 +552,28 @@ export function createCheckoutStoreSelectorFactory(): CheckoutStoreSelectorFacto
             }),
     );
 
+    const getInitializedWalletButtons = createSelector(
+        ({ customerStrategies }: InternalCheckoutSelectors) =>
+            customerStrategies.getWalletButtonsStatus,
+        (getWalletButtonsStatus) =>
+            clone((methodIds: string[]) => {
+                const initializedMethodIds = [];
+                const walletButtonsStatus = getWalletButtonsStatus(methodIds);
+
+                for (const [methodId, button] of Object.entries(walletButtonsStatus)) {
+                    if (button.isInitialized) {
+                        initializedMethodIds.push(methodId);
+                    }
+                }
+
+                if (initializedMethodIds.length === 0) {
+                    return;
+                }
+
+                return initializedMethodIds;
+            }),
+    );
+
     return memoizeOne((state: InternalCheckoutSelectors): CheckoutStoreSelector => {
         return {
             getCheckout: getCheckout(state),
@@ -573,6 +603,7 @@ export function createCheckoutStoreSelectorFactory(): CheckoutStoreSelectorFacto
             getShippingAddressFields: getShippingAddressFields(state),
             getPickupOptions: getPickupOptions(state),
             getUserExperienceSettings: getUserExperienceSettings(state),
+            getInitializedWalletButtons: getInitializedWalletButtons(state),
         };
     });
 }
--- a/packages/core/src/checkout/checkout-store-selector.ts
+++ b/packages/core/src/checkout/checkout-store-selector.ts
@@ -125,14 +125,6 @@ export default interface CheckoutStoreSelector {
     getPaymentMethods(): PaymentMethod[] | undefined;
 
     /**
-     * Gets a list of successfully initialized wallet button methodIds.
-     *
-     * @returns The list of successfully initialized wallet button methodIds if
-     * there is any, otherwise undefined.
-     */
-    getInitializedWalletButtons(methodIds: string[]): string[] | undefined;
-
-    /**
      * Gets a payment method by an id.
      *
      * The method returns undefined if unable to find a payment method with the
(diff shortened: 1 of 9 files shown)


packages/core/src/customer/customer-strategy-reducer.ts
```diff
@@ -61,6 +61,10 @@ function errorsReducer(
 
         case CustomerStrategyActionType.InitializeFailed:
             return objectMerge(errors, {
+                failedMethodIds: [
+                    ...(errors.failedMethodIds ?? []),
+                    action.meta && action.meta.methodId,
+                ],
```
bc-peng, line 67: It turns out `objectMerge` does not merge two array values. Hence I merge them with `...`.

bc-peng, line 67: It turns out `objectMerge` does not merge two array values. Hence I merged them with `...`.

packages/core/src/customer/customer-strategy-selector.ts
```diff
@@ -84,6 +92,34 @@ export function createCustomerStrategySelectorFactory(): CustomerStrategySelecto
...
+                if (state.data[methodId]) {
+                    walletButtonsStatus[methodId] = {
+                        isInitialized: true,
+                        isLoading: false,
+                    };
+                } else if (state.errors.failedMethodIds?.includes(methodId)) {
+                    walletButtonsStatus[methodId] = {
+                        isInitialized: false,
+                        isLoading: false,
+                    };
```
bc-peng, line 110: Previouly, we treated a broken wallet button as not initialised, so I'm sticking with that idea here.

packages/core/src/checkout/checkout-store-status-selector.ts
```diff
@@ -102,6 +102,14 @@ export default interface CheckoutStoreStatusSelector {
      */
     isLoadingPaymentMethod(methodId?: string): boolean;
 
+    /**
+     * Checks whether wallet buttons are loading.
+     *
+     * @param methodIds - The identifiers of the payment methods to check.
+     * @returns True if any provided wallet button method is loading, otherwise false.
+     */
+    isLoadingWalletButtons(methodIds?: string[]): boolean;
```
davidchin, line 111:
```suggestion
    isLoadingWalletButton(methodId: string[]): boolean;
```
To be consistent with other methods that accept a singular method ID?

packages/core/src/checkout/checkout-store-selector.ts
```diff
@@ -124,6 +124,14 @@ export default interface CheckoutStoreSelector {
      */
     getPaymentMethods(): PaymentMethod[] | undefined;
 
+    /**
+     * Gets a list of successfully initialized wallet button methodIds.
+     *
+     * @returns The list of successfully initialized wallet button methodIds if
+     * there is any, otherwise undefined.
+     */
+    getInitializedWalletButtons(methodIds: string[]): string[] | undefined;
```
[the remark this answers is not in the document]
bc-peng: We have `isInitialized(methodId: string): boolean;` for a customer strategy.
https://github.com/bigcommerce/checkout-sdk-js/blob/b445659272efda3f381f1ed7332d8adef5da566d/packages/core/src/customer/customer-strategy-selector.ts#L136
Perhaps I don't need to make `getInitializedWalletButtons`, just use `isInitialized` to check them individually.
What do you think?

packages/core/src/checkout/checkout-store-status-selector.ts
bc-peng: At the end of loading wallet buttons in `checkout-js`, I want to execute something like this.
```
const isLoading = isLoadingWalletButtons(methodIds);

useEffect(() => {
        if(!isLoading){
            setMethodIds(getInitializedWalletButtons(methodIds) ?? []);
        }
    }, [isLoading]);
```
Therefore, I need a way to check all buttons' loading status together.

I just wanted to confirm that this approach in `checkout-js` is good and that we are only renaming here.

packages/core/src/checkout/checkout-store-selector.ts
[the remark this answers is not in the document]
davidchin: Yes, I think you can just use `isInitialized`.

packages/core/src/checkout/checkout-store-status-selector.ts
bc-peng: @davidchin What about this one? 😝

davidchin: Could you filter out the IDs that are not initialized? e.g.:

```ts
const isLoading = methodIds.filter(isLoadingWalletButton).length > 0;
```

bc-peng: Hi DC, based on your feedback, I have updated the code, and `checkout-js` implementation looks like below.
```
    const isLoading = methodIds.filter((methodId) => isLoadingWalletButton(methodId)).length > 0;

    useEffect(() => {
        if(!isLoading){
            const initializedMethodIds = methodIds.filter((methodId) => isInitializedWalletButton(methodId));

            setMethodIds(initializedMethodIds);
        }
    }, [isLoading]);
```

bc-peng: Hi @davidchin , based on your feedback, I have updated the code, and `checkout-js` implementation looks like below.
```
    const isLoading = methodIds.filter((methodId) => isLoadingWalletButton(methodId)).length > 0;

    useEffect(() => {
        if(!isLoading){
            const initializedMethodIds = methodIds.filter((methodId) => isInitializedWalletButton(methodId));

            setMethodIds(initializedMethodIds);
        }
    }, [isLoading]);
```

animesh1987 (approved):

packages/core/src/customer/customer-strategy-reducer.ts
davidchin: Yes, it's only a shallow-merge.

packages/core/src/customer/customer-strategy-selector.ts
```diff
@@ -152,6 +153,16 @@ export function createCustomerStrategySelectorFactory(): CustomerStrategySelecto
         },
     );
 
+    const isLoadingWalletButton = createSelector(
+        (state: CustomerStrategyState) => state,
+        (state) => (methodId: string) => {
+            return !(
```
davidchin, line 159: How come we can't just use `isInitializing`? If a button fails to load, it doesn't mean it's still loading.

bc-peng: The outcome of `isLoading` will be affected by their different definitions.

- `isInitializing` is `true` if a button is actually loading.
- `isLoadingWalletButton` is `true` if a button is not initialized or failed.


There is always a tiny moment where checkout does not load anything, making `isLoading` equal `false`.  The loading skeleton will thereafter be on and off again and again.
\* screenshot of `console.log(methodIds.filter((methodId) => isInitializingCustomer(methodId); }))`.

image

`isLoadingWalletButton` can provide the best results of deciding when to show/hide loading skeletons.
\* screenshot of `console.log(methodIds.filter((methodId) => isLoadingWalletbutton(methodId); }))`.

image

I just want to avoid switching the skeleton on and off. Is there another way we can do it?

packages/core/src/customer/customer-strategy-state.ts
```diff
@@ -13,7 +13,7 @@ export interface CustomerStrategyDataState {
 export interface CustomerStrategyErrorsState {
     deinitializeError?: Error;
     deinitializeMethodId?: string;
-    initializeError?: Error;
+    initializeError?: { [key: string]: Error };
```
davidchin, line 16: Can we rename this to be plural, e.g.: `initializeErrors`?

davidchin (approved):
Nice one @bc-peng!

PR #1893, opened 2023-03-10
feat(payment): PAYPAL-000 added objectWithKebabKeys utility function to payment-integration-api package

serhii-tkachenko:
## What?
Added objectWithKebabKeys utility function to payment-integration-api package

## Why?
To make an ability to convert any object keys to kebab style if needed in other packages

## Testing / Proof
Unit tests

### DIFF
--- a/packages/payment-integration-api/src/common/utility/index.ts
+++ b/packages/payment-integration-api/src/common/utility/index.ts
@@ -1,0 +1,1 @@
+export { default as objectWithKebabCaseKeys } from './object-with-kebab-case-keys';
--- a/packages/payment-integration-api/src/common/utility/object-with-kebab-case-keys.spec.ts
+++ b/packages/payment-integration-api/src/common/utility/object-with-kebab-case-keys.spec.ts
@@ -1,0 +1,21 @@
+import objectWithKebabCaseKeys from './object-with-kebab-case-keys';
+
+const objectMock = {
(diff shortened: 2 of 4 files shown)


packages/payment-integration-api/src/common/utility/object-with-kebab-keys.ts
```diff
@@ -0,0 +1,13 @@
+import { kebabCase } from 'lodash';
+
+export default function objectWithKebabKeys(object: { [key: string]: any }) {
```
vitalii-koshovyi, line 3: Do we need this code in separate util? Is it in use more than 1 time?

And if we still need this helper I would recommend to change the naming to something like `objectWithKebabCaseKeys`💅

andriiVitvitskyi1990 (approved):

PavlenkoM (approved):

bc-nick (approved):

packages/payment-integration-api/src/common/utility/object-with-kebab-keys.ts
serhii-tkachenko: Yes, we might use this method in braintree and paypal commerce package

vitalii-koshovyi (approved):

animesh1987 (approved):

PR #1896, opened 2023-03-13
feat(payment): INT-7573 Create Access Worldpay GooglePay strategy

bc-alejandroRodriguez:
## What?
Created and modified files necessary for google pay initialization. Also unit tests were added.

## Why?
As a Merchant I want to enable Google Pay with Worldpay Access so that I can offer shoppers Google Pay as a payment method to my shoppers.

## Testing / Proof
googlepay2

image

googlepay


vmparra (approved):
LGTM 👍🏻

PR #1903, opened 2023-03-15
fix(payment): CHECKOUT-000 Expose hosted form error to exported typings

animesh1987:
## What?
As above

## Why?
We have a feature to expose errors on blur for hosted forms added for braintree form in https://github.com/bigcommerce/checkout-sdk-js/pull/1883 PR.

But since Braintree form component uses the HOC for HostedForm we need to expose the errors as part of interface.

## Testing / Proof
- Cirlce


serhii-tkachenko (approved):
LGTM 👍

TomA-R (approved):

PR #1905, opened 2023-03-16
feat(payment): PAYPAL-2003 fixed google pay button styles

andriiVitvitskyi1990:
## What?
Fixed googlepay button styles on PDP. Now it's styles applied from page builder

## Why?
Googlepay button didn't change styles before (default styles applied)

## Testing / Proof
Screenshot 2023-03-16 at 20 11 44
Screenshot 2023-03-16 at 20 11 15

@bigcommerce/checkout @bigcommerce/payments @serhii-tkachenko

### DIFF
--- a/packages/core/src/checkout-buttons/strategies/googlepay/googlepay-button-strategy.ts
+++ b/packages/core/src/checkout-buttons/strategies/googlepay/googlepay-button-strategy.ts
@@ -82,7 +82,8 @@ export default class GooglePayButtonStrategy implements CheckoutButtonStrategy {
         currencyCode?: string,
     ): HTMLElement {
         const container = document.getElementById(containerId);
-        const { buttonType, buttonColor, buyNowInitializeOptions } = buttonOptions;
+        const { buttonType, buttonColor, buyNowInitializeOptions, style } = buttonOptions;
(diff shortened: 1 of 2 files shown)


packages/core/src/checkout-buttons/strategies/googlepay/googlepay-button-strategy.ts
```diff
@@ -96,7 +97,7 @@ export default class GooglePayButtonStrategy implements CheckoutButtonStrategy {
         const googlePayButton = this._googlePayPaymentProcessor.createButton(
             handleValidButtonClick,
             buttonType,
-            buttonColor,
+            buttonColor || color,
```
bc-peng, line 100: Could we also update the related unit test(s) to cover this change?

serhii-tkachenko: @bc-peng I think we can make it in other way. Let me check some idea with @andriiVitvitskyi1990 first

serhii-tkachenko: We can make it without adding any changes in checkout sdk

PR #1907, opened 2023-03-16
refactor(payment): BOLT-000 add generic type for getPaymentMethod

PavlenkoM:
## What?
Add generic type for `getPaymentMethod` & `getPaymentMethodOrThrow` in payment-integration-api

## Why?
Because this methods return instance of `PaymentMethod` and this interface has generic type for `initializationData`
[https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/payment-integration-api/src/payment/payment-method.ts](https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/payment-integration-api/src/payment/payment-method.ts)

If we want to describe correct type for `PaymentMethod.initializationData` we need to call this methods with generic type.
`unknown` - by default, for cases when we need `PaymentMethod` but don't use `initializationData` from it

## Testing / Proof
unit tests

### DIFF
--- a/packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.ts
+++ b/packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.ts
@@ -33,6 +33,7 @@ import {
     AdyenComponent,
     AdyenComponentType,
     AdyenError,
+    AdyenPaymentMethodInitializationData,
     AdyenPaymentMethodType,
     AdyenPlaceholderData,
     AdyenV3ComponentState,
@@ -71,10 +72,9 @@ export default class Adyenv3PaymentStrategy implements PaymentStrategy {
 
         const paymentMethod = this._paymentIntegrationService
             .getState()
-            .getPaymentMethodOrThrow(options.methodId);
-        const {
-            initializationData: { environment, clientKey, paymentMethodsResponse },
-        } = paymentMethod;
+            .getPaymentMethodOrThrow<AdyenPaymentMethodInitializationData>(options.methodId);
+        const { environment, clientKey, paymentMethodsResponse } =
+            paymentMethod.initializationData || {};
 
         this._adyenClient = await this._scriptLoader.load({
             environment,
--- a/packages/adyen-integration/src/adyenv3/adyenv3.ts
+++ b/packages/adyen-integration/src/adyenv3/adyenv3.ts
@@ -881,3 +881,9 @@ export function isCardState(param: unknown): param is CardState {
         typeof (param as CardState).data.paymentMethod.encryptedExpiryMonth === 'string'
     );
 }
+
+export interface AdyenPaymentMethodInitializationData {
+    clientKey?: string;
+    environment?: string;
+    paymentMethodsResponse?: PaymentMethodsResponse;
+}
--- a/packages/squarev2-integration/src/squarev2-payment-strategy.ts
+++ b/packages/squarev2-integration/src/squarev2-payment-strategy.ts
@@ -5,11 +5,13 @@ import {
     PaymentArgumentInvalidError,
     PaymentInitializeOptions,
     PaymentIntegrationService,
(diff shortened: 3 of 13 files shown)


packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.ts
```diff
@@ -71,10 +72,9 @@ export default class Adyenv3PaymentStrategy implements PaymentStrategy {
 
         const paymentMethod = this._paymentIntegrationService
             .getState()
-            .getPaymentMethodOrThrow(options.methodId);
-        const {
-            initializationData: { environment, clientKey, paymentMethodsResponse },
-        } = paymentMethod;
+            .getPaymentMethodOrThrow<AdyenPaymentMethodInitializationData>(options.methodId);
+        const { environment, clientKey, paymentMethodsResponse } =
+            paymentMethod.initializationData || {};
```
vitalii-koshovyi, line 77: Could it return an empty object now?

packages/adyen-integration/src/adyenv3/adyenv3.ts
```diff
@@ -881,3 +881,9 @@ export function isCardState(param: unknown): param is CardState {
         typeof (param as CardState).data.paymentMethod.encryptedExpiryMonth === 'string'
     );
 }
+
+export interface AdyenPaymentMethodInitializationData {
+    clientKey?: string;
```
vitalii-koshovyi, line 886: All fields look required. Am I missing something?

packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.ts
PavlenkoM: this default `{}` for fixing issue with tsLint, because `initializationData` alwais was as optional property for [PaymentMethod](https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/payment-integration-api/src/payment/payment-method.ts) and can be undefined in some cases

packages/adyen-integration/src/adyenv3/adyenv3.ts
PavlenkoM: I get this types as the same as it used for [AdyenConfiguration](https://github.com/bigcommerce/checkout-sdk-js/blob/43b963d4f04ad461b3721b8484c8a1dbcb657f26/packages/adyen-integration/src/adyenv3/adyenv3.ts#L211)
If they should be required I will update them?

vitalii-koshovyi (approved):

animesh1987 (approved):

TomA-R (approved):

packages/squarev2-integration/src/squarev2-payment-strategy.ts
```diff
@@ -29,8 +30,17 @@ export default class SquareV2PaymentStrategy implements PaymentStrategy {
...
+            initializationData,
+        } = this._paymentIntegrationService
+            .getState()
+            .getPaymentMethodOrThrow<SquarePaymentMethodInitializationData>(methodId);
+        const { applicationId, locationId } = initializationData || {};
+
+        if (!applicationId) {
+            throw new InvalidArgumentError(
+                'Unable to proceed because "applicationId" argument is not provided.',
+            );
```
serhii-tkachenko, line 42: It might be better to use another error here, because it is not an issue with provided options from checkout js or other plcase to checkout sdk. Consequently, it is an issue with payment method configuration on CP or so.

What do you think?

```suggestion
           throw new PaymentMethodInvalidError();
```

serhii-tkachenko (approved):
Looks good to me! 👍
Check a comment that I left before merge, please.

packages/squarev2-integration/src/squarev2-payment-strategy.ts
PavlenkoM: thanks, changed

PR #1909, opened 2023-03-17
feat(payment): PAYPAL-000 added isCreditCardFormFields typeguard instead of isStoredCreditCardFormFields

serhii-tkachenko:
## What?
Added isCreditCardFormFields typeguard instead of isStoredCreditCardFormFields

## Why?
We can't be really sure that provided data is related to stored credit card form fields, because HostedStoredCardFieldOptionsMap has only optional fields. It might be better to have a typeguard where we are sure that the provided data is really what expected.

## Testing / Proof
Unit tests


packages/payment-integration-api/src/hosted-form/is-credit-card-form-fields.ts
```diff
@@ -0,0 +1,12 @@
+import HostedFieldType from './hosted-field-type';
+import { HostedCardFieldOptionsMap, HostedStoredCardFieldOptionsMap } from './hosted-form-options';
+
+export default function isCreditCardFormFields(
+    fields: HostedCardFieldOptionsMap | HostedStoredCardFieldOptionsMap,
+): fields is HostedCardFieldOptionsMap {
+    return (
+        HostedFieldType.CardNumber in fields &&
```
animesh1987, line 8: Do we need to check for cardCode as well?

serhii-tkachenko: Nope, because it is an optional param in TS interface

PR #1912, opened 2023-03-20
feat(checkout): CHECKOUT-7250 Add `floatingLabelEnabled` to `UserExperienceSettings` interface

bc-peng:
## What?
Add `floatingLabelEnabled` to the `UserExperienceSettings` interface.

## Why?
We've updated the BCapp to allow merchants to toggle floating label usage.

## Depends on
https://github.com/bigcommerce/bigcommerce/pull/51713

## Relates to
https://github.com/bigcommerce/checkout-settings-ui/pull/92
[place holder] a `checkout-js` PR uses this PR to display floating labels.

## Testing / Proof
CI.

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -89,9 +89,10 @@ export interface StoreCurrency {
     thousandsSeparator: string;
 }
 
-export type UserExperienceSettingNames = 'walletButtonsOnTop';
-
-export type UserExperienceSettings = { [key in UserExperienceSettingNames]: boolean };
+export interface UserExperienceSettings {
+    walletButtonsOnTop: boolean;
+    floatingLabelEnabled: boolean;
+}
 
 export interface CheckoutSettings {
     features: { [featureName: string]: boolean };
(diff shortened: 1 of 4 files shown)


packages/core/src/config/config.ts
```diff
@@ -89,9 +89,10 @@ export interface StoreCurrency {
     thousandsSeparator: string;
 }
 
-export type UserExperienceSettingNames = 'walletButtonsOnTop';
-
-export type UserExperienceSettings = { [key in UserExperienceSettingNames]: boolean };
+export interface UserExperienceSettings {
+    walletButtonsOnTop?: boolean;
```
animesh1987, line 93: Don't think these should be optional.

bc-peng: The `UserExperienceSettings` should be made optional instead?

bc-peng: Does this make sense?
```
checkoutUserExperienceSettings: UserExperienceSettings | null;
```

animesh1987: Think with the new change with moving `checkoutUserExperienceSettings` as part of global settings, we can safely remove null as well.

bc-peng: Yes, I have updated this PR to reflect that `checkoutUserExperienceSettings` is always available now.

animesh1987 (approved):

PR #1928, opened 2023-03-30
feat(payment): PAYPAL-2050 fixed google pay button for safari and mozilla

andriiVitvitskyi1990:
## What?
Fixed google pay button for safari and mozilla

## Why?
Google pay button didn't work on PDP in Safari and Mozilla. Safari and Mozilla showed developer issue. The reason was that we need to loadPaymentData from googleClient before any API calls. 
So we are passing to google client creation
```
paymentDataCallbacks: {
                        onPaymentDataChanged(callbackTrigger): // callback that triggers multiple times (we usecallbackTrigger  === INITIALIZE to createBuyNowCart) and then return newTransactionInfo object that contains new totalPrice and googlepay popup being updated with new price
                    },
```
Such way we are avoiding issue that doesn't allow us to make api right after click on googlepay button. 
## Testing / Proof
Tested on Dev And Integration

https://user-images.githubusercontent.com/56301104/230553647-a8740498-6e09-4550-86ba-afb5329b7587.mov

### DIFF
--- a/packages/core/src/checkout-buttons/strategies/googlepay/googlepay-button-options.ts
+++ b/packages/core/src/checkout-buttons/strategies/googlepay/googlepay-button-options.ts
@@ -1,5 +1,5 @@
-import { BuyNowCartRequestBody } from '../../../cart';
 import { ButtonColor, ButtonType } from '../../../payment/strategies/googlepay';
+import { GooglePayBuyNowInitializeOptions } from './googlepay-button-types';
 
 export interface GooglePayButtonInitializeOptions {
     /**
@@ -20,7 +20,5 @@ export interface GooglePayButtonInitializeOptions {
     /**
      * The options that are required to initialize Buy Now functionality.
      */
-    buyNowInitializeOptions?: {
-        getBuyNowCartRequestBody?(): BuyNowCartRequestBody | void;
-    };
+    buyNowInitializeOptions?: GooglePayBuyNowInitializeOptions;
 }
--- a/packages/core/src/payment/strategies/googlepay/googlepay-payment-processor.ts
+++ b/packages/core/src/payment/strategies/googlepay/googlepay-payment-processor.ts
@@ -25,8 +25,10 @@ import {
     GooglePayInitializer,
     GooglePaymentData,
     GooglePayPaymentDataRequestV2,
+    GooglePayPaymentOptions,
     GooglePaySDK,
     TokenizePayload,
+    UpdatePaymentDataRequestPayload,
 } from './googlepay';
 import { getFirstAndLastName } from './googlepay-get-first-and-last-name';
 import GooglePayScriptLoader from './googlepay-script-loader';
@@ -47,10 +49,10 @@ export default class GooglePayPaymentProcessor {
         private _requestSender: RequestSender,
     ) {}
 
-    initialize(methodId: string): Promise<void> {
+    initialize(methodId: string, googlePayClientOptions?: GooglePayPaymentOptions): Promise<void> {
         this._methodId = methodId;
 
-        return this._configureWallet();
+        return this._configureWallet(googlePayClientOptions);
     }
 
     deinitialize(): Promise<void> {
@@ -73,13 +75,25 @@ export default class GooglePayPaymentProcessor {
         });
     }
 
-    updatePaymentDataRequest(payloadToUpdate: { currencyCode: string; totalPrice: string }) {
-        const paymentDataRequest = this._getPaymentDataRequest();
+    updatePaymentDataRequest(paymentDataRequest: UpdatePaymentDataRequestPayload) {
+        const existingPaymentDataRequest = this._getPaymentDataRequest();
 
-        paymentDataRequest.transactionInfo.currencyCode = payloadToUpdate.currencyCode;
-        paymentDataRequest.transactionInfo.totalPrice = payloadToUpdate.totalPrice;
-
-        this._paymentDataRequest = paymentDataRequest;
+        this._paymentDataRequest = {
+            ...existingPaymentDataRequest,
+            ...paymentDataRequest,
+            merchantInfo: {
+                ...(existingPaymentDataRequest.merchantInfo ?? {}),
+                ...(paymentDataRequest.merchantInfo ?? {}),
+            },
+            transactionInfo: {
+                ...(existingPaymentDataRequest.transactionInfo ?? {}),
+                ...(paymentDataRequest.transactionInfo ?? {}),
+            },
+            shippingAddressParameters: {
+                ...(existingPaymentDataRequest.shippingAddressParameters ?? {}),
+                ...(paymentDataRequest.shippingAddressParameters ?? {}),
+            },
+        };
     }
 
     displayWallet(): Promise<GooglePaymentData> {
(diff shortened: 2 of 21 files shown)


serhii-tkachenko (approved):
Nice work! 👍

animesh1987:
Hey @andriiVitvitskyi1990 this is a public repo, can you please attach a screenshot or video as proof of testing.

packages/core/src/checkout-buttons/strategies/googlepay/googlepay-button-options.ts
```diff
@@ -1,5 +1,8 @@
-import { BuyNowCartRequestBody } from '../../../cart';
-import { ButtonColor, ButtonType } from '../../../payment/strategies/googlepay';
+import {
+    ButtonColor,
+    ButtonType,
+    GooglePayBuyNowInitializeOptions,
+} from '../../../payment/strategies/googlepay';
```
animesh1987, line 5: Can I ask why are we moving this interface into the payment folder?

packages/core/src/payment/strategies/googlepay/googlepay-payment-processor.ts
```diff
@@ -47,10 +48,13 @@ export default class GooglePayPaymentProcessor {
         private _requestSender: RequestSender,
     ) {}
 
-    initialize(methodId: string): Promise<void> {
+    initialize(
+        methodId: string,
+        googlePayClientOptions?: Partial<GooglePayPaymentOptions>,
```
animesh1987, line 53: Why is this using an undefined and a partial interface.

packages/core/src/payment/strategies/googlepay/googlepay.ts
```diff
@@ -7,9 +8,33 @@ import {
...
+    PAYMENT_AUTHORIZATION = 'PAYMENT_AUTHORIZATION',
+    SHIPPING_ADDRESS = 'SHIPPING_ADDRESS',
+    SHIPPING_OPTION = 'SHIPPING_OPTION',
+}
+
 export type EnvironmentType = 'PRODUCTION' | 'TEST';
 export type TokenizeType = 'AndroidPayCard' | 'CreditCard' | 'CARD';
 
+export interface GooglePayBuyNowInitializeOptions {
+    getBuyNowCartRequestBody?(): BuyNowCartRequestBody | void;
```
animesh1987, line 35: What are the scenario where the return type of this method might be a void?

packages/core/src/payment/strategies/googlepay/googlepay-payment-processor.ts
```diff
@@ -73,13 +77,25 @@ export default class GooglePayPaymentProcessor {
         });
     }
 
-    updatePaymentDataRequest(payloadToUpdate: { currencyCode: string; totalPrice: string }) {
-        const paymentDataRequest = this._getPaymentDataRequest();
+    updatePaymentDataRequest(paymentDataRequest: Partial<GooglePayPaymentDataRequestV2>) {
```
animesh1987, line 80: I believe here we can try using a union type and validate payload using a typeguard in the method.

packages/core/src/payment/strategies/googlepay/googlepay.ts
```diff
@@ -30,6 +55,11 @@ export type GooglePayCreator = BraintreeModuleCreator<GooglePayBraintreeSDK>;
 
 export interface GooglePayPaymentOptions {
     environment: EnvironmentType;
+    paymentDataCallbacks?: {
+        onPaymentDataChanged?(
```
animesh1987, line 59: Does this need to be undefined?

packages/core/src/payment/strategies/googlepay/googlepay.ts
```diff
@@ -30,6 +55,11 @@ export type GooglePayCreator = BraintreeModuleCreator<GooglePayBraintreeSDK>;
...
+            intermediatePaymentData: IntermediatePaymentData,
+        ): Promise<NewTransactionInfo | void>;
```
animesh1987, line 61: Do we need void here?

packages/core/src/checkout-buttons/strategies/googlepay/googlepay-button-options.ts
andriiVitvitskyi1990: I didn't move. I just added GooglePayBuyNowInitializeOptions interface specifically for googlepay. For Example in braintreevenmo we have the same.

packages/core/src/payment/strategies/googlepay/googlepay-payment-processor.ts
andriiVitvitskyi1990: payment processor used in two cases. One i buyNow flow - then we pass googlePayClientOptions to initialize method (we use partial because in this case we don't need environment from interface). And the second flow - regular flow , so we don't pass googlePayClientOptions to initialize method

packages/core/src/payment/strategies/googlepay/googlepay.ts
andriiVitvitskyi1990: Yes we need void here. In our case :
`paymentDataCallbacks: {
                onPaymentDataChanged: async ({
                    callbackTrigger,
                }: IntermediatePaymentData): Promise<NewTransactionInfo | void> => {
                    if (callbackTrigger !== CallbackTriggerType.INITIALIZE) {
                        return;
                    }.....`

packages/core/src/payment/strategies/googlepay/googlepay.ts
andriiVitvitskyi1990: removed void

packages/core/src/payment/strategies/googlepay/googlepay.ts
andriiVitvitskyi1990: We use paymentDataCallbacks just in buyNowFlow. So in all other cases we don't need paymentDataCallbacks from GooglePayPaymentOptions interface

andriiVitvitskyi1990: Fixed

packages/core/src/payment/strategies/googlepay/googlepay-payment-processor.ts
andriiVitvitskyi1990: Remover partial

packages/core/src/payment/strategies/googlepay/googlepay-payment-processor.ts
andriiVitvitskyi1990: There are a lot of fields to pass if we remove partial. We use this method just to update some parts of class variable only.

serhii-tkachenko: It is not a big deal if it will be empty object, because this method are used to update `existingPaymentDataRequest`. 

In this case `this._paymentDataRequest` will be equal to `existingPaymentDataRequest` that is good as well.

The main point, why we add <Partial> here, because we should have an ability to pass only the data, that we want to be updated. But if we make it as union, we should use typeguards all over the place to let Typescript know that we pass "full" `paymentDataRequest` data, what is wrong in my opinion.

What do you think?

animesh1987 (approved):

PR #1933, opened 2023-04-03
WIP chore(common): CHECKOUT-7279 Update node to 16

animesh1987:
## What?
Update to node 16

## Why?
Node 14 will be end of life at end of April, hence we should update to node 16 for any patches and support.

## Testing / Proof
- CI


.circleci/config.yml
```diff
@@ -76,6 +79,7 @@ jobs:
 
   npm_release:
     <<: *node_executor
+    resource_class: medium+
```
bc-peng, line 82: Just curious, why `sdk` needs this?

bc-peng (approved):

.circleci/config.yml
animesh1987: Seems like there are too many deprecated packages, so while running npm ci memory in circle machine is maxed out and the process gets killed. We will need to come back and fix the packages later.

PR #1946, opened 2023-04-11
fix(common): CHECKOUT-7381 Update resolver mechanism

animesh1987:
## What?
Update resolver mechanism to resolve for correct strategy for a given query.

## Why?
At the moment if we have a registry entry with the following entry `{ id : credit_card, gateway: 'gateway' }`, but even if we query the registry with following query `{ id : credit_card }`. It returns the above payment strategy which can cause issues. 

Hence we need to add extra checks if the `matches` iterator is equal to the keys for the entry.

## Testing / Proof
- Unit tests

### DIFF
--- a/packages/core/src/common/registry/resolve-id-registry.ts
+++ b/packages/core/src/common/registry/resolve-id-registry.ts
@@ -37,24 +37,31 @@ export default class ResolveIdRegistry<TType, TToken extends { [key: string]: un
 
             const result = { token: registeredToken, matches: 0, default: false };
 
-            for (const [key, value] of Object.entries(query)) {
-                if (key in query && resolverId[key] !== value) {
+            for (const [key, value] of Object.entries(resolverId)) {
+                if (key in query && query[key] !== value) {
                     result.matches = 0;
                     break;
                 }
 
-                if (key in query && resolverId[key] === value) {
+                if (key in query && query[key] === value) {
                     result.matches++;
                 }
-            }
 
-            if (resolverId.default) {
-                result.default = true;
-                results.push(result);
+                if (key === 'default' && value === true) {
+                    result.default = true;
+                }
             }
 
             results.push(result);
         });
+
+        const uniqueMatches = [...Array.from(new Set(results.map((result) => result.matches)))];
+
+        if (results.length > 1 && uniqueMatches.length === 1) {
+            throw new Error(
+                'Please enter a detailed query, since the existing ones can return the incorrect strategy',
+            );
+        }
 
         const matched = results
             .sort((a, b) => b.matches - a.matches)
--- a/packages/core/src/common/registry/resolve-id-registry.ts
+++ b/packages/core/src/common/registry/resolve-id-registry.ts
@@ -37,21 +37,23 @@ export default class ResolveIdRegistry<TType, TToken extends { [key: string]: un
 
             const result = { token: registeredToken, matches: 0, default: false };
 
-            for (const [key, value] of Object.entries(resolverId)) {
-                if (key in query && query[key] === value) {
+            for (const [key, value] of Object.entries(query)) {
+                if (key in query && resolverId[key] !== value) {
+                    result.matches = 0;
+                    break;
+                }
+
+                if (key in query && resolverId[key] === value) {
                     result.matches++;
                 }
-
-                if (key === 'default' && value === true) {
-                    result.default = true;
-                }
             }
 
-            const keys = Object.keys(resolverId).length;
-
-            if (keys === result.matches || result.default === true) {
+            if (resolverId.default) {
+                result.default = true;
                 results.push(result);
             }
+
+            results.push(result);
         });
 
         const matched = results
--- a/packages/core/src/common/registry/resolve-id-registry.ts
+++ b/packages/core/src/common/registry/resolve-id-registry.ts
@@ -29,10 +29,12 @@ export default class ResolveIdRegistry<TType, TToken extends { [key: string]: un
 
     private _resolveToken(token: string, registeredTokens: string[]): string | undefined {
         const query = this._decodeToken(token);
+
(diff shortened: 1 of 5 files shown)


packages/core/src/common/registry/resolve-id-registry.ts
```diff
@@ -45,7 +47,11 @@ export default class ResolveIdRegistry<TType, TToken extends { [key: string]: un
                 }
             }
 
-            results.push(result);
+            const keys = Object.keys(resolverId).length;
+
+            if (keys === result.matches || result.default === true) {
+                results.push(result);
+            }
```
bc-peng, line 54: I undertand this change limits matching results to those with the exact same key names or the default method, which seems too strict. What do you think?

animesh1987: I was thinking the same when I was doing that, does it make sense to just add a check for `gateway` in the keys list?

bc-peng: Are the e2e tests in `checkout-js` good with this PR?

animesh1987: I haven't tried running them as of now. But I suppose should be ok.

davidchin: The PR description doesn't seem to match with what the actual problem described in #1917? The query is always going to have all the keys, e.g.: 
```json
{"id":"credit_card", "gateway":"barclaycard", "type":"PAYMENT_TYPE_HOSTED"}
```
instead of 
```json
{"id":"credit_card"}
```
The resolve ID, however, may not contain all the keys. So it could be `{"id":"credit_card", "gateway":"bluesnapdirect"}` or `{"type":"PAYMENT_TYPE_HOSTED"}` or `{"id":"credit_card"}`. 

Based on that problem definition, I think what we need is change the `for` loop in `resolveToken` to something like this?
```ts
for (const [key, value] of Object.entries(resolverId)) {
    if (key in query && query[key] !== value) {
        result.matches = 0;
        break;
    }

    if (key in query && query[key] === value) {
        result.matches++;
    }

    // ...
}
```

With this, if a strategy is specifically registered for `{"id":"credit_card", "gateway":"bluesnapdirect"}`, it won't match with `{"id":"credit_card", "gateway":"barclaycard", "type":"PAYMENT_TYPE_HOSTED"}` as the `gateway` field is incompatible. But if the strategy is registered for `{"id":"credit_card"}`, it will still match as there is no incompatible field.

packages/core/src/common/registry/resolve-id-registry.ts
```diff
@@ -48,6 +55,14 @@ export default class ResolveIdRegistry<TType, TToken extends { [key: string]: un
             results.push(result);
         });
 
+        const uniqueMatches = [...Array.from(new Set(results.map((result) => result.matches)))];
+
+        if (results.length > 1 && uniqueMatches.length === 1) {
```
bc-peng, line 60: `results` always contains all registeredTokens, hence it should be always greater than 1. See L55.
I think you want to do the same check on the `matched`, L66.

bc-peng: On a second thought, I would find the biggest matches value and how many tokens have that value.
If the number is greater than one, throw an error then.
My concern is that `uniqueMatches.length === 1` might not cover all equal matches cases.

animesh1987: Yes I think I am checking for the case where there are more than one `registeredTokens` in result. 

I am looking for testing the case where there are multiple tokens but for `uniqueMatches` only have a length of 1. 

Right now I think it works well for the existing test cases. Do we want to test for if `uniqueMatches > 1`?

bc-peng: `results.length` equals `registeredTokens.length`.

Is it possible that we have a result like this?
```
[{ matches: 2 }, { matches: 2 }, { matches: 1 }, { matches: 1 }]
```

davidchin: I'd move the check down after the sort. i.e.:

```ts
const matchedResults = results
    .sort((a, b) => b.matches - a.matches)
    .filter((result) => result.matches > 0);

if (matchedResults.length > 1 && matchedResults[0].matches === matchedResults[1].matches) {
    if (getEnvironment() === 'development') {
        getDefaultLogger().warn('The provided query matches at least two strategies with the same specificity. This warning can be resolved by making their resolve ID more specific.');
    }
}

const matched = matchedResults[0];

if (matched && matched.token) {
    return matched.token;
}
```

Also, I'd make it warning in development rather than an error.

animesh1987: Thanks @davidchin that seems to be good, but I am still on the fence regarding the error vs warn. Since warnings can be ignored and can result in issues if pushed to prod. Thoughts @bc-peng?

bc-peng: Thank you both.
We may want to eliminate uncertainty in production, so an error seems more appropriate to me.

davidchin: I suggested warning because it's not really an error if it resolves to the right strategy even if there are more than one match. But if you think error is more appropriate, I'm also happy with it. 🙂  Just make sure you do manual testing for payment methods that don't have e2e tests.

animesh1987: Thanks for the input, I will update and do the necessary testing 😄

bc-peng (approved):
Great job. I loved the way you implemented this change.

animesh1987:
> Great job. I loved the way you implemented this change.

It was a team effort 😄

PR #1962, opened 2023-04-26
fix(payment): PAYPAL-2036 added shippingAddressEditable option to Braintree payment configuration

serhii-tkachenko:
## What?
Added `shippingAddressEditable` option to Braintree payment configuration.

`shippingAddressEditable` is optional, because it is `true` by default on Braintree side

## Why?
To remove an ability to change shipping address by shoppers who use braintree in payment section of checkout page

## Testing / Proof
Unit tests
Manual tests

Before:
https://user-images.githubusercontent.com/25133454/234507334-8419ef8d-8660-4fd4-9d9b-d5484a67ecba.mov

After:
https://user-images.githubusercontent.com/25133454/234507271-46fd0aa1-8cf2-4a74-989b-80eb4bec95e4.mov

### DIFF
--- a/packages/core/src/payment/strategies/braintree/braintree-payment-processor.ts
+++ b/packages/core/src/payment/strategies/braintree/braintree-payment-processor.ts
@@ -38,6 +38,7 @@ export interface PaypalConfig {
     currency: string;
     locale: string;
     offerCredit?: boolean;
+    shippingAddressEditable?: boolean;
     shippingAddressOverride?: BraintreeShippingAddressOverride;
     shouldSaveInstrument?: boolean;
 }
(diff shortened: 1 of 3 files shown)


packages/core/src/payment/strategies/braintree/braintree-payment-processor.ts
```diff
@@ -38,6 +38,7 @@ export interface PaypalConfig {
     currency: string;
     locale: string;
     offerCredit?: boolean;
+    shippingAddressEditable?: boolean;
```
animesh1987, line 41: Does this need to be optional?

serhii-tkachenko: Yes, we can skip adding this option for other strategies

animesh1987 (approved):
Thanks for updating the PR description.

PR #1968, opened 2023-05-04
feat(payment): BOLT-484 Add styling for Bolt Button on PDP

PavlenkoM:
## What?
Add styling for Bolt Button on PDP

## Why?
Add customisation options for Bolt smart payment button on Product Deteiles Page

## Testing / Proof
Unit test & manual testing

### DIFF
--- a/packages/bolt-integration/src/bolt-button-strategy.ts
+++ b/packages/bolt-integration/src/bolt-button-strategy.ts
@@ -1,3 +1,5 @@
+import { stringify } from 'query-string';
+
 import {
     CheckoutButtonInitializeOptions,
     CheckoutButtonStrategy,
@@ -6,7 +8,13 @@ import {
     PaymentIntegrationService,
 } from '@bigcommerce/checkout-sdk/payment-integration-api';
 
-import { BoltHostWindow, BoltPaymentMethod } from './bolt';
+import {
+    BoltButtonStyleOptions,
+    BoltHostWindow,
+    BoltPaymentMethod,
+    StyleButtonShape,
+    StyleButtonSize,
+} from './bolt';
 import { WithBoltButtonInitializeOptions } from './bolt-button-initialize-options';
 import BoltScriptLoader from './bolt-script-loader';
 
@@ -21,7 +29,7 @@ export default class BoltButtonStrategy implements CheckoutButtonStrategy {
         options: CheckoutButtonInitializeOptions & WithBoltButtonInitializeOptions,
     ): Promise<void> {
         const { bolt, containerId, methodId } = options;
-        const { buyNowInitializeOptions } = bolt || {};
+        const { buyNowInitializeOptions, style } = bolt || {};
 
         if (!methodId) {
             throw new InvalidArgumentError(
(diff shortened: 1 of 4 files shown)


packages/bolt-integration/src/bolt-button-strategy.ts
```diff
@@ -100,20 +116,67 @@ export default class BoltButtonStrategy implements CheckoutButtonStrategy {
...
+            variant: 'ppc',
+            ...(buttonHeight && { height: buttonHeight }),
+            ...(buttonBorderRadius && { border_radius: buttonBorderRadius }),
+        });
+
+        return `https://${domainUrl}/v1/checkout_button?${urlParameters}`;
+    }
+
+    private getButtonHeight(buttonSize?: StyleButtonSize): number | undefined {
+        if (!buttonSize) {
```
serhii-tkachenko, line 148: May we pass default `buttonHeight` in cases if `buttonSize` is not provided? In this case we have control on button size instead of Bolt.

What do you think?

packages/bolt-integration/src/bolt-button-strategy.ts
```diff
@@ -100,20 +116,67 @@ export default class BoltButtonStrategy implements CheckoutButtonStrategy {
...
+            default:
+                return 40;
+        }
+    }
+
+    private getButtonBorderRadius(
+        buttonShape?: StyleButtonShape,
+        buttonHeight?: number,
+    ): number | undefined {
+        if (!buttonShape) {
```
serhii-tkachenko, line 169: Same question as above

packages/bolt-integration/src/bolt-button-strategy.ts
```diff
@@ -100,20 +116,67 @@ export default class BoltButtonStrategy implements CheckoutButtonStrategy {
...
+        const buttonHeight = this.getButtonHeight(style?.size);
+        const buttonBorderRadius = this.getButtonBorderRadius(style?.shape, buttonHeight);
+        const urlParameters = stringify({
+            publishable_key: publishableKey,
+            variant: 'ppc',
+            ...(buttonHeight && { height: buttonHeight }),
+            ...(buttonBorderRadius && { border_radius: buttonBorderRadius }),
+        });
+
+        return `https://${domainUrl}/v1/checkout_button?${urlParameters}`;
```
serhii-tkachenko, line 144: [Optional to change]

Here is another way how this piece can be done:
```
return stringifyUrl({
    currentUrl:  `https://${domainUrl}/v1/checkout_button`,
    query: {
        publishable_key: publishableKey,
        variant: 'ppc',
        height: buttonHeight,
        border_radius: buttonBorderRadius,
    },
});
```

serhii-tkachenko (approved):
Nice work! 👍

packages/bolt-integration/src/bolt-button-strategy.ts
PavlenkoM: This button also can be configured by merchant on the Bolt side. So I think in case if in some reasons we won't have configs from BC side, merchant can control this button from Bolt side, instead of overwriting by default options.

packages/bolt-integration/src/bolt-button-strategy.ts
PavlenkoM: done

PR #1970, opened 2023-05-05
refactor(payment): PAYPAL-2414 added PayPalCommerceIntegrationService method creation

bc-dronov:
## What?
Payment strategies have been refactored by moving  PayPalCommerceIntegrationService method in the separate file.

## Why?
To avoid code duplication in several places and make it easier to maintain in the future.

## Testing / Proof
All tests has been passed.

### DIFF
--- a/packages/paypal-commerce-integration/src/paypal-commerce-alternative-methods/create-paypal-commerce-alternative-methods-button-strategy.ts
+++ b/packages/paypal-commerce-integration/src/paypal-commerce-alternative-methods/create-paypal-commerce-alternative-methods-button-strategy.ts
@@ -1,37 +1,19 @@
-import { createFormPoster } from '@bigcommerce/form-poster';
-import { createRequestSender } from '@bigcommerce/request-sender';
-import { getScriptLoader } from '@bigcommerce/script-loader';
-
 import {
     CheckoutButtonStrategyFactory,
     toResolvableModule,
 } from '@bigcommerce/checkout-sdk/payment-integration-api';
 
-import {
-    PayPalCommerceIntegrationService,
-    PayPalCommerceRequestSender,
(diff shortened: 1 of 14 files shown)


packages/paypal-commerce-integration/src/paypal-commerce-alternative-methods/create-paypal-commerce-alternative-methods-button-strategy.ts
```diff
@@ -1,37 +1,19 @@
...
-        createFormPoster(),
-        paymentIntegrationService,
-        new PayPalCommerceRequestSender(createRequestSender({ host: getHost() })),
-        new PayPalCommerceScriptLoader(getScriptLoader()),
-    );
-
-    return new PayPalCommerceAlternativeMethodsButtonStrategy(
+> = (paymentIntegrationService) =>
+    new PayPalCommerceAlternativeMethodsButtonStrategy(
         paymentIntegrationService,
```
andriiVitvitskyi1990, line 14: As i see we have paymentIntegrationService as first argument in every provider. Maybe we can move it to the same mutual file and return it from there with  `new PayPalCommerceIntegrationService`. Is it makes sense?)

serhii-tkachenko: I don't think we should, because `paymentIntegrationService` is a prop that comes from another place. It could be better to keep strategy initialisation as simple as possible.

serhii-tkachenko (approved):
Nice work! Looks good to me 👍

bc-nick (approved):

andriiVitvitskyi1990 (approved):

PR #1971, opened 2023-05-08
feat(payment): INT-7516 SquareV2: Throw a `PaymentExecuteError` if tokenization fails

mauricio-sg:
## What? [INT-7516](https://bigcommercecloud.atlassian.net/browse/INT-7516)
1. Validate only blacklisted fields.
2. Move `PaymentExecuteError` to the `payment-integration-api` package.
3. Throw a `PaymentExecuteError` instead of a plain `Error` if tokenization fails.

## Why?
Because we only rely on the validation of blacklisted fields, tokenization errors are going to increase. So developers should be able to catch this error, which is more specific, and display a more user-friendly message.

Q: Why do we only validate blacklisted fields?
A: Because only blacklisted fields have shown stable behavior when triggering events. Square is aware of this and plans to fix it at some point.

## Testing / Proof
https://user-images.githubusercontent.com/4843328/236933218-6e5c1a90-75fe-4503-8f45-0413e1e81bd1.mov

## Dependency of
👉https://github.com/bigcommerce/checkout-js/pull/1268

[INT-7516]: https://bigcommercecloud.atlassian.net/browse/INT-7516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vmparra (approved):
LGTM 👍🏻

mauricio-sg:
💚 Tested, evidence can be found on the ticket.

animesh1987 (approved):

PR #1973, opened 2023-05-11
fix(payment): PAYPAL-2370 updated incoming parameters for ACH

bc-nick:
## What?
...

## Why?
...

## Testing / Proof
...

### DIFF
--- a/packages/braintree-integration/src/braintree-paypal-ach/braintree-paypal-ach-payment-strategy.ts
+++ b/packages/braintree-integration/src/braintree-paypal-ach/braintree-paypal-ach-payment-strategy.ts
@@ -24,7 +24,7 @@ import { WithBraintreePaypalAchPaymentInitializeOptions } from './braintree-payp
 
 export default class BraintreePaypalAchPaymentStrategy implements PaymentStrategy {
     private usBankAccount?: BraintreeBankAccount;
-    private mandateText? = '';
+    private getMandateText?: () => string | undefined;
 
     constructor(
         private paymentIntegrationService: PaymentIntegrationService,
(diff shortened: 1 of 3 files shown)


packages/braintree-integration/src/braintree-paypal-ach/braintree-paypal-ach-payment-strategy.ts
```diff
@@ -24,7 +24,7 @@ import { WithBraintreePaypalAchPaymentInitializeOptions } from './braintree-payp
 
 export default class BraintreePaypalAchPaymentStrategy implements PaymentStrategy {
     private usBankAccount?: BraintreeBankAccount;
-    private mandateText? = '';
+    private getMandateText?: () => string | undefined;
```
serhii-tkachenko, line 27: I think we can skip adding `undefined`, because getMandateText is an optional param/variable.

packages/braintree-integration/src/braintree-paypal-ach/braintree-paypal-ach-payment-strategy.ts
```diff
@@ -79,16 +79,18 @@ export default class BraintreePaypalAchPaymentStrategy implements PaymentStrateg
             throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
         }
 
-        if (!this.mandateText) {
+        const mandateText = this.getMandateText && this.getMandateText();
```
serhii-tkachenko, line 82: It might be better to check this.getMandateText type. It should be a function, otherwise, this line of code will have an error.

serhii-tkachenko (approved):
LGTM! 👍

PR #1984, opened 2023-05-19
fix(analytics): DATA-11047 Data Layer Updates - Field Additions and Changes

bc-rmalyavc:
## What? [DATA-11047](https://bigcommercecloud.atlassian.net/browse/DATA-11047)
- Add `product_id` field type to string for line_items
- Add `retail_price` field for physical and digital line items

## Why?
This change is supposed to fix the BODL payload for checkout events. As per [this document](https://docs.google.com/document/d/13qlYa9gtuVcjs2iBZ-jWP74W1GPRQw9J496t7QYbbJg/edit#heading=h.a4ql1e2fxt52) `product_id` field should be string and also `retail_price` field should be there for line_items

## Testing / Proof
Manually tested on dev store

[DATA-11047]: https://bigcommercecloud.atlassian.net/browse/DATA-11047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -212,10 +212,16 @@ export default class BodlEmitterService implements BodlService {
...
             };
         });
 
-        return [...customItems, ...physicalAndDigitalItems, ...giftCertificateItems];
+        return [...customItems, ...physicalAndDigitalItems, ...giftCertificateItems].map(
+            (item, index) => ({
+                ...item,
+                index,
+            }),
+        );
```
solofeed, line 224:
```suggestion
        return [
            ...customItems,
            ...physicalAndDigitalItems,
            ...giftCertificateItems
          ].map((item, index) => ({ ...item, index }));
       );
```

solofeed (approved):
LGTM! 👍 

Reminder: Merge this PR only after releasing backend PR

_add this pr to description please_

animesh1987:
hey @bc-rmalyavc can you please remove the dist folder from this PR.

bc-rmalyavc:
Hey @animesh1987,
Removed dist files. Thanks for the catch

packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -212,10 +212,16 @@ export default class BodlEmitterService implements BodlService {
...
                 category_names: item.categoryNames || [],
+                retail_price: item.retailPrice,
             };
         });
 
-        return [...customItems, ...physicalAndDigitalItems, ...giftCertificateItems];
+        return [...customItems, ...physicalAndDigitalItems, ...giftCertificateItems].map(
+            (item) => ({
+                ...item,
+                product_id: String(item.product_id),
```
bc-peng, line 222: I cannot access the doc in the description.
This change makes the produce id as a string.
Is it intentional?

bc-peng: I cannot access the doc in the description.
This change makes the product id as a string.
Is it intentional?

bc-rmalyavc: @bc-peng,
Yes, it's a requirement for BODL. Generally it's because of gift certificates. They don't have product_id, so in BODL context, gift_certificate_id becomes product_id. This field has UUID format, which is string. So it has been decided, that product_id will be always string for BODL events

bc-rmalyavc (commented):
@bc-peng,
Yes, it's a requirement for BODL. Generally it's because of gift certificates. They don't have product_id, so in BODL context, gift_certificate_id becomes product_id. This field has UUID format, which is string. So it has been decided, that product_id will be always string for BODL events

bc-peng (approved):
lgtm👍

animesh1987:
Hey @bc-rmalyavc this is a public repository, so unless the doc in the PR description is to be shared externally I would say maybe better to remove it and explain the `why` question better as to what problem we need to solve.

packages/core/src/cart/line-item.ts
```diff
@@ -57,6 +57,7 @@ export interface LineItem {
     couponAmount: number;
     listPrice: number;
     salePrice: number;
+    retailPrice: number;
```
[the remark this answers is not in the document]
bc-rmalyavc: @animesh1987,
I have updated the endpoint on backend side in a way that it should be always number. So that if retail price is not provided for a product, then 0 will be returned.
See
https://github.com/bigcommerce/bigcommerce/pull/52634/files#diff-b956c4c7fb4801b21e2e3c4cb2a725dfb819dd4b4c33a762270f13b4c3a6ec86R64
and
https://github.com/bigcommerce/bigcommerce/pull/52634/files#diff-16a680b18279d5d71064084b8613e1845784d3351afad2899c6a992dcc5ae7f1R69.

Also just updated the `Why` section. Should be clear enough now.

PR #1990, opened 2023-05-22
feat(payment): PAYPAL-2451 ACH vaulting

bc-nick:
## What?

Vaulting the Braintree ACH payment method

## Why?

The ability to use the stored data to place an order

## Testing / Proof

All test passed


packages/braintree-integration/src/braintree-paypal-ach/braintree-paypal-ach-payment-strategy.ts
```diff
@@ -25,6 +29,7 @@ import { WithBraintreePaypalAchPaymentInitializeOptions } from './braintree-payp
 export default class BraintreePaypalAchPaymentStrategy implements PaymentStrategy {
     private usBankAccount?: BraintreeBankAccount;
     private getMandateText?: () => string;
+    private paymentMethod?: PaymentMethod;
```
serhii-tkachenko, line 32: I don't really see any use cases why we should create this variable here

packages/braintree-integration/src/braintree-paypal-ach/braintree-paypal-ach-payment-strategy.ts
```diff
@@ -69,7 +74,103 @@ export default class BraintreePaypalAchPaymentStrategy implements PaymentStrateg
             throw new PaymentArgumentInvalidError(['payment']);
         }
 
-        const { paymentData } = payment;
+        if (!this.paymentMethod) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+        }
```
serhii-tkachenko, line 79: This code can be removed

packages/braintree-integration/src/braintree-paypal-ach/braintree-paypal-ach-payment-strategy.ts
```diff
@@ -69,7 +74,103 @@ export default class BraintreePaypalAchPaymentStrategy implements PaymentStrateg
...
+            } = this.paymentMethod;
+
+            if (!isVaultingEnabled) {
+                throw new InvalidArgumentError(
+                    'Vaulting is disabled but a vaulted instrument was being used for this transaction',
+                );
+            }
+
+            return null;
+        }
```
serhii-tkachenko, line 173: In this case, to avoid using `this.paymentMethod` we can write code a bit different:
If you check OrderPaymentRequestBody interface, you can see that you have access to payment `methodId` and `gatewayId`. So instead of creating and using `this.paymentMethod` + handle the error all over the place we can get specific paymentMethod from checkout state.

```suggestion
        const {
            methodId,
            paymentData = {},
        } = payment;

        if (isVaultedInstrument(paymentData)) {
            const state = this.paymentIntegrationService.getState();
            const { config } = state.getPaymentMethodOrThrow(methodId);

            if (!config.isVaultingEnabled) {
                throw new InvalidArgumentError(
                    'Vaulting is disabled but a vaulted instrument was being used for this transaction',
                );
            }

            return null;
        }
```

packages/braintree-integration/src/braintree-paypal-ach/braintree-paypal-ach-payment-strategy.ts
```diff
@@ -69,7 +74,103 @@ export default class BraintreePaypalAchPaymentStrategy implements PaymentStrateg
...
+                set_as_default_stored_instrument: shouldSetAsDefaultInstrument || null,
+                device_info: sessionId || null,
+                ach_account: {
+                    routing_number: routingNumber,
+                    last4: accountNumber.substr(-4),
+                    token: nonce,
+                    email: email || null,
+                },
+            },
+        };
```
serhii-tkachenko, line 146: Is it possible to use billing address from state rather from checkout object?

```suggestion
        const billingAddress = state.getBillingAddressOrThrow();

        const { shouldSaveInstrument, shouldSetAsDefaultInstrument, routingNumber, accountNumber } =
            paymentData;

        const paymentPayload = {
            formattedPayload: {
                vault_payment_instrument: shouldSaveInstrument || null,
                set_as_default_stored_instrument: shouldSetAsDefaultInstrument || null,
                device_info: sessionId || null,
                ach_account: {
                    routing_number: routingNumber,
                    last4: accountNumber.substr(-4),
                    token: nonce,
                    email: billingAddress.email || null,
                },
            },
        };
```

serhii-tkachenko (approved):
Nice work! 👍

PR #2005, opened 2023-05-31
fix(payment): CHECKOUT-7498 Update registry key as a combination

animesh1987:
## What?
Update registry key as a combination of methodId and gatewayId.

## Why?
If two payment methods have the same methodId, toggling between them causes issue where clicking on the second method doesn't initialise the new payment method since strategy registry checks based on methodId only.

Update the key to be a combination of methodId and gatewayId.

## Testing / Proof
- Tests


packages/core/src/payment/payment-strategy-action-creator.spec.ts
```diff
@@ -249,13 +253,13 @@ describe('PaymentStrategyActionCreator', () => {
             expect(actions).toEqual([
                 {
                     type: PaymentStrategyActionType.InitializeRequested,
-                    meta: { methodId: method.id },
+                    meta: { key },
```
davidchin, line 256: Is there a reason what it can't be `{ methodId: method.id, gatwayId: method.gateway }`?

animesh1987: Since we changed the implementation in the action-creator to be a combined value of `method.id + method.gateway`, have kept this consistent that way. Otherwise we will need to pass the object from action-creator as well. And subsequently apply the logic in reducer/selector?

davidchin: Yeah I'm suggesting to have `{ methodId: method.id, gatwayId: method.gateway }` all the way through. Currently, action creators need to know internally how the data is reduced and stored - as they need to know they are stored in a flat dictionary and the format of each key (`method.id + method.gateway` instead of `method.gateway + method.id` or `method.id + '.' + method.gateway`). Ideally we want to avoid that by simply passing the identifiers.

davidchin (approved):
Nice one! 👍

PR #2008, opened 2023-06-02
feat(payment): PI-120 Autofilling of holderName to be configurable

bc-ania:
## What?

Autofilling of holderName to be configurable.
Checkbox is available in Adyen Settings in CP allowing merchants to avoid auto populating of billing name to the card holder name

Checkbox state is enabled by default for both existing and new merchants

## Why?

At the moment, Merchants using Adyen will see that the credit card holder name is being autofilled from the Shoppers selected billing address at checkout.

 We send shopperName from billing address, holderName (from card) is sent directly as is to card form (for Adyen V3)

## Testing / Proof

Screenshot 2023-06-13 at 16 30 11

### DIFF
--- a/packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.spec.ts
+++ b/packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.spec.ts
@@ -482,6 +482,71 @@ describe('AdyenV3PaymentStrategy', () => {
                 expect(additionalActionComponent.unmount).toHaveBeenCalledTimes(1);
             });
 
+            it('prefills holderName with billingAddress data if prefillCardHolderName is true', async () => {
+                jest.spyOn(
+                    paymentIntegrationService.getState(),
+                    'getPaymentMethodOrThrow',
+                ).mockReturnValue({
+                    ...getAdyenV3(),
+                    initializationData: {
+                        ...getAdyenV3().initializationData,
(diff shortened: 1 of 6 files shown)


packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.ts
```diff
@@ -358,11 +361,15 @@ export default class Adyenv3PaymentStrategy implements PaymentStrategy {
         return new Promise((resolve, reject) => {
             const billingAddress = this._paymentIntegrationService.getState().getBillingAddress();
 
+            const { prefillCardHolderName } = paymentMethod.initializationData;
```
vitalii-koshovyi, line 364: I believe we need to cover it with tests

packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.spec.ts
```diff
@@ -477,6 +477,16 @@ describe('AdyenV3PaymentStrategy', () => {
                 expect(additionalActionComponent.unmount).toHaveBeenCalledTimes(1);
             });
 
+            it('calls paymentComponent with prefillCardHolderName in initializationData', () => {
```
vitalii-koshovyi, line 480: this test should be skipped, it doesn't test anything

vitalii-koshovyi (approved):

PR #2011, opened 2023-06-08
feat(payment): CHECKOUT-7504 Enable recaptcha for apple pay

animesh1987:
## What?
- Enable recaptcha for applepay, so stores can have apple pay and recaptcha enabled at the same time.
- Remove code in core package related to apple pay buttons

## Why?
Enable recaptcha for apple pay buttons, so merchants can benefit from added spam protection store wide. 

## Testing / Proof
- CI 
- Screencast (TBD)


packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
```diff
@@ -473,9 +473,12 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
                 );
             }
 
+            await this._paymentIntegrationService.verifyCheckoutSpamProtection();
```
davidchin, line 476: What would it look like when visible captcha is triggered?

animesh1987: So far in my testing, I haven't seen the visible captcha being triggered 🤔 Is there a way we can force trigger it?

animesh1987: Checked offline and tested with user agent, once the challenge is visible. It completely breaks apple pay sheet and requires safari to be force quit. Hence will need to try another approach.

packages/core/src/payment-integration/default-payment-integration-service.ts
```diff
@@ -192,4 +194,12 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
 
         return this._storeProjection.getState();
     }
+
+    async verifyCheckoutSpamProtection(): Promise<PaymentIntegrationSelectors> {
+        await this._store.dispatch(
```
davidchin, line 199: Maybe check if `shouldExecuteSpamCheck` is `true` first before trigger the action, similar to the logic we have here: https://github.com/bigcommerce/checkout-sdk-js/blob/605b50164333e6d37065ecd63f15a625a34e3155/packages/core/src/payment/payment-strategy-action-creator.ts#L55-L56

davidchin (approved):
Fantastic work! 👍

bc-peng (approved):
Great job on completing such a massive task! 🎉

TomA-R (approved):

PR #2014, opened 2023-06-09
fix(payment): PAYPAL-2602 Changed getting the billing address

bc-nick:
## What?

Changed getting the billing address

## Why?

Shoppers should use our native billing address form on the checkout page. 

## Testing / Proof

All tests passed


andriiVitvitskyi1990 (approved):

vitalii-koshovyi (approved):

packages/payment-integration-api/src/payment/payment.ts
```diff
@@ -78,12 +78,6 @@ export interface WithBankAccountInstrument {
     firstName?: string;
     lastName?: string;
     businessName?: string;
-    address1: string;
-    address2: string;
-    city: string;
-    countryCode: string;
-    postalCode: string;
-    stateOrProvinceCode: string;
     shouldSaveInstrument?: boolean;
```
animesh1987, line 81: Can this potentially break other integrations?

bc-nick: @animesh1987 no, this is only for Braintree ACH payment method

animesh1987 (approved):

packages/payment-integration-api/src/payment/payment.ts
animesh1987: Just a note for future, if this is only used by braintree shall we move this in a braintree package?

PR #2020, opened 2023-06-15
feat(checkout): CHECKOUT-7536 Load extensions

bc-peng:
## What?
- Retrieve all extension configurations on a store.

- Integrate them seamlessly into the existing checkout state.

## Why?
Efficiently manage and utilize extensions.

## Testing / Proof
- CI checks.
- Manual testing.
   image

### DIFF
--- a/packages/core/src/extension/extension-action-creator.ts
+++ b/packages/core/src/extension/extension-action-creator.ts
@@ -1,0 +1,43 @@
+import { createAction, createErrorAction, ThunkAction } from '@bigcommerce/data-store';
+import { Observable, Observer } from 'rxjs';
+
+import { InternalCheckoutSelectors } from '../checkout';
+import { RequestOptions } from '../common/http-request';
+
+import { ExtensionAction, ExtensionActionType } from './extension-actions';
+import { ExtensionRequestSender } from './extension-request-sender';
+
+export class ExtensionActionCreator {
+    constructor(private _requestSender: ExtensionRequestSender) {}
+
+    loadExtensions(
+        options?: RequestOptions,
+    ): ThunkAction<ExtensionAction, InternalCheckoutSelectors> {
+        return (store) =>
+            Observable.create((observer: Observer<ExtensionAction>) => {
+                const state = store.getState();
+                const cart = state.cart.getCartOrThrow();
+
+                observer.next(createAction(ExtensionActionType.LoadExtensionsRequested));
+
+                this._requestSender
+                    .loadExtensions({
+                        ...options,
+                        params: { ...options?.params, cartId: cart.id },
+                    })
+                    .then((response) => {
+                        const extensions = response.body;
+
+                        observer.next(
+                            createAction(ExtensionActionType.LoadExtensionsSucceeded, extensions),
+                        );
+                        observer.complete();
+                    })
+                    .catch((response) => {
+                        observer.error(
+                            createErrorAction(ExtensionActionType.LoadExtensionsFailed, response),
+                        );
+                    });
+            });
+    }
+}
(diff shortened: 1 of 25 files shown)


animesh1987 (commented):
Nice work overall. Had a question, have we thought about moving `extension` package to it's own separate package out of core?

bc-peng:
> Nice work overall. Had a question, have we thought about moving `extension` package to it's own separate package out of core?

Thank you for your suggestion.
I'm not sure I see the benefits, given that the majority of the files are already defined here.
If we decide to make it a package, we'll need to figure out how to provide access to the checkout service and state similar to the payment integration API. It may be tricky but doable.

packages/core/src/extension/extension-action-creator.ts
```diff
@@ -0,0 +1,43 @@
...
+export class ExtensionActionCreator {
+    constructor(private _requestSender: ExtensionRequestSender) {}
+
+    loadExtensions(
+        options?: RequestOptions,
+    ): ThunkAction<ExtensionAction, InternalCheckoutSelectors> {
+        return (store) =>
+            Observable.create((observer: Observer<ExtensionAction>) => {
+                const state = store.getState();
+                const cart = state.cart.getCartOrThrow();
```
animesh1987, line 19: Do we need to pass cartId for extensions? Aren't they something that belong at store level?

packages/core/src/extension/extension-request-sender.ts
```diff
@@ -0,0 +1,28 @@
...
+import { Extension } from './extension';
+
+export const EXTENSIONS_API_URL = '/api/storefront/checkout-extensions';
+
+export class ExtensionRequestSender {
+    constructor(private _requestSender: RequestSender) {}
+
+    loadExtensions({ timeout, params }: RequestOptions = {}): Promise<Response<Extension[]>> {
+        return this._requestSender.get(EXTENSIONS_API_URL, {
+            timeout,
```
animesh1987, line 19: What does this timeout do?

packages/core/src/extension/extension-request-sender.ts
```diff
@@ -0,0 +1,28 @@
...
+            headers: {
+                Accept: ContentType.JsonV1,
+                'X-API-INTERNAL': INTERNAL_USE_ONLY,
```
animesh1987, line 22: I remember there was discussion offline, do we need to add this?

packages/core/src/extension/extension-request-sender.ts
bc-peng: I think we offer `timeout`for almost all requests, it is a standard option in `@bigcommerce/request-sender`.
> Provide this option if you want to cancel or time out the request. If the timeout object completes before the request, the request will be cancelled.

packages/core/src/extension/extension-request-sender.ts
bc-peng: Removed.

packages/core/src/extension/extension-action-creator.ts
bc-peng: I put it here to match with loading cart and payment methods.
Removed.
On another note, when we render an extension, we'll pass `cartId` and `extensionId` to an iframe as params.

packages/core/src/extension/extension-request-sender.ts
animesh1987: Think seems like a copy paste issue, as the client will need to pass `timeout` as options, which we rarely do 🤔

packages/core/src/extension/extension-action-creator.ts
animesh1987: Yes that can we done at a different layer, if we don't need information as part of action creation we should avoid it.

animesh1987 (approved):
Nice work 👍

PR #2028, opened 2023-06-22
feat(checkout): CHECKOUT-7537 Render extensions

bc-peng:
## What?
- Retrieve all extension configs from the checkout state.
- Render an extension’s iframe based its config.

## Why?
Efficiently utilize extensions.

## Testing / Proof
- CI checks.

### DIFF
--- a/packages/core/src/checkout/checkout-service.ts
+++ b/packages/core/src/checkout/checkout-service.ts
@@ -1396,7 +1396,7 @@ export default class CheckoutService {
     renderExtension(container: string, region: ExtensionRegion): Promise<CheckoutSelectors> {
         const action = this._extensionActionCreator.renderExtension(container, region);
 
-        return this._dispatch(action, { queueId: 'renderExtension' });
+        return this._dispatch(action, { queueId: 'extensions' });
     }
 
     /**
--- a/packages/core/src/checkout/checkout-service.ts
+++ b/packages/core/src/checkout/checkout-service.ts
@@ -20,7 +20,7 @@ import {
     ExecutePaymentMethodCheckoutOptions,
     GuestCredentials,
 } from '../customer';
-import { ExtensionActionCreator, ExtensionRegions } from '../extension';
+import { ExtensionActionCreator, ExtensionRegion } from '../extension';
 import { FormFieldsActionCreator } from '../form';
 import { CountryActionCreator } from '../geography';
 import { OrderActionCreator, OrderRequestBody } from '../order';
@@ -1393,7 +1393,7 @@ export default class CheckoutService {
      * @param region The name of an area where the extension should be presented.
      * @returns A promise that resolves to the current state.
      */
-    renderExtension(container: string, region: ExtensionRegions): Promise<CheckoutSelectors> {
+    renderExtension(container: string, region: ExtensionRegion): Promise<CheckoutSelectors> {
         const action = this._extensionActionCreator.renderExtension(container, region);
 
         return this._dispatch(action, { queueId: 'renderExtension' });
--- a/packages/core/src/checkout/checkout-service.ts
+++ b/packages/core/src/checkout/checkout-service.ts
@@ -20,7 +20,7 @@ import {
     ExecutePaymentMethodCheckoutOptions,
     GuestCredentials,
 } from '../customer';
-import { ExtensionActionCreator } from '../extension';
+import { ExtensionActionCreator, ExtensionRegions } from '../extension';
 import { FormFieldsActionCreator } from '../form';
 import { CountryActionCreator } from '../geography';
 import { OrderActionCreator, OrderRequestBody } from '../order';
@@ -1368,7 +1368,7 @@ export default class CheckoutService {
      * Loads a list of extensions available for checkout.
      *
      * ```js
-     * const state = service.loadExtensions();
+     * const state = await service.loadExtensions();
      *
      * console.log(state.data.getExtensions());
      * ```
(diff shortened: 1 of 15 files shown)


packages/core/src/extension/extension-iframe.ts
```diff
@@ -0,0 +1,42 @@
+import { Cart } from '@bigcommerce/checkout-sdk/payment-integration-api';
+
+import { InvalidExtensionConfigError } from './errors';
+import { Extension } from './extension';
+
+export class ExtensionIframe {
+    private _iframe: HTMLIFrameElement;
+
+    constructor(private _containerId: string, private _extension: Extension, cart: Cart) {
```
animesh1987, line 9: Why not have _ for cart as well, seems like a private member?

bc-peng: Since we only pass`extensionId` and `cartId` to an extension here, should I update it to `cartId` instead?

animesh1987: Yes I would say so, less is better :D We can always enrich with more data.

bc-peng: Thanks. This area has been updated accordingly.

packages/core/src/extension/extension-action-creator.ts
```diff
@@ -34,4 +37,38 @@ export class ExtensionActionCreator {
...
     }
+
+    renderExtension(
+        container: string,
+        region: ExtensionRegions,
+    ): ThunkAction<ExtensionAction, InternalCheckoutSelectors> {
+        return (store) =>
+            Observable.create((observer: Observer<ExtensionAction>) => {
+                const state = store.getState();
+                const cart = state.cart.getCartOrThrow();
```
animesh1987, line 48:
```suggestion
                const { id: cartId } = state.cart.getCartOrThrow();
```

packages/core/src/extension/extension-action-creator.ts
```diff
@@ -34,4 +37,38 @@ export class ExtensionActionCreator {
...
+    renderExtension(
+        container: string,
+        region: ExtensionRegions,
+    ): ThunkAction<ExtensionAction, InternalCheckoutSelectors> {
+        return (store) =>
+            Observable.create((observer: Observer<ExtensionAction>) => {
+                const state = store.getState();
+                const { id: cartId } = state.cart.getCartOrThrow();
+                const extensions = state.extensions.getExtensions();
+                const extension = extensions?.filter((e) => e.region === region)[0];
```
animesh1987, line 50: I know right now we only support one extension, but this will need to be changed if we have two exensions both for different regions. 
Maybe we need to add a method called `extensions.getExtensionByRegion` which takes a region as an argument and returns the extension for that region:
`state.extensions.getExtensionByRegion(region: ExtensionRegion): Extension`

bc-peng: Thanks. I have added `state.extensions.getExtensionByRegion(region: ExtensionRegion): Extension` and renamed `ExtensionRegions` to `ExtensionRegion`.

packages/core/src/extension/extension-selector.ts
```diff
@@ -19,6 +20,11 @@ export function createExtensionSelectorFactory(): ExtensionSelectorFactory {
         (data) => () => data,
     );
 
+    const getExtensionByRegion = createSelector(
+        (state: ExtensionState) => state.data,
+        (data) => (region: ExtensionRegion) => data?.filter((e) => e.region === region)[0],
```
animesh1987, line 25: Can we use `find` here, it will return the first element that is truthy.

bc-peng: Good point. Code updated. Thanks.

animesh1987 (approved):
Nice work @bc-peng. LGTM 👍 , @davidchin can you please take a look at this as well please.

packages/core/src/checkout/checkout-service.ts
```diff
@@ -1384,6 +1384,21 @@ export default class CheckoutService {
...
+     *
+     * @alpha
+     * @param container The ID of a container which the extension should be inserted.
+     * @param region The name of an area where the extension should be presented.
+     * @returns A promise that resolves to the current state.
+     */
+    renderExtension(container: string, region: ExtensionRegion): Promise<CheckoutSelectors> {
+        const action = this._extensionActionCreator.renderExtension(container, region);
+
+        return this._dispatch(action, { queueId: 'renderExtension' });
```
davidchin, line 1399:
```suggestion
        return this._dispatch(action, { queueId: 'extensions' });
```

packages/core/src/extension/extension-action-creator.ts
```diff
@@ -34,4 +37,37 @@ export class ExtensionActionCreator {
...
+                }
+
+                observer.next(createAction(ExtensionActionType.RenderExtensionRequested));
+
+                try {
+                    const iframe = new ExtensionIframe(container, extension, cartId);
+
+                    iframe.attach();
+
+                    observer.next(createAction(ExtensionActionType.RenderExtensionSucceeded));
```
davidchin, line 64: Do we want to resolve the async action only when the iframe is rendered successfully? At the moment, it is resolved as soon as the iframe is attached, meaning that it might not be actually rendered.

bc-peng: Thanks DC for your review.
`iframe.attach();` is still WIP. In the next PR, by the time `iframe.attach()` finishes, the iframe should have adjusted its own size based on the iframe content.

packages/core/src/checkout/checkout-service.ts
bc-peng: `extensions` is already used as `loadExtensions` 's queueId.
Shall I remove or rename that one?

davidchin: What I meant is you can reuse the same queue ID. If you do that, you can ensure that these actions are executed sequentially to prevent possible race conditions.

PR #2032, opened 2023-06-27
fix(order): DATA-11056 Populate category names for BODL purchase event

bc-rmalyavc:
## What? [DATA-11056](https://bigcommercecloud.atlassian.net/browse/DATA-11056)
- Include categories when loading order by default
- Update `BODL` line item transform logic to handle `line.item.categories`
- Update unit tests

## Why?
1. `BODL` only looked for `categoryNames` in line items. This works for checkout, since the `checkout` request includes `categoryNames` by default. However, for the `order` request, there is no `categoryNames` key in line items, but there is `categories` key instead with different structure. So we need to handle it on BODL end. 2. Another part is that categories were not included for `order` request by default. They were included only for the `affirm-payment-strategy` though. BODL should work with any payment method, and it should include categories for the `purchase` event. So that we need to always include categories when loading order.

## Testing / Proof
- Manually on dev store
- Unit tests
image

[DATA-11056]: https://bigcommercecloud.atlassian.net/browse/DATA-11056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/order/order-params.ts
+++ b/packages/core/src/order/order-params.ts
@@ -1,8 +1,0 @@
-export enum OrderIncludes {
-    DigitalItemsCategories = 'lineItems.digitalItems.categories',
-    PhysicalItemsCategories = 'lineItems.physicalItems.categories',
-}
-
-export default interface OrderParams {
-    include?: OrderIncludes[];
-}
--- a/packages/core/src/order/order-request-sender.ts
+++ b/packages/core/src/order/order-request-sender.ts
@@ -13,7 +13,6 @@ import { OrderTaxProviderUnavailableError } from './errors';
 import InternalOrderRequestBody from './internal-order-request-body';
 import { InternalOrderResponseBody } from './internal-order-responses';
 import Order from './order';
-import OrderParams from './order-params';
 
 export interface SubmitOrderRequestOptions extends RequestOptions {
     headers?: {
@@ -24,10 +23,7 @@ export interface SubmitOrderRequestOptions extends RequestOptions {
 export default class OrderRequestSender {
     constructor(private _requestSender: RequestSender) {}
 
-    loadOrder(
-        orderId: number,
-        { timeout, params }: RequestOptions<OrderParams> = {},
-    ): Promise<Response<Order>> {
+    loadOrder(orderId: number, { timeout }: RequestOptions = {}): Promise<Response<Order>> {
         const url = `/api/storefront/orders/${orderId}`;
         const headers = {
             Accept: ContentType.JsonV1,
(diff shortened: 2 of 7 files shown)


animesh1987:
Hey @bc-rmalyavc can you please remove the dist folder please

packages/core/src/order/order-request-sender.ts
```diff
@@ -37,13 +33,15 @@ export default class OrderRequestSender {
...
+            'lineItems.physicalItems.categories',
             'lineItems.digitalItems.socialMedia',
             'lineItems.digitalItems.options',
+            'lineItems.digitalItems.categories',
         ];
 
         return this._requestSender.get(url, {
             params: {
-                include: joinIncludes([...include, ...((params && params.include) || [])]),
+                include: joinIncludes([...include]),
```
bc-peng, line 44:
```suggestion
                include,
```
It appears that a merging is no longer needed.

bc-rmalyavc: @bc-peng,
I have done so in the first place, but this function actually joins the array into comma separated string, so it is still needed

bc-peng: Oh, I see🤔. The unnecessary merging is this one.

```suggestion
                include: joinIncludes(include),
```

bc-rmalyavc: Good point, thanks 👍 
Updated

bc-peng (approved):
Nice work. 🚢

packages/core/src/order/order-params.ts
```diff
@@ -1,8 +0,0 @@
-export enum OrderIncludes {
```
animesh1987, line 1: Hey @bc-rmalyavc can I ask why this file is removed?

animesh1987:
Hey @bc-rmalyavc there seems to be changes unrelated to bodl here? Is that really required?

packages/core/src/order/order-params.ts
bc-rmalyavc: This file was not used anywhere anymore. Previously it was used in the order-request-sender itself, so that categories could be added via params. Since now categories are included by default, these `OrderIncludes` are not needed anymore

bc-rmalyavc:
@animesh1987,
`Hey @bc-rmalyavc there seems to be changes unrelated to bodl here? Is that really required?`
1. So the main purpose of this PR was to make BODL populate category names for the purchase event. So one change is that BODL now handles another key `categories` in the items payload.
2. Another change is in order-request-sender. Previously categories were not included for the order request by default. So I had to add these default includes into request-sender
3. Finally after I have included categories includes by default, there was an affirm-payment-strategy, that used order-request-sender and added categories includes via params. It was not needed anymore, so I've cleaned it up too.

PR #2034, opened 2023-06-29
feat(checkout): CHECKOUT-7534 Add fees field

bc-mackxu:
## What?
Introduce fees field for checkout 

## Why?
Able to show the fees on the summary detail

## Testing / Proof
CI checks.

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/checkout/checkout.ts
+++ b/packages/core/src/checkout/checkout.ts
@@ -42,7 +42,7 @@ export default interface Checkout {
     updatedTime: string;
     payments?: CheckoutPayment[];
     channelId: number;
-    fees?: Fee[];
+    fees: Fee[];
 }
 
 export interface CheckoutRequestBody {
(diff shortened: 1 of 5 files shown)


bc-peng (approved):

packages/core/src/checkout/checkout.ts
```diff
@@ -41,6 +42,7 @@ export default interface Checkout {
     updatedTime: string;
     payments?: CheckoutPayment[];
     channelId: number;
+    fees?: Fee[];
```
animesh1987, line 45: Is this an optional field?

bc-mackxu: Thanks for reviewing. Confirmed we will always return the field regardless of whether the experiment is enabled or disabled.

animesh1987 (approved):

PR #2042, opened 2023-07-03
fix(order): DATA-11056 Populate category names for BODL purchase event

bc-rmalyavc:
## What? [DATA-11056](https://bigcommercecloud.atlassian.net/browse/DATA-11056)
- Include categories when loading order by default
- Update `BODL` line item transform logic to handle `line.item.categories`
- Update unit tests

## Why?
1. `BODL` only looked for `categoryNames` in line items. This works for checkout, since the `checkout` request includes `categoryNames` by default. However, for the `order` request, there is no `categoryNames` key in line items, but there is `categories` key instead with different structure. So we need to handle it on BODL end. 2. Another part is that categories were not included for `order` request by default. They were included only for the `affirm-payment-strategy` though. BODL should work with any payment method, and it should include categories for the `purchase` event. So that we need to always include categories when loading order.

## Testing / Proof
- Manually on dev store
- Unit tests
image

[DATA-11056]: https://bigcommercecloud.atlassian.net/browse/DATA-11056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/bodl/bodl-emitter-service.ts
+++ b/packages/core/src/bodl/bodl-emitter-service.ts
@@ -1,4 +1,4 @@
-import { LineItemMap } from '../cart';
+import { LineItem, LineItemMap } from '../cart';
 import { CheckoutSelectors, CheckoutStoreSelector } from '../checkout';
 import { MissingDataError, MissingDataErrorType } from '../common/error/errors';
 
@@ -241,6 +241,18 @@ export default class BodlEmitterService implements BodlService {
             ...lineItems.physicalItems,
             ...lineItems.digitalItems,
         ].map((item) => {
+            const getCategoryNames = (lineItem: LineItem): string[] => {
+                if (Array.isArray(lineItem.categoryNames)) {
+                    return lineItem.categoryNames;
+                } else if (Array.isArray(lineItem.categories)) {
+                    return Array.prototype.concat
+                        .apply([], lineItem.categories)
+                        .map(({ name }) => name);
+                }
+
+                return [];
+            };
+
             let itemAttributes;
 
             if (item.options && item.options.length) {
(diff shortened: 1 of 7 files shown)


packages/core/src/bodl/bodl-emitter-service.ts
```diff
@@ -241,6 +241,18 @@ export default class BodlEmitterService implements BodlService {
...
             ...lineItems.digitalItems,
         ].map((item) => {
+            const getCategoryNames = (lineItem: LineItem): string[] => {
+                if (Array.isArray(lineItem.categoryNames)) {
+                    return lineItem.categoryNames;
+                } else if (Array.isArray(lineItem.categories)) {
+                    return Array.prototype.concat
+                        .apply([], lineItem.categories)
+                        .map(({ name }) => name);
+                }
```
animesh1987, line 251: One question why do we need `apply` here? Can we not map over `lineItem.categories` directly?

bc-rmalyavc: `lineItems.categories` is an array of with nested arrays of objects. It looks something like `[[{ name: "Shop All" }], [{ name: "Kitchen" }]]` when we get it from the API. So with this I first concat them into one flat array of category objects and then map the result array to have only names

animesh1987: Interesting since this https://github.com/bigcommerce/checkout-sdk-js/blob/44fa6198236c5492df3c9bffabbc769a474c0fb5/packages/core/src/payment/strategies/affirm/affirm-payment-strategy.ts#L314 just maps over a list. 

So for above case can we use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat or a lodash method.

bc-rmalyavc: @animesh1987,
Updated. I previously intended to use the `flat` method however, I would need to change `compilerOptions` and replace `es6` with `es2019`. I didn't feel confident doing it, so used that old school approach. I didn't think about lodash though. Used it now

animesh1987 (approved):

PR #2045, opened 2023-07-05
fix(checkout): PI-00 fix pending promise for bluesnapdirect APMs iframe

vitalii-koshovyi:
## What?
Fix of the unnecessary browser alert before redirect

## Why?
Before redirect browser alert appears, because there was pending promise in background I believe

## Testing / Proof
Before fix:
image
After fix:
Redirect works without browser alert
 @bigcommerce/team-payments


serhii-tkachenko (approved):
LGTM! 👍

PR #2050, opened 2023-07-07
feat(checkout): CHECKOUT-7538 Introduce extension messenger and command handler

bc-peng:
## What?

- Establish a two-way communication mechanism between checkout-js and an extension.
- The message from an extension should always include the extension ID.
- Define command handling mechanism and its related interfaces.

## Why?
`checkout-js` will be able to carry out the corresponding actions when receving a command from an extension.

## Testing / Proof
- CI.


packages/core/src/extension/errors/invalid-extension-config-error.ts
bc-peng, line 1: Delete as a result of using `ResizableIframeCreator`.

packages/core/src/extension/extension-messenger.ts
bc-peng, line 1: This is supposed to be used on both host and enxtension sides.

packages/core/src/extension/extension-post-event.ts
```diff
@@ -0,0 +1,26 @@
+export type ExtensionCommand = 'RELOAD_CHECKOUT' | 'SHOW_LOADING_INDICATOR';
+
+export type HostCommand = 'HOST_COMMAND';
```
davidchin, line 3: What's a `HostCommand`? If it's intended to be issued by the host to the extension rather than the extension to the host, it'd be good to be explicit about what action you want to perform. Also, to make it clearer who's the recipient of a command, I'd split the `ExtensionPostEvent` into two, one is exclusively for the extension -> host, the other for the host -> extension.

packages/core/src/checkout/checkout-service.ts
```diff
@@ -1399,6 +1404,37 @@ export default class CheckoutService {
...
+     * @param event - A message event that strictly adheres to the ExtensionPostEvent type definition.
+     * @param extensionId - The ID of the extension to which the message will be sent. If no extension ID is provided,
+     * the message will be sent to the host.
+     * @returns A promise that resolves to the current state.
+     */
+    postExtensionMessage(
+        event: ExtensionPostEvent,
+        extensionId?: string,
+    ): Promise<CheckoutSelectors> {
+        const action = this._extensionActionCreator.postExtensionMessage(event, extensionId);
```
davidchin, line 1433: Is this the "broadcast" method that you mentioned - to notify extension state changes? If that's the case, I'm not sure if there's a need to have a method that UCO needs to call. Could we just set up the change-detection subscriber internally and broadcast to all extensions interested in the change?

packages/core/src/extension/extension-messenger.ts
```diff
@@ -0,0 +1,108 @@
...
+
+            if (!iframe?.contentWindow) {
+                throw new ExtensionNotFoundError(
+                    `Unable to post due to no extension rendered for ID: ${extensionId}.`,
+                );
+            }
+
+            this._poster.setTarget(iframe.contentWindow);
+            this._poster.setTargetOrigin(extension.url);
+        } else {
```
davidchin, line 54: As I mentioned earlier, I think it's best to avoid changing the target. In this case, it's not very obvious that omitting the `extensionId` actually means posting to a different origin/target.

packages/core/src/extension/extension-messenger.ts
```diff
@@ -0,0 +1,108 @@
...
+            // Post to host
+            this._poster.setTarget(window.parent);
+            this._poster.setTargetOrigin(window.document.referrer);
```
davidchin, line 57: I'm not sure what you're trying to do here, but `referrer` could be empty depending on the referrer policy.

packages/core/src/extension/extension.ts
```diff
@@ -5,7 +5,6 @@ export interface Extension {
     url: string;
 }
 
-export enum ExtensionRegion {
-    ShippingShippingAddressFormBefore = 'shipping.shippingAddressForm.before',
-    ShippingShippingAddressFormAfter = 'shipping.shippingAddressForm.after',
-}
+export type ExtensionRegion =
```
davidchin, line 8: Is there a particular reason why it's changed from enum to union type?

packages/core/src/checkout/checkout-service.ts
```diff
@@ -1399,6 +1404,37 @@ export default class CheckoutService {
...
 
+    /**
+     * Registers command handlers. Works on both host and extension sides.
+     *
+     * @alpha
+     * @param handlers - a set of handlers for checkout extension commands
+     * @returns A promise that resolves to the current state.
+     */
+    handleExtensionCommand(handlers: ExtensionCommandHandlers): Promise<CheckoutSelectors> {
+        const action = this._extensionActionCreator.handleExtensionCommand(handlers);
```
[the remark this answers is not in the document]
bc-peng: I have changed it. Because I couldn't find another way to setup extensions inside the messenger, I'm requesting it as a parameter here.
```
handleExtensionCommand(extensions: Extension[], handlers: ExtensionCommandHandlers): void
```

To remove handlers, simply use this function again with a new set of desired handlers, which I recommend as an efficient way to set up multiple handlers at once. Do we want to add/remove hander one at a time?

packages/core/src/common/iframe/iframe-event-poster.ts
```diff
@@ -76,4 +76,8 @@ export default class IframeEventPoster<TEvent> {
     setTarget(window: Window) {
         this._targetWindow = window;
     }
+
+    setTargetOrigin(targetOrigin: string) {
```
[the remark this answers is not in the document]
bc-peng: Good idea. This file is untouched now. I have created below inside the messenger.
```
private _posters: { [extensionId: string]: IframeEventPoster<ExtensionOriginEvent> }
```

packages/core/src/extension/extension-action-creator.ts
```diff
@@ -70,4 +76,48 @@ export class ExtensionActionCreator {
...
+    handleExtensionCommand(
+        handlers: ExtensionCommandHandlers,
+    ): ThunkAction<ExtensionAction, InternalCheckoutSelectors> {
+        return (store) =>
+            Observable.create((observer: Observer<ExtensionAction>) => {
+                const extensions = store.getState().extensions.getExtensions();
+
+                if (extensions && extensions.length > 0) {
+                    this._extensionMessenger = new ExtensionMessenger(
+                        new IframeEventPoster<ExtensionPostEvent>('*'),
```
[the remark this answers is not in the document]
bc-peng: Sure, we are able to create a new iframe event poster with all details now.
```
if (!this._posters[extensionId]) {
      this._posters[extensionId] = new IframeEventPoster(extension.url, iframe.contentWindow);
}
```

packages/core/src/extension/extension-post-event.ts
bc-peng: `HostCommand` was created for client-side use.
Since we are spliting them now, I have removed it and renamed `ExtensionPostEvent` as `ExtensionOriginEvent`.

packages/core/src/checkout/checkout-service.ts
bc-peng: Removed it for now.
Will circle back to `postMessage` once we get to the "broadcast" ticket.

packages/core/src/extension/extension-messenger.ts
bc-peng: Sure, now the post method is updated as below:
```
post(extensionId: string, event: ExtensionOriginEvent): void
```
`target` and `origin` are set at a poster's instantiation

packages/core/src/extension/extension-messenger.ts
bc-peng: This part is removed and will be deliverd in Animesh's PR.

packages/core/src/checkout/checkout-service.spec.ts
```diff
@@ -1488,11 +1488,29 @@ describe('CheckoutService', () => {
...
     });
+
+    describe('#handleExtensionCommand()', () => {
+        it('handles extension command', async () => {
+            const checkoutServiceMock: any = checkoutService;
+            const listenMock = jest.fn();
+
+            checkoutServiceMock._extensionMessenger = {
+                listen: listenMock,
+            };
```
bc-peng, line 1506: A workaround to test `extensionMessenger` as a private property. 😹

packages/core/src/extension/extension.ts
bc-peng: We are experiencing an issue that checkout-js does not recognize SDK's enum value (enum value undefined) during compile time. Converting it to union types is a workaround for now.
cc @animesh1987

packages/core/src/checkout/checkout-service.ts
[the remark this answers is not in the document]
davidchin: > To remove handlers, simply use this function again with a new set of desired handlers

I can't see `ExtensionMessenger` making any `window.removeEventListener` call? 🤔 We need a way to remove command handlers to avoid potential memory leaks. For example, we could end up with a leak if a component is unmounted and it still has a function registered as a command handler.

packages/core/src/extension/extension.ts
animesh1987: Actually I was trying to find a solution for it, can you try using 
```
export const enum ExtensionRegion {
    ShippingShippingAddressFormBefore = 'shipping.shippingAddressForm.before',
    ShippingShippingAddressFormAfter = 'shipping.shippingAddressForm.after',
}
```

I believe this should be able to resolve the issue.

davidchin: Hrm interesting, perhaps it's exported as a type rather than a value. Perhaps we can try exporting it explicitly in bundle index file rather than relying on API extractor?

packages/core/src/extension/host-origin-event.ts
```diff
@@ -0,0 +1,4 @@
+// TODO - will finalize in the following PR
```
davidchin, line 1: Maybe just remove it for now because it's not used. We can review how it gets used in the following PR?

packages/core/src/checkout/checkout-service.spec.ts
davidchin: Why don't you just inject an instance of `ExtensionMessenger` when constructing `checkoutService` in the `beforeEach`?

packages/core/src/extension/extension-messenger.ts
```diff
@@ -0,0 +1,101 @@
...
+import { bindDecorator as bind } from '../common/utility';
+
+import { ExtensionNotFoundError } from './errors';
+import { Extension } from './extension';
+import { ExtensionCommandHandlers } from './extension-command-handler';
+import { ExtensionOriginEvent } from './extension-origin-event';
+import { HostOriginEvent } from './host-origin-event';
+
+export class ExtensionMessenger {
+    _isListening = false;
```
davidchin, line 12: Missing `private` keyword?

packages/core/src/extension/host-origin-event.ts
bc-peng: It's used in the post method of `extensionMessenger` as that method posts messages to an extension.
```
post(extensionId: string, event: HostOriginEvent): void {
```

packages/core/src/extension/extension-messenger.ts
bc-peng: 😅 Thanks for pointing this out.

packages/core/src/extension/extension-messenger.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export class ExtensionMessenger {
+    private _isListening = false;
+    private _extensions: Extension[] | undefined;
+    private _extensionCommandHandlers: ExtensionCommandHandlers = {};
+    private _posters: { [extensionId: string]: IframeEventPoster<HostOriginEvent> } = {};
+
+    listen(extensions: Extension[], extensionCommandHandlers: ExtensionCommandHandlers): void {
+        this._extensions = extensions;
+        this._extensionCommandHandlers = extensionCommandHandlers;
```
davidchin, line 19: It looks like a non-TS client can listen to / post any arbitrary command not defined in `ExtensionCommand`? It would be good to restrict that so that only known commands can be registered.

packages/core/src/extension/extension-messenger.ts
```diff
@@ -0,0 +1,101 @@
...
+
+        if (this._isListening) {
+            return;
+        }
+
+        this._isListening = true;
+
+        window.addEventListener('message', this._handleMessage);
```
davidchin, line 27: Is there a particular reason why you chose to listen to messages directly rather than using `IframeEventListener`?

packages/core/src/extension/extension.ts
animesh1987: @davidchin yes that's correct its an issue with Typescript as far as I know. Hence I suggested to try and add a `const` keyword above and that might solve the issue.

bc-peng: `export const enum ExtensionRegion` is working. I'll apply it all enums.

packages/core/src/checkout/checkout-service.spec.ts
bc-peng: I was trying to avoid modifying the constructor to inject an instance of extensionMessenger as the constructor mostly receives actionCreators instead. Now I have injected extensionMessenger, this workaround is no longer needed.

packages/core/src/extension/extension-messenger.ts
bc-peng: Sure, I've implemented a filter to only handle enum ExtensionCommand In the extensionMessenger's `listen()` method.

packages/core/src/extension/extension-messenger.ts
bc-peng: I thought it might be too costly to have a separate `IframeEventListener` and command handlers for each extension at beginning. Now I have replaced it with the existing `IframeEventListener`.

packages/core/src/checkout/checkout-service.ts
```diff
@@ -1399,6 +1405,30 @@ export default class CheckoutService {
...
     }
 
+    /**
+     * Registers an extension's command handlers.
+     *
+     * @alpha
+     * @param extensionId - The ID of an extension sending the commands.
+     * @param handlers - A set of handlers for the extension commands.
+     */
+    addExtensionCommandHandlers(extensionId: string, handlers: ExtensionCommandHandlers): void {
```
davidchin, line 1415: I think it's better to return a function that could unsubscribe the handler, similar to the existing `subscribe` method. e.g.:

```ts
listenExtensionCommand(extensionId: string, commandType: ExtensionCommand, commandHandler, ExtensionCommandHandler): () => void
```

This makes it easier to unsubscribe a handler, as calling the return value can specifically remove such handler. Otherwise, clients need to hold a reference to the handler and therefore can't use inline anonymous function as the handler.

packages/core/src/extension/extension-messenger.ts
```diff
@@ -0,0 +1,96 @@
...
+import { HostOriginEvent } from './host-origin-event';
+
+export class ExtensionMessenger {
+    private _extensions: Extension[] | undefined;
+    private _listeners: { [extensionId: string]: IframeEventListener<ExtensionOriginEventMap> } =
+        {};
+    private _posters: { [extensionId: string]: IframeEventPoster<HostOriginEvent> } = {};
+
+    listen(
+        extensions: Extension[],
```
davidchin, line 16: I'm thinking you can just inject `ReadableCheckoutStore`  as a constructor parameter of `ExtensionMessenger`. That way you don't need to pass `extensions`. You can just call the following to get the available extensions.

```ts
const { data: { getExtensions } } = this.store.getState();
const extensions = getExtensions();
```

packages/core/src/extension/extension-messenger.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        const listener = this._listeners[extensionId];
+
+        listener.listen();
+
+        Object.entries(extensionCommandHandlers).forEach(([command, handler]) => {
+            const extensionCommand = stringToExtensionCommand(command);
+
+            if (extensionCommand && handler) {
+                listener.addListener(extensionCommand, handler);
```
davidchin, line 36: If we take one handler instead and return unsubscribe method, it'd look something like this:

```ts
listener.addListener(extensionCommand, handler);

return () => listener.removeListener(extensionCommand, handler);
```

packages/core/src/checkout/checkout-service.ts
bc-peng: I have updated it as:
```
listenExtensionCommand(
        extensionId: string,
        command: ExtensionCommand,
        handler: ExtensionCommandHandler,
    ): () => void
```
To entirely remove the command listening, I also added a method in the `checkoutService` to stop listening for an extension.
```
stopListenExtensionCommand(extensionId: string): void
```

packages/core/src/extension/extension-messenger.ts
bc-peng: Sure. Now the constructor is:
```
constructor(
        private _store: CheckoutStore,
        private _listeners: { [extensionId: string]: IframeEventListener<ExtensionOriginEventMap> },
        private _posters: { [extensionId: string]: IframeEventPoster<HostOriginEvent> },
    ) {}
```

packages/core/src/extension/extension-messenger.ts
bc-peng: I did follow this and add a unit test to cover it. Thanks.

packages/core/src/checkout/checkout-service.ts
```diff
@@ -1399,6 +1406,33 @@ export default class CheckoutService {
...
+        return this._extensionMessenger.listen(extensionId, command, handler);
+    }
+
+    /**
+     * Stops handling commands for an extension.
+     *
+     * @alpha
+     * @param extensionId - The ID of the extension that originally sent the commands.
+     */
+    stopListenExtensionCommand(extensionId: string): void {
```
davidchin, line 1432: Maybe we could skip this method for now unless we really need it. `IframeEventListener#stopListen` doesn't actually remove the references to the handlers previously registered. So we need to do a bit of work to actually do what we want.

packages/core/src/extension/extension-messenger.ts
```diff
@@ -0,0 +1,111 @@
...
+import { Extension } from './extension';
+import { ExtensionCommandHandler } from './extension-command-handler';
+import { ExtensionCommand, ExtensionOriginEventMap } from './extension-origin-event';
+import { HostOriginEvent } from './host-origin-event';
+
+export class ExtensionMessenger {
+    private _extensions: Extension[] | undefined;
+
+    constructor(
+        private _store: CheckoutStore,
```
davidchin, line 15: Preferably use the read-only version, i.e.: `ReadableCheckoutStore`

bc-peng: Yes, `ReadableCheckoutStore` is used in ways I was not aware of before.

packages/core/src/checkout/checkout-service.ts
bc-peng: Removed. We can circle back to it once we have the use case.

davidchin (approved):
Fantastic work @bc-peng, LGTM! 👏

PR #2052, opened 2023-07-10
fix(payment): PAYPAL-2719 fix amazon pay button for buy now flow

bc-nick:
## What?

Fix Amazon Pay button for buy now flow on PDP

## Why?

To keep up-to-date data during the quota creation process

## Testing / Proof

https://github.com/bigcommerce/checkout-sdk-js/assets/99336044/999b12ee-4570-4b08-b912-0c5f2134b4c4

### DIFF
--- a/packages/core/src/checkout-buttons/strategies/amazon-pay-v2/amazon-pay-v2-button-strategy.ts
+++ b/packages/core/src/checkout-buttons/strategies/amazon-pay-v2/amazon-pay-v2-button-strategy.ts
@@ -1,5 +1,4 @@
 import { CartRequestSender } from '../../../cart';
-import BuyNowCartRequestBody from '../../../cart/buy-now-cart-request-body';
 import { BuyNowCartCreationError } from '../../../cart/errors';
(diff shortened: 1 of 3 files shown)


bc-peng:
Can I know if you had the chance to test the recent changes on the cart page?
The reason I ask is that this code is shared there, and we had a previous incident where the cart page was breaking without our knowledge. It would be great if we could double-check to ensure it is working as expected.

bc-peng:
Can I know if you had the chance to test the changes on the cart page?
The reason I ask is that this code is shared there, and we had a previous incident where the cart page was breaking without our knowledge. It would be great if we could double-check to ensure it is working as expected.

bc-nick:
> Can I know if you had the chance to test the changes on the cart page? The reason I ask is that this code is shared there, and we had a previous incident where the cart page was breaking without our knowledge. It would be great if we could double-check to ensure it is working as expected.

@bc-peng these changes only for Buy Now Flow on PDP and don't affect the general logic of the button elsewhere. i tested on the integration store and ensure everything works as expected

bc-peng (approved):

PR #2056, opened 2023-07-13
feat(checkout): CHECKOUT-7595 Add client extension service

animesh1987:
## What?
Add client extension service

## Why?
Add a client extension service that is shipped out as a separate bundle that extensions can import and use for messaging to and from with checkout.

## Testing / Proof
- CI

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/extension/extension-client.ts
+++ b/packages/core/src/extension/extension-client.ts
@@ -1,5 +1,5 @@
 export interface InitializeExtensionServiceOptions {
-    extensionId: number;
+    extensionId: string;
     parentOrigin: string;
 }
 
@@ -28,7 +28,7 @@ export const enum ExtensionPostEventType {
 
 export interface BaseEventPayload {
     payload: {
-        extensionId?: number;
+        extensionId?: string;
     };
 }
 
--- a/packages/core/src/extension/extension-client.ts
+++ b/packages/core/src/extension/extension-client.ts
@@ -3,7 +3,7 @@ export interface InitializeExtensionServiceOptions {
     parentOrigin: string;
 }
 
-export const enum ExtensionListenEventType {
+export enum ExtensionListenEventType {
     BroadcastCart = 'BROADCAST_CART',
     ShippingCountryChange = 'SHIPPING_COUNTRY_CHANGE',
 }
@@ -20,7 +20,7 @@ export interface ExtensionListenEventMap {
     };
 }
 
-export const enum ExtensionPostEventType {
+export enum ExtensionPostEventType {
     FRAME_LOADED = 'FRAME_LOADED',
     RELOAD_CHECKOUT = 'RELOAD_CHECKOUT',
     SHOW_LOADING_INDICATOR = 'SHOW_LOADING_INDICATOR',
--- a/packages/core/src/extension/extension-client.ts
+++ b/packages/core/src/extension/extension-client.ts
@@ -4,27 +4,27 @@ export interface InitializeExtensionServiceOptions {
 }
 
 export enum ExtensionListenEventType {
-    BroadcastCart = 'BROADCAST_CART',
+    CheckoutLoaded = 'CHECKOUT_LOADED',
     ShippingCountryChange = 'SHIPPING_COUNTRY_CHANGE',
 }
 
-export interface BroadcastCartEvent {
-    type: ExtensionListenEventType.BroadcastCart;
+export interface CheckoutLoadedEvent {
+    type: ExtensionListenEventType.CheckoutLoaded;
 }
 
 export interface ShippingCountryChangeEvent {
-    type: ExtensionListenEventType.BroadcastCart;
+    type: ExtensionListenEventType.CheckoutLoaded;
     payload: {
         countryCode: string;
     };
 }
 
 export interface ExtensionListenEventMap {
-    [ExtensionListenEventType.BroadcastCart]: BroadcastCartEvent;
+    [ExtensionListenEventType.CheckoutLoaded]: CheckoutLoadedEvent;
     [ExtensionListenEventType.ShippingCountryChange]: ShippingCountryChangeEvent;
 }
 
-export enum ExtensionPostEventType {
+export enum ExtensionCommandType {
     FRAME_LOADED = 'FRAME_LOADED',
     RELOAD_CHECKOUT = 'RELOAD_CHECKOUT',
     SHOW_LOADING_INDICATOR = 'SHOW_LOADING_INDICATOR',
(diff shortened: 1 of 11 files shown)


packages/core/src/extension/extension-client.ts
bc-peng, line 1: Commands orginated from the extension side are defined [here](https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/extension/extension-origin-event.ts#L1) as part of the host side validation process.

Commands originated from the host side are new things. I think we only need a file for the only two commands now, `BROADCAST_CART` and `SHIPPING_COUNTRY_CHANGE`.

packages/core/src/extension/extension-service.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    post(event: ExtensionPostEvent): void {
+        if (!this._extensionId) {
+            return;
+        }
+
+        this._eventPoster.setTarget(window.parent);
+
+        if (!Object.values(ExtensionPostEventType).includes(event.type)) {
```
bc-peng, line 37: Once it is loaded as `const enum`, this line may need to be updated.

packages/core/src/extension/extension-service.ts
```diff
@@ -0,0 +1,56 @@
...
+
+        const payload = {
+            ...event.payload,
+            extensionId: this._extensionId,
+        };
+
+        this._eventPoster.post({ ...event, payload });
+    }
+
+    addListener(eventType: ExtensionListenEventType, callback: () => void = noop): void {
```
bc-peng, line 49: We can return `() => void` here as DC suggested to me before to remove the listener. 
I did it here: https://github.com/bigcommerce/checkout-sdk-js/blob/5c5e2e5ad1a6b49912febcfbfcd39e5d684399c0/packages/core/src/extension/extension-messenger.ts#L40.

packages/core/src/extension/extension-client.ts
animesh1987: Discussed offline that we can keep a separate type file for clients.

packages/core/src/extension/extension-service.ts
```diff
@@ -0,0 +1,58 @@
...
+
+        this._eventListener.listen();
+    }
+
+    post(event: ExtensionPostEvent): void {
+        if (!this._extensionId) {
+            return;
+        }
+
+        this._eventPoster.setTarget(window.parent);
```
bc-peng, line 35: We may move this line to `constructor` so that the window does not need to be set every time.

packages/core/src/extension/extension-client.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export interface ExtensionListenEventMap {
+    [ExtensionListenEventType.BroadcastCart]: {
+        type: ExtensionListenEventType.BroadcastCart;
+    };
+    [ExtensionListenEventType.ShippingCountryChange]: {
+        type: ExtensionListenEventType.ShippingCountryChange;
+        payload: {
+            countryCode: string;
+        };
```
bc-peng, line 19: Maybe we can give it a type name, because the host side will use it eventually.

bc-peng (approved):

packages/core/src/extension/extension-service.ts
```diff
@@ -0,0 +1,58 @@
...
+    private _extensionId?: string;
+
+    constructor(
+        private _eventListener: IframeEventListener<ExtensionListenEventMap>,
+        private _eventPoster: IframeEventPoster<ExtensionPostEvent>,
+    ) {
+        this._eventPoster.setTarget(window.parent);
+    }
+
+    initialize(extensionId: string) {
```
davidchin, line 22:
```suggestion
    initialize(extensionId: string): void {
```

packages/core/src/extension/extension-service.ts
```diff
@@ -0,0 +1,58 @@
...
+
+        const payload = {
+            ...event.payload,
+            extensionId: this._extensionId,
+        };
+
+        this._eventPoster.post({ ...event, payload });
+    }
+
+    addListener(eventType: ExtensionListenEventType, callback: () => void = noop): () => void {
```
davidchin, line 49: It seems we need to standardise the naming. There is `Command`, `ListenEvent`, `PostEvent`, `OriginEvent`, `BroadcastEvent` etc... I'm not exactly sure what they are and whether they need to be categorised differently.

packages/core/src/extension/extension-service.ts
```diff
@@ -0,0 +1,58 @@
...
+        if (!Object.values(ExtensionListenEventType).includes(eventType)) {
+            return noop;
```
davidchin, line 51: Should it raise an exception instead of failing silently?

animesh1987: This is on the client side so I am not sure if we throw an exception? Probably makes sense to throw an exception to improve the developer experience. Will change it accordingly :D

packages/core/src/extension/extension-service.ts
```diff
@@ -0,0 +1,58 @@
...
+        if (!extensionId) {
+            throw new Error('Extension Id not found.');
+        }
+
+        this._extensionId = extensionId;
+
+        this._eventListener.listen();
+    }
+
+    post(event: ExtensionPostCommand): void {
```
bc-peng, line 32:
```suggestion
    post(event: ExtensionCommand): void {
```
Because the client side only sends command, `post` in the name seems redundant.

packages/core/src/extension/extension-client.ts
```diff
@@ -0,0 +1,54 @@
...
+    CheckoutLoaded = 'CHECKOUT_LOADED',
+    ShippingCountryChange = 'SHIPPING_COUNTRY_CHANGE',
+}
+
+export interface CheckoutLoadedEvent {
+    type: ExtensionListenEventType.CheckoutLoaded;
+}
+
+export interface ShippingCountryChangeEvent {
+    type: ExtensionListenEventType.CheckoutLoaded;
```
bc-peng, line 16:
```suggestion
    type: ExtensionListenEventType.ShippingCountryChange;
```

packages/core/src/extension/extension-client.ts
```diff
@@ -0,0 +1,54 @@
...
+export interface ExtensionListenEventMap {
+    [ExtensionListenEventType.CheckoutLoaded]: CheckoutLoadedEvent;
+    [ExtensionListenEventType.ShippingCountryChange]: ShippingCountryChangeEvent;
+}
+
+export enum ExtensionCommandType {
+    FRAME_LOADED = 'FRAME_LOADED',
+    RELOAD_CHECKOUT = 'RELOAD_CHECKOUT',
+    SHOW_LOADING_INDICATOR = 'SHOW_LOADING_INDICATOR',
+}
```
bc-peng, line 31: Do we still have `export const enum ExtensionCommandType` for `checkout-js`?

packages/core/src/extension/extension-client.ts
```diff
@@ -0,0 +1,54 @@
+export interface InitializeExtensionServiceOptions {
+    extensionId: string;
+    parentOrigin: string;
+}
+
+export enum ExtensionListenEventType {
```
davidchin, line 6:
```suggestion
export enum ExtensionEventType {
```

packages/core/src/extension/extension-client.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export interface ShippingCountryChangeEvent {
+    type: ExtensionListenEventType.CheckoutLoaded;
+    payload: {
+        countryCode: string;
+    };
+}
+
+export interface ExtensionListenEventMap {
```
davidchin, line 22:
```suggestion
export interface ExtensionEventMap {
```

packages/core/src/extension/extension-client.ts
animesh1987: I think so at least for now, unless we converge one section 🤔 Can combine that as a last stage fast follow I think.

packages/core/src/extension/extension-client.ts
```diff
@@ -0,0 +1,54 @@
...
+    CheckoutLoaded = 'CHECKOUT_LOADED',
+    ShippingCountryChange = 'SHIPPING_COUNTRY_CHANGE',
+}
+
+export interface CheckoutLoadedEvent {
+    type: ExtensionEventType.CheckoutLoaded;
+}
+
+export interface ShippingCountryChangeEvent {
+    type: ExtensionEventType.CheckoutLoaded;
```
bc-peng, line 16:
```suggestion
    type: ExtensionEventType.ShippingCountryChange;
```

davidchin (approved):
Nice work @animesh1987! 👏

bc-peng (approved):

PR #2064, opened 2023-07-18
feat(payment): PAYPAL-000 added dispatch method to payments integration service

serhii-tkachenko:
## What?
Added dispatch method to payments integration service

## Why?
To have an ability to call payment provider specific actions in different integration packages

## Testing / Proof
Unit tests

### DIFF
--- a/packages/core/src/payment-integration/default-payment-integration-service.ts
+++ b/packages/core/src/payment-integration/default-payment-integration-service.ts
@@ -1,3 +1,5 @@
+import { Action } from '@bigcommerce/data-store';
+
 import {
     BillingAddressRequestBody,
     BuyNowCartRequestBody,
@@ -47,6 +49,12 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
         private _spamProtectionActionCreator: SpamProtectionActionCreator,
     ) {
         this._storeProjection = this._storeProjectionFactory.create(this._store);
+    }
+
+    async dispatch(action: Action) {
+        await this._store.dispatch(action);
+
+        return this._storeProjection.getState();
     }
 
     createHostedForm(host: string, options: HostedFormOptions): HostedForm {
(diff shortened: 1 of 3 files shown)


packages/core/src/payment-integration/default-payment-integration-service.ts
```diff
@@ -49,6 +51,12 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
         this._storeProjection = this._storeProjectionFactory.create(this._store);
     }
 
+    async dispatch(action: Action) {
```
animesh1987, line 54: This feels a bit odd, don't think we want to expose a generic dispatch method. Is there a way you can create a specific action needed? 

Another question is does core package need to know about this action at all? Can this be achieved by an API call in paypal package?

serhii-tkachenko: I think it might be great to have such method to reduce amount of methods I payment integration package specific for each payment provider.

We can't go with API call in paypal package, because we need to update checkout state with data, that we get from paypal.

animesh1987: I think the point of the service is to avoid such generic methods, may I ask what is the use case we are trying to solve. Which bit in state needs to be updated?

serhii-tkachenko: We need to have an ability to update customers and instruments states with PayPal Connect profile data

serhii-tkachenko: I am thinking on creating 2 different methods:
1) updateCustomer method, that will be responsible for updating customer state;
2) updateInstruments method, that will be responsible for updating vaulted instruments state;

In this case we will have 2 generic but meaningful methods and anyone will understand how to use them. What do you think @animesh1987?

vitalii-koshovyi: @serhii-tkachenko yeah, we should avoid just exposing dispatch to the payment-integrations-service. I believe `payment-integrations-service` it's like an inner SDK, so adding 2 different methods seems like a better idea to me.

serhii-tkachenko: Ok, I will close the PR because we should find another solution for that

PR #2066, opened 2023-07-19
feat(orders): ORDERS-5715 add OrderFee interface, add fees field to Order interface

matt-evangelidis:
## What?
This PR adds an `OrderFee` interface and a `fees` field to the `Order` interface in `order.ts`.

## Why?
This adds some interfaces for fees specific to the Orders domain to provide separation between Checkout and Orders, as each domain's definitions of fees are likely to drift further apart over time.

## Testing / Proof
N/A

@bigcommerce/team-checkout @bigcommerce/team-orders

### DIFF
--- a/packages/core/src/order/order.ts
+++ b/packages/core/src/order/order.ts
@@ -34,6 +34,15 @@ export default interface Order {
     taxes: Tax[];
     taxTotal: number;
     channelId: number;
+    fees: OrderFee[];
+}
+
+export interface OrderFee {
+    id: number;
+    type: string;
+    customerDisplayName: string;
+    cost: number;
+    source: string;
 }
 
 export type OrderPayments = Array<GatewayOrderPayment | GiftCertificateOrderPayment>;
--- a/packages/core/src/order/orders.mock.ts
+++ b/packages/core/src/order/orders.mock.ts
@@ -52,6 +52,7 @@ export function getOrder(): Order {
         ],
         taxTotal: 3,
         channelId: 1,
+        fees: [],
     };
 }
 


packages/core/src/order/order.ts
```diff
@@ -34,6 +34,15 @@ export default interface Order {
     taxes: Tax[];
     taxTotal: number;
     channelId: number;
+    fees?: OrderFee[];
```
animesh1987, line 37: Is this supposed to an optional?

animesh1987:
@matt-evangelidis  For commit validation the scope needs to be `order` compared to orders 😄

packages/core/src/order/order.ts
matt-evangelidis: I marked it as optional because fees are only included by default when coming from `api/storefront/orders/` if `PROJECT-5577.support_order_level_fee` is on.

I suppose that the field itself should always be in the payload, so it should be fine to have as mandatory.

animesh1987: I think that makes sense to mark it mandatory.

matt-evangelidis: Marked as mandatory now.

animesh1987 (approved):
Nicely done, looks good. Can you please check if checkout js build works fine with this change before merging.

matt-evangelidis:
> Nicely done, looks good. Can you please check if checkout js build works fine with this change before merging.

image

matt-evangelidis:
@animesh1987 I don't have write access to this repo. Are you alright to merge?

PR #2069, opened 2023-07-20
refactor(checkout): CHECKOUT-7624 Update extension-related naming

bc-peng:
## What?
Revise the names of extension-related types, interfaces, methods.

## Why?
Align better with our extension context and ensure consistency.

## Testing / Proof
- CI checks.


packages/core/src/extension/extension-iframe.ts
```diff
@@ -18,6 +18,14 @@ export class ExtensionIframe {
...
 
     async attach(): Promise<void> {
+        if (
+            document
+                .querySelector(`[data-extension-id="${this._extension.id}"]`)
+                ?.querySelector('iframe')
+        ) {
+            return;
+        }
+
```
bc-peng, line 28: During the UI implementation, I noticed that sometimes the iframe gets reattached when the React state updates. Therefore, adding a check here to ensure that only one iframe attached for an extension.

animesh1987 (approved):
Nice work 👏

PR #1967, opened 2023-05-02
feat(payment): INT-7676 GooglePay: Add payment strategy

mauricio-sg:
## What? [INT-7676](https://bigcommercecloud.atlassian.net/browse/INT-7676)

1. Add Google Pay payment strategy for Worldpay Access
2. Refactor existing logic for:
    - Authorize Net
    - Checkout Com
    - Cybersource / BNZ
    - Orbital
    - Stripe v3 / Stripe UPE

```mermaid
classDiagram
    PaymentStrategy <|.. GooglePayPaymentStrategy
    GooglePayPaymentStrategy *-- GooglePayPaymentProcessor
    GooglePayPaymentProcessor *-- GooglePayGateway
    GooglePayGateway <|-- GooglePayAuthorizeNetGateway
    GooglePayGateway <|-- GooglePayCheckoutComGateway
    GooglePayAdditionalActionProcessable <|.. GooglePayCheckoutComGateway
    GooglePayGateway <|-- GooglePayCybersourceGateway
    GooglePayGateway <|-- GooglePayOrbitalGateway
    GooglePayGateway <|-- GooglePayStripeGateway
    GooglePayGateway <|-- GooglePayWorldpayAccessGateway

    class PaymentStrategy {
        <<interface>>
        +initialize() Promise~void~
        +execute() Promise~void~
        +finalize() Promise~void~
        +deinitialize() Promise~void~
    }

    class GooglePayPaymentStrategy {
        -_paymentButton : HTMLElement | undefined
        -_clickListener : #40;#40;event: MouseEvent#41; => unknown#41; | undefined
        -_methodId : GooglePayKey | undefined
        -_paymentIntegrationService : PaymentIntegrationService
        -_googlePayPaymentProcessor : GooglePayPaymentProcessor

        -_addPaymentButton(walletButton: string, callbacks: Object): void
        -_handleClick(callbacks: Object): (event: MouseEvent) => unknown
        -_interactWithPaymentSheet(): Promise~void~
        -_getMethodId(): GooglePayKey
    }
    class GooglePayPaymentProcessor {
        -_paymentsClient : GooglePaymentsClient | undefined
        -_baseRequest : GooglePayGatewayBaseRequest
        -_baseCardPaymentMethod : GooglePayBaseCardPaymentMethod | undefined
        -_cardPaymentMethod : GooglePayCardPaymentMethod | undefined
        -_paymentDataRequest : GooglePayPaymentDataRequest | undefined
[...]

PR #2072, opened 2023-07-20
fix(payment): PAYPAL-2634 fix GP button on the checkout page for unsupported browser

bc-nick:
## What?
...

## Why?
...

## Testing / Proof
...


vitalii-koshovyi (commented):
I believe this PR should be merged after this one https://github.com/bigcommerce/checkout-sdk-js/pull/1967

packages/core/src/payment/strategies/googlepay/googlepay-payment-processor.ts
```diff
@@ -179,9 +182,13 @@ export default class GooglePayPaymentProcessor {
...
                         })
-                        .then((response) => {
-                            if (response.result) {
-                                return;
+                        .then(({ result: isReadyToPay }) => {
+                            if (isReadyToPay) {
+                                return isReadyToPay;
+                            }
+
+                            if (this._isInAppUnsupportedBrowser('instagram')) {
```
vitalii-koshovyi, line 190: Does problem occurs only with `instagram` user agent? I think would be better to do it for all in-app browsers

bc-nick: @vitalii-koshovyi I didn't check it for other in-app browsers. But i will check what happens if we use facebook for example and if we have the same issue i will make some changes and let you know. Thanks for the point!

animesh1987:
Also will be good for @mauricio-sg to have a look for this PRs context https://github.com/bigcommerce/checkout-sdk-js/pull/1967.

vitalii-koshovyi:
@animesh1987 I believe we need to wait for https://github.com/bigcommerce/checkout-sdk-js/pull/1967 merge and after that we can merge this PR

animesh1987 (approved):

PR #2074, opened 2023-07-21
feat(payment): PAYPAL-2773 added an ability to create shipping strategy in integration package

serhii-tkachenko:
## What?
Automatically register shipping strategies exported from payment integration packages and surface their initialisation options through the `CheckoutService` interface.

For example, if the entry point of the `braintreeacceleratedcheckout` package (i.e.: `index.ts`), exports a shipping strategy factory function (i.e.: `createBraintreeAcceleratedCheckoutShippingStrategy`), it will automatically be registered by an instance of `Registry<ShippingStrategy>` inside the `core` package. The factory will be registered against the `id` of a payment method as specified in the `braintree` package.

```ts
export default toResolvableModule(
    createBraintreeAcceleratedCheckoutShippingStrategy,
    [{ id: 'braintreeacceleratedcheckout' }]
);
```

## Why?
So that shipping strategies can be created in integration packages.

## Testing / Proof
Unit tests
Manual tests

Screenshots of generated files:
Screenshot 2023-07-21 at 16 43 46
Screenshot 2023-07-21 at 16 43 56

Lets imagine that this is a BraintreeAcceleratedCheckoutShipping component presented on the screen:

https://github.com/bigcommerce/checkout-sdk-js/assets/25133454/26903c0b-ce17-4328-a02d-9de544943c75

### DIFF
--- a/packages/braintree-integration/src/braintree-accelerated-checkout/braintree-accelerated-checkout-shipping-strategy.ts
+++ b/packages/braintree-integration/src/braintree-accelerated-checkout/braintree-accelerated-checkout-shipping-strategy.ts
@@ -1,0 +1,58 @@
+import {
+    AddressRequestBody,
+    InvalidArgumentError,
+    PaymentIntegrationService,
+    ShippingInitializeOptions,
+    ShippingRequestOptions,
+    ShippingStrategy,
+} from '@bigcommerce/checkout-sdk/payment-integration-api';
+
+// TODO: this strategy will be updated in one of the future PR
+export default class BraintreeAcceleratedCheckoutShippingStrategy implements ShippingStrategy {
+    constructor(private paymentIntegrationService: PaymentIntegrationService) {}
+
+    initialize(options: ShippingInitializeOptions): Promise<void> {
+        const { methodId } = options;
+
+        if (!methodId) {
+            throw new InvalidArgumentError(
+                'Unable to initialize payment because "options.methodId" argument is not provided.',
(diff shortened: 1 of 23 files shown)


animesh1987:
Hey @serhii-tkachenko thanks for raising this, this is something we can leave in core package for now in my opinion the strategy that is. We don't want to be leveraging payment-integration-api package to support shipping side of things as well since the name itself suggests payment integration API and not shipping. 

This is something if we do in future will have to add in mechanism to support shipping integrations and not use the payment integration framework to support the usecases.

serhii-tkachenko:
> Hey @serhii-tkachenko thanks for raising this, this is something we can leave in core package for now in my opinion the strategy that is. We don't want to be leveraging payment-integration-api package to support shipping side of things as well since the name itself suggests payment integration API and not shipping.
> 
> This is something if we do in future will have to add in mechanism to support shipping integrations and not use the payment integration framework to support the usecases.

Should we build shipping-integration-api and use it in our packages?

animesh1987:
Think that can increase the scope of the project in my opinion, plus this is something we in checkout will want to take ownership of. Since we don't have plans for this at the moment I would say you can add the strategy in core package?

serhii-tkachenko:
> Think that can increase the scope of the project in my opinion, plus this is something we in checkout will want to take ownership of. Since we don't have plans for this at the moment I would say you can add the strategy in core package?

Yes, we can.

PR #2075, opened 2023-07-21
feat(payment): PAYPAL-2731 added PaymentProviderCustomer state

serhii-tkachenko:
## What?
Added PaymentProviderCustomer state

## Why?
To be able to save provider specific data and use those data in separate strategies.

## Testing / Proof
Unit tests
Manual tests

// Info: the screenshots a bit out dated after the review, but it works almost the same (naming changed)
Here are some screenshots/proofs that it works (hardcoded only for example):
Screenshot 2023-07-24 at 11 51 26
Screenshot 2023-07-24 at 11 51 44

### DIFF
--- a/packages/core/src/payment-integration/default-payment-integration-service.ts
+++ b/packages/core/src/payment-integration/default-payment-integration-service.ts
@@ -20,6 +20,7 @@ import { DataStoreProjection } from '../common/data-store';
 import { CustomerActionCreator, CustomerCredentials } from '../customer';
 import { HostedFormFactory } from '../hosted-form';
 import { OrderActionCreator } from '../order';
+import { PaymentProviderCustomerActionCreator } from '../payment-provider-customer';
 import PaymentActionCreator from '../payment/payment-action-creator';
 import PaymentMethodActionCreator from '../payment/payment-method-action-creator';
 import { ConsignmentActionCreator } from '../shipping';
@@ -45,6 +46,7 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
         private _cartRequestSender: CartRequestSender,
         private _storeCreditActionCreator: StoreCreditActionCreator,
         private _spamProtectionActionCreator: SpamProtectionActionCreator,
+        private _paymentProviderCustomerActionCreator: PaymentProviderCustomerActionCreator,
     ) {
         this._storeProjection = this._storeProjectionFactory.create(this._store);
     }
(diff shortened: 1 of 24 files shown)


packages/payment-integrations-test-utils/src/test-utils/payment-integration-service.mock.ts
```diff
@@ -73,6 +74,7 @@ const PaymentIntegrationServiceMock = jest
             selectShippingOption,
             applyStoreCredit,
             verifyCheckoutSpamProtection,
+            updateProviderCustomerData,
```
animesh1987, line 77: Can we call it `updatePaymentProviderCustomerInformation` ? This does seem a bit verbose though 🤔

animesh1987: Or can just call it `updatePaymentProviderCustomer`

packages/core/src/payment-integration/default-payment-integration-service.ts
```diff
@@ -207,4 +209,16 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
 
         return this._storeProjection.getState();
     }
+
+    async updateProviderCustomerData<T>(
+        providerCustomerData: T,
+    ): Promise<PaymentIntegrationSelectors> {
+        await this._store.dispatch(
```
animesh1987, line 216: Is this supposed to be async?

packages/payment-integrations-test-utils/src/test-utils/payment-integration-service.mock.ts
serhii-tkachenko: `updatePaymentProviderCustomer` looks good to me

packages/core/src/payment-integration/default-payment-integration-service.ts
serhii-tkachenko: It does not matter which action get called, however `this._store.dispatch` returns a Promise. So yes, we should use `await` here.

PavlenkoM (approved):

packages/core/src/payment-provider-customer/payment-provider-customer-actions-creator.ts
```diff
@@ -0,0 +1,20 @@
...
+    PaymentProviderCustomerType,
+} from './payment-provider-customer-actions';
+
+export default class PaymentProviderCustomerActionCreator {
+    updatePaymentProviderCustomer<T = object>(
+        providerCustomerData: T,
+    ): Observable<PaymentProviderCustomerAction> {
+        return of(
+            createAction(
+                PaymentProviderCustomerType.PaymentProviderCustomerCollected,
```
animesh1987, line 15:
```suggestion
                PaymentProviderCustomerType.UpdatePaymentProviderCustomer,
```

packages/core/src/payment-provider-customer/payment-provider-customer-state.ts
```diff
@@ -0,0 +1,7 @@
+export default interface PaymentProviderCustomerState<T = any> {
```
animesh1987, line 1: Do we need to use `any` here?

serhii-tkachenko: To avoid generic prop drilling through CheckoutStoreState

packages/core/src/payment-provider-customer/payment-provider-customer-actions-creator.ts
serhii-tkachenko: Thanks, it looks much meaningful. Updated.

animesh1987 (approved):

PR #2085, opened 2023-07-31
refactor(checkout): CHECKOUT-7634 Improvements for extension

bc-peng:
## What?

- Task 1: Make `getExtensionByRegion` a public selector to enable `checkout-js` consumption.
- Task 2: Move extension experiment check to `ExtensionService#renderExtension`.
- Task 3: Map command type to handler signature in `handleExtensionCommand`.

## Why?
Improve code structure and developer experience.

## Testing / Proof
- CI checks.

### DIFF
--- a/packages/core/src/extension/extension-messenger.ts
+++ b/packages/core/src/extension/extension-messenger.ts
@@ -93,18 +93,21 @@ export class ExtensionMessenger {
     }
 
     private _validateCommand<T extends keyof ExtensionCommandMap>(command: string): T {
-        switch (command) {
-            case 'RELOAD_CHECKOUT':
-                return ExtensionCommandType.ReloadCheckout as T;
-
-            case 'SHOW_LOADING_INDICATOR':
-                return ExtensionCommandType.ShowLoadingIndicator as T;
-
-            case 'SET_IFRAME_STYLE':
-                return ExtensionCommandType.SetIframeStyle as T;
-
-            default:
-                throw new UnsupportedExtensionCommandError();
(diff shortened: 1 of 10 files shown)


packages/core/src/extension/extension-messenger.ts
```diff
@@ -93,16 +92,16 @@ export class ExtensionMessenger {
...
-                return ExtensionCommandType.ReloadCheckout;
+                return ExtensionCommandType.ReloadCheckout as T;
 
             case 'SHOW_LOADING_INDICATOR':
-                return ExtensionCommandType.ShowLoadingIndicator;
+                return ExtensionCommandType.ShowLoadingIndicator as T;
 
             case 'SET_IFRAME_STYLE':
-                return ExtensionCommandType.SetIframeStyle;
+                return ExtensionCommandType.SetIframeStyle as T;
```
bc-peng, line 104: `as` is a side effect of `const num`. 😂

animesh1987: Generics are hard, especially on our present typescript version 🤦

animesh1987: On another side note can we use typeguard to check of the generic is of a particular kind?

bc-peng: I didn't think I could but I added a `tyepguard<T>` to replace `as T`.
Please check the update.

animesh1987: Nicely done 👏

animesh1987 (approved):
Nice work 👏

PR #2086, opened 2023-07-31
feat(payment): PAYPAL-000 created storage package

serhii-tkachenko:
## What?
Created "storage" package and added BrowserStorage there.

## Why?
To be able to use it in different integration packages

## Testing / Proof
Unit tests
CI


PavlenkoM (approved):

animesh1987:
Hey @serhii-tkachenko don't think we should move this to payment integration service. maybe a better idea is to move this a separate package?

serhii-tkachenko:
> Hey @serhii-tkachenko don't think we should move this to payment integration service. maybe a better idea is to move this a separate package?

Hmm.. I think you are right. Let me update it

packages/storage/README.md
```diff
@@ -0,0 +1,11 @@
+# storage
```
animesh1987, line 1: Can you please update the readme to be a bit more detailed, compared to boilerplate one.

animesh1987 (approved):
Looks good, can you please address one comment before merging.

packages/storage/README.md
serhii-tkachenko: Sure. Give me some time to update it

PR #2088, opened 2023-08-01
CHECKOUT-7611: Notify extension when consignment changes

davidchin:
## What?
Notify extensions that have subscribed when there is a change to a consignment.

## Why?
Extensions can subscribe and react to the event. e.g.:

```ts
extensionService.addListener(
    ExtensionEventType.ConsignmentsChanged, 
    ({ payload: { consignments, previousConsignments } }) => {
        consignments.forEach((consignment) => {
            const previousConsignment = previousConsignments.find(({ id }) => id === consignment.id);

            Object.keys(consignment.address).forEach((key) => {
                if (consignment.address[key] === previousConsignment?.address[key]) {
                    return;
                }

                console.log(`${key}: '${previousConsignment?.address[key]}' -> '${consignment.address[key]}'`);
            });
        });
    }
);
```

## Testing / Proof
https://github.com/bigcommerce/checkout-sdk-js/assets/667603/0d6aa5f1-ecf6-4c1c-8028-c35f163c5191

@bigcommerce/team-checkout


packages/core/src/extension/extension-service.ts
```diff
@@ -47,12 +53,30 @@ export default class ExtensionService {
     }
 
     addListener(eventType: ExtensionEventType, callback: () => void = noop): () => void {
```
bc-peng, line 55: Perhaps we can rename this method as `subscribe` because an extension may want to listen for events from another extension or replies from `checkout-js`.

packages/core/src/extension/extension-client.ts
```diff
@@ -4,25 +4,21 @@ export interface InitializeExtensionServiceOptions {
 }
 
 export enum ExtensionEventType {
-    CheckoutLoaded = 'CHECKOUT_LOADED',
-    ShippingCountryChange = 'SHIPPING_COUNTRY_CHANGE',
-}
-
-export interface CheckoutLoadedEvent {
-    type: ExtensionEventType.CheckoutLoaded;
+    ShippingCountryChanged = 'EXTENSION:SHIPPING_COUNTRY_CHANGED',
```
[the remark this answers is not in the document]
davidchin: I was going to mention this in the channel. 😄 I think it's best if we namespace all the messages that we post, to minimise the chance of event name collision. For example, events for hosted fields have `HOSTED_FIELD:` prefix.

packages/core/src/extension/extension-service.ts
davidchin: I thought we agree that we "handle" commands and "listen" to events? If that's the case, I think `addListener` is a more appropriate name?

bc-peng: Make sense, we should follow this agreement.

packages/core/src/extension/extension-event-broadcaster.ts
```diff
@@ -0,0 +1,94 @@
...
+    private _unsubscribers: { [type: string]: ExtensionChangeUnsubscriber } = {};
+    private _listeners: { [id: string]: IframeEventListener<ExtensionInternalCommandMap> } = {};
+
+    constructor(
+        private _store: ReadableCheckoutStore,
+        private _extensionMessenger: ExtensionMessenger,
+        private _subscribers: Record<ExtensionEventType, ExtensionChangeSubscriber>,
+    ) {}
+
+    listen(): void {
```
animesh1987, line 25: Just wondering what will be the use case for this 🤔

packages/core/src/extension/subscribers/subscribe-shipping-country-change.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import { ExtensionChangeSubscriber } from './extension-change-subscriber';
+
+export const subscribeShippingCountryChange: ExtensionChangeSubscriber = (
+    store: ReadableCheckoutStore,
+    broadcaster: ExtensionEventBroadcaster,
+) => {
+    let countryCodes: Record<string, string> = getShippingCountryCodes(store.getState());
+
+    return store.subscribe(
```
animesh1987, line 13: This looks good, but as I see we need to have a consignment before sending the event for country change? If that's the case we might need to communicate accordingly 🤔

packages/core/src/extension/extension-event-broadcaster.ts
davidchin: It's called when the first extension is rendered, see https://github.com/bigcommerce/checkout-sdk-js/blob/dcbdcaa5d48947d5dd6c0460837eae87eddcd237/packages/core/src/checkout/checkout-service.ts#L1405-L1412

packages/core/src/extension/subscribers/subscribe-shipping-country-change.ts
davidchin: Hrm, I thought we want to broadcast changes that have been committed, not uncommitted changes. If we're looking to broadcaster the latter, we have to change the approach where we listen for UI events instead.

animesh1987 (approved):

animesh1987:
Nice work @davidchin 👏 Simply amazing.

animesh1987 (approved):
Thanks for the refactor in the PR itself, VoTW worked well 😄

bc-peng (approved):
Kudos ⭐️ to you.

PR #2089, opened 2023-08-01
feat(payment): PI-516 Move Stripe V3 to package

bc-sebastianszafraniec:
## What?
...

## Why?
...

## Testing / Proof
...


packages/core/src/payment/strategies/stripev3/index.ts
serhii-tkachenko, line 1: I think it might be better to keep this file, just leave the first line to be able to get any interface from './stripev3' file

packages/stripe-integration/src/stripev3/stripev3.mock.ts
```diff
@@ -276,3 +278,24 @@ export function getStripeShippingAddressGuestUserWithoutAddress(): StripeConfirm
         },
     };
 }
+
+export function getBillingAddress(): BillingAddress {
```
serhii-tkachenko, line 282: You can get this mock from payment-integration-test-utils package

packages/stripe-integration/src/stripev3/stripev3.ts
```diff
@@ -0,0 +1,945 @@
...
+}
+
+export interface IndividualCardElementOptions {
+    cardCvcElementOptions: CardCvcElementOptions;
+    cardExpiryElementOptions: CardExpiryElementOptions;
+    cardNumberElementOptions: CardNumberElementOptions;
+    zipCodeElementOptions?: ZipCodeElementOptions;
+}
+
+export default function isIndividualCardElementOptions(
```
serhii-tkachenko, line 921: As far as I see this file contains enums and interfaces, so I think, we can move this type guard method to another file

packages/payment-integration-api/src/billing/is-billing-address-like.ts
```diff
@@ -0,0 +1,9 @@
+import { Address } from '../address';
+
+import BillingAddress from './billing-address';
+
+export default function isBillingAddressLike(address: Address): address is BillingAddress {
+    const billingAddress = address as BillingAddress;
+
+    return typeof billingAddress.id !== 'undefined';
+}
```
serhii-tkachenko, line 9: It might be better to check all required billing address fields to make sure that it is a billing address. Also if its possible, try to avoid using 'as' operator.

vitalii-koshovyi (approved):
Good job!

packages/payment-integration-api/src/billing/is-billing-address-like.ts
bc-sebastianszafraniec: The only required field in BillingAddress is „id”, „email” is optional. So if this function is checking if provided address argument is type of BillingAddress, I think that checking „id” is sufficient.
In term of assertion looks like it is necessary here.

packages/stripe-integration/src/stripev3/stripev3.ts
bc-sebastianszafraniec: I moved isIndividualCardElementOptions to another file

packages/stripe-integration/src/stripev3/stripev3.mock.ts
bc-sebastianszafraniec: done

packages/core/src/payment/strategies/stripev3/index.ts
bc-sebastianszafraniec: done

serhii-tkachenko (approved):
Nice work!

packages/payment-integration-api/src/billing/is-billing-address-like.ts
```diff
@@ -0,0 +1,9 @@
+import { Address } from '../address';
+
+import BillingAddress from './billing-address';
+
+export default function isBillingAddressLike(address: Address): address is BillingAddress {
+    const billingAddress = address as BillingAddress;
```
animesh1987, line 6: Can we avoid using as here?

packages/stripe-integration/README.md
```diff
@@ -0,0 +1,11 @@
+# stripe-integration
+
+This library was generated with [Nx](https://nx.dev).
```
animesh1987, line 3: Can we please add a better readme here.

packages/stripe-integration/src/stripev3/is-individual-card-element-options.ts
```diff
@@ -0,0 +1,17 @@
+import { IndividualCardElementOptions } from './stripev3';
+
+export default function isIndividualCardElementOptions(
+    individualCardElementOptions: unknown,
+): individualCardElementOptions is IndividualCardElementOptions {
+    return (
+        Boolean(
+            (individualCardElementOptions as IndividualCardElementOptions).cardNumberElementOptions,
```
animesh1987, line 8: Avoid as please.

packages/stripe-integration/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "include": [],
+    "references": [
+        {
+            "path": "./tsconfig.lib.json"
+        },
+        {
+            "path": "./tsconfig.spec.json"
+        }
+    ],
+    "compilerOptions": {
```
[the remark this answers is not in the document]
animesh1987: This file should be an extension of 
```
{
    "extends": "../../tsconfig.base.json"
}

```

packages/payment-integration-api/src/payment/payment-method-actions.ts
```diff
@@ -0,0 +1,51 @@
+import { Action } from '@bigcommerce/data-store';
+
+import PaymentMethod from './payment-method';
```
animesh1987, line 3: Is there a reason why we are adding these here?

packages/payment-integration-api/src/billing/is-billing-address-like.ts
bc-sebastianszafraniec: Serhii already asked about it, but looks like assertion is necessary here.

packages/stripe-integration/src/stripev3/is-individual-card-element-options.ts
bc-sebastianszafraniec: The same in this case, I think "as" is necessary here.

packages/payment-integration-api/src/payment/payment-method-actions.ts
bc-sebastianszafraniec: I added this file because some types are used in stripev3-payment-strategy.spec.ts file. 
Should I keep the whole file as it is now or leave just types which are used?

packages/stripe-integration/README.md
bc-sebastianszafraniec: I edited readme file and added some more informations.

packages/payment-integration-api/src/billing/is-billing-address-like.ts
animesh1987: We can check for properties in the object, that is much better than using assertion. Since using assertion is an override, you can call a `string as number` and it will be fine.

packages/stripe-integration/src/stripev3/is-individual-card-element-options.ts
animesh1987: Same as above.

packages/payment-integration-api/src/payment/payment-method-actions.ts
animesh1987: I just had a look at the file, you can possibly rewrite the test to avoid this. I say since these are internals for core package and should not be exported or used outside.

bc-sebastianszafraniec: Tests are rewrited so the was no need to import this file

packages/stripe-integration/src/stripev3/is-individual-card-element-options.ts
bc-sebastianszafraniec: Done

packages/payment-integration-api/src/billing/is-billing-address-like.ts
bc-sebastianszafraniec: Done

animesh1987 (approved):
Nice work 👏

PR #2097, opened 2023-08-07
feat(payment): PAYPAL-2500 captured device info for Braintree Apple Pay and Braintree Google Pay

bc-nick:
## What?

Added `deviceSessionId` to the payment payload for the GP and AP strategies

## Why?

`Braintree Fraud Protection` requires deviceData to work correctly.
To do this we need to pass `deviceSessionId` and make sure that the device data captured in the transaction journal.

## Testing / Proof

Tested with a Braintree sandbox account

From transaction journal
Screenshot 2023-08-09 at 18 39 35

Data Device Captured equals `true` for both (GP and AP) payment method

### DIFF
--- a/packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
+++ b/packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
@@ -4,6 +4,8 @@ import {
     Cart,
     Checkout,
     InvalidArgumentError,
+    MissingDataError,
+    MissingDataErrorType,
     NotInitializedError,
     NotInitializedErrorType,
     OrderFinalizationNotRequiredError,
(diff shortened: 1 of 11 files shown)


packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
```diff
@@ -235,4 +247,44 @@ export default class ApplePayPaymentStrategy implements PaymentStrategy {
             );
         }
     }
+
+    private async _getBraintreeApplePaySessionId(applePayPaymentMethod: PaymentMethod) {
+        if (applePayPaymentMethod.initializationData?.gateway === ApplePayGatewayType.BRAINTREE) {
```
serhii-tkachenko, line 252: Please move this condition to the place where you are calling _getBraintreeApplePaySessionId method

PR #2106, opened 2023-08-09
chore(common): CHECKOUT-7612 Upgrade to node v18

animesh1987:
## What?
Update to node 18

## Why?
- With node 16 being EOL https://nodejs.org/en/blog/announcements/nodejs16-eol earlier than expected we should update node version and keep SLA for repo.

## Testing / Proof
- CI

@bigcommerce/team-checkout


bc-peng:
DA DA , a gentle reminder.
https://github.com/bigcommerce/checkout-sdk-js/blob/master/.nvmrc

bc-peng (approved):
Node 1️⃣ 🎱, let's go.

PR #2113, opened 2023-08-14
feat(checkout): PI-520 Stripe UPE - Provide AVS results

bc-ania:
## What?
Jira issue: https://bigcommercecloud.atlassian.net/browse/PI-520
As a merchant, I want Address Line and Zip/Post Code responses for failures to manually review orders and see if I want to accept them. Currently, we only provide when they match and not when they don’t - Even though Stripe has the information and sends it via webhooks. 

## Why?
SportsShoes have requested this information, it's part of their internal fraud review process. Orders can be flagged for review for a number of reasons (e.g. high value order, high risk items etc). Having AVS details helps the staff to decide what to do with the order. 

## Testing / Proof
...

### DIFF
--- a/packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
+++ b/packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
@@ -579,9 +579,15 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
 
     private _mapStripeAddress(address?: Address): AddressOptions {
         if (address) {
-            const { city, countryCode: country, postalCode } = address;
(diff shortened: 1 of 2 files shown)


packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
serhii-tkachenko, line 1: It might be better to cover your changes with tests

vitalii-koshovyi: agree with Serhii here👍

serhii-tkachenko (approved):
Nice work!

animesh1987 (approved):

PR #2116, opened 2023-08-15
CHECKOUT-7652: Only trigger command handler if command matches its extension ID

davidchin:
## What?
* Fix a problem where a command handler registered for one extension could be triggered by a command issued by another extension.
* Remove duplicate `ExtensionCommand` interface (defined in `extension-client.ts`).
* Improve `ExtensionCommand` type by removing `extensionId`, as extension clients are not required to provide it.
* Replace `ExtensionCommandType.FrameLoaded` command with `ExtensionInternalCommandType.ResizeIframe` command, as the event/command is an internal one that should not be issued by extension clients.
* Create a separate `ResizableIframeCreator` class within `extensions` module instead of reusing a similar one from `embedded-checkout` module to decouple the two.
* Add a namespace for all extension commands to avoid potential conflicts. They are now prefixed with `EXTENSION:`.
* Fix the export of `ExtensionCommandType` and `ExtensionEventType` enums so they can be imported by extension clients.

## Why?
To fix a bug, refine SDK API and improve maintainability.

## Testing / Proof
https://github.com/bigcommerce/checkout-sdk-js/assets/667603/f4bee89e-b04a-461a-a4f7-038bcfe7df9c

@bigcommerce/team-checkout


packages/core/src/extension/extension-messenger.ts
```diff
@@ -94,22 +110,11 @@ export class ExtensionMessenger {
         return extension;
     }
 
-    private _validateCommand<T extends keyof ExtensionCommandMap>(command: string): T {
-        if (this._isExtensionCommandType<T>(command)) {
+    private _validateCommand<T extends keyof ExtensionCommandMap>(command: T): T {
+        if (Object.values(ExtensionCommandType).includes(command)) {
```
animesh1987, line 114: @bc-peng I can't remember why we had the generic typeguard here before? This change looks good, can you please check as well.

packages/core/src/extension/extension-service.ts
```diff
@@ -33,23 +33,19 @@ export default class ExtensionService {
         this._extensionId = extensionId;
 
         this._eventListener.listen();
+        this._commandPoster.setContext({ extensionId });
+        this._internalCommandPoster.post({
+            type: ExtensionInternalCommandType.ResizeIframe,
```
animesh1987, line 38: Are we changing the `FRAME_LOADED` command to `ResizeIframe`?

packages/core/src/extension/extension-service.ts
```diff
@@ -33,23 +33,19 @@ export default class ExtensionService {
...
+            throw new Error(`${command.type} is not supported.`);
         }
 
-        const payload = {
-            ...event.payload,
-            extensionId: this._extensionId,
-        };
-
-        this._eventPoster.post({ ...event, payload });
+        this._commandPoster.post(command);
```
animesh1987, line 48: Not sure if I missed another change, does the client need to pass the extensionId explicitly?

packages/core/src/extension/extension-messenger.ts
bc-peng: We need a typeguard here because of `const`, and DC removed it in this PR.
image

packages/core/src/extension/extension-service.ts
animesh1987: Or they don't need to pass extensionId at all now?

packages/core/src/extension/extension-messenger.ts
animesh1987: Perfect 👏

davidchin: @animesh1987 I'm also exporting `ExtensionCommandType` explicitly so it can be imported as a regular object. Previously, we're only exporting it as a TS enum (`const enum`) so clients can't use it as a value.

packages/core/src/extension/extension-service.ts
davidchin: Yes, because `FRAME_LOADED` is an event, not a command. Therefore I renamed it to be a command.

packages/core/src/extension/extension-service.ts
davidchin: I think as clients you don't need to pass it, see https://github.com/bigcommerce/checkout-extension-example/blob/4943cf88922f96174fa05fece849032f502eeb65/src/commands/loading-indicator.js#L2-L5

packages/core/src/extension/extension-service.ts
animesh1987: I think we just need to check if we were making listening to FrameLoaded specifically on checkout side, @bc-peng do you remember?

davidchin: If my understanding is correct, Checkout doesn't listen for it. Instead, it's listened by `ResizeableIframeCreator`.

bc-peng: https://github.com/bigcommerce/checkout-sdk-js/blob/01378de2d1e69efa38214d863178457943eead2b/packages/core/src/embedded-checkout/resizable-iframe-creator.ts#L68
We display an iframe upon receiving it.

bc-peng (approved):
Great job, I realized the gap here. 🙇🏻‍♂️

packages/core/src/extension/extension-service.ts
animesh1987: Yes, that's correct. We were adding that in the service in order to keep the client side simple.  Ahh I see we now set it as extension context, is that correct?

animesh1987 (commented):
Nice work DC, as usual fantastic effort in refactoring and making things more structured and clearer. 🚢

animesh1987 (approved):

PR #2131, opened 2023-08-18
feat(payment): PAYPAL-2843 captured device info for Braintree Apple Pay

bc-nick:
## What?

Added `deviceSessionId` to the payment payload for the AP **payment** strategy

## Why?

To do this we need to pass `deviceSessionId` and make sure that the device data captured in the transaction journal.

## Testing / Proof

Tested with a Braintree sandbox account

Screenshot 2023-08-18 at 17 57 21


packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
```diff
@@ -235,4 +248,27 @@ export default class ApplePayPaymentStrategy implements PaymentStrategy {
             );
         }
     }
+
+    private async _getDataCollector() {
```
vitalii-koshovyi, line 252:
```suggestion
    private async _getDeviceData() {
```
looks better to me

vitalii-koshovyi (approved):

packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
```diff
@@ -235,4 +248,27 @@ export default class ApplePayPaymentStrategy implements PaymentStrategy {
...
+        const state = await this._paymentIntegrationService.loadPaymentMethod(
```
serhii-tkachenko, line 253: It could be better to load braintree in initialization method

packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
```diff
@@ -235,4 +248,27 @@ export default class ApplePayPaymentStrategy implements PaymentStrategy {
...
+            ApplePayGatewayType.BRAINTREE,
+        );
+
+        const braintreePaymentMethod: PaymentMethod = state.getPaymentMethodOrThrow(
+            ApplePayGatewayType.BRAINTREE,
+        );
+
+        if (!braintreePaymentMethod.clientToken || !braintreePaymentMethod.initializationData) {
```
serhii-tkachenko, line 261: I don't think we should stop user to make a transaction if the error occurs on getting braintree device data. Hence to avoid blocking the flow, it might be better to return empty device data, if there are no `clientToken` or `initializationData`.

packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
bc-nick: Moved load braintree in initialization method

packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
bc-nick: Thanks. Done

serhii-tkachenko (approved):
Looks good to me

packages/apple-pay-integration/src/create-apple-pay-payment-strategy.ts
```diff
@@ -12,11 +17,16 @@ const createApplePayPaymentStrategy: PaymentStrategyFactory<ApplePayPaymentStrat
...
+    const hostWindow = window;
 
     return new ApplePayPaymentStrategy(
         createRequestSender({ host: getHost() }),
         paymentIntegrationService,
         new ApplePaySessionFactory(),
+        new BraintreeIntegrationService(
+            new BraintreeScriptLoader(getScriptLoader(), hostWindow),
+            hostWindow,
+        ),
```
bc-peng, line 29: I can see `BraintreeIntegrationService` is instaniated to support `getDeviceData()` only.
Does it make sense to create a `BraintreeIntegrationService` only when the gateway is `braintree`?

packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
```diff
@@ -235,4 +253,27 @@ export default class ApplePayPaymentStrategy implements PaymentStrategy {
             );
         }
     }
+
+    private async _getDeviceData() {
```
bc-peng, line 257: Can you move this function to a braintree related pakcage since it is obvious that this is the braintree way of collecting data?

packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
```diff
@@ -209,9 +219,17 @@ export default class ApplePayPaymentStrategy implements PaymentStrategy {
         promise: ApplePayPromise,
     ) {
         const { token } = event.payment;
+
+        let deviceSessionId: string | undefined;
+
+        if (paymentMethod.initializationData?.gateway === ApplePayGatewayType.BRAINTREE) {
+            deviceSessionId = await this._getDeviceData();
```
bc-peng, line 226: Perhaps we can instantiate a `braintreeIntegrationService` here and get device data.

packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
```diff
@@ -58,7 +61,14 @@ export default class ApplePayPaymentStrategy implements PaymentStrategy {
...
         this._shippingLabel = options.applepay?.shippingLabel || DefaultLabels.Shipping;
         this._subTotalLabel = options.applepay?.subtotalLabel || DefaultLabels.Subtotal;
-        await this._paymentIntegrationService.loadPaymentMethod(methodId);
+
+        const state = await this._paymentIntegrationService.loadPaymentMethod(methodId);
+
+        const paymentMethod: PaymentMethod = state.getPaymentMethodOrThrow(methodId);
+
+        if (paymentMethod.initializationData?.gateway === ApplePayGatewayType.BRAINTREE) {
+            await this._paymentIntegrationService.loadPaymentMethod(ApplePayGatewayType.BRAINTREE);
```
animesh1987, line 70: We already fetch paymentMethod on line 65, Is that not sufficient?

animesh1987:
This looks ok, but I have questions and doubts around shall we have a separate strategy for braintree and apple pay. Similar to how google pay is being implemented?

bc-nick:
> This looks ok, but I have questions and doubts around shall we have a separate strategy for braintree and apple pay. Similar to how google pay is being implemented? Since now we have conditional logic that caters to Braintree, can see we will have another request from another gateway and that adds another class in dependencies and conditional logic in strategy.

@animesh1987 it is a good point, thanks! yes, it is possible to get another request from another gateway. We can think about separating similar to google pay but first of all we need to close this feature request for all places where it is needed.

packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
bc-nick: Did not understand a little bit. But we can't do it because the core logic is already placed in related package.

packages/apple-pay-integration/src/create-apple-pay-payment-strategy.ts
bc-nick: do you mean something like this https://github.com/bigcommerce/checkout-sdk-js/pull/2131#discussion_r1305002849 ?

bc-peng: Yes, I do.

packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
bc-peng: I believe you should be able to move this function to a Braintree package and end up with:
```
getBraintreeDeviceData(loadPaymentMethod, braintreeIntegrationService): DeviceData
```

packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
bc-peng: Yes, by doing so, we can remove `private _braintreeIntegrationService: BraintreeIntegrationService` from the constructor.

bc-peng (approved):

PR #2136, opened 2023-08-22
feat(payment): PAYPAL-2729 updated ExecutePaymentMethodCheckoutOptions with an optional email field

serhii-tkachenko:
## What?
Updated ExecutePaymentMethodCheckoutOptions with an optional email field

## Why?
To be able to use customer's email in customer strategy execution method.
It might be helpful for cases when we need to call execution method right after sign up.

## Testing / Proof
Unit tests
Manual tests

### DIFF
--- a/packages/core/src/customer/customer-request-options.ts
+++ b/packages/core/src/customer/customer-request-options.ts
@@ -151,6 +151,7 @@ export interface BaseCustomerInitializeOptions extends CustomerRequestOptions {
  *
  */
 export interface ExecutePaymentMethodCheckoutOptions extends CustomerRequestOptions {
+    email?: string;
     checkoutPaymentMethodExecuted?(data?: CheckoutPaymentMethodExecutedOptions): void;
     continueWithCheckoutCallback?(): void;
 }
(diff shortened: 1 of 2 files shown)


animesh1987:
Hey @serhii-tkachenko can you please explain what problem this is trying to solve?

serhii-tkachenko:
> Hey @serhii-tkachenko can you please explain what problem this is trying to solve?

I am passing customer's email address to use it for triggering PayPal connect authentication flow. I can't get it in any other way, because we are not updating checkout state after user sign up

animesh1987:
> > Hey @serhii-tkachenko can you please explain what problem this is trying to solve?
> 
> I am passing customer's email address to use it for triggering PayPal connect authentication flow. I can't get it in any other way, because we are not updating checkout state after user sign up

Just wanted to check can we not get it from the customer object in checkout state?

serhii-tkachenko:
I checked it earlier and customer object was undefined if we try to get it after sign up. But I can double check it one more time, to be sure that it works as I think

serhii-tkachenko:
We with @animesh1987 double checked checkout state (customer object) after customer's sign up on integration and production envs and the state was updated successfully. It seems like we have an issue on local env, but it's not related to this PR, so I close this one because we don't need this implementation anymore.

PR #2145, opened 2023-08-23
feat(payment): PAYPAL-2728 create shipping strategy for Braintree AXO

PavlenkoM:
## What?
Create shipping strategy for Braintree Accelerated checkout

## Why?
To check logged in user and show OTP modal for PayPal Connect users

## Testing / Proof
Unit test and manually tested

https://github.com/bigcommerce/checkout-sdk-js/assets/9430298/dfb0a7d0-8186-4744-bccc-f8a41b765652

### DIFF
--- a/packages/core/src/shipping/create-shipping-strategy-registry.ts
+++ b/packages/core/src/shipping/create-shipping-strategy-registry.ts
@@ -1,17 +1,27 @@
 import { RequestSender } from '@bigcommerce/request-sender';
 import { getScriptLoader } from '@bigcommerce/script-loader';
 
+import {
+    BraintreeHostWindow,
+    BraintreeIntegrationService,
+    BraintreeScriptLoader,
+} from '@bigcommerce/checkout-sdk/braintree-utils';
+
+import { BillingAddressActionCreator, BillingAddressRequestSender } from '../billing';
 import { CheckoutRequestSender, CheckoutStore } from '../checkout';
 import { Registry } from '../common/registry';
 import { PaymentMethodActionCreator, PaymentMethodRequestSender } from '../payment';
+import { PaymentProviderCustomerActionCreator } from '../payment-provider-customer';
 import { createAmazonPayV2PaymentProcessor } from '../payment/strategies/amazon-pay-v2';
 import { StripeScriptLoader } from '../payment/strategies/stripe-upe';
+import { SubscriptionsActionCreator, SubscriptionsRequestSender } from '../subscription';
 
 import ConsignmentActionCreator from './consignment-action-creator';
 import ConsignmentRequestSender from './consignment-request-sender';
 import ShippingStrategyActionCreator from './shipping-strategy-action-creator';
 import { ShippingStrategy } from './strategies';
 import { AmazonPayV2ShippingStrategy } from './strategies/amazon-pay-v2';
+import { BraintreeAcceleratedCheckoutShippingStrategy } from './strategies/braintree';
 import { DefaultShippingStrategy } from './strategies/default';
 import { StripeUPEShippingStrategy } from './strategies/stripe-upe';
 
@@ -30,6 +40,14 @@ export default function createShippingStrategyRegistry(
         new PaymentMethodRequestSender(requestSender),
     );
     const scriptLoader = getScriptLoader();
+    const subscriptionsActionCreator = new SubscriptionsActionCreator(
+        new SubscriptionsRequestSender(requestSender),
+    );
+    const billingAddressActionCreator = new BillingAddressActionCreator(
+        new BillingAddressRequestSender(requestSender),
+        subscriptionsActionCreator,
+    );
+    const braintreeHostWindow: BraintreeHostWindow = window;
 
     registry.register(
         'amazonpay',
@@ -51,6 +69,22 @@ export default function createShippingStrategyRegistry(
                 new StripeScriptLoader(scriptLoader),
                 consignmentActionCreator,
                 paymentMethodActionCreator,
+            ),
+    );
+
+    registry.register(
+        'braintreeacceleratedcheckout',
+        () =>
+            new BraintreeAcceleratedCheckoutShippingStrategy(
+                store,
+                billingAddressActionCreator,
+                consignmentActionCreator,
+                paymentMethodActionCreator,
+                new PaymentProviderCustomerActionCreator(),
+                new BraintreeIntegrationService(
+                    new BraintreeScriptLoader(getScriptLoader(), braintreeHostWindow),
+                    braintreeHostWindow,
+                ),
             ),
     );
 
--- a/packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
+++ b/packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
@@ -1,0 +1,221 @@
+import {
+    BraintreeConnectAddress,
+    BraintreeConnectVaultedInstrument,
+    BraintreeIntegrationService,
+} from '@bigcommerce/checkout-sdk/braintree-utils';
+import { BrowserStorage } from '@bigcommerce/checkout-sdk/storage';
+
+import { AddressRequestBody } from '../../../address';
+import { BillingAddressActionCreator } from '../../../billing';
+import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
+import {
+    InvalidArgumentError,
+    MissingDataError,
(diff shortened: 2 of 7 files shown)


packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
```diff
@@ -0,0 +1,219 @@
...
+        this.braintreeIntegrationService.initialize(clientToken, initializationData);
+
+        const braintreeConnect = await this.braintreeIntegrationService.getBraintreeConnect(
+            cart?.id,
+        );
+
+        const customerEmail = customer?.email || billingAddress?.email;
+
+        if (!customerEmail) {
+            return Promise.resolve(state);
```
serhii-tkachenko, line 120: Actually, we don't need to return anything here and in the code below in this function

serhii-tkachenko (approved):
Nice work! 👌

packages/core/src/shipping/create-shipping-strategy-registry.ts
```diff
@@ -54,6 +76,19 @@ export default function createShippingStrategyRegistry(
             ),
     );
 
+    registry.register(
```
animesh1987, line 79: Can you please update a test for this too.

packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
```diff
@@ -0,0 +1,219 @@
...
+
+    constructor(
+        private store: CheckoutStore,
+        private billingAddressActionCreator: BillingAddressActionCreator,
+        private consignmentActionCreator: ConsignmentActionCreator,
+        private paymentMethodActionCreator: PaymentMethodActionCreator,
+        private paymentProviderCustomerActionCreator: PaymentProviderCustomerActionCreator,
+        private braintreeIntegrationService: BraintreeIntegrationService,
+    ) {
+        this.browserStorage = new BrowserStorage('paypalConnect');
```
animesh1987, line 39:
```suggestion
        this._browserStorage = new BrowserStorage('paypalConnect');
```

PavlenkoM: done

packages/core/src/shipping/create-shipping-strategy-registry.ts
PavlenkoM: done

packages/core/src/shipping/create-shipping-strategy-registry.ts
```diff
@@ -30,6 +40,18 @@ export default function createShippingStrategyRegistry(
...
+    );
+    const billingAddressActionCreator = new BillingAddressActionCreator(
+        new BillingAddressRequestSender(requestSender),
+        subscriptionsActionCreator,
+    );
+    const braintreeHostWindow: BraintreeHostWindow = window;
+    const braintreeIntegrationService = new BraintreeIntegrationService(
+        new BraintreeScriptLoader(getScriptLoader(), braintreeHostWindow),
+        braintreeHostWindow,
+    );
```
bc-peng, line 54: Could you please move this to line 88?

packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+        try {
+            await this.store.dispatch(this.paymentMethodActionCreator.loadPaymentMethod(methodId));
+            await this.runAuthenticationFlowOrThrow(methodId);
+        } catch (error) {
+            // Info: we should not throw any error here to avoid
+            // customer stuck on shipping step due to the payment provider
+            // custom flow
+        }
```
bc-peng, line 84: What if an error occurs during `runAuthenticationFlowOrThrow`, can a shopper proceed to the payment and check out?

packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
bc-peng: DC corrected me this before. 😂
>We don't really use `_` as a convention for private members in checkout-js because it's not published as a library.
https://github.com/bigcommerce/checkout-js/pull/1392#discussion_r1275649501

packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
bc-peng: I mean will the payment step work if the shipping encounters an error?

PavlenkoM: Yes, the regular flow will work in this case.
The main idea of this empty `catch` is to skip all fails of AXO flow and not to lock checkout flow for shoppers.
And payment step also has logic for such case.

packages/core/src/shipping/create-shipping-strategy-registry.ts
PavlenkoM: I am not sure that keep so big construction as an argument it is a good idea. As for me it is less comfortable for reading. But if you think that it will be better, no problem, it moved.

packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
```diff
@@ -0,0 +1,219 @@
...
+
+        return Promise.resolve(this.store.getState());
+    }
+
+    private shouldRunAuthenticationFlow(): boolean {
```
animesh1987, line 89:
```suggestion
    private _shouldRunAuthenticationFlow(): boolean {
```

can you please add _ for all private members

PavlenkoM: oh, OK
done

packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
```diff
@@ -0,0 +1,221 @@
...
+            this._paymentProviderCustomerActionCreator.updatePaymentProviderCustomer({
+                authenticationState,
+                addresses,
+                instruments,
+            }),
+        );
+
+        if (addresses.length > 0) {
+            await this._store.dispatch(
+                this._billingAddressActionCreator.updateAddress({
```
animesh1987, line 150: I see we are updating billing address here already, what happens for the case when this setting is off?
Screenshot 2023-08-31 at 10 35 45 pm

packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
```diff
@@ -0,0 +1,221 @@
...
+            company: address.company || '',
+            address1: address.streetAddress,
+            address2: address.extendedAddress || '',
+            city: address.locality,
+            stateOrProvince: address.region,
+            stateOrProvinceCode: address.region,
+            country: this._getCountryNameByCountryCode(address.countryCodeAlpha2, countries),
+            countryCode: address.countryCodeAlpha2,
+            postalCode: address.postalCode,
+            phone: '',
```
animesh1987, line 186: What happens in the case for stores that have phone number as a required property?

packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
PavlenkoM: In this case we are also preset billing address as it is AXO flow, but after this user can disable this check mark and select address which he wants, as in general checkout flaw

packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
PavlenkoM: In this case will be shown form where user will add a phone number, because we don't change the basic checkout flaw, we only complement it
https://github.com/bigcommerce/checkout-js/blob/master/packages/core/src/app/shipping/ShippingAddressForm.tsx#L80

animesh1987 (approved):

bc-peng (approved):

PR #2146, opened 2023-08-24
fix(payment): CHECKOUT-6883 Pass store credit to apple pay sheet

animesh1987:
## What?
- Pass store credit to be displayed in apple pay sheet.
- Remove apple pay payment strategy in core package.

## Why?
If an order contains an applied store credit value, and ‘Apple Pay' is selected at the payment method at checkout we should reflect the correct details in apple pay sheet.

## Testing / Proof
- Store credit applied
Screenshot 2023-08-24 at 10 23 28 pm

- Store credit removed
Screenshot 2023-08-24 at 10 24 06 pm

@bigcommerce/team-checkout


bc-peng (approved):
LGTM. Nice work.👏

PR #2152, opened 2023-08-30
fix(payment): PAYPAL-2877 captured device info for Braintree GooglePay

bc-nick:
## What?

Added deviceSessionId to the payment payload for the GooglePay payment strategy

## Why?

To do this we need to pass deviceSessionId and make sure that the device data captured in the transaction journal.

## Testing / Proof

Screenshot 2023-08-30 at 13 23 39

Verified for the next GP strategies: payment, customer and button

### DIFF
--- a/packages/braintree-utils/src/index.ts
+++ b/packages/braintree-utils/src/index.ts
@@ -37,6 +37,7 @@ export {
     getPaypalCheckoutMock,
     getPayPalCheckoutCreatorMock,
     getBraintreeLocalPaymentMock,
+    getDeviceDataMock,
 } from './mocks/braintree.mock';
 
 export { BRAINTREE_SDK_STABLE_VERSION, BRAINTREE_SDK_ALPHA_VERSION } from './sdk-verison';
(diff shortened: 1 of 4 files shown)


animesh1987:
Hey 👋 we have recently done a lot of refactoring for google pay, can you please create a separate braintree strategy heere https://github.com/bigcommerce/checkout-sdk-js/tree/master/packages/google-pay-integration/src/factories/payment.

Example PR: https://github.com/bigcommerce/checkout-sdk-js/pull/1967

bc-nick:
> Hey 👋 we have recently done a lot of refactoring for google pay, can you please create a separate braintree strategy heere https://github.com/bigcommerce/checkout-sdk-js/tree/master/packages/google-pay-integration/src/factories/payment.
> 
> Example PR: #1967

@animesh1987 hi! Yes, thought about it. But i think would be better to create a separate task for this after this a small fix. Because it would take more time. As well as for this PR https://github.com/bigcommerce/checkout-sdk-js/pull/2131

animesh1987:
I don't think it will take that long, as most of the work is moving boilerplate code, I believe it's better to move it now than add a fix here. As this is a good opportunity and going forward checkout team can be removed as a dependency here.

bc-nick:
> I don't think it will take that long, as most of the work is moving boilerplate code, I believe it's better to move it now than add a fix here. As this is a good opportunity and going forward checkout team can be removed as a dependency here.

@animesh1987 okay, we can avoid a lot of similar cases if we move it now. I'll get on it then for GP with braintree gateway. Thank you for the clarification

PR #2162, opened 2023-09-04
Revert "fix(payment): PAYPAL-000 fixed the issue with braintree initialization in apple pay payment strategy"

serhii-tkachenko:
Reverts bigcommerce/checkout-sdk-js#2161

Due to the issue with apple pay e2e tests in checkout js project


serhii-tkachenko:
There 2 apple pay PRs will be reverted in one PR, because they can't be reverted authomatically

PR #2169, opened 2023-09-07
feat(payment): PI-734 fix error message for empty card data in Mollie

bc-wpietrzak:
## What?
Fixed error message that appears when user tries to place order using card (via Mollie) without filling in card data.

## Why?
For better user experience.

## Testing / Proof
Error message before fix:
Screenshot 2023-09-07 at 09 26 54

Error message after fix:
Screenshot 2023-09-07 at 09 25 37


serhii-tkachenko (approved):
Nice work!

PR #2179, opened 2023-09-13
feat(payment): PI-811 Make an ability to use stripeLinkAuthenticatedA…

bc-sebastianszafraniec:
…ction in integration checkout-sdk packages

## What?
Add `stripeLinkAuthenticatedA` to `PaymentIntegrationService`

## Why?
To get ability to use `stripeLinkAuthenticatedA` from `PaymentIntegrationService`.
This method was added during work on moving StripeUPE customer strategy to package. Place in code where stripeLinkAuthenticated method was needed: 
Zrzut ekranu 2023-09-13 o 10 14 20

## Testing / Proof
Zrzut ekranu 2023-09-13 o 08 33 48

### DIFF
--- a/packages/core/src/payment-integration/default-payment-integration-service.ts
+++ b/packages/core/src/payment-integration/default-payment-integration-service.ts
@@ -1,3 +1,5 @@
+import { createAction } from '@bigcommerce/data-store';
+
 import {
     BillingAddressRequestBody,
     BuyNowCartRequestBody,
@@ -17,7 +19,7 @@ import { BillingAddressActionCreator } from '../billing';
 import { CartRequestSender } from '../cart';
 import { CheckoutActionCreator, CheckoutStore } from '../checkout';
 import { DataStoreProjection } from '../common/data-store';
-import { CustomerActionCreator, CustomerCredentials } from '../customer';
+import { CustomerActionCreator, CustomerActionType, CustomerCredentials } from '../customer';
 import { HostedFormFactory } from '../hosted-form';
 import { OrderActionCreator } from '../order';
 import {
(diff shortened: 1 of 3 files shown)


PavlenkoM (approved):

packages/core/src/payment-integration/default-payment-integration-service.ts
```diff
@@ -248,4 +250,14 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
 
         return this._storeProjection.getState();
     }
+
+    async stripeLinkAuthenticatedAction(
+        authenticated: boolean | undefined,
+    ): Promise<PaymentIntegrationSelectors> {
+        this._store.dispatch(
+            createAction(CustomerActionType.StripeLinkAuthenticated, authenticated),
```
animesh1987, line 258: In past this was done in a way since payment integrations were inside core package, can we implement this in such a way that we don't need to have provider specific methods here?

cc @serhii-tkachenko

bc-sebastianszafraniec: The action which is dispatched by stripeLinkAuthenticatedAction method is used in this place https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/customer/customer-reducer.ts#L48 and then used with combineReducers https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/customer/customer-reducer.ts#L30 mehtod which comes from @bigcommerce/data-store’ so I think that this method is neccessary.
But I think that maybe I can make it more generic. Something like this:

    async dispatchAction<TPayload, TMeta, TType extends string>(
        type: TType,
        payload?: TPayload,
        meta?: TMeta,
    ): Promise<PaymentIntegrationSelectors> {
        this._store.dispatch(createAction(type, payload, meta));

    }

serhii-tkachenko: We should not add dispatch action method, because it can cause some inconsistency between payment integrations and core. 

@animesh1987 it is a problem that we talked earlier about payment provider specific state. So in this case we should build extra branch for a state like `getPaymentProviderAuthentication(methodId: string)` and `updatePaymentProviderAuthentication(methodId: string, value: boolean)`. What do you think?

It might take some time on stripe refactoring, but it can resolve the issue that you faced in integration package.

As a reference to add new functionality for a state you can use this PR:
https://github.com/bigcommerce/checkout-sdk-js/pull/2075

But better to wait @animesh1987 to confirm, that this will be the best way to handle this.

animesh1987: We don't want to run into a state where we have a umbrella dispatch method, as that can cause a lot of confusion and give access to state where it should not be allowed.

I do agree with @serhii-tkachenko here we can use the state in the PR linked by Serhii or add something similar but not tied to a payment provider.

PR #2181, opened 2023-09-14
refactor(payment): PAYPAL-000 created new Braintree SDK as a part of future refactoring

serhii-tkachenko:
## What?
Created new Braintree SDK as a part of future refactoring

## Why?
It might be better to create separate class to keep BraintreeSdk interactions in one place. Also, we should keep our methods in those class isolated from any side effects to make it easier to debug and maintain.

## Testing / Proof
Unit tests


andriiVitvitskyi1990 (approved):

vitalii-koshovyi (approved):
LGTM!

bc-dronov (approved):

bc-nick (approved):

PR #2185, opened 2023-09-19
fix(payment): PAYPAL-2928 fixed the issue with braintree buttons rendering on PDP page

serhii-tkachenko:
## What?
Fixed the issue with braintree buttons rendering on PDP page.

## Why?
There was a problem with getting `storeConfig` to get access to feature flags. Therefore we should change `getStoreConfigOrThrow` to more calm `getStoreConfig` to get store config if it's available. It does not brake anything, because the reason why we need storeConfig is to understand whenever to load Braintree Alpha version for PayPal Connect that used only on Checkout Page.

## Testing / Proof
Unit tests
Manual tests

Before: 
There are console errors and the braintree buttons failed to render
Screenshot 2023-09-19 at 16 56 33

After:
There are no console errors and the braintree buttons rendered sucessfuly
Screenshot 2023-09-19 at 17 46 33


PavlenkoM (approved):

bc-nick (approved):
Will keep this case in mind for **apple-pay** strategy with **braintree** gateway. Thanks!

PR #2187, opened 2023-09-20
feat(extension): CHECKOUT-7733 Enhance Extension Auto-Resizing

bc-peng:
## What?
- Add two initialization options to allow an extension to resize itself better:
   -  `taggedElementId` : specify an element ID that is guaranteed to be visible on the checkout page.
   - `fixedHeight`: specify a number that defines the fixed height for an extension.
*When both options are specified, `fixedheight` takes priority.
- Add a semi-handshake mechanism at an extension's initial loading stage, guaranteeing visibility and usability.

## Why?
An extension's presentation can be precisely controlled.

## Testing / Proof
- CI checks.
- Manual testing. 
  This extension is consistenly changing height and colour.

  https://github.com/bigcommerce/checkout-sdk-js/assets/88361607/76006d32-c8b8-46a2-95c8-2a3ca5c20b6a


packages/core/src/extension/iframe-resizer-methods.ts
```diff
@@ -0,0 +1,43 @@
...
+                    }
+
+                    reject(new Error(`Element Id not found.`));
+                } else {
+                    window.parentIFrame.setHeightCalculationMethod('bodyOffset');
+
+                    resolve();
+                }
+            } else {
+                setTimeout(waitingForIframeResizer, 100);
```
bc-peng, line 37: `Window.parentIFrame` may not be available until a few ms during the extension side loading phase, hence retrying every 100ms.

This promise guarantees that the following usage of 'Window.parentIFrame' will not be delayed.

packages/core/src/extension/resizable-iframe-creator.ts
```diff
@@ -58,9 +58,10 @@ export default class ResizableIframeCreator {
 
                     const iframes = iframeResizer(
                         {
+                            autoResize: false,
                             scrolling: false,
                             sizeWidth: false,
-                            heightCalculationMethod: 'bodyScroll',
+                            heightCalculationMethod: 'bodyOffset',
```
bc-peng, line 64: According to [iframeResizer doc](https://github.com/davidjbradshaw/iframe-resizer/blob/master/docs/parent_page/options.md#widthcalculationmethod), `bodyOffset` and `taggedElement` are the only two methods that will not introudce screen flicker in some browsers.

Therefore, use `bodyOffset` as the default option.

packages/core/src/extension/iframe-resizer-methods.ts
```diff
@@ -0,0 +1,43 @@
...
+    parentIFrame: {
+        autoResize: (isEnabled: boolean) => void;
+        setHeightCalculationMethod: (id: string) => void;
+    };
+}
+
+export function isIframeResizerWindow(window: Window): window is IframeResizerWindow {
+    const iframeResizerWindow: IframeResizerWindow = window as IframeResizerWindow;
+
+    return !!iframeResizerWindow.parentIFrame;
```
animesh1987, line 11: Can't you do this
```suggestion
    return 'parentIFrame' in window;
```

packages/core/src/extension/iframe-resizer-methods.ts
```diff
@@ -0,0 +1,43 @@
...
+                    window.parentIFrame.setHeightCalculationMethod('bodyOffset');
+
+                    resolve();
+                }
+            } else {
+                setTimeout(waitingForIframeResizer, 100);
+            }
+        };
+
+        waitingForIframeResizer();
```
animesh1987, line 41: Is this triggered or useful as we return the promise earlier?

packages/core/src/extension/iframe-resizer-methods.ts
```diff
@@ -0,0 +1,43 @@
...
+                    const element = document.getElementById(taggedElementId);
+
+                    if (element) {
+                        element.setAttribute('data-iframe-height', '');
+                        window.parentIFrame.setHeightCalculationMethod('taggedElement');
+
+                        resolve();
+                    }
+
+                    reject(new Error(`Element Id not found.`));
```
animesh1987, line 30:
```suggestion
                    reject(new Error(`Element not found.`));
```

bc-peng: Updated.

packages/core/src/extension/iframe-resizer-methods.ts
bc-peng: This method is removed. No more pooling for `iframeResizer`.

packages/core/src/extension/iframe-resizer-methods.ts
bc-peng: Updated.

packages/core/src/extension/extension-service.ts
```diff
@@ -34,10 +35,23 @@ export default class ExtensionService {
...
+                    payload: { extensionId },
+                },
+                {
+                    successType: ExtensionInternalEventType.ExtensionReady,
+                    errorType: ExtensionInternalEventType.ExtensionFailed,
+                },
+            );
+        } catch (event) {
+            if (this._isExtensionFailedEvent(event)) {
+                throw new Error('Extension failed to initialize within 60 seconds');
```
bc-peng, line 52: When the host side fails to get the`iframeResizer`'s `init` message from the client side  within 60 seconds, it will send back `ExtensionInternalEventType.ExtensionFailed` and trigger this error.

packages/core/src/extension/initialize-extension-service.ts
```diff
@@ -8,14 +8,19 @@ import { ExtensionCommand, ExtensionCommandContext } from './extension-commands'
 import { ExtensionEventMap } from './extension-events';
 import { ExtensionInternalCommand } from './extension-internal-commands';
 import ExtensionService from './extension-service';
+import { iframeResizerSetup } from './iframe-resizer-setup';
 
 export interface InitializeExtensionServiceOptions {
     extensionId: string;
     parentOrigin: string;
+    taggedElementId?: string;
+    fixedHeight?: number;
```
bc-peng, line 17: Add two new optiosn.
When both are provided, we will prioritize the `fixedHeight` option.

packages/core/src/extension/iframe-resizer-methods.ts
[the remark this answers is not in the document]
bc-peng: Hi @davidchin , I have updated the PR accrodingly:

- ✅ Modify `ExtensionService#initialize` to return a Promise.
- ✅ Use the success/failed events of `IframeEventPoster` to avoid polling for `window.parentIFrame`.
- ✅ Ensure the Promise is resolved only when the host side receives `iframeResizer`'s initialization message.
- ✅ Add `taggedElementId` and `fixedHeight` to the initialization options.

packages/core/src/extension/iframe-resizer-setup.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export function isIframeResizerWindow(window: Window): window is IframeResizerWindow {
+    return 'parentIFrame' in window;
+}
+
+export function iframeResizerSetup(
+    taggedElementId: string | undefined,
+    fixedHeight: number | undefined,
+): void {
+    if (isIframeResizerWindow(window)) {
```
davidchin, line 17:
```suggestion
    if (!isIframeResizerWindow(window)) {
        return;
    }
```

Or if it's expected that `window` is an iframe resizer window, throw an exception here so it doesn't fail silently.

packages/core/src/extension/extension-service.ts
bc-peng: Do we want to proceed this further, such as refreshing the page?

packages/core/src/extension/resizable-iframe-creator.ts
```diff
@@ -41,9 +47,13 @@ export default class ResizableIframeCreator {
         // Instead, listen to the `load` inside the iframe and let the parent frame know when it happens.
         return new Promise((resolve, reject) => {
             const timeout = window.setTimeout(() => {
+                this._extensionMessenger.post(this._extensionId, {
```
davidchin, line 50: Can we just create a new instance of `IframeEventPoster` and use it to post the `ExtensionInternalEventType.ExtensionFailed` event and `ExtensionInternalEventType.ExtensionReady` event? That way we don't have to retrofit `ExtensionMessenger` to post both `ExtensionInternalEvent` and `ExtensionEvent`. That change could be problematic because callers of `ExtensionMessenger#post` method expect to post one or the other event, not both.

packages/core/src/extension/resizable-iframe-creator.ts
```diff
@@ -58,16 +68,27 @@ export default class ResizableIframeCreator {
...
                         iframe,
                     );
 
-                    teardown();
                     resolve(iframes[iframes.length - 1]);
                 }
+
+                if (
+                    typeof event.data === 'string' &&
+                    /\[iFrameSizer]iFrameResizer.*init/.test(event.data)
```
davidchin, line 84: Do you know all the possible values of `event.data`? It's better to list them out explicitly rather than using regexp.

bc-peng: The `event.data` format is a string, `\[iFrameSizer\]iFrameResizer[seq of the iframe]:[height]:[width]:init`.
`init` can be other values, such as `size`, `reset`. 
To confirm the successful initialization of `iframeResizer`, we just need to look for `init`.

davidchin: Oh okay interesting. I didn't know about that. Does `iframe-resizer` provide a public API to listen for the `init` event? For example, it seems there's `onInit` callback that we can use? (https://github.com/davidjbradshaw/iframe-resizer/blob/master/docs/parent_page/events.md#oninit)

packages/core/src/extension/extension-service.ts
animesh1987: Can we just return to the client that something failed so they can do a retry at their end?

packages/core/src/extension/extension-iframe.ts
```diff
@@ -29,8 +31,26 @@ export class ExtensionIframe {
...
+        };
+        const failedCallback = () => {
+            iframeResizerCallback(ExtensionInternalEventType.ExtensionFailed);
+        };
+
+        this._iframe = await iframeCreator.createFrame(
+            this._url.toString(),
+            this._containerId,
+            initCallback,
+            failedCallback,
```
bc-peng, line 52: This change is for testing purpose.

packages/core/src/extension/extension-service.ts
bc-peng: I have changed the error message to `Extension failed to load within 60 seconds; please reload and try again.`. A dev should be able to know what to do next. 😆

packages/core/src/extension/iframe-resizer-setup.ts
bc-peng: I've thrown an error because it's anticipated to be an iframe resizer window.

packages/core/src/extension/resizable-iframe-creator.ts
bc-peng: `ExtensionMessenger` won't be used here. A one-time use `IframeEventPoster` will be created in the callback function.

packages/core/src/extension/resizable-iframe-creator.ts
bc-peng: `initCallback` can be used as a `iframeResizer`'s initialization option.
As a result, we now rely on `iframeResizer` to notify an extension when it is ready.

davidchin (approved):
Nice work!

packages/core/src/extension/extension-iframe.ts
```diff
@@ -29,8 +31,26 @@ export class ExtensionIframe {
         }
 
         const iframeCreator = new ResizableIframeCreator();
+        const iframeResizerCallback = (type: ExtensionInternalEventType) => {
+            const poster = createExtensionEventPoster<ExtensionInternalEvent>(this._extension);
```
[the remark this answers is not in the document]
bc-peng: Thank you for the review, DC. We are reusing the same poster here. Both callback functions uses the same `iframeResizerCallback` to post a message.
```
        const iframeResizerCallback = (type: ExtensionInternalEventType) => {
            const poster = createExtensionEventPoster<ExtensionInternalEvent>(this._extension);

            poster.post({
                type,
            });
        };
        const initCallback = () => {
            iframeResizerCallback(ExtensionInternalEventType.ExtensionReady);
        };
        const failedCallback = () => {
            iframeResizerCallback(ExtensionInternalEventType.ExtensionFailed);
        };
```

PR #2192, opened 2023-09-20
feat(payment): PAYPAL-2947 added fields to mandate interface

andriiVitvitskyi1990:
## What?
Added mandateText field to order mandate interface

## Why?
To apply new feature

## Testing / Proof
...

### DIFF
--- a/packages/core/src/order/order.ts
+++ b/packages/core/src/order/order.ts
@@ -66,6 +66,9 @@ export interface GatewayOrderPayment extends OrderPayment {
     mandate?: {
         id: string;
         url?: string;
+        mandateText?: {
+            [key: string]: string;
+        };
     };
 }
 


bc-nick (approved):

packages/core/src/order/order.ts
```diff
@@ -66,6 +66,9 @@ export interface GatewayOrderPayment extends OrderPayment {
     mandate?: {
         id: string;
         url?: string;
+        mandateText?: {
```
serhii-tkachenko, line 69: I think, you can rename it to `text` since in is in `mandate` object

```suggestion
        text?: {
```

packages/core/src/order/order.ts
```diff
@@ -66,6 +66,9 @@ export interface GatewayOrderPayment extends OrderPayment {
...
+            [key: string]: string;
```
serhii-tkachenko, line 70: Why do you need and `object` with `key: value` instead of a `string`?

packages/core/src/order/order.ts
andriiVitvitskyi1990: The structure that we receive from back-end is `payments.mandate.mandateText` where `mandateText` is
```
{
account_holder_name: 'Paypal-Ratepay',
iban: '43434344',
payment_reference: 'GD78789789',
}
```

bc-peng (approved):
LGTM.👏

PR #2201, opened 2023-09-25
fix(payment): PI-758 check if Stripe container exists before mount

PavlenkoM:
## What?
Check if container exists before mounting payment fields inside.

## Why?
Because of error when payment provider was switched before current provider had been loaded

## Testing / Proof

https://github.com/bigcommerce/checkout-sdk-js/assets/9430298/18ed0b4f-a2d6-42e0-bb4f-e90fa558ecb6


serhii-tkachenko (approved):
Nice work!

animesh1987 (approved):

PR #2193, opened 2023-09-20
feat(payment): PI-517 Move Stripe UPE customer and payment strategy t…

bc-sebastianszafraniec:
…o package

There are also included changes from 4 PR’s related to StripeUPE, which were merged during work on this PR:
PI-758 - https://github.com/bigcommerce/checkout-sdk-js/pull/2201/
PI-889 - https://github.com/bigcommerce/checkout-sdk-js/pull/2213
PI-976 - https://github.com/bigcommerce/checkout-sdk-js/pull/2236
PI-924 - https://github.com/bigcommerce/checkout-sdk-js/pull/2242

Checkout-js part: https://github.com/bigcommerce/checkout-js/pull/1585

Related PR:  https://github.com/bigcommerce/checkout-sdk-js/pull/2330

## What?
Create StripevUPE package in checkout-sdk-js and move to it StripeUPE payment and customer strategies.

## Why?
For better separation between payment integrations and core Checkout SDK

## Testing / Proof
Manually tested

@bigcommerce/team-checkout @bigcommerce/team-payments

PR #2202, opened 2023-09-26
feat(payment): PI-845 Add PaymentProviderAuthentication state

bc-sebastianszafraniec:
## What?
Added PaymentProviderAuthentication state

## Why?
To be able to save provider authentication data and use those data in separate strategies.

## Testing / Proof
Manutal testing
Unit tests
Zrzut ekranu 2023-09-27 o 09 16 32

### DIFF
--- a/packages/payment-integration-api/src/index.ts
+++ b/packages/payment-integration-api/src/index.ts
@@ -138,6 +138,7 @@ export {
 } from './payment';
 export { default as PaymentIntegrationSelectors } from './payment-integration-selectors';
 export { default as PaymentIntegrationService } from './payment-integration-service';
+export { PaymentProviderAuthentication } from './payment-provider-authentication';
 export {
     BraintreeAcceleratedCheckoutCustomer,
     PaymentProviderCustomer,
--- a/packages/payment-integration-api/src/payment-provider-authentication/payment-provider-authentication.ts
+++ b/packages/payment-integration-api/src/payment-provider-authentication/payment-provider-authentication.ts
@@ -1,0 +1,3 @@
+export interface PaymentProviderAuthentication {
+    isAuthenticated?: boolean;
+}
--- a/packages/core/src/checkout/checkout-store-state.ts
+++ b/packages/core/src/checkout/checkout-store-state.ts
@@ -10,6 +10,7 @@ import { CountryState } from '../geography';
 import { OrderState } from '../order';
 import { OrderBillingAddressState } from '../order-billing-address';
 import { PaymentMethodState, PaymentState, PaymentStrategyState } from '../payment';
+import { PaymentProviderAuthenticationState } from '../payment-provider-authentication';
 import { PaymentProviderCustomerState } from '../payment-provider-customer';
 import { InstrumentState } from '../payment/instrument';
 import { RemoteCheckoutState } from '../remote-checkout';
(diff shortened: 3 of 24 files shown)


packages/payment-integration-api/src/payment-provider-authentication/payment-provider-authentication.ts
```diff
@@ -0,0 +1,3 @@
+export interface PaymentProviderAuthentication {
+    isAuthenticated?: boolean;
```
animesh1987, line 2: Is this supposed to be undefined?

serhii-tkachenko (approved):
Looks good to me

packages/payment-integration-api/src/payment-provider-authentication/payment-provider-authentication.ts
animesh1987: Can we please remove the undefined.

packages/payment-integration-api/src/index.ts
```diff
@@ -138,6 +138,7 @@ export {
 } from './payment';
 export { default as PaymentIntegrationSelectors } from './payment-integration-selectors';
 export { default as PaymentIntegrationService } from './payment-integration-service';
+export { PaymentProviderAuthentication } from './payment-provider-authentication';
 export {
     BraintreeAcceleratedCheckoutCustomer,
     PaymentProviderCustomer,
```
animesh1987, line 144: Just out of curiosity can we not use `PaymentProviderCustomer` here  for the authentication flow? 

cc @serhii-tkachenko

serhii-tkachenko: As far as I see Sebastian is not using PaymentProviderCustomer for Authentication flow. May I ask you @animesh1987 to describe what do you mean by that?

bc-sebastianszafraniec: @animesh1987  Can you please clarify what you mean?

packages/payment-integration-api/src/payment-provider-authentication/payment-provider-authentication.ts
bc-sebastianszafraniec: I can set default isAuthenticated value to false. But I am wondering as it is used only in StripeUPE, would it not be confusing to someone having it set to false by default? Then somebody might think  "Should I have this set to true to have authentication?". But maybe I am wrong?

packages/payment-integration-api/src/index.ts
animesh1987: I meant to say was that can we try and use the `PaymentProviderCustomer` for the flow, since now we are going to be having two different flows for different payment providers? Which makes checks in clients difficult, since we know at a given time we are only going to support one accelerated checkout solution.

bc-sebastianszafraniec: @animesh1987 I used PaymentProviderCustomer in this commit https://github.com/bigcommerce/checkout-sdk-js/pull/2193/commits/d40f1eda20441b3c0f3aea9f0a8c4146f0162838 
Can you please take a look? It is is ok, than this PR will be closed and I will add necessary changes in PR PI-517

animesh1987: Thanks @bc-sebastianszafraniec looks ok to me at a glance. I was expecting something similar as well. 

@serhii-tkachenko can you also take a look please.

serhii-tkachenko: Yes, we talked about this solution with @bc-sebastianszafraniec in DM. I like the way how it looks.

bc-sebastianszafraniec: I am closing this PR. Necessary changes will be added here https://github.com/bigcommerce/checkout-sdk-js/pull/2193

PR #2206, opened 2023-09-27
feat(payment): PAYPAL-2984 added an ability to change BT AXO component styling through strategy initialisation options

serhii-tkachenko:
## What?
Added an ability to change BT AXO component styling through strategy initialisation options

## Why?
Previously external developers didn't have an ability to change PayPal Connect Components styling from the UI

## Testing / Proof
Unit tests
Manual tests

Screenshot 2023-10-03 at 11 41 43

Screenshot 2023-09-29 at 16 08 43


andriiVitvitskyi1990 (approved):

packages/braintree-utils/src/index.ts
```diff
@@ -1,37 +1,8 @@
...
-    BraintreeConnectPhone,
-    BraintreeConnectProfileData,
-    onPaymentStartData,
-    StartPaymentError,
-    LocalPaymentsPayload,
-    LocalPaymentInstance,
-    BraintreeHostWindow,
-} from './braintree';
-
+export * from './braintree';
```
bc-nick, line 1: are you sure we need to export all from `./braintree`? i mean a lot of redundant types are exported too and never be used in another place

bc-dronov (approved):

packages/braintree-utils/src/index.ts
serhii-tkachenko: This `*` is returning only exported interfaces (`export interface FooBar { }`) from the file. Therefore it should provide only those interfaces that should be useful in other places, otherwise we should ask ourself: why do we export an interface that is not in use anywhere?

So yes, I think we can use `*` to export all interfaces in the file and save some time on missed interface exports for future development, while working with `./braintree` file.

bc-nick (approved):

PR #2209, opened 2023-09-29
chore(common): PAYPAL-000 updated pull request template

serhii-tkachenko:
## What?
Updated pull request template

## Why?
Because github are tagging codeowners, so it is not necessary to tag checkout and core payments team if the changes related to paypal or integration teams code

## Testing / Proof
There is nothing to test


serhii-tkachenko:
Discussed internaly with @animesh1987. We will keep pull request template as it is for now

PR #2210, opened 2023-10-04
feat(payment): PAYPAL-2507 polling mechanism removed from PPCP APM

bc-dronov:
## What?
Polling mechanism removed from PPCP APM.

## Why?
Because we don't need this method anymore.

## Testing / Proof
Screenshot 2023-10-04 at 22 04 42

### DIFF
--- a/packages/paypal-commerce-integration/src/paypal-commerce-alternative-methods/paypal-commerce-alternative-methods-payment-strategy.spec.ts
+++ b/packages/paypal-commerce-integration/src/paypal-commerce-alternative-methods/paypal-commerce-alternative-methods-payment-strategy.spec.ts
@@ -251,7 +251,6 @@ describe('PayPalCommerceAlternativeMethodsPaymentStrategy', () => {
                     label: 'pay',
                 },
                 createOrder: expect.any(Function),
-                onClick: expect.any(Function),
                 onApprove: expect.any(Function),
                 onCancel: expect.any(Function),
                 onError: expect.any(Function),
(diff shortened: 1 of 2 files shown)


vitalii-koshovyi (approved):

serhii-tkachenko:
I don't really sure that it is only one way how to resolve this issue... Why don't you move `handleClick` implementation to the `onCreateOrder` callback?

bc-dronov:
> I don't really sure that it is only way how to resolve this issue... Why don't you move `handleClick` implementation to the `onCreateOrder` callback?

Thanks for your comment! Yes, there one more solution I found and the draft looks like this:
screenshot_2023-10-03_at_17 44 32

But the main reason why I chose other solution is that `createOrder` method must return a value and only after it's done we can execute `handleClick` function. If we try to call a `handleClick` function before value is returned we could face with some async loop issues. 
For example: Calling `createOrder` method => waiting for result from backend => executing `handleClick` function => `handleClick` calling `initializePollingMechanism` => method `initializePollingMechanism` is a loop, which calls itself depends on `isOrderPending` => `isOrderPending` will be always true because we still didn't return a value from `createOrder` function => and then `initializePollingMechanism` calling again until we get out of timer

So according to that issue I decided to make a simple solution which minimally affects logic of APM's.

serhii-tkachenko:
We talked about this issue during sprint planning. May I ask you to double check if we need pooling mechanism at all? I am asking because we found that pooling mechanism is not necessary for Braintree LPMs implementation and it might use the same PayPal implementation under the hood. So it might great if we does not need polling mechanism for PP Commerce provider and we can simplify handleClick method.

serhii-tkachenko (approved):
Nice work!

PR #2216, opened 2023-10-10
feat(payment): PAYPAL-3098 added sessionId to local storage even for unrecognised guest users for Gary flow Braintree AXO

serhii-tkachenko:
## What?
Added sessionId to local storage even for unrecognised guest users for Gary flow Braintree AXO

## Why?
To call `lookup` method for each cases when the user probably might use PayPal Connect. This fix fixes the behaviour with Braintree PayPal Connect Credit Card form for guest users who leave and return to the checkout page after passing customer step.

## Testing / Proof
Unit tests
Manual tests

Before:

https://github.com/bigcommerce/checkout-sdk-js/assets/25133454/0e93b646-cbf2-4f2b-907b-9ef45b417efa

After:

https://github.com/bigcommerce/checkout-sdk-js/assets/25133454/6e16f6df-0aac-43b6-9999-fb0747be156e

### DIFF
--- a/packages/braintree-integration/src/braintree-accelerated-checkout/braintree-accelerated-checkout-utils.spec.ts
+++ b/packages/braintree-integration/src/braintree-accelerated-checkout/braintree-accelerated-checkout-utils.spec.ts
@@ -168,7 +168,7 @@ describe('BraintreeAcceleratedCheckoutUtils', () => {
             await subject.initializeBraintreeConnectOrThrow(methodId);
             await subject.runPayPalConnectAuthenticationFlowOrThrow();
 
-            expect(browserStorage.removeItem).toHaveBeenCalledWith('sessionId');
+            expect(browserStorage.setItem).toHaveBeenCalledWith('sessionId', cart.id);
             expect(paymentIntegrationService.updatePaymentProviderCustomer).toHaveBeenCalledWith({
                 authenticationState: BraintreeConnectAuthenticationState.UNRECOGNIZED,
                 addresses: [],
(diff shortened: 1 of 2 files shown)


bc-donfran:
Hey @serhii-tkachenko!

With regards to the browserstorage being used is there a reason we couldn't use cookies can't be used. Or is there any protection in place to ensure that the token is shortlived and any sensitive data in it gets removed after the period is over.

serhii-tkachenko:
> Hey @serhii-tkachenko!
> 
> With regards to the browserstorage being used is there a reason we couldn't use cookies can't be used. Or is there any protection in place to ensure that the token is shortlived and any sensitive data in it gets removed after the period is over.

Hello @bc-donfran 

We use browserstorage here to match the implementation in other places. There is no reason why do we use browserstorage instead of cookies, it is a result of the discussions that we had with checkout team.

The token is a cart id, so it should be expired right after cart expiration or the purchase complete. Also we are removing token from browser storage right after the customer clicks on "Place order" button.

bc-donfran:
Hey @serhii-tkachenko!

Thank you for a bit more context this. I think it is fine for this to happen due to the quick expiration. But in the event, where an abandon cart situation happens, how do you deal with this?

serhii-tkachenko:
> Hey @serhii-tkachenko!
> 
> 
> 
> Thank you for a bit more context this. I think it is fine for this to happen due to the quick expiration. But in the event, where an abandon cart situation happens, how do you deal with this?

Let me add a bit more context on this. We are using sessionId (cartId) to detect that the customer triggered PayPal Connect OTP before in the same session. So when the customer refresh the page, then PP OTP check email one more time (and trigger OTP) due to security reasons. 

I didn't think about the flow with abandon carts... but as far as I understand, we are using this session only for one condition.

So yeah, I understand your point in the way to use cookies instead of local storage. We can update our implementation to cookies based in separate PR, since it is not a main goal of current task.

Is it ok for you @bc-donfran

bc-donfran:
Yup that would be perfect. There are some security implications with local storage that I'm really worried about. But please do tag new on the other PR. For now, since the data in local storage is extremely short lived this should be okay for now. Thank you so much for understanding Serhii

andriiVitvitskyi1990 (approved):

bc-dronov (approved):

bc-nick (approved):

PR #2230, opened 2023-10-18
feat(payment):PI-659 added hosted form for storefront account vaulting

vitalii-koshovyi:
## What?
Added hosted form for storefront account vaulting

## Why?
Due to the need to render hosted fields on the Shopper account payments page

## Testing / Proof
https://github.com/bigcommerce/storefront-account-payments/assets/79574476/48edb7e2-4fa8-4fd7-8c58-1dcc6ff2afcc


packages/core/src/hosted-form/hosted-form-vaulting.ts
```diff
@@ -0,0 +1,42 @@
...
+    company?: string;
+    first_name: string;
+    last_name: string;
+    email: string;
+    phone?: string;
+    state_or_province_code?: string;
+}
+
+export interface HostedFormVaultingInstrumentForm {
+    billingAddress: HostedFormVaultingBillingAddress;
```
animesh1987, line 34: Is there a reason we have a mix of camel case and snake case?

packages/core/src/payment/storefront-vaulting-request-sender.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import {
+    HostedFormVaultingData,
+    HostedFormVaultingInstrumentForm,
+} from '../hosted-form/hosted-form-vaulting';
+
+export default class StorefrontVaultingRequestSender {
+    constructor(private _requestSender: RequestSender) {}
+
+    async submitPaymentMethod(
```
animesh1987, line 11: Shall we call this `submitPaymentInstrument` since payment method means something else in the context of SDK.

vitalii-koshovyi: @animesh1987 yeah, no problem. I'll rename it.

packages/core/src/hosted-form/hosted-form-vaulting.ts
vitalii-koshovyi: @animesh1987 the reason was to use common practice from the checkout SDK by naming in the camel case, but endpoint expects to get fields in the snake case. 
Yeah, I agree with you would be better to rename everything to the snake case I believe.

packages/core/src/hosted-form/iframe-content/hosted-input.ts
```diff
@@ -52,6 +54,10 @@ export default class HostedInput {
             HostedFieldEventType.SubmitRequested,
             this._paymentHandler.handle,
         );
+        this._eventListener.addListener(
+            HostedFieldEventType.VaultingRequested,
```
davidchin, line 58: Where does this event get emitted?

packages/core/src/hosted-form/iframe-content/hosted-input-events.ts
```diff
@@ -53,12 +57,25 @@ export type HostedInputEvent =
...
 }
 
+export interface HostedInputVaultingSucceededEvent {
+    type: HostedInputEventType.VaultingSucceeded;
+}
+
+export interface HostedInputVaultingErrorEvent {
+    type: HostedInputEventType.VaultingFailed;
+    payload?: {
+        errors?: string[];
```
davidchin, line 75: For payment submission, we return a rich error object to the parent window as the submission could potentially trigger human verification. I assume for vaulting, we need a similar form of protection to prevent carding attack. If that's the case, `string[]` might not be the most suitable interface. 

PS: I can't see where the vaulting submission is triggered so I can only assume we have or plan to have reCaptcha verification.

packages/core/src/hosted-form/iframe-content/hosted-input-vaulting-handler.ts
```diff
@@ -0,0 +1,62 @@
...
+                type: HostedInputEventType.VaultingFailed,
+            });
+        }
+
+        const { default_instrument, cardholder_name, ...billingAddress } = fields;
+
+        const [expiry_month, expiry_year] = values.cardExpiry ? values.cardExpiry.split('/') : [];
+
+        try {
+            await this._vaultingRequestSender.submitPaymentMethod(data, {
```
davidchin, line 40: Also, it seems we're missing the verification nonce?

packages/core/src/hosted-form/iframe-content/hosted-input-events.ts
vitalii-koshovyi: @davidchin yeah, it would be done in the future

packages/core/src/hosted-form/iframe-content/hosted-input-vaulting-handler.ts
vitalii-koshovyi: I believe this endpoint doesn't require verification nonce.

packages/core/src/hosted-form/hosted-form-vaulting.ts
vitalii-koshovyi: Yeah, great idea!

packages/core/src/hosted-form/iframe-content/hosted-input-vaulting-handler.ts
vitalii-koshovyi: So as far as I understand we need to modify endpoint and add nonce check?

packages/core/src/hosted-form/iframe-content/hosted-input.ts
vitalii-koshovyi: I’m not sure that I understood correctly about this one.
I understand that you recommended to move client logic to the SDK as we have for payment hosted fields. So in this case I need to create a vaulting/payment strategy(some separate package) to expose interface to working with hosted fields with client. Am I correct?
In this case I'll add SDK to microapp and would import only vaulting package to the microapp.

packages/core/src/hosted-form/iframe-content/hosted-input-vaulting-handler.ts
vitalii-koshovyi: It's a former endpoint, but I'm not sure if we can add captcha to it. So yeah, we'll keep it in mind.
cc @bc-akravets

packages/core/src/hosted-form/iframe-content/hosted-input.ts
davidchin: > In this case I'll add SDK to microapp and would import only vaulting package to the microapp.

Yes, that'd be my recommendation. Export an interface from the SDK. When it's called, it'll emit the `HostedFieldEventType.VaultingRequested` event to submit the form to the vaulting endpoint.

packages/core/src/hosted-form/hosted-form-vaulting.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    detach(): void {
+        this._eventListener.stopListen();
+
+        this._fields.forEach((field) => {
+            field.detach();
+        });
+    }
+
+    async submit(payload: {
```
davidchin, line 90: It looks like the fields don't have a separate class implementation. Instead, they expose a separate `submit` method for posting to the vaulting endpoint. Could the form also take a similar approach so that the design is consistent with the fields (e.g.: `submitStoredCard`)? We could revisit the design in the future if there's a greater need to create entirely separate implementations for vaulting.

packages/core/src/hosted-form/create-hosted-form-vaulting-service.ts
```diff
@@ -0,0 +1,13 @@
...
+import HostedFormVaultingService from './hosted-form-vaulting-service';
+
+/**
+ * Creates an instance of `HostedFormVaultingService`.
+ *
+ *
+ * @param host - Host url string parameter.
+ * @returns An instance of `HostedFormVaultingService`.
+ */
+export default function createHostedFormVaultingService(host: string) {
```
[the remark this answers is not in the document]
vitalii-koshovyi: ✅

packages/core/src/hosted-form/hosted-form-vaulting.ts
vitalii-koshovyi: It makes sense! 👍
Previously I thought to do it in this way, but somehow I though that it is a bad idea to create a checkout store instance just for creation of the hosted form. But after your comment I re-though it and there is [nothing wrong with it ](https://github.com/bigcommerce/checkout-sdk-js/pull/2230/files#diff-6a1d91e54da7e5712a131b154877196ab8f7a45fc918b26d26321671005b607eR14). So I [added it to the HostedForm class](https://github.com/bigcommerce/checkout-sdk-js/pull/2230/files#diff-40e7580dc4c696da6802c1ee6190abb57a12f19ce78664a5430915d0a92e8756R118) to avoid entirely separate implementations for vaulting.

packages/core/src/hosted-form/create-hosted-form-stored-card-service.ts
```diff
@@ -0,0 +1,17 @@
...
+import HostedFormStoredCardService from './hosted-form-stored-card-service';
+
+/**
+ * Creates an instance of `HostedFormStoredCardService`.
+ *
+ *
+ * @param host - Host url string parameter.
+ * @returns An instance of `HostedFormStoredCardService`.
+ */
+export default function createHostedFormStoredCardService(host: string) {
```
davidchin, line 13:
```suggestion
export default function createStoredCardHostedFormService(host: string) {
```

I think it reads more naturally this way.

packages/core/src/hosted-form/hosted-form-stored-card-service.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import HostedForm from './hosted-form';
+import HostedFormFactory from './hosted-form-factory';
+import HostedFormOptions from './hosted-form-options';
+import {
+    HostedFormStoredCardData,
+    HostedFormStoredCardInstrumentFields,
+} from './hosted-form-stored-card-type';
+
+export default class HostedFormStoredCardService {
```
davidchin, line 11:
```suggestion
export default class StoredCardHostedFormService {
```

davidchin (approved):
Really nice work. I just have one more comment regarding the naming. But other than that, LGTM!

PR #2232, opened 2023-10-18
fix(payment): PI-924 Add loader while waiting order confirmation from Stripe

PavlenkoM:
## What?
Add polling mechanism to wait while Stripe will confirm order by webhook

## Why?
Polling mechanism is needed in case when payment was successfully confirmed by Stripe.confirmPayment request on FE side, but has some delay to confirm payment on BE side, and BC order status mismatch with Stripe order status.
In this case the second payment request returns an error and we need to add some delay to wait confirmation by webhook on BE side and wait while BC order status will be changed.

PR for checkout-js: [https://github.com/bigcommerce/checkout-js/pull/1564](https://github.com/bigcommerce/checkout-js/pull/1564)

## Testing / Proof
Orders status updates immediately.

https://github.com/bigcommerce/checkout-sdk-js/assets/9430298/add170ab-8947-4f92-a291-4352226838d8

Order confirmed by Stripe but has some delay to be updated on BC side

https://github.com/bigcommerce/checkout-sdk-js/assets/9430298/030e2b39-c2ea-4e77-bdef-b7a0c270111e

Order was confirmed by Stripe but BC order status wasn't change.

https://github.com/bigcommerce/checkout-sdk-js/assets/9430298/d687e414-ce28-4f69-8e53-2b3c8115317e

### DIFF
--- a/packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
+++ b/packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
@@ -1,6 +1,7 @@
 import { includes, some } from 'lodash';
 
 import { PaymentMethodFailedError } from '@bigcommerce/checkout-sdk/payment-integration-api';
+import { LoadingIndicator } from '@bigcommerce/checkout-sdk/ui';
 
 import { isHostedInstrumentLike } from '../..';
 import { Address } from '../../../address';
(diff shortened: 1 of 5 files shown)


packages/core/src/payment/strategies/stripe-upe/stripe-upe-payment-strategy.ts
```diff
@@ -664,4 +678,69 @@ export default class StripeUPEPaymentStrategy implements PaymentStrategy {
         stripeElement.mount(`#${containerId}`);
         this._isMounted = true;
     }
+
+    private async _checkOrderStatus(): Promise<string | undefined> {
```
vitalii-koshovyi, line 682: I believe we need to cover it with unit tests

PavlenkoM: I added unit test for new logic

PR #2234, opened 2023-10-19
feat(payment): PAYPAL-2877 Add deviceSessionId to submitPayment

bc-nick:
## What?

Add deviceSessionId to submitPayment

## Why?

To do this we need to pass deviceSessionId and make sure that the device data captured in the transaction journal.

## Testing / Proof

Screenshot 2023-10-19 at 17 04 44

### DIFF
--- a/packages/google-pay-integration/src/gateways/google-pay-braintree-gateway.spec.ts
+++ b/packages/google-pay-integration/src/gateways/google-pay-braintree-gateway.spec.ts
@@ -186,8 +186,8 @@ describe('GooglePayBraintreeGateway', () => {
         });
     });
 
-    describe('#getDeviceSessionId', () => {
-        it('get deviceSessionId', async () => {
+    describe('#extraPaymentData', () => {
+        it('get extraPaymentData', async () => {
             await googlePayBraintreeGateway.initialize(getBraintree);
 
             const extraData = await googlePayBraintreeGateway.extraPaymentData();


packages/google-pay-integration/src/google-pay-payment-strategy.ts
```diff
@@ -69,11 +69,12 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
         await this._paymentIntegrationService.submitOrder();
 
         const nonce = await this._googlePayPaymentProcessor.getNonce(payment.methodId);
+        const deviceSessionId = await this._googlePayPaymentProcessor.getDeviceSessionId();
 
         try {
             await this._paymentIntegrationService.submitPayment({
                 ...payment,
-                paymentData: { nonce },
+                paymentData: { nonce, deviceSessionId },
```
serhii-tkachenko, line 77: I am wondering if we have another payment data param in one of the other provider, then we should create almost the same as you did in this PR. On the other hand, we can think about creating generic method to avoid some problems in the future.

My idea is to create a method, that will be responsible to get all provider specific payment data in one method like: getPaymentData, that you can simply adjust in different providers.

```
const extraPaymentData = await this._googlePayPaymentProcessor.getPaymentData() || {};

...
paymentData: { nonce, ...extraPaymentData },
...
```

What do you think?

packages/google-pay-integration/src/google-pay-payment-strategy.ts
```diff
@@ -69,11 +69,12 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
         await this._paymentIntegrationService.submitOrder();
 
         const nonce = await this._googlePayPaymentProcessor.getNonce(payment.methodId);
+        const extraData = await this._googlePayPaymentProcessor.getExtraData();
```
serhii-tkachenko, line 72: It might be better to update the name to `extraPaymentData` to understand where it should be used

packages/google-pay-integration/src/google-pay-payment-processor.ts
```diff
@@ -63,6 +63,16 @@ export default class GooglePayPaymentProcessor {
...
 
+    async getDeviceSessionId(): Promise<string | undefined> {
+        return this._gateway.getDeviceSessionId();
+    }
+
+    async getExtraData() {
+        return {
+            deviceSessionId: await this._gateway.getDeviceSessionId(),
+        };
+    }
```
serhii-tkachenko, line 74: I don't really sure that this is a right way... getDeviceSessionId is relevant only for Braintree gateway, so I don't see any arguments why should we keep this method in the processor file.

```suggestion
    getExtraPaymentData() {
        return this._gateway.getExtraPaymentData();
    }
```

We as integration developers should be able to provide specific to our provider payment data (like formattedData for example), so the data should be build on gateway side, but not in the abstract class like this place.

Also, there should be an interface of the returned method's value.

packages/google-pay-integration/src/gateways/google-pay-gateway.ts
```diff
@@ -119,6 +119,10 @@ export default class GooglePayGateway {
         return Promise.resolve(nonce);
     }
 
+    getDeviceSessionId(): Promise<undefined | string> {
+        return Promise.resolve(undefined);
+    }
```
serhii-tkachenko, line 124: Don't think we need this method here

packages/google-pay-integration/src/gateways/google-pay-braintree-gateway.ts
```diff
@@ -101,6 +101,10 @@ export default class GooglePayBraintreeGateway extends GooglePayGateway {
         return nonce;
     }
 
+    async getDeviceSessionId() {
+        return this._getBraintreeDeviceData();
+    }
```
serhii-tkachenko, line 106: This method should be changed to:
```suggestion
    async getExtraPaymentData() {
        return {
            deviceSessionId: await this._getBraintreeDeviceData();
        };
    }
```

serhii-tkachenko (approved):

animesh1987 (approved):

PR #2236, opened 2023-10-24
feat(payment): PI-976 Stripe UPE payment payload refactoring

PavlenkoM:
## What?
Add single method for payment request payload

## Why?
Refactoring and types fixes

## Testing / Proof
Manually tested and unit test

### DIFF
--- a/packages/core/src/payment/payment.ts
+++ b/packages/core/src/payment/payment.ts
@@ -211,7 +211,8 @@ interface StripeV3Intent {
     confirm: boolean;
 }
 
-interface StripeUPEIntent {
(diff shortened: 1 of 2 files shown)


serhii-tkachenko (approved):

animesh1987 (commented):
Changes look good, do we need to update the tests or add new tests to cover the change?

PavlenkoM:
> Changes look good, do we need to update the tests or add new tests to cover the change?

I don't think that we need to add tests, because logic didn't change, only removed duplication, so if nothing is broken - the same tests should pass

animesh1987 (approved):

PR #2241, opened 2023-10-30
feat(payment): PI-846 Quickly selecting different payment methods on …

bc-sebastianszafraniec:
…Payment step of Checkout sometimes results in error

## What?
Check if container exists before mounting payment fields inside.

## Why?
Quickly selecting different payment methods on Payment step of Checkout sometimes results in an error.

## Testing / Proof
I was not able to reproduce the bug on dev or int store. Fix must be tested on merchant store.

### DIFF
--- a/packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.ts
+++ b/packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.ts
@@ -309,7 +309,7 @@ export default class Adyenv3PaymentStrategy implements PaymentStrategy {
                 );
             }
 
-            additionalActionComponent.mount(`#${containerId}`);
+            this._mountElement(additionalActionComponent, containerId);
 
             if (onLoad) {
                 onLoad(() => {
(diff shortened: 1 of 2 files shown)


packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.ts
```diff
@@ -482,4 +485,12 @@ export default class Adyenv3PaymentStrategy implements PaymentStrategy {
 
         return errors;
     }
+
+    private _mountElement(adyenComponent: AdyenComponent, containerId: string): void {
+        if (!document.getElementById(containerId)) {
+            return;
```
serhii-tkachenko, line 491: Do we need to show something to users to let them know why don't they see the form (or the element)?

bc-sebastianszafraniec: At this moment the error which is occurring comes from error in Adyen `mount` method which is called here :
https://github.com/bigcommerce/checkout-sdk-js/blob/b875ed2d0959641dc2e88ffcc073759e5a1970ae/packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.ts#L413
`Mount` method is looking for container and when container is not found then customer see error from catch block below.
The reported bug occurs when clicking Adyen payment method and then immediately click on other provider payment method. With such a quick clicks the Adyen payment fields will not be loaded anyway so there is no reason to inform customer about it.

serhii-tkachenko (approved):

vitalii-koshovyi (approved):

PR #2243, opened 2023-11-01
refactor(payment): PAYPAL-3164 updated paypal sdk script loading in a new way

serhii-tkachenko:
## What?
We changed a way how we load paypal sdk script to try to avoid zoid issues that occur due to the multiple script loading with different configurations.

## Why?
Prev implementation had extra step with loading paypal script from CDN that we used to trigger paypal sdk script loading process. So the reason why we update that is to reduce amount of steps to get paypal sdk working properly, plus it might reduce amount of time spent on wallet buttons rendering process.

## Testing / Proof
Unit tests
Manual tests
QA regression


andriiVitvitskyi1990 (approved):

bc-nick (approved):

bc-donfran (approved):
Looks good from a security persepective!

PR #2252, opened 2023-11-08
fix(checkout): Prevent Vite From Creating Syntax Errors

### LINKED ISSUE #2138: Unexpected token process.env.NODE_ENV
I'm currently using Vite, but during rollup im getting an unexpected token error. This is regarding the warning text in file create-checkout-service.ts on line 95.

```javascript
getDefaultLogger().warn(
    'Note that the development build is not optimized. To create a production build, set process.env.NODE_ENV to `production`.',
);
```

When I remove the  "process.env.NODE_ENV" text it will build just fine.

Magnesium38:
## What?
This adds a[ zero-width space character](https://en.wikipedia.org/wiki/Zero-width_space) into the `process.env.NODE_ENV` warning text so that it is not statically replaced. This is how the [Vite documentation](https://vitejs.dev/guide/env-and-mode.html#production-replacement) recommends fixing this issue (see the first bullet point under "Production Replacement").

Fixes #2138 

## Why?
When using Vite, env variables like `process.env.NODE_ENV` are statically replaced, even inside of strings. This means that that instead of a well-formed warning string being produced, invalid Javascript is actually created.

image

## Testing / Proof

https://stackblitz.com/edit/vitejs-vite-t7xh32?file=main.js

Baseline reproduction of the issue. If you use devtools you can see the relevant error being thrown and jumping into the source you'll see the malformed replacement as seen in the screenshot above.

### DIFF
--- a/packages/core/src/checkout/create-checkout-service.ts
+++ b/packages/core/src/checkout/create-checkout-service.ts
@@ -92,7 +92,7 @@ export default function createCheckoutService(options?: CheckoutServiceOptions):
 
     if (getEnvironment() !== 'production') {
         getDefaultLogger().warn(
-            'Note that the development build is not optimized. To create a production build, set process.env.NODE_ENV to `production`.',
+            'Note that the development build is not optimized. To create a production build, set process\u200b.env.NODE_ENV to `production`.',
         );
     }
 


bc-0dp:
@Magnesium38 I think there might also be something going on here:
`Uncaught SyntaxError: missing ) after argument list[@bigcommerce_checkout-sdk.js:47037:254](http://localhost:5173/node_modules/.vite/deps/@bigcommerce_checkout-sdk.js?t=1699520596263&v=0a5e6856)`

Magnesium38:
@bc-0dp That's the same error that this is aiming to fix!

It might be easier to see if we ask github to attempt to syntax highlight it as typescript. I've also taken the long line and run it through prettier to make it easier to see where the statements are.

```ts
"https:" !== document.location.protocol &&
  R().warn(
    "The BigCommerce Checkout SDK should not be used on a non-HTTPS page",
  ),
  "production" !== A() &&
    R().warn(
      "Note that the development build is not optimized. To create a production build, set "development" to `production`.",
    );
```

You can see that it's creating a string ` "Note that the development build is not optimized. To create a production build, set "` and then there's the text `development`, and then another string ``" to `production`."`` It's illegal JS to have a string and an identifier next to each other, which is why it's saying something about a missing `)`.

With the fix in place, that same line would produce the following, which Github can syntax highlight correctly as it's valid JS.

```ts
"https:" !== document.location.protocol &&
  R().warn(
    "The BigCommerce Checkout SDK should not be used on a non-HTTPS page",
  ),
  "production" !== A() &&
    R().warn(
      "Note that the development build is not optimized. To create a production build, set process\u200b.env.NODE_ENV to `production`.",
    );
```

Thanks for responding so quick though. Let me know if there's anything I could do to help this get this landed faster!

bc-0dp:
@Magnesium38 I am unfortunately not one of the maintainers but I pinged @bigcommerce/team-checkout and a member will have a look.

Once again, thanks a bunch! 

Community ❤️

animesh1987 (approved):

animesh1987:
Hey @Magnesium38 thanks for putting this fix in, change looks good. Unfortunately we are in a release freeze + stabilisation period for a couple of weeks, will get it merged as soon as we can.

bc-0dp:
@animesh1987 can we review this one now we're out of release freeze?

animesh1987 (approved):

animesh1987:
@bc-0dp lost track of this, thanks for the ping. Merged.

PR #2253, opened 2023-11-10
[POC] feat(payment): PAYPAL-3229 added PaymentProvidersTracker for analytics

serhii-tkachenko:
## What?
Added PaymentProvidersTracker for analytics to cover Analytic gaps that BC can't cover on BE side.

## Why?
As of now PayPal is not aware if any other than PayPal Credit Card/Connect payment method is used in their conversion tracking. PayPal is interested in detecting other than CC methods switch, so called Alternative Payment Methods switch to exclude any attempt to use an APM from PayPal Connect conversion calculations. 

## Testing / Proof
TBD


packages/core/src/analytics/payment-providers-tracker/create-payment-providers-tracker.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+export default function createPaymentProvidersTracker(
+    checkoutService: CheckoutService,
+): PaymentProvidersTrackerService {
+    const state = checkoutService.getState();
+    const paymentMethods = state.data.getPaymentMethods() || [];
+
+    console.log('createPaymentProvidersTracker paymentMethods', paymentMethods);
+
+    if (paymentMethods?.length > 0) {
```
serhii-tkachenko, line 30: Don't really sure on this approach, but we will test it later

packages/core/src/analytics/payment-providers-tracker/noop-payment-providers-tracker.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export default class NoopPaymentProvidersTracker implements PaymentProvidersTrackerService {
+    checkoutBegin(): void {}
+    clickPayButton(): void {}
+    customerEmailEntry(): void {}
+    customerSuggestionInit(): void {}
+    customerSuggestionExecute(): void {}
+    customerPaymentMethodExecuted(): void {}
+    exitCheckout(): void {}
+    orderPurchased(): void {}
```
BC-ACherednichenko, line 11: Do we really need all these methods/events? or maybe we can focus on the `selectedPaymentMethod` only?

serhii-tkachenko: This is a mock for PaymentProvidersTracker (a copy of bodl service UI part), so we can use this methods in the future. Also for BraintreeConnect tracker we will use only 2 methods: `selectPaymentOption`, `walletButtonClick` events

serhii-tkachenko: So in our case we have a focus on what PayPal asked for

packages/core/src/analytics/payment-providers-tracker/payment-providers-tracker.ts
```diff
@@ -0,0 +1,77 @@
...
+
+import PaymentProvidersTrackerService from './payment-providers-tracker-service';
+import { BraintreeConnectTracker } from './trackers';
+import { EventsPayload } from './types';
+
+export default class PaymentProvidersTracker implements PaymentProvidersTrackerService {
+    private braintreeConnectTracker?: BraintreeConnectTracker;
+
+    constructor () {
+        this.braintreeConnectTracker = new BraintreeConnectTracker();
```
animesh1987, line 11: Can we have a tracker service per provider? Since here we have the `braintreeConnectTracker` which means this could be tied to BT, but in future we could have different `tracker` and then we will end up adding conditional logic possibly.

packages/core/src/analytics/payment-providers-tracker/payment-providers-tracker.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    paymentRejected() {
+        console.log('PaymentProvidersTracker: paymentRejected');
+    }
+
+    selectedPaymentMethod(paymentOption?: string) {
+        console.log('PaymentProvidersTracker: selectedPaymentMethod', paymentOption);
+
+        this.braintreeConnectTracker?.selectedPaymentMethod(paymentOption);
```
animesh1987, line 57: Same as pervious comment, not sure if this is the right place.

packages/core/src/analytics/payment-providers-tracker/noop-payment-providers-tracker.ts
animesh1987: I agree with @BC-ACherednichenko that we should start with the basic methods now and not copy BODL here. Maybe we can keep the interface to only two methods for now.

packages/core/src/analytics/payment-providers-tracker/payment-providers-tracker.ts
serhii-tkachenko: I think we might have trackers for different providers in the future. So I created abstract class `PaymentProvidersTracker` that can call other trackers whenever it needed.

```
this.braintreeConnectTracker = new BraintreeConnectTracker();
this.paypalCommerceConnectTracker = new PaypalCommerceConnectTracker();
this.stripeTracker = new StripeTracker();
```

PR #2256, opened 2023-11-16
feat(payment): PI-853 BluesnapDirect added Sepa payment strategy

vitalii-koshovyi:
## What?
 BluesnapDirect SEPA implemantation
https://github.com/bigcommerce/checkout-js/pull/1545
## Why?
Due the Bluesnap requirement

## Testing / Proof
image

### DIFF
--- a/packages/core/src/order/order-request-body.ts
+++ b/packages/core/src/order/order-request-body.ts
@@ -1,5 +1,6 @@
 import {
-    BlueSnapDirectEcpInstrument,
+    EcpInstrument,
+    SepaInstrument,
     WithAccountCreation,
     WithBankAccountInstrument,
 } from '@bigcommerce/checkout-sdk/payment-integration-api';
@@ -40,13 +41,14 @@ export default interface OrderRequestBody {
 
 export type OrderPaymentInstrument =
     | WithBankAccountInstrument
+    | EcpInstrument
+    | SepaInstrument
     | CreditCardInstrument
     | HostedInstrument
     | HostedCreditCardInstrument
     | HostedVaultedInstrument
     | NonceInstrument
     | VaultedInstrument
-    | BlueSnapDirectEcpInstrument
     | (CreditCardInstrument & WithDocumentInstrument)
     | (CreditCardInstrument & WithCheckoutcomFawryInstrument)
     | (CreditCardInstrument & WithCheckoutcomSEPAInstrument)
--- a/packages/bluesnap-direct-integration/src/bluesnap-direct-sepa-payment-strategy.spec.ts
+++ b/packages/bluesnap-direct-integration/src/bluesnap-direct-sepa-payment-strategy.spec.ts
@@ -1,0 +1,113 @@
+import {
+    OrderFinalizationNotRequiredError,
+    OrderRequestBody,
+    PaymentArgumentInvalidError,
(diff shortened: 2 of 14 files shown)


packages/core/src/order/order-request-body.ts
```diff
@@ -47,6 +48,7 @@ export type OrderPaymentInstrument =
     | NonceInstrument
     | VaultedInstrument
     | BlueSnapDirectEcpInstrument
+    | BlueSnapDirectSepaInstrument
```
animesh1987, line 51: How come we are adding payment method specific types here?

bc-peng: Yes, it's concerning. Can we find a workaround?

vitalii-koshovyi: I believe BluesnapEcp was first one, I've added SEPA by inertia🌝
Yeah, I'll check how we can improve situation here. 
Thanks for pointing out🙌

serhii-tkachenko (approved):
Nice work!

bc-peng (approved):
LGTM.

PavlenkoM (approved):

packages/core/src/order/order-request-body.ts
vitalii-koshovyi: @animesh1987  @bc-peng  guys, sorry my workaround was wrong, I've just tuned a bit type guard.
We still need payment method specific type in order request body. And I don't see suitable workaround at the moment. It already has CheckoutCom, Mollie, Paypal, Bolt specific payment instruments.

bc-peng: Perhaps the naming makes it confusing.
The actual content of `BlueSnapDirectSepaInstrument` is:
```
export interface BlueSnapDirectSepaInstrument {
    firstName: string;
    lastName: string;
    iban: string;
    shopperPermission: boolean;
}
```
We don't have a similar type before because `iban` rarely appears in the work.

Could we create a more reusable interface here that supports `iban`?

For example, not sure `IbanInstrument` is OK, but you'll get my idea.
```
export interface IbanInstrument {
    firstName: string;
    lastName: string;
    iban: string;
    shopperPermission: boolean;
}

export interface BlueSnapDirectSepaPayload {
    sepa_direct_debit: IbanInstrument;
}
```

vitalii-koshovyi: @bc-peng thank you for your input, I'll add requested changes.
I'm fully agree with you that we need to have generic interfaces for such payment methods, but it's not always possible😢

I want to align general approach with you guys.
We already have SEPA implementation for the checkout.com and the interface is a bit different here from what we need to have in the Bluesnap Sepa implementation: https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/payment-integration-api/src/payment/payment.ts#L104.

cc @animesh1987

animesh1987: Thanks for trying different approaches @vitalii-koshovyi, can we try and formalise the same way for Bluesnap as `checkoutcom`.  I suppose this has some time pressure so seems ok, but we need to come up with better ways to consolidate the interfaces. 
Maybe use generics. 

@serhii-tkachenko do you want to have a review and we can work on it in the doc that you had created?

PR #2258, opened 2023-11-21
feat(payment): CHECKOUT-7901 Add method to load button payment method config in parallel

animesh1987:
## What?
- Add method to load button payment method config in parallel
- Refactor exisiting payment strategies to leverage pre-fetched configs

## Why?
Right now we have method to fetch button payment method config one at a time in sequence, which are blocking and cause a delay in rendering the buttons. 

Add a method to fetch button configs in parallel so buttons can be rendered quicker.

## Testing / Proof
- CI
- Manual
 
### DIFF
--- a/packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
+++ b/packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
@@ -76,11 +76,14 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
         this._onAuthorizeCallback = onPaymentAuthorize;
         this._onError = onError;
 
-        await this._paymentIntegrationService.loadPaymentMethod(methodId);
+        let state = this._paymentIntegrationService.getState();
 
-        const state = this._paymentIntegrationService.getState();
-
-        this._paymentMethod = state.getPaymentMethodOrThrow(methodId);
+        try {
+            this._paymentMethod = state.getPaymentMethodOrThrow(methodId);
+        } catch (_e) {
+            state = await this._paymentIntegrationService.loadPaymentMethod(methodId);
+            this._paymentMethod = state.getPaymentMethodOrThrow(methodId);
+        }
 
         await this._paymentIntegrationService.verifyCheckoutSpamProtection();
 
--- a/packages/braintree-integration/src/braintree-paypal/braintree-paypal-customer-strategy.ts
+++ b/packages/braintree-integration/src/braintree-paypal/braintree-paypal-customer-strategy.ts
@@ -62,7 +62,11 @@ export default class BraintreePaypalCustomerStrategy implements CustomerStrategy
             );
         }
 
+        console.log('going to init: SDK load method', methodId);
+
         await this.paymentIntegrationService.loadPaymentMethod(methodId);
+
+        console.log('going to init: SDK load method finished', methodId);
 
         const state = this.paymentIntegrationService.getState();
         const storeConfig = state.getStoreConfigOrThrow();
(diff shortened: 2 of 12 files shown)


vitalii-koshovyi (approved):
Well done!

vitalii-koshovyi:
but need to fix tests🤗

bc-nick (approved):
LGTM

packages/braintree-integration/src/braintree-paypal/braintree-paypal-customer-strategy.ts
```diff
@@ -62,12 +62,17 @@ export default class BraintreePaypalCustomerStrategy implements CustomerStrategy
...
         }
 
-        await this.paymentIntegrationService.loadPaymentMethod(methodId);
+        let state = this.paymentIntegrationService.getState();
+        let paymentMethod: PaymentMethod<BraintreeInitializationData>;
+
+        try {
+            paymentMethod = state.getPaymentMethodOrThrow(methodId);
+        } catch (_e) {
+            state = await this.paymentIntegrationService.loadPaymentMethod(methodId);
```
bc-peng, line 71: It seems like we'll send another request for the same method if `getPaymentMethodOrThrow(methodId)` is executed in the middle of the first try.
Perhaps we can avoid this by only initializing a method after `loadPaymentMethodByIds` is done in `checkout-js`. `loadPaymentMethod(methodId)` should read from checkout store whenever possible.

bc-peng: As a result, we don't need to change the existing payment methods if we make `loadPaymentMethod` do nothing when a method is already loaded.

packages/core/src/payment/payment-method-action-creator.ts
```diff
@@ -21,6 +21,47 @@ const APPLEPAYID = 'applepay';
...
+                            ),
+                        );
+                        observer.complete();
+                    })
+                    .catch((response) => {
+                        observer.error(
+                            createErrorAction(
+                                PaymentMethodActionType.LoadPaymentMethodsFailed,
+                                response,
+                            ),
```
bc-peng, line 59: We just noticed this error handling will change the current UX when loading a payment method fails.
It appears that when one method fails, all methods cannot work properly, right?

packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
```diff
@@ -76,11 +76,14 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
...
-
-        const state = this._paymentIntegrationService.getState();
+        let state = this._paymentIntegrationService.getState();
 
-        this._paymentMethod = state.getPaymentMethodOrThrow(methodId);
+        try {
+            this._paymentMethod = state.getPaymentMethodOrThrow(methodId);
+        } catch (_e) {
+            state = await this._paymentIntegrationService.loadPaymentMethod(methodId);
+            this._paymentMethod = state.getPaymentMethodOrThrow(methodId);
```
bc-peng, line 85: What if it fails again?

packages/core/src/payment/payment-method-action-creator.ts
animesh1987: Yes, for this when one fails all might need to be failed. Isn't the experience similar right now?

packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
animesh1987: Then should be similar experience as of now, as this statement wasn't wrapped in try catch before as well.

packages/core/src/payment/payment-method-action-creator.ts
bc-peng: The current implementation handles each failure separately and adjusts the layout accordingly.

animesh1987: I think this should work similar way, since we still have the fallback in each strategy so that effectively works as a retry.

bc-peng:
If this PR is going to break `checkout-js` e2e tests, we may want to work out the adoption there before merging this one, otherwise, all following SDK PRs will be blocked at `checkout-js`.

bc-peng (approved):

PR #2277, opened 2023-12-06
feat(payment): PAYPAL-3337 PPCP GooglePay setGatewayIdentifier added

bc-dronov:
## What?
We need to add dynamic gateway name from PPCP api: `allowedPaymentMethods[0].tokenizationSpecification.parameters.gateway`. 

## Why?
Because now it works only in sandbox mode because of hardcoded `paypalsb` name.

## Testing / Proof
All tests have been passed

### DIFF
--- a/packages/google-pay-integration/src/google-pay-paypal-commerce/google-pay-paypal-commerce-gateway.ts
+++ b/packages/google-pay-integration/src/google-pay-paypal-commerce/google-pay-paypal-commerce-gateway.ts
@@ -54,6 +54,12 @@ export default class GooglePayPaypalCommerceGateway extends GooglePayGateway {
         await this.paypalCommerceScriptLoader.getPayPalSDK(paymentMethod, currency);
 
         this.googlepayConfig = await this.paypalCommerceScriptLoader.getGooglePayConfigOrThrow();
+
+        const { allowedPaymentMethods } = this.googlepayConfig;
+
+        this.setGatewayIdentifier(
+            allowedPaymentMethods[0]?.tokenizationSpecification?.parameters?.gateway,
+        );
     }
 
     getPaymentGatewayParameters(): GooglePayPayPalCommerceGatewayParameters {
--- a/packages/google-pay-integration/src/gateways/google-pay-gateway.ts
+++ b/packages/google-pay-integration/src/gateways/google-pay-gateway.ts
@@ -236,6 +236,10 @@ export default class GooglePayGateway {
         return this._gatewayIdentifier;
     }
 
+    protected setGatewayIdentifier(gateway?: string) {
+        this._gatewayIdentifier = gateway || this.getGatewayIdentifier();
+    }
+
     private _isShippingAddressRequired(): boolean {
         const { getCartOrThrow, getStoreConfig, getShippingAddress } =
             this._paymentIntegrationService.getState();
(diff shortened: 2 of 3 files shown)


packages/google-pay-integration/src/google-pay-paypal-commerce/google-pay-paypal-commerce-gateway.ts
```diff
@@ -54,6 +55,13 @@ export default class GooglePayPaypalCommerceGateway extends GooglePayGateway {
         await this.paypalCommerceScriptLoader.getPayPalSDK(paymentMethod, currency);
 
         this.googlepayConfig = await this.paypalCommerceScriptLoader.getGooglePayConfigOrThrow();
+
+        const { allowedPaymentMethods } = this.googlepayConfig;
+
+        this.setGatewayIdentifier(
+            allowedPaymentMethods[0]?.tokenizationSpecification?.parameters?.gateway ||
+                this.defaultGateway,
```
bc-nick, line 63: We can avoid applying `this.defaultGateway` variable here if you make some update for `setGatewayIdentifier` method:

```
protected setGatewayIdentifier(gateway?: string) {
    this._gatewayIdentifier = gateway || this._gatewayIdentifier;
}
```

What do you think?

andriiVitvitskyi1990 (approved):

packages/google-pay-integration/src/google-pay-paypal-commerce/google-pay-paypal-commerce-gateway.ts
bc-dronov: @bc-nick Thanks for comment! But I would like this method to keep a simple `setter` logic. Especially it's a parent class.
I think we can keep logic with default variables outside of this method.

bc-nick: @bc-dronov you already set this variable here https://github.com/bigcommerce/checkout-sdk-js/blob/de326bd05ee167b5636ee87e78d96abd1312ee86/packages/google-pay-integration/src/google-pay-paypal-commerce/google-pay-paypal-commerce-gateway.ts#L29

so you do not need to pass it to `setGatewayIdentifier` because parent class already has it

bc-dronov: @serhii-tkachenko Agree, we should don't forget to keep it simple.

bc-dronov: @bc-nick I think we don't need to change parent's method logic because of child's class. If it would be possible, I would call the `super` method like:
`super(this.defaultGateway, service)`
But unfortunately `super` must be called before accessing  `this`.

bc-nick: To make it more easily there is another way. @bc-dronov you can do the same logic here https://github.com/bigcommerce/checkout-sdk-js/blob/de326bd05ee167b5636ee87e78d96abd1312ee86/packages/google-pay-integration/src/google-pay-paypal-commerce/google-pay-paypal-commerce-gateway.ts#L74

For example: 

```
return {
            gateway: allowedPaymentMethods[0]?.tokenizationSpecification?.parameters?.gateway || this.getGatewayIdentifier(),
            gatewayMerchantId,
        };
```

serhii-tkachenko (approved):

PR #2294, opened 2023-12-20
feat(payment): PI-1126 TD Bank 3DS flow implementation

PavlenkoM:
## What?
TD Bank 3DS flow

## Why?
Add 3DS challenge for TD Bank payment startegy

## Testing / Proof
Unit test and manually tested

### DIFF
--- a/packages/td-bank-integration/src/td-online-mart-payment-strategy.ts
+++ b/packages/td-bank-integration/src/td-online-mart-payment-strategy.ts
@@ -1,3 +1,5 @@
+import { FormPoster } from '@bigcommerce/form-poster';
+
 import {
     InvalidArgumentError,
     isHostedInstrumentLike,
@@ -16,7 +18,13 @@ import {
     PaymentStrategy,
 } from '@bigcommerce/checkout-sdk/payment-integration-api';
 
-import { FieldType, TDCustomCheckoutSDK, TdOnlineMartElement } from './td-online-mart';
+import { isTdOnlineMartAdditionalAction } from './isTdOnlineMartAdditionalAction';
+import {
+    FieldType,
+    TDCustomCheckoutSDK,
+    TdOnlineMartElement,
+    TdOnlineMartThreeDSErrorBody,
+} from './td-online-mart';
 import TDOnlineMartScriptLoader from './td-online-mart-script-loader';
 
 export default class TDOnlineMartPaymentStrategy implements PaymentStrategy {
@@ -28,6 +36,7 @@ export default class TDOnlineMartPaymentStrategy implements PaymentStrategy {
     constructor(
         private paymentIntegrationService: PaymentIntegrationService,
         private tdOnlineMartScriptLoader: TDOnlineMartScriptLoader,
+        private formPoster: FormPoster,
     ) {}
 
     async initialize(options: PaymentInitializeOptions): Promise<void> {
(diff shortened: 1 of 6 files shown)


packages/td-bank-integration/src/td-online-mart.ts
```diff
@@ -32,3 +34,16 @@ interface CreateTokenError {
...
+/* eslint-disable @typescript-eslint/naming-convention */
+export interface TdOnlineMartThreeDSErrorBody {
+    errors?: Array<{ code: string }>;
+    three_ds_result?: {
+        formUrl: string;
+        threeDSSessionData: string;
+        creq: string;
+    };
+}
+/* eslint-enable @typescript-eslint/naming-convention */
```
serhii-tkachenko, line 47: [optional] It might be better to create another interface for error object

```suggestion
interface TdOnlineMartThreeDSError {
    code: string;
}

/* eslint-disable @typescript-eslint/naming-convention */
export interface TdOnlineMartThreeDSErrorBody {
    errors?: TdOnlineMartThreeDSError[];
    three_ds_result?: {
        formUrl: string;
        threeDSSessionData: string;
        creq: string;
    };
}
/* eslint-enable @typescript-eslint/naming-convention */
```

packages/td-bank-integration/src/td-online-mart-payment-strategy.ts
```diff
@@ -147,4 +160,29 @@ export default class TDOnlineMartPaymentStrategy implements PaymentStrategy {
...
+    private async processWithAdditionalAction(error: unknown): Promise<void> {
+        if (!isTdOnlineMartAdditionalAction(error)) {
+            throw error;
+        }
+
+        const { three_ds_result: threeDSResult }: TdOnlineMartThreeDSErrorBody = error.body;
+        const { formUrl, threeDSSessionData, creq } = threeDSResult || {};
+
+        if (!formUrl || !threeDSSessionData || !creq) {
+            throw new PaymentArgumentInvalidError(['formUrl', 'threeDSSessionData', 'sreq']);
```
serhii-tkachenko, line 173:
```suggestion
            throw new PaymentArgumentInvalidError(['formUrl', 'threeDSSessionData', 'creq']);
```

serhii-tkachenko: Also, I don't really sure that it is right error to show for the customer, because the message is: 
"Make sure the following fields are provided correctly:"
I don't think that user can fix that on their end.. So it's better to do some search and throw another message.

What do you think?

packages/td-bank-integration/src/td-online-mart.ts
PavlenkoM: I don't think that we need to do this. Because this interface won't be used in other places and hasn't difficult structure to split it to another interface.

serhii-tkachenko (approved):

packages/td-bank-integration/src/td-online-mart-payment-strategy.ts
PavlenkoM: In this case users can't fix this issue on their end. Because these options are broadcast from TDBank over BigPay, and the user can’t change its values. But understanding what goes wrong for a shopper, by simple screenshot for support, will help us more accurately and faster understand and fix bugs.

PR #2299, opened 2024-01-02
feat(payment): PAYPAL-3408 updated paypal commerce sdk configuration with PayPal Connect component

serhii-tkachenko:
## What?
Updated paypal commerce sdk configuration with PayPal Connect component

## Why?
To be able to use PayPal Connect SDK for PayPal Commerce Accelerated checkout implementation

## Testing / Proof
Unit tests
Manual tests


andriiVitvitskyi1990 (approved):

packages/paypal-commerce-integration/src/paypal-commerce-script-loader.ts
```diff
@@ -137,14 +146,22 @@ export default class PayPalCommerceScriptLoader {
...
+                    ...enableConnectComponent,
                 ],
                 currency: currencyCode,
                 intent,
                 ...(isDeveloperModeApplicable && { 'buyer-country': buyerCountry }),
             },
             attributes: {
                 'data-partner-attribution-id': attributionId,
-                'data-client-token': clientToken,
+                ...(shouldEnableConnectComponent
```
bc-nick, line 157: Should we do an update here `packages/google-pay-integration/src/google-pay-paypal-commerce/google-pay-paypal-commerce-script-loader.ts` too?

serhii-tkachenko: We don't need to do it yet, because for PPCP AXO beta we will disable wallet buttons on checkout page due to some issues with configuration. Also we will create paypal commerce utils package which can be use as a single source of truth for paypal sdk configuration (no need to do extra work for now)

bc-nick (approved):

packages/paypal-commerce-integration/src/paypal-commerce-script-loader.ts
serhii-tkachenko: Thanks for your question

PR #2302, opened 2024-01-03
feat(payment): PI-1318 TDBank add  browser info

PavlenkoM:
## What?
Add browser info to payments request payload

## Why?
Need this info for 3DS flow on TD Bank side

## Testing / Proof
unity tests and manually tested
Screenshot 2024-01-03 at 17 42 32

### DIFF
--- a/packages/td-bank-integration/src/td-online-mart-payment-strategy.ts
+++ b/packages/td-bank-integration/src/td-online-mart-payment-strategy.ts
@@ -1,6 +1,7 @@
 import { FormPoster } from '@bigcommerce/form-poster';
 
 import {
+    getBrowserInfo,
     InvalidArgumentError,
     isHostedInstrumentLike,
     isVaultedInstrument,
@@ -23,6 +24,7 @@ import { isTdOnlineMartAdditionalAction } from './isTdOnlineMartAdditionalAction
 import {
     FieldType,
     TDCustomCheckoutSDK,
+    TdOnlineMartBrowserInfo,
     TdOnlineMartElement,
     TdOnlineMartThreeDSErrorBody,
 } from './td-online-mart';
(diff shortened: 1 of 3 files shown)


bc-sebastianszafraniec (approved):

packages/td-bank-integration/src/td-online-mart-payment-strategy.ts
```diff
@@ -112,13 +119,28 @@ export default class TDOnlineMartPaymentStrategy implements PaymentStrategy {
...
+    private getTDBrowserInfo(): TdOnlineMartBrowserInfo {
+        const { time_zone_offset, ...defaultBrowserInfo } = getBrowserInfo();
+
+        return {
+            ...defaultBrowserInfo,
+            accept_header:
+                'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3',
+            javascript_enabled: true,
+            time_zone: time_zone_offset,
+            user_agent: window.navigator.userAgent,
```
TomA-R, line 139: Sending this in the JSON payload seems a bit odd to me:
- `User-Agent` and `Accept` should already be sent as header with the request
- Forcing `Accept` to a static value is probably not desired for the 3DS check
- javascript_enabled can probably be assumed to always be `true` in the backend, since the checkout process requires it

TomA-R (changes requested):

packages/td-bank-integration/src/td-online-mart-payment-strategy.ts
PavlenkoM: Hi @TomA-R 
Thanks for info about headers. The main idea was to keep all needed data in one place, but if you think that better to collect some data on BE side I removed fields for `Accept`, `User_Agent` and `js-enabled`.

TomA-R: Thanks @PavlenkoM, best to collect that from the actual requests 👍

TomA-R (approved):

PR #2308, opened 2024-01-05
feat(payment): PAYPAL-3409 updated PaymentProviderCustomer with PPCP AXO related interface

serhii-tkachenko:
## What?
Updated PaymentProviderCustomer with PPCP AXO related interface

## Why?
To be able to save PayPalCommerce related PaymentProviderCustomer data into checkout state

## Testing / Proof
Unit tests
Ci tests
Manual test in separate PR with the feature, that uses this interface


animesh1987:
Is there a reason why we are adding these interfaces in integration package?

serhii-tkachenko:
> Is there a reason why we are adding these interfaces in integration package?

Yes. We should have an ability to save PaymentProviderCustomer based on PPCP AXO data provided from PayPal after OTP

animesh1987 (approved):

PR #2321, opened 2024-01-11
fix(checkout): DATA-11638 Fix payloads for Segment events

bc-rmalyavc:
## What? [DATA-11638](https://bigcommercecloud.atlassian.net/browse/DATA-11638)
- Make `product_id` field a string for `Checkout Started` and `Order Completed` events
- Rename `orderId` field to `order_id` for `Order Completed` event
- Make `order_id` field a string  for `Order Completed` event
- Rename `shippingMethod` field to `shipping_method` for `Checkout Step Completed` event
- Rename `paymentMethod` field to `payment_method` for `Checkout Step Completed` event
- Do not populate `brand` field if brand is missing for a product
- Assign default value `single-product-option` for the `variant` field in products payload in order to make it consistent with other events. [Related PR](https://github.com/bigcommerce/bigcommerce/pull/56279)
- Update unit tests

## Why?
There are several inconsistencies between payload we generate and Segment specification. So that we need to adjust the existing logic to make the payload consistent with the specifications.

## Testing / Proof
- Manually tested on dev store
- Unit tests

[DATA-11638]: https://bigcommercecloud.atlassian.net/browse/DATA-11638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/analytics/analytics-step-tracker.ts
+++ b/packages/core/src/analytics/analytics-step-tracker.ts
@@ -6,7 +6,7 @@ import {
     ExtraItemsData,
 } from '@bigcommerce/checkout-sdk/analytics';
 
-import { LineItemMap } from '../cart';
+import { DigitalItem, LineItemMap, PhysicalItem } from '../cart';
 import { Checkout, CheckoutService } from '../checkout';
 import { InvalidArgumentError } from '../common/error/errors';
 import { ShopperCurrency, StoreProfile } from '../config';
@@ -152,7 +152,7 @@ export default class AnalyticsStepTracker implements StepTracker {
             isPayloadSizeLimitReached(payload)
         ) {
             sendGoogleAnalytics('transaction', {
-                '&ti': payload.orderId,
+                '&ti': payload.order_id,
                 '&ta': payload.affiliation,
                 '&tr': payload.revenue,
                 '&ts': payload.shipping,
@@ -162,7 +162,7 @@ export default class AnalyticsStepTracker implements StepTracker {
             });
             payload.products.forEach((product) => {
                 sendGoogleAnalytics('item', {
-                    '&ti': payload.orderId,
+                    '&ti': payload.order_id,
                     '&in': product.name,
                     '&ic': product.sku,
                     '&iv': `${product.category}`,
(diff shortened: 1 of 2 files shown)


solofeed (approved):
⚠️ It might be breaking change for current clients

packages/core/src/analytics/analytics-step-tracker.ts
```diff
@@ -421,15 +421,18 @@ export default class AnalyticsStepTracker implements StepTracker {
...
                 sku: item.sku,
                 price: item.salePrice,
                 image_url: item.imageUrl,
                 name: item.name,
                 quantity: item.quantity,
-                brand: itemsData[item.productId] ? itemsData[item.productId].brand : '',
                 category: itemsData[item.productId] ? itemsData[item.productId].category : '',
-                variant: (itemAttributes || []).join(', '),
+                variant: (itemAttributes || []).join(', ') || 'single-product-option',
+                ...(itemsData[item.productId] &&
```
animesh1987, line 432: Indentation?

packages/core/src/analytics/analytics-step-tracker.ts
```diff
@@ -421,15 +421,18 @@ export default class AnalyticsStepTracker implements StepTracker {
...
+                product_id: String(item.productId),
                 sku: item.sku,
                 price: item.salePrice,
                 image_url: item.imageUrl,
                 name: item.name,
                 quantity: item.quantity,
-                brand: itemsData[item.productId] ? itemsData[item.productId].brand : '',
                 category: itemsData[item.productId] ? itemsData[item.productId].category : '',
-                variant: (itemAttributes || []).join(', '),
+                variant: (itemAttributes || []).join(', ') || 'single-product-option',
```
animesh1987, line 431: Can you please refactor this to move to a separate function, and also break it into multiple lines since it can be a bit hard to read.

bc-rmalyavc: Hey @animesh1987,
Refactored this part a bit and also moved all this transform logic to a separate function, so it should be more straightforward now. Please, check again.

packages/core/src/analytics/analytics-step-tracker.ts
bc-rmalyavc: Not sure what do you mean. Eslint doesn't indicate any issues

animesh1987 (approved):

PR #2327, opened 2024-01-17
feat(payment): PAYPAL-3470 Update paypal-commerce-sdk configuration to use PayLater Messages separately

bc-nick:
## What?

Update paypal-commerce-sdk configuration

## Why?

To use PayLater Messages separately

## Testing / Proof

Manual

### DIFF
--- a/packages/paypal-commerce-utils/src/paypal-commerce-sdk.ts
+++ b/packages/paypal-commerce-utils/src/paypal-commerce-sdk.ts
@@ -11,6 +11,8 @@ import {
     PayPalAxoSdk,
     PayPalCommerceHostWindow,
     PayPalCommerceInitializationData,
+    PayPalCommerceSdkNamespaces,
+    PayPalMessagesSdk,
     PayPalSdkConfig,
 } from './paypal-commerce-types';
 
@@ -40,6 +42,26 @@ export default class PayPalCommerceSdk {
         }
 
         return this.window.paypalAxo;
+    }
+
+    async getPayPalMessages(
+        paymentMethod: PaymentMethod<PayPalCommerceInitializationData>,
+        currencyCode: string,
+    ): Promise<PayPalMessagesSdk> {
+        if (!this.window.paypalMessages) {
+            const paypalSdkMessagesConfig = this.getPayPalSdkMessagesConfiguration(
+                paymentMethod,
+                currencyCode,
+            );
+
+            await this.loadPayPalSdk(paypalSdkMessagesConfig);
+        }
+
+        if (!this.window.paypalMessages) {
+            throw new PaymentMethodClientUnavailableError();
+        }
+
+        return this.window.paypalMessages;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


andriiVitvitskyi1990 (approved):

packages/paypal-commerce-utils/src/paypal-commerce-sdk.ts
```diff
@@ -42,6 +44,26 @@ export default class PayPalCommerceSdk {
...
+                paymentMethod,
+                currencyCode,
+            );
+
+            await this.loadPayPalSdk(paypalSdkMessagesConfig);
+        }
+
+        if (!this.window.paypalMessages) {
+            throw new PaymentMethodClientUnavailableError();
+        }
```
bc-dronov, line 62: We can optimize it by combining `if` checks. That way we can avoid redundant condition check.
 
```suggestion
      if (!this.window.paypalMessages) {
        const paypalSdkMessagesConfig = this.getPayPalSdkMessagesConfiguration(
            paymentMethod,
            currencyCode,
        );

      await this.loadPayPalSdk(paypalSdkMessagesConfig);
      
      if (!this.window.paypalMessages) {
          throw new PaymentMethodClientUnavailableError();
      }
  }
```

bc-nick: We need to use the second `if` operator because for some reason the instance might be loaded with the wrong namespace and we need to check it

bc-dronov: Yes, we don't remove the second `if`, we just move it inside the first `if`, so after that we can avoid redundant condition check.

bc-nick: Misunderstood a little bit. It makes sense. I did that for the `getPayPalAxo` function as well. Thanks!

packages/paypal-commerce-utils/src/paypal-commerce-types.ts
```diff
@@ -37,15 +37,27 @@ export interface PayPalCommerceInitializationData {
...
+/**
+ *
+ * PayPal Commerce Sdk Namespaces
+ *
+ */
+export enum PayPalCommerceSdkNamespaces {
+    PaypalAxo = 'paypalAxo',
+    PaypalMessages = 'paypalMessages',
+    PaypalConnect = 'paypalConnect',
+}
```
serhii-tkachenko, line 49: This is not quite right to set `paypalConnect` as a namespace because this is a part of PayPalAxo implementation, but not a part of PP SDK initialization.

packages/paypal-commerce-utils/src/paypal-commerce-sdk.ts
```diff
@@ -90,7 +112,36 @@ export default class PayPalCommerceSdk {
...
+            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+        }
+
+        const { intent, clientId, merchantId, attributionId } = initializationData;
+
+        return {
+            options: {
+                'client-id': clientId,
+                'merchant-id': merchantId,
+                commit: true,
```
serhii-tkachenko, line 138: I don't really sure that `commit` param should be true for messages... As far as I know, it should be false if we use banners outside checkout page. I might be better to check paypal documentation and update commit param with right value for banners

packages/paypal-commerce-utils/src/paypal-commerce-sdk.ts
```diff
@@ -90,7 +112,36 @@ export default class PayPalCommerceSdk {
...
+                components: ['messages'],
+                currency: currencyCode,
+                intent,
```
serhii-tkachenko, line 141: Also might be better to check `intent` option too.

packages/paypal-commerce-utils/src/paypal-commerce-types.ts
bc-nick: But we add it as the value of the `data-namespace` attribute in the `script`

bc-nick: You are right. I confused it with `paypalAxo`

bc-dronov (approved):

PR #2339, opened 2024-01-22
feat(payment): PAYPAL-3419 added PayPalCommerceAcceleratedCheckout shipping strategy

serhii-tkachenko:
## What?
Added PayPalCommerceAcceleratedCheckout shipping strategy

## Why?
To be able to authenticate user with PayPal Connect and get related data to use in checkout flow

## Testing / Proof
Unit tests
Manual tests
CI

### DIFF
--- a/packages/core/src/shipping/strategies/paypal-commerce/paypal-commerce-accelerated-checkout-shipping-strategy.ts
+++ b/packages/core/src/shipping/strategies/paypal-commerce/paypal-commerce-accelerated-checkout-shipping-strategy.ts
@@ -60,20 +60,20 @@ export default class PayPalCommerceAcceleratedCheckoutShippingStrategy implement
             );
         }
 
-        if (!this._shouldRunAuthenticationFlow()) {
-            return Promise.resolve(this._store.getState());
-        }
+        const isAuthenticatedUser = !this._shouldRunAuthenticationFlow();
 
-        try {
-            await this._store.dispatch(
-                this._paymentMethodActionCreator.loadPaymentMethod(methodId),
-            );
+        if (!isAuthenticatedUser) {
+            try {
+                await this._store.dispatch(
+                    this._paymentMethodActionCreator.loadPaymentMethod(methodId),
+                );
 
-            await this._initializePayPalSdk(methodId, styles);
-            await this._runAuthenticationFlowOrThrow(methodId);
-        } catch (error) {
-            // Info: we should not throw any error here to avoid customer stuck on
-            // shipping step due to the payment provider custom flow
+                await this._initializePayPalSdk(methodId, styles);
+                await this._runAuthenticationFlowOrThrow(methodId);
+            } catch (error) {
+                // Info: we should not throw any error here to avoid customer stuck on
+                // shipping step due to the payment provider custom flow
+            }
         }
 
         return Promise.resolve(this._store.getState());
--- a/packages/core/src/shipping/strategies/paypal-commerce/paypal-commerce-accelerated-checkout-shipping-strategy.ts
+++ b/packages/core/src/shipping/strategies/paypal-commerce/paypal-commerce-accelerated-checkout-shipping-strategy.ts
@@ -3,6 +3,7 @@ import {
     PayPalCommerceAcceleratedCheckoutUtils,
     PayPalCommerceConnectAuthenticationState,
     PayPalCommerceConnectStylesOption,
+    PayPalCommerceSdk,
 } from '@bigcommerce/checkout-sdk/paypal-commerce-utils';
 
 import { AddressRequestBody } from '../../../address';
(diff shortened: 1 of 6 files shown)


andriiVitvitskyi1990 (approved):

bc-dronov (approved):

packages/core/src/shipping/strategies/paypal-commerce/paypal-commerce-accelerated-checkout-shipping-strategy.ts
```diff
@@ -0,0 +1,175 @@
...
+
+        // Load payment method
+        // Load run auth flow
+        try {
+            await this._store.dispatch(
+                this._paymentMethodActionCreator.loadPaymentMethod(methodId),
+            );
+            await this._runAuthenticationFlowOrThrow(methodId, styles);
+        } catch (error) {
+            // Info: we should not throw any error here to avoid
```
animesh1987, line 74: What happens in the case where the payment method fails to load?

packages/core/src/shipping/strategies/paypal-commerce/paypal-commerce-accelerated-checkout-shipping-strategy.ts
```diff
@@ -0,0 +1,175 @@
...
+        // Should run Authentication flow
+        if (!this._shouldRunAuthenticationFlow()) {
+            return Promise.resolve(this._store.getState());
+        }
+
+        // Load payment method
+        // Load run auth flow
+        try {
+            await this._store.dispatch(
+                this._paymentMethodActionCreator.loadPaymentMethod(methodId),
```
animesh1987, line 70: Do you need to call `loadPaymentMethod` here? This will trigger an API call which can slow down the render. In this case the `getPaymentMethod` should work 🤔

packages/core/src/shipping/strategies/paypal-commerce/paypal-commerce-accelerated-checkout-shipping-strategy.ts
```diff
@@ -0,0 +1,175 @@
...
+    ): Promise<InternalCheckoutSelectors> {
+        const { methodId, styles } = options || {};
+
+        if (!methodId) {
+            throw new InvalidArgumentError(
+                'Unable to proceed because "methodId" argument is not provided.',
+            );
+        }
+
+        // Should run Authentication flow
```
animesh1987, line 61: Do we need all these comments?

animesh1987: For reference https://bpoplauschi.github.io/2021/01/20/Clean-Code-Comments-by-Uncle-Bob-part-2.html

packages/core/src/shipping/strategies/paypal-commerce/paypal-commerce-accelerated-checkout-shipping-strategy.ts
serhii-tkachenko: PP AXO SDK will not be loaded, but it does not brake default flow

packages/core/src/shipping/strategies/paypal-commerce/paypal-commerce-accelerated-checkout-shipping-strategy.ts
serhii-tkachenko: Let me double check this part

packages/core/src/shipping/strategies/paypal-commerce/paypal-commerce-accelerated-checkout-shipping-strategy.ts
serhii-tkachenko: I will clean all comments in the file

packages/core/src/shipping/strategies/paypal-commerce/paypal-commerce-accelerated-checkout-shipping-strategy.ts
serhii-tkachenko: We should load payment method to get access to payment method data. PP AXO Authentication flow should run only when customer lands on Shipping step (after cart update or reloads a page after passing customer step), so in this case we will not have payment method to get it from state

animesh1987: Is this only fired for logged in shopper or is fired for guest shopper as well? I ask since for guest shopper we load the payment method in customer step and set that in state. So in that case we can read from state, and load method if not present in state.

serhii-tkachenko: It fires for customer who was authenticated in the same session before. So this is rather an edge case to get and autofill checkout form with PP Connect data

animesh1987 (approved):

PR #2366, opened 2024-02-19
refactor(payment): PAYPAL-3647 updated PaymentInstrumentPayload interface with generic

serhii-tkachenko:
## What?
Updated PaymentInstrumentPayload interface with generic

## Why?
This feature gives developers an ability to provide custom formatted payload interface to payment submit method from provider specific package. It should help us to keep all custom interfaces inside our own packages and remove all provider specific code from core type packages (payment integration api).

## Testing / Proof
I have build the project in different ways using `npm run build`, `npm run bundle:watch`, `npm run bundle-dts`, different commands to check lint issues and tests and they are all passed successfully

And CI tests as well

### DIFF
--- a/packages/payment-integration-api/src/payment/payment.ts
+++ b/packages/payment-integration-api/src/payment/payment.ts
@@ -3,14 +3,14 @@ import { Omit } from '../util-types';
 
 import PaymentAdditionalAction from './payment-additional-action';
 
-export default interface Payment<FP = unknown> {
+export default interface Payment<T = unknown> {
     methodId: string;
     gatewayId?: string;
-    paymentData?: PaymentInstrumentPayload<FP> & PaymentInstrumentMeta;
+    paymentData?: PaymentInstrumentPayload<T> & PaymentInstrumentMeta;
     additionalAction?: PaymentAdditionalAction;
 }
 
-export type PaymentInstrumentPayload<FP = unknown> =
+export type PaymentInstrumentPayload<T = unknown> =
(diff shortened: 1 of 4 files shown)


andriiVitvitskyi1990 (approved):

bc-nick (approved):

PavlenkoM (approved):

vitalii-koshovyi (approved):
👍 Nice idea. Should help to cleanup the mess with FormattedPayload types a bit.

packages/payment-integration-api/src/payment/payment.ts
```diff
@@ -3,14 +3,14 @@ import { Omit } from '../util-types';
 
 import PaymentAdditionalAction from './payment-additional-action';
 
-export default interface Payment {
+export default interface Payment<FP = unknown> {
```
animesh1987, line 6: Why FP?

serhii-tkachenko: FP - is like formatted payload type

animesh1987: Seems fine to name it `T` even since it's a generic. But this line confuses me then https://github.com/bigcommerce/checkout-sdk-js/pull/2366/files#diff-5c53fae9c62715282b70567f8e100092944dc0f3335e998b9b38fd7dacf0f77eR44. We passing a generic, but it can be it's own type as well.

animesh1987 (approved):

PR #2374, opened 2024-02-21
feat(checkout): PI-1655 Update payment integration service to include more methods from the core package

bc-wpietrzak:
## What?
Updated `PaymentIntegrationService`, so the following methods/errors from the core package can be used:
- `forgetCheckout` method
- `getContextConfig` method
- `OrderFinalizationNotCompletedError` error

## Why?
Needed for moving Afterpay payment strategy to a separate package.

## Testing / Proof
Screenshot 2024-02-21 at 20 22 59


packages/core/src/payment-integration/default-payment-integration-service.ts
```diff
@@ -110,6 +110,12 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
         return this._storeProjection.getState();
     }
 
+    async loadPaymentMethods(options?: RequestOptions): Promise<PaymentIntegrationSelectors> {
```
serhii-tkachenko, line 113: It's ok to add this method, but why do we need to load full list of payment methods for one integration?

bc-wpietrzak: I tried to remove this method or use `loadPaymentMethod` instead, but both cases caused cc form fields to appear if there's an error:
Screenshot 2024-02-22 at 14 33 38
When `loadPaymentMethods` is used, the fields do not appear after error.

serhii-tkachenko: I think its more a bug then a feature..

vitalii-koshovyi: I don't understand it too. I believe we're loading whole list of payment methods, when payments step is opened.
So why do we need to load them again?
Could we take those data from the state?

PavlenkoM: @bc-wpietrzak 
Can you please check what the difference between responses for all payment methods(`loadPaymentMethods`) and only for afterpay(`loadPaymentMethod(methodId)`) on error?
Maybe we can change something on BE side to prevent refetch for all payment methods and request only afterpay if needed?

bc-wpietrzak: Here's response for `loadPaymentMethods`:

Screenshot 2024-02-26 at 10 59 14

And for `loadPaymentMethod(methodId)`:

Screenshot 2024-02-26 at 11 26 47

PavlenkoM: So, responses look very similar? Both have the same display name, and I think that it is a correct display name.
What is the reason that you can't use` loadPaymentMethod(method)`?

bc-wpietrzak: As I stated in my first response to Serhii, using `loadPaymentMethod(method)` causes form fields to render when there's an error + the display name changes to `Afterpay`, even though we get correct display name from the BE. I'd have to check if there's something on checkout-js side that may need adjusting to prevent that from happening. Currently Afterpay and Clearpay are the only providers, where `loadPaymentMethods` is used - that's why I assumed it's intentional, because of the behavior described above.

PavlenkoM: - Should we hide `Afterpay` from the list of available payment methods after any `submitPayment` error occurs?
- If we still need to show `Afterpay` after error, will be better to fix bug with "display name" instead of reload all payment methods.

serhii-tkachenko (approved):
LGTM 👍

packages/core/src/payment-integration/default-payment-integration-service.ts
animesh1987: I think we still need to think if we can avoid loading all payment methods, since that comes with it's own problem with performance/API latency. As @PavlenkoM said we probably can try other approach to solve this problem.

bc-wpietrzak: @animesh1987 Sure, I'll make sure to work on that while moving the Afterpay payment strategy. Shall I remove `loadPaymentMethods` from this PR or can I leave it as it is?

serhii-tkachenko: @bc-wpietrzak lets remove it then.

PavlenkoM (approved):

vitalii-koshovyi (approved):

packages/core/src/payment-integration/default-payment-integration-service.ts
animesh1987: Thanks @bc-wpietrzak 🙇

animesh1987 (approved):
Nice work @bc-wpietrzak 👏

PR #2426, opened 2024-03-24
fix(embedded-checkout): Handle www redirects on embed checkout links

### LINKED ISSUE #2425: Embedded checkout fails to load due to www redirect.
**Describe the bug**
When the `embedCheckout` function is passed a url like `https://store.com/`, the underlying `ResizableIframeCreator` is only listening for messages that [come from the same origin](https://github.com/bigcommerce/checkout-sdk-js/blob/b4e9bee3c0fa999210cca52089f21469679b0663/packages/core/src/embedded-checkout/resizable-iframe-creator.ts#L54). If the store has a www redirect and the iframe ends up loading from `https://www.store.com/`, the message will be ignored.

**To Reproduce**
1. Have a store that redirects from the bare domain to the www subdomain
2. Have an embedded checkout url that does not include the www subdomain
3. Attempt to embed the checkout
4. Checkout fails to load after ignoring the message posted from the unexpected domain

**Expected behavior**
The checkout is loaded successfully.

**Additional context**
I saw that the `IframeEventListener` already [handles the possibility of the www subdomain](https://github.com/bigcommerce/checkout-sdk-js/blob/b4e9bee3c0fa999210cca52089f21469679b0663/packages/core/src/common/iframe/iframe-event-listener.ts#L16-L19). I'll have a PR up shortly to adjust the `ResizableIframeCreator` in a similar way.

Magnesium38:
## What?
Adds support in the `ResizeableIframeCreator` for bare and www versions of the domain passed to `embedCheckout` .

Fixes #2425 

## Why?
If the embed checkout url that is passed to `embedCheckout` is a bare url missing the www subdomain whilst the store loads with a redirect to the www subdomain, the resulting `MessageEvent` that is posted will be ignored by the `ResizeableIframeCreator`.

I used the [`IframeEventListener`'s handling](https://github.com/bigcommerce/checkout-sdk-js/blob/b4e9bee3c0fa999210cca52089f21469679b0663/packages/core/src/common/iframe/iframe-event-listener.ts#L16-L19) of this functionality as a baseline for this change.

## Testing / Proof
The test that has been included as part of this PR fails without the associated change.

### DIFF
--- a/packages/core/src/embedded-checkout/resizable-iframe-creator.spec.ts
+++ b/packages/core/src/embedded-checkout/resizable-iframe-creator.spec.ts
@@ -184,4 +184,18 @@ describe('ResizableIframeCreator', () => {
             );
         }
     });
+
+    it('inserts the iframe even if the message comes from the www version of the url', async () => {
+        setTimeout(() => {
+            eventEmitter.emit('message', {
+                origin: 'http://www.mybigcommerce.com',
+                data: { type: EmbeddedCheckoutEventType.FrameLoaded },
+            });
+        });
+
+        const frame = await iframeCreator.createFrame(url, 'checkout');
+
+        expect(frame.tagName).toBe('IFRAME');
+        expect(frame.src).toEqual(url);
+    });
 });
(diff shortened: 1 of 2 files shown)


animesh1987 (approved):

animesh1987:
@Magnesium38 thanks for raising this fix 👍

PR #2444, opened 2024-04-03
feat(payment): PAYPAL-3850 Removed BraintreeScriptLoader from core and transferred related code

bc-nick:
## What?

Removed `BraintreeScriptLoader` from core and transferred related `code` to the `braintree-utils` package.

## Why?

Now we have only one place from where all related scripts to the `braintree` are loaded

## Testing / Proof

Manual testing of Braintree integration


bc-dronov (approved):

packages/core/src/payment/strategies/braintree/braintree-sdk-creator.ts
```diff
@@ -15,13 +7,22 @@ import {
...
+
+import {
+    NotInitializedError,
+    NotInitializedErrorType,
+    UnsupportedBrowserError,
+} from '../../../common/error/errors';
+import { StoreConfig } from '../../../config';
+import { PaypalHostWindow } from '../paypal';
 
+// TODO: move BraintreeSDKCreator to the braintree-utils package, take a look at the BraintreeIntegrationService
```
serhii-tkachenko, line 25: We don't need BraintreeSdkCreator so we should not move it to braintree utils

serhii-tkachenko: Please remove this comment to avoid misunderstanding in the future

packages/braintree-utils/src/visacheckout.ts
serhii-tkachenko, line 1: We have types.ts file created in braintree-utils package (braintree.ts and paypal.ts files are deprecated and will be removed in the future). Therefore we should decide how do we want to manage types in our team:
1. Hold all the types inside types.ts file;
2. Create separate file for each braintree module and hold all related interfaces there

packages/braintree-utils/src/types.ts
```diff
@@ -26,15 +27,6 @@ export interface BraintreeModule {
     teardown(): Promise<void>;
 }
 
-/**
- *
- * Braintree Window
- *
- */
-export interface BraintreeWindow extends Window {
```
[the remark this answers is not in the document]
serhii-tkachenko: BraintreeHostWindow will be removed in the future. This BraintreeWindow interface is a new place where all refactored braintree modules will be located for a while, until they all be refactored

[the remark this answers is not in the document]
bc-nick: We do not need to use it anymore. Instead we should to use the generic interface with the list of instants here: https://github.com/bigcommerce/checkout-sdk-js/pull/2444/files#diff-7ee590679477de93cd76ab37c88315ed14aafd6b5f02e09122d53c2222ef0009R75

And i have updated line here: https://github.com/bigcommerce/checkout-sdk-js/pull/2444/files#diff-7ee590679477de93cd76ab37c88315ed14aafd6b5f02e09122d53c2222ef0009R1078

That's how we got rid of the extra method

packages/braintree-utils/src/visacheckout.ts
bc-nick: We can make appropriate refactoring for these files in the future. I mean for `braintree.ts`, `paypal.ts` and `visacheckout.ts`. I thought for now we could add `visacheckout.ts` import, same for `braintree.ts`, `paypal.ts`.

packages/braintree-utils/.eslintrc.json
```diff
@@ -5,7 +5,8 @@
         {
             "files": ["*.ts", "*.tsx"],
             "rules": {
-                "@typescript-eslint/naming-convention": "off"
+                "@typescript-eslint/naming-convention": "off",
+                "@typescript-eslint/no-explicit-any": "off"
```
serhii-tkachenko, line 9: It might be better to remove this rule (we can keep it for spec files, but I on't think it's a good idea to set it for all ts files in package)

bc-nick: We have the same rule in another packages for `ts` files.I went in the same way

serhii-tkachenko: It is probably a mistake. Lets remove it if

serhii-tkachenko (approved):

animesh1987 (approved):

PR #2445, opened 2024-04-04
CHECKOUT-8121: Remove queue ID for executing strategies to minimise concurrent requests

davidchin:
## What?
Remove the queue IDs used to execute payment, shipping, and customer strategies to minimise the likelihood of [conflict](https://kibana.bigcommerce.net/goto/dfc2e710-f228-11ee-8e61-2dd683c10ce8) due to concurrent requests. By removing the queue IDs, the calls will be executed sequentially.

## Why?
These strategies could trigger mutations to cart/checkout. They may conflict with each other if they are triggered concurrently.

## Testing / Proof
CircleCI

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/checkout/checkout-service.spec.ts
+++ b/packages/core/src/checkout/checkout-service.spec.ts
@@ -855,9 +855,7 @@ describe('CheckoutService', () => {
             await checkoutService.initializePayment(options);
 
             expect(paymentStrategyActionCreator.initialize).toHaveBeenCalledWith(options);
-            expect(store.dispatch).toHaveBeenCalledWith(action, {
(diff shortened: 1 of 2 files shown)


animesh1987:
Hey @davidchin thanks for raising this, I had a question to get benefits of these the custom checkouts for example TRR will have to update their checkout codebase with the SDK bump?

davidchin:
Yes, that's correct.

animesh1987 (approved):

davidchin:
Upon further testing, I realised this change won't really work because strategies won't be able to trigger any action unless the action is triggered in a separate queue.

PR #2453, opened 2024-04-10
feat(payment): PI-1606 add validateCheckout and loadPaymentMethods to Payment Integration Service

bc-wpietrzak:
## What?
Updated `PaymentIntegrationService` to include:

- `validateCheckout` - so the `validate` method from the core package can be used
- `loadPaymentMethods` - to load available payments methods after `forgetCheckout` method is used

## Why?
Needed for moving Afterpay payment strategy to a separate package.

## Testing / Proof
Unit + manual tests.


bc-peng (approved):
LGTM.

PR #2455, opened 2024-04-10
feat(payment): PAYPAL-3850 Update braintree scripts by passing additional integrity and crossorigin attributes to script tag

bc-nick:
## What?

Update braintree scripts by passing additional `integrity` and `crossorigin` attributes to script tag that is loaded from the BraintreeScriptLoader

## Why?

According to PCI v4.0

## Testing / Proof

Braintree manual testing

### DIFF
--- a/packages/braintree-utils/src/braintree-script-loader.ts
+++ b/packages/braintree-utils/src/braintree-script-loader.ts
@@ -1,11 +1,11 @@
-import { LoadScriptOptions, ScriptLoader } from '@bigcommerce/script-loader';
+import { ScriptLoader } from '@bigcommerce/script-loader';
 
 import {
     PaymentMethodClientUnavailableError,
     StoreConfig,
 } from '@bigcommerce/checkout-sdk/payment-integration-api';
 
-import { BraintreeIntegrityValues, BraintreeSdkFileName } from './braintree';
+import { BraintreeIntegrityValues } from './braintree';
 import { BRAINTREE_SDK_ALPHA_VERSION, BRAINTREE_SDK_STABLE_VERSION } from './sdk-verison';
 import {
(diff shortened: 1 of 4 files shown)


packages/braintree-utils/src/braintree-script-loader.ts
```diff
@@ -153,4 +177,106 @@ export default class BraintreeScriptLoader {
         // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
         return braintreeModule as T;
     }
+
+    // Info: for generating an integrity hash, you must use an appropriate hash generator https://www.srihash.org/
+    private getIntegrityValuesByFileName(fileName: string): BraintreeIntegrityValues {
```
serhii-tkachenko, line 182: Instead of passing file name we can stick to module, therefore we don't need to have BraintreeSdkFileName enum at all

packages/braintree-utils/src/braintree-script-loader.ts
```diff
@@ -46,103 +47,126 @@ export default class BraintreeScriptLoader {
...
+            fileName,
+            hash
+                ? {
+                      async: true,
+                      attributes: {
+                          integrity: hash,
+                          crossorigin: 'anonymous',
+                      },
+                  }
+                : undefined,
```
serhii-tkachenko, line 158: It is not necessary to create another method for this... you can update loadBraintreeModuleOrThrow instead

serhii-tkachenko: Or you can probably create a method called `getScriptOptions` and pass a `moduleName` inside and also use this method in `loadBraintreeModuleOrThrow`

serhii-tkachenko (approved):

PR #2462, opened 2024-04-15
fix(checkout): CHECKOUT-8183 set shouldSaveAddress to false for existing customer addresses while loadCheckout

bc-shawnwang:
TODO: 
- [x] wrapped with experiment 

## What?
set shouldSaveAddress to false for existing customer addresses while loadCheckout

## Why?
so sending consignments/billing address update request, for an existing address, will set shouldSaveAddress to false, get rid of create address again.

## Testing / Proof

in checkout.js (do not do any changes) use this changed sdk, print address in AddressSelect:

Screenshot 2024-04-15 at 2 42 07 PM

before change this field is missed, so sending request to backend, it will be by default true, which causes problem.

it's already set shouldSaveAddress to false.

### DIFF
--- a/packages/core/src/checkout/checkout-action-creator.spec.ts
+++ b/packages/core/src/checkout/checkout-action-creator.spec.ts
@@ -68,7 +68,9 @@ describe('CheckoutActionCreator', () => {
     describe('#loadCheckout', () => {
         it('emits action to notify loading progress', async () => {
             const { id } = getCheckout();
-            const actions = await actionCreator.loadCheckout(id).pipe(toArray()).toPromise();
+            const actions = await from(actionCreator.loadCheckout(id)(store))
+                .pipe(toArray())
+                .toPromise();
 
             expect(checkoutRequestSender.loadCheckout).toHaveBeenCalledWith(id, undefined);
 
@@ -95,8 +97,7 @@ describe('CheckoutActionCreator', () => {
             const { id } = getCheckout();
             const errorHandler = jest.fn((action) => of(action));
 
-            const actions = await actionCreator
-                .loadCheckout(id)
+            const actions = await from(actionCreator.loadCheckout(id)(store))
                 .pipe(catchError(errorHandler), toArray())
                 .toPromise();
 
@@ -127,8 +128,9 @@ describe('CheckoutActionCreator', () => {
             );
 
             const errorHandler = jest.fn((action) => of(action));
-            const actions = await actionCreator
-                .loadCheckout('b20deef40f9699e48671bbc3fef6ca44dc80e3c7')
+            const actions = await from(
+                actionCreator.loadCheckout('b20deef40f9699e48671bbc3fef6ca44dc80e3c7')(store),
+            )
                 .pipe(catchError(errorHandler), toArray())
                 .toPromise();
 
@@ -151,15 +153,15 @@ describe('CheckoutActionCreator', () => {
             );
         });
 
-        it('calls loadConfig in parallel', () => {
+        it('does not call loadCheckout until loadConfig resolves', () => {
             const { id } = getCheckout();
 
             jest.spyOn(configActionCreator, 'loadConfig').mockReturnValue(new Promise(() => {}));
 
-            actionCreator.loadCheckout(id).toPromise();
+            from(actionCreator.loadCheckout(id)(store)).toPromise();
 
             expect(configActionCreator.loadConfig).toHaveBeenCalled();
-            expect(checkoutRequestSender.loadCheckout).toHaveBeenCalled();
+            expect(checkoutRequestSender.loadCheckout).not.toHaveBeenCalled();
         });
     });
 
--- a/packages/core/src/checkout/checkout-action-creator.ts
+++ b/packages/core/src/checkout/checkout-action-creator.ts
@@ -5,10 +5,10 @@ import { catchError } from 'rxjs/operators';
 import { throwErrorAction } from '../common/error';
 import { MissingDataError, MissingDataErrorType } from '../common/error/errors';
 import { RequestOptions } from '../common/http-request';
-import { ConfigActionCreator } from '../config';
+import { ConfigActionCreator, StoreConfig } from '../config';
 import { FormFieldsActionCreator } from '../form';
 
-import { CheckoutRequestBody } from './checkout';
+import Checkout, { CheckoutRequestBody } from './checkout';
 import { CheckoutActionType, LoadCheckoutAction, UpdateCheckoutAction } from './checkout-actions';
 import CheckoutRequestSender from './checkout-request-sender';
 import InternalCheckoutSelectors from './internal-checkout-selectors';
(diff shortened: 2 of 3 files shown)


packages/core/src/checkout/checkout-action-creator.ts
```diff
@@ -31,11 +31,12 @@ export default class CheckoutActionCreator {
...
                 defer(() =>
-                    this._checkoutRequestSender
-                        .loadCheckout(id, options)
-                        .then(({ body }) =>
-                            createAction(CheckoutActionType.LoadCheckoutSucceeded, body),
-                        ),
+                    this._checkoutRequestSender.loadCheckout(id, options).then(({ body }) => {
+                        return createAction(
+                            CheckoutActionType.LoadCheckoutSucceeded,
+                            this._transformCheckout(body),
```
bc-shawnwang, line 37: should we also transform in `loadDefaultCheckout()`? I don't know when this is called instead of `loadCheckout()` though.

animesh1987: Yes, this transform function should be added to `loadDefaultCheckout` as well. Also maybe need to rename `_transformCheckout` to something like `_transformCustomerAddresses` 🤔

bc-shawnwang: 👍

animesh1987: Also can you please write a unit test for this case.

bc-shawnwang: I think the unit test is covered? so I changed the mock, because https://github.com/bigcommerce/checkout-sdk-js/blob/b445659272efda3f381f1ed7332d8adef5da566d/packages/core/src/checkout/checkout-action-creator.spec.ts#L85, after called the actionCreator.loadCheckout, it checked the response with expected one. so it should be false now.

packages/core/src/checkout/checkout-action-creator.ts
```diff
@@ -20,28 +20,33 @@ export default class CheckoutActionCreator {
...
-                this._configActionCreator.loadConfig({
-                    ...options,
-                    useCache: true,
-                    params: { ...options?.params, checkoutId: id },
-                }),
-                this._formFieldsActionCreator.loadFormFields({ ...options, useCache: true }),
-                defer(() =>
-                    this._checkoutRequestSender.loadCheckout(id, options).then(({ body }) => {
+    loadCheckout(id: string, options?: RequestOptions): ThunkAction<LoadCheckoutAction, InternalCheckoutSelectors> {
+        return (store) => {
```
bc-shawnwang, line 24: this change will still break core test

also manual test will fail

image

bc-shawnwang: need to find a way to access store config in the function

bc-shawnwang: I made another commit (https://github.com/bigcommerce/checkout-sdk-js/pull/2462/commits/448fb6a7015845f149fb7dcf9ba89a88692827a8),

change a bit on observable operations, from 

concat( of(), merge(..., ... , defer()) ) to 
concat(of(), merge(..., ... ), defer()) 

> refer to loadDefaultCheckout.

now manually test on checkout page, it works normally,
but for test there is still one failure, so I'm not sure the change is correct.

packages/core/src/checkout/checkout-action-creator.ts
```diff
@@ -20,27 +20,42 @@ export default class CheckoutActionCreator {
...
+                        ...options,
+                        useCache: true,
+                        params: { ...options?.params, checkoutId: id },
+                    }),
+                    this._formFieldsActionCreator.loadFormFields({ ...options, useCache: true }),
+                    defer(() => {
+                        const state = store.getState();
+                        const config = state.config.getStoreConfigOrThrow();
+
+                        console.log(config);
```
bc-shawnwang, line 41: TODO: to use config to determine if we should transform body

packages/core/src/checkout/checkout-action-creator.ts
animesh1987: I will say we should have different test cases for logged in customer vs not logged in customer.

bc-shawnwang: not logged in shopper doesn't have addresses in `checkout.customer`. So they don't even have address selection. so our change doesn't affect that flow.

packages/core/src/checkout/checkout-action-creator.ts
```diff
@@ -34,22 +34,23 @@ export default class CheckoutActionCreator {
...
-                            .loadCheckout(id, options)
-                            .then(({ body }) => {
-                                return createAction(
-                                    CheckoutActionType.LoadCheckoutSucceeded,
-                                    this._transformCustomerAddresses(body),
-                                );
-                            });
-                    }),
                 ),
+                defer(() => {
```
bc-shawnwang, line 38: before the change, while manual test, I go to checkout page and it crashes:
image

I tried to make a change from
contact(of(), merge(..., ..., defer()))
to 
contact(of(), merge(..., ...), defer())

then manual test works fine. but the core test fails on one case. So I don't really know how to deal with it.

packages/core/src/checkout/checkout-action-creator.ts
```diff
@@ -20,27 +20,41 @@ export default class CheckoutActionCreator {
...
-                    params: { ...options?.params, checkoutId: id },
-                }),
-                this._formFieldsActionCreator.loadFormFields({ ...options, useCache: true }),
-                defer(() =>
-                    this._checkoutRequestSender
+    loadCheckout(
+        id: string,
+        options?: RequestOptions,
+    ): ThunkAction<LoadCheckoutAction, InternalCheckoutSelectors> {
+        return (store) => {
```
bc-shawnwang, line 27: made the change so we can access store

packages/core/src/checkout/checkout-action-creator.spec.ts
```diff
@@ -151,15 +153,15 @@ describe('CheckoutActionCreator', () => {
             );
         });
 
-        it('calls loadConfig in parallel', () => {
+        it('does not call loadCheckout until loadConfig resolves', () => {
```
bc-shawnwang, line 156: as discussed with @animesh1987, after changed the loadCheckout, the loading is in sequence, like loadDefaultCheckout, so have to change this test.

packages/core/src/checkout/checkout-action-creator.ts
```diff
@@ -117,7 +136,28 @@ export default class CheckoutActionCreator {
...
-            return this.loadCheckout(checkout.id, options);
+            return this.loadCheckout(checkout.id, options)(store);
+        };
+    }
+
+    private _shouldTransformCustomerAddress(storeConfig: StoreConfig): boolean {
+        return (
+            storeConfig.checkoutSettings.features[
+                'CHECKOUT-8183.set_shouldSaveAddress_false_for_existing_address'
+            ] ?? false
```
bc-peng, line 147:
```suggestion
            ] ?? true
```
For the reason that when we remove the experiment value, any existing custom checkout with this SDK change will behave as expected.

bc-peng (approved):

PR #2469, opened 2024-04-18
feat(payment): PAYPAL-4051 Apple Pay throws an error in checkout due to attempting to load Braintree information

bc-nick:
## What?

Added `try {} catch {}` to solve the applepay issue with braintree initialization process when credit card feature is disabled

## Why?

To get the collect data the credit card feature must to be enabled otherwise the request for getting braintree data will be failed. In this case we can not be able to use braintree integration for collecting data because we do not have clientToken for correct initialization

This is a temporary solution for a quick fix

## Testing / Proof

Manual testing

### DIFF
--- a/packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
+++ b/packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
@@ -556,6 +556,8 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
 
     private async _initializeBraintreeIntegrationService() {
         try {
+            // TODO: temporary solution to initialize the braintree payment method to get a clientToken to create a braintree client instance
+            // TODO: this approach should be removed in the future
             await this._paymentIntegrationService.loadPaymentMethod(ApplePayGatewayType.BRAINTREE);
 
             const state = this._paymentIntegrationService.getState();
--- a/packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
+++ b/packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
@@ -543,11 +543,11 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
     }
 
     private async _getBraintreeDeviceData() {
-        const braintreePaymentMethod: PaymentMethod = this._paymentIntegrationService
+        const braintreePaymentMethod = this._paymentIntegrationService
             .getState()
-            .getPaymentMethodOrThrow(ApplePayGatewayType.BRAINTREE);
+            .getPaymentMethod(ApplePayGatewayType.BRAINTREE);
 
-        if (braintreePaymentMethod.clientToken) {
+        if (braintreePaymentMethod?.clientToken) {
             const data = await this._braintreeIntegrationService.getDataCollector();
 
             return data.deviceData;
@@ -557,25 +557,25 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
     private async _initializeBraintreeIntegrationService() {
         try {
             await this._paymentIntegrationService.loadPaymentMethod(ApplePayGatewayType.BRAINTREE);
+
+            const state = this._paymentIntegrationService.getState();
+
+            const storeConfig = state.getStoreConfigOrThrow();
+
+            const braintreePaymentMethod: PaymentMethod = state.getPaymentMethodOrThrow(
+                ApplePayGatewayType.BRAINTREE,
+            );
+
+            if (!braintreePaymentMethod.clientToken || !braintreePaymentMethod.initializationData) {
+                throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+            }
+
+            this._braintreeIntegrationService.initialize(
+                braintreePaymentMethod.clientToken,
+                storeConfig,
+            );
         } catch (_) {
-            return;
+            return noop();
         }
-
-        const state = this._paymentIntegrationService.getState();
-
-        const storeConfig = state.getStoreConfigOrThrow();
-
-        const braintreePaymentMethod: PaymentMethod = state.getPaymentMethodOrThrow(
-            ApplePayGatewayType.BRAINTREE,
-        );
-
-        if (!braintreePaymentMethod.clientToken || !braintreePaymentMethod.initializationData) {
-            throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
-        }
-
-        this._braintreeIntegrationService.initialize(
-            braintreePaymentMethod.clientToken,
-            storeConfig,
-        );
     }
 }
--- a/packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
+++ b/packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
@@ -575,7 +575,7 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
                 storeConfig,
             );
         } catch (_) {
-            return noop();
+            // we don't need to do anything in this block
         }
     }
 }
--- a/packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
+++ b/packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
@@ -543,15 +543,25 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
     }
 
     private async _getBraintreeDeviceData() {
-        const data = await this._braintreeIntegrationService.getDataCollector();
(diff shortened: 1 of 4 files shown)


packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
```diff
@@ -543,29 +543,45 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
...
+        let braintreePaymentMethod: PaymentMethod;
 
-        return data.deviceData;
+        try {
+            braintreePaymentMethod = this._paymentIntegrationService
+                .getState()
+                .getPaymentMethodOrThrow(ApplePayGatewayType.BRAINTREE);
+        } catch (_) {
+            return;
+        }
```
serhii-tkachenko, line 554: It is not necessary to use try catch here. You can just remove orThrow.

```suggestion
            const braintreePaymentMethod = this._paymentIntegrationService
                .getState()
                .getPaymentMethod(ApplePayGatewayType.BRAINTREE);
```

packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
```diff
@@ -267,29 +268,45 @@ export default class ApplePayPaymentStrategy implements PaymentStrategy {
...
+        let braintreePaymentMethod: PaymentMethod;
 
-        return data.deviceData;
+        try {
+            braintreePaymentMethod = this._paymentIntegrationService
+                .getState()
+                .getPaymentMethodOrThrow(ApplePayGatewayType.BRAINTREE);
+        } catch (_) {
+            return;
+        }
```
serhii-tkachenko, line 279: It is not necessary to use try catch here. You can just remove orThrow.
```suggestion
            braintreePaymentMethod = this._paymentIntegrationService
                .getState()
                .getPaymentMethod(ApplePayGatewayType.BRAINTREE);
```

serhii-tkachenko (changes requested):
Please add a video proof and unit tests

packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
```diff
@@ -267,29 +268,39 @@ export default class ApplePayPaymentStrategy implements PaymentStrategy {
...
+            if (!braintreePaymentMethod.clientToken || !braintreePaymentMethod.initializationData) {
+                return;
+            }
+
+            this._braintreeIntegrationService.initialize(
+                braintreePaymentMethod.clientToken,
+                storeConfig,
+            );
+        } catch (_) {
+            return noop();
```
serhii-tkachenko, line 303: It is not necessary to add `noop` here

packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
```diff
@@ -543,29 +543,39 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
...
+            if (!braintreePaymentMethod.clientToken || !braintreePaymentMethod.initializationData) {
+                throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+            }
+
+            this._braintreeIntegrationService.initialize(
+                braintreePaymentMethod.clientToken,
+                storeConfig,
+            );
+        } catch (_) {
+            return noop();
```
serhii-tkachenko, line 578: It is not necessary to add `noop` here as well

packages/apple-pay-integration/src/apple-pay-payment-strategy.ts
bc-nick: Added comment to prevent lint error for empty block

packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
bc-nick: The same

serhii-tkachenko (approved):
Thanks for a fix. LGTM! 👍

packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
```diff
@@ -543,29 +543,39 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
...
 
     private async _getBraintreeDeviceData() {
-        const data = await this._braintreeIntegrationService.getDataCollector();
+        const braintreePaymentMethod = this._paymentIntegrationService
+            .getState()
+            .getPaymentMethod(ApplePayGatewayType.BRAINTREE);
 
-        return data.deviceData;
+        if (braintreePaymentMethod?.clientToken) {
+            const data = await this._braintreeIntegrationService.getDataCollector();
```
animesh1987, line 551: Is there a way we can avoid BT calls?

packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
```diff
@@ -543,29 +543,39 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
...
+            return data.deviceData;
+        }
     }
 
     private async _initializeBraintreeIntegrationService() {
-        const state = await this._paymentIntegrationService.loadPaymentMethod(
-            ApplePayGatewayType.BRAINTREE,
-        );
+        try {
+            await this._paymentIntegrationService.loadPaymentMethod(ApplePayGatewayType.BRAINTREE);
```
animesh1987, line 559: This seemed to be called in initialize path, is there a reason why we need to load `braintree` payment method.

packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
bc-nick: We can not avid it because we need to initialize BT to collect device data based on merchant configuration

packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
animesh1987: > https://github.com/bigcommerce/checkout-sdk-js/pull/2469#discussion_r1576177822

Is this the same as here? Since we might be doing the same thing twice? Also maybe add a TODO with a JIRA for plans to fix this i.e. avoid making this call and maybe pass the data as part of apple pay config load.

packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
```diff
@@ -543,29 +543,39 @@ export default class ApplePayCustomerStrategy implements CustomerStrategy {
...
+            if (!braintreePaymentMethod.clientToken || !braintreePaymentMethod.initializationData) {
+                throw new MissingDataError(MissingDataErrorType.MissingPaymentMethod);
+            }
+
+            this._braintreeIntegrationService.initialize(
+                braintreePaymentMethod.clientToken,
+                storeConfig,
+            );
+        } catch (_) {
+            // we don't need to do anything in this block
```
bc-peng, line 578: I see `// we don't need to do anything in this block` appears twice in the PR.
Could you please explain what would happen if an error occurs in the middle?

packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
bc-nick: @animesh1987 it's not the same. This is the best way to fix the issue at this point. It's a quick fix. I've already talked to our team about it, and it's going to take some time to fix it on the backend side. When another way will be found, the Braintree initialization process will be removed. At the moment we have some issue with generating `clientToken` on the backend side to pass it to the applepay configuration.

packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
bc-nick: @bc-peng This will mean that the braintree payment method cannot be initialized because we have enabled the appropriate feature in the CP

packages/apple-pay-integration/src/apple-pay-customer-strategy.ts
animesh1987: That seems ok, can we please add a TODO with JIRA to address the issue above so we have a plan.

bc-nick: @animesh1987 added TODO without JIRA. We need to create a ticket for backend developers with description. But they are aware of this problem

bc-peng: Please create a Jira ticket for tracking purpose.

bc-nick: @animesh1987 @bc-peng i have updated PR with adding JIRA ticket

animesh1987 (approved):

PR #2471, opened 2024-04-19
refactor(payment): PI-00 bundle watch LIBRARY_VERSION fix

vitalii-koshovyi:
## What?
Fix of the bundle:watch wrong library version problem

## Why?
For local development, we have a case when hosted fields are trying to load a non-existing version of the SDK.
This happens when you already have a commit in the branch and the LIBRARY_VERSION variable is bumped by 1 minor version.

For example, the current version of the SDK defined in the package.json is v1.585.1.
I've been working in the feature branch and created a new commit.
I'm trying to test this feature, but I can't because hosted fields are trying to load SDK with the version v1.586.0.
This happens because on the [line]

(https://github.com/bigcommerce/checkout-sdk-js/blob/master/scripts/webpack/get-next-version.js#L11)

we're trying to compare execSync('git describe').toString().trim() === `v${packageJson.version}`

`git describe` returns 'v1.585.1-1-balablas3' and packageJson.version is v1.585.1

So, we are going to the [line](https://github.com/bigcommerce/checkout-sdk-js/blob/master/scripts/webpack/get-next-version.js#L16) where, inside of the function, we're bumping SDK version by 1 minor version.

To avoid it, I suggest adding an env variable, so that we can check if the bundler is in the watch mode. And if it is, we just use the version from  package.json

## Testing / Proof
Before:
image


PavlenkoM (approved):

bc-peng (approved):
Thanks for the useful fix.

animesh1987 (approved):

PR #2478, opened 2024-04-23
refactor(payment): PAYPAL-4070 updated shipping initialise options with braintreefastlane and paypalcommercefastlane options

serhii-tkachenko:
## What?
Updated shipping initialise options with braintreefastlane and paypalcommercefastlane options

## Why?
To be able to pass provider specific options from UI

## Testing / Proof
Unit tests
CI

Checkout and checkout-sdk builds passed successfully:
Screenshot 2024-04-23 at 12 36 49
Screenshot 2024-04-23 at 12 36 42
Screenshot 2024-04-23 at 12 33 18

### DIFF
--- a/packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
+++ b/packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
@@ -31,10 +31,8 @@ import { PaymentProviderCustomerActionCreator } from '../../../payment-provider-
 import { CardInstrument } from '../../../payment/instrument';
 import { UntrustedShippingCardVerificationType } from '../../../payment/instrument/instrument';
 import ConsignmentActionCreator from '../../consignment-action-creator';
-import { ShippingRequestOptions } from '../../shipping-request-options';
+import { ShippingInitializeOptions, ShippingRequestOptions } from '../../shipping-request-options';
 import ShippingStrategy from '../shipping-strategy';
-
-import BraintreeAcceleratedCheckoutShippingInitializeOptions from './braintree-accelerated-checkout-shipping-initialize-options';
 
 export default class BraintreeAcceleratedCheckoutShippingStrategy implements ShippingStrategy {
     private _browserStorage: BrowserStorage;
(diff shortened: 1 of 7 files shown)


andriiVitvitskyi1990 (approved):

bc-dronov (approved):

packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
```diff
@@ -90,7 +86,7 @@ export default class BraintreeAcceleratedCheckoutShippingStrategy implements Shi
                 this._paymentMethodActionCreator.loadPaymentMethod(methodId),
             );
 
-            await this._runAuthenticationFlowOrThrow(methodId, styles);
+            await this._runAuthenticationFlowOrThrow(methodId, braintreefastlane?.styles);
```
animesh1987, line 89: Is this async call only in case for fastlane or for all cases?

packages/core/src/shipping/strategies/paypal-commerce/paypal-commerce-fastlane-shipping-strategy.ts
```diff
@@ -75,7 +71,7 @@ export default class PayPalCommerceFastlaneShippingStrategy implements ShippingS
 
                 this._isFastlaneEnabled = !!paymentMethod.initializationData.isFastlaneEnabled;
 
-                await this._initializePayPalSdk(methodId, styles);
+                await this._initializePayPalSdk(methodId, paypalcommercefastlane?.styles);
```
animesh1987, line 74: Same question as before.

packages/core/src/shipping/strategies/paypal-commerce/paypal-commerce-fastlane-shipping-strategy.ts
```diff
@@ -75,7 +71,7 @@ export default class PayPalCommerceFastlaneShippingStrategy implements ShippingS
...
                 await this._runAuthenticationFlowOrThrow(methodId);
```
animesh1987, line 75: Can we to `Promise.all` so we are not calling these in sequence but in parallel.

packages/core/src/shipping/strategies/braintree/braintree-accelerated-checkout-shipping-strategy.ts
serhii-tkachenko: For both Braintree Fastlane and Braintree Connect experiences. 
P.s.: Braintree Connect will be removed in the future

packages/core/src/shipping/strategies/paypal-commerce/paypal-commerce-fastlane-shipping-strategy.ts
serhii-tkachenko: Same answer as above. It should be called only for Fastlane related methods

packages/core/src/shipping/strategies/paypal-commerce/paypal-commerce-fastlane-shipping-strategy.ts
serhii-tkachenko: We should not call this in parallel, because _runAuthenticationFlowOrThrow should be called only after PayPal Sdk initialization

animesh1987 (approved):

PR #2485, opened 2024-04-26
feat(payment): PAYPAL-3520 Max capture amount failure

bc-nick:
## What?

Handling an specific error related to max capture failure

## Why?

To show the error message with an appropriate information for users

## Testing / Proof

Manual testing

Tested with mock data

https://github.com/bigcommerce/checkout-sdk-js/assets/99336044/9968f684-3e2f-4b00-abf4-7a976fb4aa81


packages/braintree-integration/src/braintree-paypal/braintree-paypal-payment-strategy.ts
```diff
@@ -316,4 +338,14 @@ export default class BraintreePaypalPaymentStrategy implements PaymentStrategy {
 
         throw new PaymentMethodFailedError(error.message);
     }
+
+    private isProviderError(error: unknown): boolean {
+        if (isPaypalCommerceProviderError(error)) {
+            const paypalProviderError = error?.errors?.filter((e: any) => e.provider_error) || [];
+
+            return paypalProviderError[0].provider_error?.code === '2046';
```
bc-dronov, line 346: Maybe it would be better to create a separate constant for '2046' error? How do you think?

bc-dronov (approved):

packages/braintree-integration/src/braintree-paypal/braintree-paypal-payment-strategy.ts
bc-nick: Sorry for the late response. I think it is not necessary, we use this code only here and nowhere else

packages/braintree-integration/src/braintree-paypal/create-braintree-paypal-payment-strategy.ts
```diff
@@ -9,7 +9,8 @@ import {
     PaymentStrategyFactory,
     toResolvableModule,
 } from '@bigcommerce/checkout-sdk/payment-integration-api';
-import { Overlay } from '@bigcommerce/checkout-sdk/ui';
+import { LOADING_INDICATOR_STYLES } from '@bigcommerce/checkout-sdk/paypal-commerce-utils';
```
serhii-tkachenko, line 12: We should not use PayPal Commerce specific code in Braintree integration. We have 2 ways to solve this issue:
1) move `LOADING_INDICATOR_STYLES` into `ui` package and use it from there (but leave ppcp refactoring for another PR);
2) make a copy of `LOADING_INDICATOR_STYLES` and paste it somewhere in braintree related package;

packages/braintree-integration/src/braintree-paypal/braintree-paypal-payment-strategy.ts
```diff
@@ -316,4 +484,35 @@ export default class BraintreePaypalPaymentStrategy implements PaymentStrategy {
 
         throw new PaymentMethodFailedError(error.message);
     }
+
+    private isProviderError(error: unknown): boolean {
+        if (isPaypalCommerceProviderError(error)) {
```
serhii-tkachenko, line 489: We should not use paypal commerce related code inside brainteree strategy. Please create another typeguard for that.

packages/braintree-integration/src/braintree-paypal/braintree-paypal-payment-strategy.ts
```diff
@@ -257,11 +297,139 @@ export default class BraintreePaypalPaymentStrategy implements PaymentStrategy {
...
+            if (onError) {
+                onError(error);
+            }
 
-        if (this.braintreeHostWindow.paypal && isMessageContainerAvailable) {
+            throw error;
+        }
+    }
+
+    private renderPayPalMessages(initializationData?: any) {
```
serhii-tkachenko, line 421: `any` type is forbidden for any new code in checkout sdk

packages/braintree-integration/src/braintree-paypal/braintree-paypal-payment-strategy.ts
```diff
@@ -95,6 +115,22 @@ export default class BraintreePaypalPaymentStrategy implements PaymentStrategy {
...
+                this.paypalButtonRender?.close();
+
+                await this.loadPaypalCheckoutInstance();
+
+                await new Promise((_resolve, reject) => {
+                    if (onError && typeof onError === 'function') {
+                        onError(new Error('INSTRUMENT_DECLINED'));
+                    }
+
+                    reject();
```
serhii-tkachenko, line 130: Why do we need to reject promise in this case?

packages/braintree-integration/src/braintree-paypal/braintree-paypal-payment-strategy.ts
```diff
@@ -138,12 +178,17 @@ export default class BraintreePaypalPaymentStrategy implements PaymentStrategy {
...
+        if (token) {
             const state = await this.paymentIntegrationService.loadPaymentMethod(methodId);
 
             this.paymentMethod = state.getPaymentMethod(methodId);
 
-            return Promise.resolve({ ...payment, paymentData: this.formattedPayload(nonce) });
+            return Promise.resolve({
+                ...payment,
+                paymentData: this.formattedPayload(token),
+            });
```
serhii-tkachenko, line 191: It is not necessary to make `Promise.resolve` here

```suggestion
            return {
                ...payment,
                paymentData: this.formattedPayload(token),
            };
```

packages/braintree-integration/src/braintree-paypal/braintree-paypal-payment-strategy.ts
```diff
@@ -257,11 +297,139 @@ export default class BraintreePaypalPaymentStrategy implements PaymentStrategy {
...
+        } catch (error) {
+            if (onPaymentError) {
+                onPaymentError(error);
+            }
+
+            throw error;
+        }
+    }
+
+    private async tokenizePayment(
```
serhii-tkachenko, line 405:
```suggestion
    private async tokenizePaymentOrThrow(
```

packages/braintree-integration/src/braintree-paypal/braintree-paypal-payment-strategy.ts
```diff
@@ -247,8 +286,9 @@ export default class BraintreePaypalPaymentStrategy implements PaymentStrategy {
 
             await this.braintreeIntegrationService.getPaypalCheckout(
                 paypalCheckoutConfig,
-                () => {
-                    this.renderPayPalMessages(braintreeOptions?.bannerContainerId);
+                (braintreePaypalCheckout) => {
+                    this.renderPayPalMessages(initializationData);
+                    this.renderPayPalButton(braintreePaypalCheckout);
```
serhii-tkachenko, line 291: It looks a bit complicated.. because when you read the code you can see that `loadPaypalCheckoutInstance` method might be called up to 3 times, and it is not quite clear to me because we don't really render PayPal message or Button in all cases. 

We can put it on hold and refactor in the future, just want to let you know that here is a place, where the code should be updated.

serhii-tkachenko (changes requested):

packages/braintree-integration/src/braintree-paypal/braintree-paypal-payment-strategy.ts
bc-nick: Made decision to not pass this parameter, moved it to higher level

serhii-tkachenko (approved):
LGTM 👍

PR #2502, opened 2024-05-10
feat(payment): PAYPAL-2624 Removing paypa pro payment method

bc-nick:
## What?

Removing paypa pro payment method

## Why?

This implementation is deprecated for `optimise one-page checkout`, because if we already have `one-page checkout` with `paypal-pro` connected and we decide to switch to `optimise one-page checkout`, this provider will be disabled by default

## Testing / Proof

Verified by switching from `one-page` checkout to optimised `one-page checkout`


serhii-tkachenko (changes requested):
I don't really sure that we should do this now, since it might brake PayPal Payments Pro for current merchants who have this integration on their stores.

animesh1987:
@serhii-tkachenko maybe we can do an AQI to check for the merchants that use this integration?

serhii-tkachenko:
PayPal Payments Pro still can be onboarded when Feature_PaypalPaymentsProUS is enabled. Also I talked with Nick and he said, that we have merchants who still use this payment provider.

animesh1987:
> PayPal Payments Pro still can be onboarded when Feature_PaypalPaymentsProUS is enabled. Also I talked with Nick and he said, that we have merchants who still use this payment provider.

Thanks. Is that available in OOPC? I know PaypalPro is available for legacy checkout, but legacy checkout doesn't use SDK integration.

serhii-tkachenko:
> Is that available in OOPC?

@animesh1987 Yes. You can get an ability to setup PayPal Payments Pro for OOPC by enabling Feature_PaypalPaymentsProUS in Assets manager

PR #2504, opened 2024-05-13
feat(payment): PAYPAL-4123 Update address request body by adding digitalItems to lineItems

bc-nick:
## What?

Update address request body by adding digitalItems to lineItems

## Why?

If we try to place order with digital item we get an error by clicking on Pay button in the GP modal window
We need to update request body by adding digitalItems to lineItems to avoid this error

## Testing / Proof

Before

https://github.com/bigcommerce/checkout-sdk-js/assets/99336044/f6056cfc-7f1d-437c-b0af-aa08e848fbd8

After

https://github.com/bigcommerce/checkout-sdk-js/assets/99336044/f45b44de-4340-426f-8042-6367092dcb90

### DIFF
--- a/packages/core/src/shipping/consignment-action-creator.spec.ts
+++ b/packages/core/src/shipping/consignment-action-creator.spec.ts
@@ -957,6 +957,10 @@ describe('consignmentActionCreator', () => {
                             itemId: 'custom',
                             quantity: 1,
                         },
+                        {
+                            itemId: '667',
+                            quantity: 1,
+                        },
                     ],
                 },
                 options,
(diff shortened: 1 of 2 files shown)


bc-dronov (approved):

richapatel1510 (approved):
LGTM 👍

animesh1987 (changes requested):
Hey @bc-nick this might cause issues with regular consignments, since we don't need digital items.

I will suggest we probably need to find another way to fix this.

serhii-tkachenko:
Yeah... when customer is buying digital item with buy now cart, then we should not use consignment for this case at all

PR #2506, opened 2024-05-14
feat(payment): PAYPAL-4123 Avoiding calling of updateAddress for digitalItems

bc-nick:
## What?

Avoiding calling of `updateAddress` for `digitalItems`

## Why?

If we try to place order with digital item we get an error by clicking on Pay button in the GP modal window
We need to get rid of `updateAddress` request for  `digitalItems` on PDP

## Testing / Proof

Before

https://github.com/bigcommerce/checkout-sdk-js/assets/99336044/d6ff1cb2-f25e-48c5-b103-714f024acdb2

After

https://github.com/bigcommerce/checkout-sdk-js/assets/99336044/bfa573de-00b0-4ccd-8531-30804bdb1f42

### DIFF
--- a/packages/google-pay-integration/src/google-pay-button-strategy.ts
+++ b/packages/google-pay-integration/src/google-pay-button-strategy.ts
@@ -170,7 +170,7 @@ export default class GooglePayButtonStrategy implements CheckoutButtonStrategy {
             await this._paymentIntegrationService.updateBillingAddress(billingAddress);
         }
 
-        if (shippingAddress) {
+        if (shippingAddress && !this._buyNowCart?.lineItems.digitalItems.length) {
             await this._paymentIntegrationService.updateShippingAddress(shippingAddress);
         }
 
--- a/packages/google-pay-integration/src/google-pay-button-strategy.spec.ts
+++ b/packages/google-pay-integration/src/google-pay-button-strategy.spec.ts
@@ -163,6 +163,7 @@ describe('GooglePayButtonStrategy', () => {
 
                 expect(paymentIntegrationService.loadCheckout).not.toHaveBeenCalled();
                 expect(paymentIntegrationService.createBuyNowCart).toHaveBeenCalled();
+                expect(paymentIntegrationService.updateShippingAddress).not.toHaveBeenCalled();
             });
         });
 


bc-dronov (approved):

animesh1987 (dismissed):

packages/google-pay-integration/src/google-pay-button-strategy.ts
```diff
@@ -170,7 +170,7 @@ export default class GooglePayButtonStrategy implements CheckoutButtonStrategy {
             await this._paymentIntegrationService.updateBillingAddress(billingAddress);
         }
 
-        if (shippingAddress) {
+        if (shippingAddress && !this._buyNowCart?.lineItems.digitalItems.length) {
```
serhii-tkachenko, line 173: Should updateShippingAddress being called if customer has both digital and physical items in the cart?

animesh1987: Good point, missed this. I think in this case we will not call update address at all which leads to other issues.

bc-nick: Good point. In this case it seems like we should call. I will double check

bc-nick: But at the same time how is it possible?

bc-nick: We work with one item on PDP

bc-nick: I get it. We are talking about cart) I will double check

animesh1987: I agree we are talking about PDP, but this button is rendered in cart summary and cart page as well. I think this is something @serhii-tkachenko and I discussed a while ago around adding different strategies for buy-now buttons. That should simplify the logic immensely in the long term. 
Something to think about.

bc-nick: @serhii-tkachenko @animesh1987  In this case, everything works as expected. if we try to checkout from the cart with both items our checkout only works for the “Buy Now” feature, which means that if we have an update address, it will be updated successfully

Verified on dev store

serhii-tkachenko: Oh I see, you are using this._buyNowCart in the condition... Should be ok then.

serhii-tkachenko (approved):

PR #2511, opened 2024-05-16
feat(payment): PI-1578 Refactor the existing `sage-pay` payment strategy in Checkout SDK

bc-ania:
## What?
Refactor the existing sage-pay (Opayo by Elavon) payment strategy in Checkout SDK 

## Why?
To use the new payment integration API

## Testing / Proof
Screenshot 2024-05-20 at 15 44 29

https://github.com/bigcommerce/checkout-sdk-js/assets/130754705/4434ed52-47d8-4766-a978-f45ce0a0d42e


packages/core/src/payment/create-payment-strategy-registry.ts
```diff
@@ -523,17 +523,17 @@ export default function createPaymentStrategyRegistry(
...
+    //     PaymentStrategyType.SAGE_PAY,
+    //     () =>
+    //         new SagePayPaymentStrategy(
+    //             store,
+    //             orderActionCreator,
+    //             paymentActionCreator,
+    //             hostedFormFactory,
+    //             formPoster,
+    //         ),
+    // );
```
serhii-tkachenko, line 536: Do you still need this code?

bc-ania: No, forgot to delete it :)

packages/sagepay-integration/src/sage-pay-payment-strategy.ts
```diff
@@ -88,24 +80,21 @@ export default class SagePayPaymentStrategy extends CreditCardPaymentStrategy {
...
+    async finalize(options?: PaymentRequestOptions): Promise<void> {
+        const state = this.paymentIntegrationService.getState();
+        const order = state.getOrder();
 
-        if (order && state.payment.getPaymentStatus() === paymentStatusTypes.FINALIZE) {
-            return this._store.dispatch(
-                this._orderActionCreator.finalizeOrder(order.orderId, options),
-            );
+        if (order && state.getPaymentStatus() === PaymentStatusTypes.FINALIZE) {
+            await this.paymentIntegrationService.finalizeOrder(options);
```
serhii-tkachenko, line 88: By looking for prev implementation I see that `super.finalize()` should not be called for current condition. So it might be better to update this code with:
```suggestion
            return this.paymentIntegrationService.finalizeOrder(options);
```

vitalii-koshovyi (approved):

packages/sagepay-integration/src/sage-pay-payment-strategy.ts
bc-ania: Looks like I can't do it, I get error 'finalize' in 'SagePayPaymentStrategy' is not assignable to the same property in base type 'CreditCardPaymentStrategy', and it has to return Promise<void>

bc-peng (approved):
LGTM.

PR #2523, opened 2024-05-27
feat(payment): PI-2132 Call Klarna order initialization endpoint befo… …re showing Klarna pop-up window

bc-sebastianszafraniec:
## What?
Adding call Klarna order initialization endpoint before showing Klarna pop-up window

## Why?
To update Klarna checkout session.

## Testing / Proof
Tested manually

### DIFF
--- a/packages/klarna-integration/src/klarnav2/klarnav2-payment-strategy.ts
+++ b/packages/klarna-integration/src/klarnav2/klarnav2-payment-strategy.ts
@@ -1,12 +1,9 @@
 /* eslint-disable @typescript-eslint/naming-convention */
-import { RequestSender } from '@bigcommerce/request-sender';
 import { includes } from 'lodash';
 
 import {
     Address,
     BillingAddress,
-    ContentType,
-    INTERNAL_USE_ONLY,
     InvalidArgumentError,
     MissingDataError,
     MissingDataErrorType,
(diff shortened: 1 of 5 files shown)


packages/klarna-integration/src/klarnav2/klarnav2-payment-strategy.ts
```diff
@@ -257,4 +262,25 @@ export default class KlarnaV2PaymentStrategy {
             );
         });
     }
+
+    private async klarnaOrderInitialization(methodId: string): Promise<void> {
```
PavlenkoM, line 266: let's move this method to the [klarnav2-token-updater.ts](https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/klarna-integration/src/klarnav2/klarnav2-token-updater.ts)
and rename to something like `setKlarnaToken`

packages/klarna-integration/src/klarnav2/klarnav2-payment-strategy.ts
```diff
@@ -257,4 +262,25 @@ export default class KlarnaV2PaymentStrategy {
...
+        const state = this.paymentIntegrationService.getState();
+        const cartId = state.getCartOrThrow().id;
+        const clientToken = state.getPaymentMethodOrThrow(methodId).clientToken;
```
PavlenkoM, line 269:
```suggestion
        const { id: cartId } = state.getCartOrThrow();
        const { clientToken } = state.getPaymentMethodOrThrow(methodId);
```
let's keep the same code style

PavlenkoM (approved):

PR #2525, opened 2024-05-28
CHECKOUT-8284 Create new hosted-form package

richapatel1510:
## What?
Create a new hosted-form-v2 package for generating hosted form fields in manual order. 

## Why?
So that manual orders UI can import it and generate the hosted form fields (without importing the entire sdk in future). And eventually we can modify [My Account](https://github.com/bigcommerce/storefront-account-payments) and checkout to use the new implementation and remove the old one.

## Testing / Proof

- CI checks

### DIFF
--- a/packages/hosted-form-v2/README.md
+++ b/packages/hosted-form-v2/README.md
@@ -1,0 +1,11 @@
+# Hosted Form
+
+This library is for creating a hosted form that can be used outside of the checkout page.
+
+## Running unit tests
+
+Run `nx test hosted-form-v2` to execute the unit tests via [Jest](https://jestjs.io).
+
+## Running lint
+
+Run `nx lint hosted-form-v2` to execute the lint via [ESLint](https://eslint.org/).
--- a/packages/hosted-form-v2/src/hosted-field.ts
+++ b/packages/hosted-form-v2/src/hosted-field.ts
@@ -1,0 +1,186 @@
+import { values } from 'lodash';
+import { fromEvent } from 'rxjs';
+import { switchMap, take } from 'rxjs/operators';
+
+import { DetachmentObserver } from './common/dom';
+import { mapFromPaymentErrorResponse } from './common/errors';
+import { IframeEventListener, IframeEventPoster } from './common/iframe';
+import { parseUrl } from './common/url';
+import {
+    InvalidHostedFormConfigError,
+    InvalidHostedFormError,
+    InvalidHostedFormValueError,
+} from './errors';
+import { HostedFieldEvent, HostedFieldEventType } from './hosted-field-events';
+import HostedFieldType from './hosted-field-type';
+import { HostedFieldStylesMap } from './hosted-form-options';
+import HostedFormOrderData from './hosted-form-order-data';
+import {
+    HostedInputEventMap,
+    HostedInputEventType,
+    HostedInputSubmitErrorEvent,
+    HostedInputSubmitSuccessEvent,
+    HostedInputValidateEvent,
+} from './iframe-content';
+
+export const RETRY_INTERVAL = 60 * 1000;
+export const LAST_RETRY_KEY = 'lastRetry';
+
+export default class HostedField {
+    private _iframe: HTMLIFrameElement;
+
+    constructor(
+        private _type: HostedFieldType,
+        private _containerId: string,
+        private _placeholder: string,
+        private _accessibilityLabel: string,
+        private _styles: HostedFieldStylesMap,
+        private _eventPoster: IframeEventPoster<HostedFieldEvent>,
+        private _eventListener: IframeEventListener<HostedInputEventMap>,
+        private _detachmentObserver: DetachmentObserver,
+    ) {
+        this._iframe = document.createElement('iframe');
+
+        this._iframe.src = `/checkout/payment/hosted-field?version=${LIBRARY_VERSION}`;
+        this._iframe.style.border = 'none';
+        this._iframe.style.height = '100%';
+        this._iframe.style.overflow = 'hidden';
+        this._iframe.style.width = '100%';
+    }
+
+    getType(): HostedFieldType {
+        return this._type;
+    }
+
+    async attach(): Promise<void> {
+        const container = document.getElementById(this._containerId);
+
+        if (!container) {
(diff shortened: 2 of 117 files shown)


packages/hosted-form-v2/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This library is for creating a hosted form that can be used outside of the checkout page.
+
+## Running unit tests
+
+Run `nx test hosted-form` to execute the unit tests via [Jest](https://jestjs.io).
+
+## Running lint
+
+Run `nx lint hosted-form` to execute the lint via [ESLint](https://eslint.org/).
```
bc-peng, line 11: These appear to be inaccurate.

packages/hosted-form-v2/src/stored-card-hosted-form-service.ts
bc-peng, line 1: We may not want to handle a stored card right now, perhaps just name this class `HostedFormService`.

packages/hosted-form-v2/src/index.ts
```diff
@@ -0,0 +1,6 @@
+export { initializeHostedInput, notifyInitializeError } from './iframe-content';
```
davidchin, line 1: Can you provide an example of how other packages could consume this package?

richapatel1510: I think it is exported to be used in bigpay https://github.com/bigcommerce/bigpay/blob/1d0e8fc28ad1f17779662d8ab603d8ed04137b8e/app/views/payments/onsite/hosted_form_fields/show.html.erb#L2
https://github.com/bigcommerce/bigpay/blob/1d0e8fc28ad1f17779662d8ab603d8ed04137b8e/app/views/payments/onsite/hosted_form_fields/error.html.erb#L4

These exported functions are not used directly in checkout js, other than for e2e tests there to generate iframe 
https://github.com/bigcommerce/checkout-js/blob/918be1808209440f53b3053a8464abfbbe553260/packages/adyen-integration/e2e/support/hostedField.ejs#L10

davidchin: To clarify, I'm interest to know, once this package is done, how manual order can consume it to create hosted fields. And how it will eventually replace `hosted-form` v1 in the `core` package.

richapatel1510: Yes, that is still a open question that we need help with as mentioned in slack: `as to from where do we expose the HostedFormFieldFactory, do we create a new service in the  core package and export from there OR do we create a new HostedFormService in hosted-form-v2 and export it from core package because right now all we are able to use should be exported via core afaik?`

davidchin: I think the intention is to publish this as a standalone package so that manual orders UI can import it without bringing in all the unnecessary code of Checkout SDK. And eventually we can modify [My Account](https://github.com/bigcommerce/storefront-account-payments) and checkout to use the new implementation and remove the old one.

We can probably still achieve the same thing without creating a new monorepo package. But the monorepo setup can help enforce boundary. Having said that, we can defer the work if you feel it's quicker to modify the existing code in `core` instead.

No matter which approach we take, we still need to do some refactoring so that `HostedFormFieldFactory` doesn't depend on `CheckoutStore`: https://github.com/bigcommerce/checkout-sdk-js/blob/a6e895a8f1fe9a5e1dd849b345c800953d06a389/packages/core/src/hosted-form/hosted-form-factory.ts#L21 This is because it brings in a lot of other dependencies, which makes it hard to code split.

packages/hosted-form-v2/src/hosted-form-service.ts
```diff
@@ -0,0 +1,22 @@
...
+        const form = this._hostedFormFactory.create(this._host, options);
+
+        return form.attach().then(() => {
+            this._hostedForm = form;
+        });
+    }
+
+    deinitialize() {
+        if (this._hostedForm) {
+            this._hostedForm.detach();
```
davidchin, line 19: Should we unset `_hostedForm` after the detach call? i.e.:

```ts
this._hostedForm = undefined;
```

Otherwise, we will be holding onto a reference of an object that is no longer needed.

packages/hosted-form-v2/src/hosted-field.ts
```diff
@@ -0,0 +1,186 @@
...
+            return;
+        }
+
+        this._iframe.parentElement.removeChild(this._iframe);
+        this._eventListener.stopListen();
+    }
+
+    async submitForm(
+        fields: HostedFieldType[],
+        data: HostedFormOrderData,
```
davidchin, line 114: It depends what `HostedFormOrderData` will eventually contain. But if it varies depending on the context (checkout vs manual orders vs my account), it'd be better if we're using generic rather than a single interface.

Do you know if there's any data that needs to be submitted from the host window to BigPay?

richapatel1510: Makes sense, can use generic type here.

I am not sure about what data needs to be submitted for manual-orders. @bc-peng Do you know?

bc-peng: For the payload, we follow the new payment endpoint's YAML file.
https://github.com/bigcommerce/api-schema/blob/26a229103c0a285dea3e1850595814fe655e090a/bigpay/payments.yaml#L37
If we resue this method to submit all payment methods' payment, the `data` type will have many variations.

richapatel1510: @davidchin We will possibly need a new method to submit manual order payment and the method for existing payments can stay same, so I guess the type can be as it is in the core package for `checkoutjs` and `storefront-account-payments`. Can bring it over when we work on changes for using this package in those places.  
For manual order, the new submit method will be implemented as a part of [CHECKOUT-8275](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-8275)

davidchin: Ok, in that case, I will review the code around payment submission in the follow-up PR. 🙂

davidchin (approved):
LGTM! Look forward to the next iteration 👍

animesh1987 (approved):

PR #2528, opened 2024-06-03
CHECKOUT-8284: Fix build issue affecting `hosted-form-v2` package

davidchin:
## What?
Fix the build issue that is preventing `api-extractor` from compiling types defined in `hosted-form-v2` package. The problem is due to `api-extractor` not able to reference monorepo packages, such as `@bigcommerce/hosted-form-v2`. To fix this, I've created a new `build-dts` task within `hosted-form-v2` so that it can call `api-extractor` without referencing the monorepo package from `core`.

## Why?
This is an incremental step towards making `hosted-form-v2` a standalone package that can be used for Checkout, Manual Orders, and My Account.

## Testing / Proof
CircleCI

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/project.json
+++ b/packages/core/project.json
@@ -63,7 +63,8 @@
                 "commands": [
                     "tsc --outDir ../../temp --declaration --emitDeclarationOnly",
                     "api-extractor run --config api-extractor/checkout-sdk.json & api-extractor run --config api-extractor/checkout-button.json & api-extractor run --config api-extractor/embedded-checkout.json & api-extractor run --config api-extractor/internal-mappers.json",
-                    "rm -rf ../../temp"
+                    "rm -rf ../../temp",
+                    "nx run hosted-form-v2:build-dts"
                 ]
             }
         },
--- a/packages/core/src/bundles/checkout-sdk.ts
+++ b/packages/core/src/bundles/checkout-sdk.ts
@@ -1,5 +1,4 @@
 export { createTimeout } from '@bigcommerce/request-sender';
-
 export { createCheckoutService } from '../checkout';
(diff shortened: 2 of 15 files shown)


packages/core/project.json
```diff
@@ -63,7 +63,8 @@
                 "commands": [
                     "tsc --outDir ../../temp --declaration --emitDeclarationOnly",
                     "api-extractor run --config api-extractor/checkout-sdk.json & api-extractor run --config api-extractor/checkout-button.json & api-extractor run --config api-extractor/embedded-checkout.json & api-extractor run --config api-extractor/internal-mappers.json",
-                    "rm -rf ../../temp"
+                    "rm -rf ../../temp",
+                    "nx run hosted-form-v2:build-dts"
```
animesh1987, line 67: Just out of curiosity, shall we add this command here as we won't need to generate these files for all build since we won't be changing hosted-form-v2 as regularly. Maybe we can keep the type generation as a manual process for now?

davidchin: @animesh1987 Good call out. I think eventually we want to move the command out of `core`. But right now, the compilation of the script is still coupled with `core` and its webpack config. In the other words, the build process for `hosted-form-v2` is not entirely independent from `core` and further work is needed to fully separate them. That's why I think it's better to keep the DTS generation together for now. In terms of the build speed concern, NX should have caching in place to prevent unnecessary rebuild if the source files are unchanged.

richapatel1510 (approved):
👏 👏

packages/core/project.json
animesh1987: I think for the builds like this NX loses the caching, something we can verify 👍

animesh1987 (approved):

PR #2533, opened 2024-06-05
feat(checkout): PAYPAL-4200 provided initial state config for store creation on createCheckoutButtonInitializer

serhii-tkachenko:
## What?
Updated `createCheckoutButtonInitializer` with default initialization config param for `createCheckoutStore` function

## Why?
There was an issue that host option provided to `createCheckoutButtonInitializer` did not use in checkout api requests. Default request sender uses window.location.origin param as default when getHost method from store state returns empty string. The main problem is Headless stores do not have a chance to change default host, so checkout api calls fails due to the wrong domain origin, however payment api requests work as expected.

## Testing / Proof
Unit tests
Manual tests

Before this fix paypal commerce request used host option provided into createCheckoutButtonInitializer, while checkout-settings request used `window.location.origin`:
Screenshot 2024-06-05 at 10 34 22
Screenshot 2024-06-05 at 10 34 41

After the fix, all requests uses the same host and api calls do not fail:
Screenshot 2024-06-05 at 10 35 30

P.s.: This is a part of Headless eWallet Buttons initialitive

### DIFF
--- a/packages/core/src/checkout-buttons/create-checkout-button-initializer.spec.ts
+++ b/packages/core/src/checkout-buttons/create-checkout-button-initializer.spec.ts
@@ -1,11 +1,21 @@
 import { createFormPoster } from '@bigcommerce/form-poster';
 import { createRequestSender } from '@bigcommerce/request-sender';
 
+import { createCheckoutStore } from '../checkout';
+
 import CheckoutButtonInitializer from './checkout-button-initializer';
 import createCheckoutButtonInitializer from './create-checkout-button-initializer';
 
 jest.mock('@bigcommerce/form-poster');
 jest.mock('@bigcommerce/request-sender');
+jest.mock('../checkout', () => ({
+    createCheckoutStore: jest.fn(() => ({
+        getState: jest.fn().mockReturnValue({}),
+        subscribe: jest.fn(),
+    })),
+}));
+jest.mock('../payment-integration');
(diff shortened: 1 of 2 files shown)


bc-peng:
> Default request sender uses window.location.origin param as default 

If possible, Could you please show me where this happens?

serhii-tkachenko:
Hello @bc-peng 

When you get access to checkout by using `createCheckoutButtonsInitializer` you can pass custom host as a property, then this custom host passed to request sender inside a function and after that, request sender used in `CheckoutButtonStrategyRegistry` and `PaymentMethodRequestSender` (thats why payment requests work as expected). However, button strategies uses `PaymentIntegrationService` which uses request sender with host that it gets from store state: https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/payment-integration/create-payment-integration-service.ts#L55 which is empty. So when request sender gets empty host option, it will send a request to the url that passed to it (https://github.com/bigcommerce/request-sender-js/blob/e65f4779cd740d17f8a803f04c14ead4ec88e43c/src/request-sender.ts#L109). For example, if we do not pass host to request sender and try to make a request with `/test` url, then it will make a request to `https://<your_domain>/test`. In this case, to fix the issue with empty host we should pass the host as a part of checkout initial state to `createCheckoutStore` method.

animesh1987 (approved):

PR #2540, opened 2024-06-11
fix(payment): PI-2161 [TDOnlineMart] fix issue with vaulted card payment when mixed products in the cart

bc-wpietrzak:
## What?
Added additional check for digital items in the cart to decide whether or not to add verification nonce to `submitPayment` payload in TD Bank.

## Why?
With both digital and physical items in the cart, the `trustedShippingAddress` value is set to true, so the verification nonce wasn't added to the `submitPayment` payload. However, verification nonce seems to be required for payments when there are digital items in the cart.

## Testing / Proof
Unit + manual testing

Before:

Screenshot 2024-06-12 at 09 51 13

After:

Screenshot 2024-06-12 at 09 49 37

### DIFF
--- a/packages/td-bank-integration/src/td-online-mart-payment-strategy.spec.ts
+++ b/packages/td-bank-integration/src/td-online-mart-payment-strategy.spec.ts
@@ -10,6 +10,7 @@ import {
     PaymentIntegrationService,
 } from '@bigcommerce/checkout-sdk/payment-integration-api';
 import {
+    getCart,
     getInstruments,
     PaymentIntegrationServiceMock,
 } from '@bigcommerce/checkout-sdk/payment-integrations-test-utils';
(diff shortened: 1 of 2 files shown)


packages/td-bank-integration/src/td-online-mart-payment-strategy.spec.ts
```diff
@@ -368,6 +376,38 @@ describe('TDOnlineMartPaymentStrategy', () => {
                 );
             });
 
+            it('call submitPayment with verification nonce when there are both physical and digital items in the cart', async () => {
```
vitalii-koshovyi, line 379: should we add test when do we have at least one digitalItem?

vitalii-koshovyi: in the test description I see `both physical and digital items in the cart` but in the reality you set `lineItems: { digitalItems: [] },`
So please improve test or the title😄

bc-wpietrzak: Actually I've set `lineItems: { digitalItems: [] }` in `beforeEach` (so every test would have no digital items as default), but for this particular test I set `getCartOrThrow` to mock return `getCart()` (line 389), which contains both physical and digital items

vitalii-koshovyi: ok

vitalii-koshovyi (approved):

PR #2547, opened 2024-06-21
feat(payment): PAYPAL-4148 added styles to fastlane

andriiVitvitskyi1990:
## What?
Added styles for fastlane

## Why?
To be able to customise fastlane component

## Testing / Proof
Unit test
Screenshot 2024-06-28 at 13 53 52


packages/braintree-integration/src/braintree-fastlane/braintree-fastlane-payment-strategy.ts
```diff
@@ -19,10 +21,7 @@ import {
     PaymentRequestOptions,
     PaymentStrategy,
 } from '@bigcommerce/checkout-sdk/payment-integration-api';
-import {
-    isPayPalFastlaneCustomer,
-    PayPalCommerceInitializationData,
-} from '@bigcommerce/checkout-sdk/paypal-commerce-utils';
+import { isPayPalFastlaneCustomer } from '@bigcommerce/checkout-sdk/paypal-commerce-utils';
```
serhii-tkachenko, line 24: It might be better to find another method somewhere in Braintree package

andriiVitvitskyi1990: fixed

packages/braintree-utils/src/utils/get-valid-braintree-fastlane-styles.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    const fastlaneStyles: BraintreeFastlaneStylesOption = {};
+
+    getValidRootSettings(styles, fastlaneStyles);
+    getValidInputSettings(styles, fastlaneStyles);
+    getValidToggleSettings(styles, fastlaneStyles);
+    getValidTextStyles(styles, fastlaneStyles);
+
+    return Object.keys(fastlaneStyles).length ? fastlaneStyles : {};
```
serhii-tkachenko, line 17: We should avoid making such side effects and keep function as pure as possible. I recommend to return an object from each function that you use and place the result into fastlaneStyles object.

```suggestion
    if (!styles) {
        return undefined;
    }

    const rootStyles = getValidRootSettings(styles);
    const ... = getValidInputSettings(styles);
    const ... = getValidToggleSettings(styles);
    const ... = getValidTextStyles(styles);

    return {
        ...(rootStyles || {}),
        and so on
    };
```

serhii-tkachenko (approved):
Nice work! 👍

PR #2549, opened 2024-06-26
CHECKOUT-8275: Implement submit payment method for manual order

richapatel1510:
## What?
Implement `submitManualOrderPayment` method in HostedFormService.

## Why?
So that manual-order-ui can use this method to submit payment to bigpay with the hosted field values.

## Testing / Proof

- CI checks


packages/hosted-form-v2/src/hosted-field.ts
```diff
@@ -180,7 +175,9 @@ export default class HostedField {
             .map((link) => link.href);
     }
 
-    private _isSubmitErrorEvent(event: any): event is HostedInputSubmitErrorEvent {
-        return event.type === HostedInputEventType.SubmitFailed;
+    private _isSubmitManualOrderErrorEvent(
+        event: any,
```
davidchin, line 179:
```suggestion
        event: unknown,
```

Preferably avoid `any` as `unknown` forces you to perform more checks to make sure `event` is an object with `type` key. i.e.: 

```ts
if (typeof event !== 'object' || event !== null || 'type' in event === false) {
    return false;
}
```

packages/hosted-form-v2/src/hosted-form-factory.ts
```diff
@@ -23,6 +23,7 @@ export default class HostedFormFactory {
                 new HostedField(
                     type,
                     fieldOptions.containerId,
+                    fieldOptions.orderId,
```
davidchin, line 26: Considering `orderId` is identical for all fields, it's better to pass it at the `options` level rather than `fieldOptions` level so the caller doesn't need to repeat.

packages/hosted-form-v2/src/hosted-form-service.ts
```diff
@@ -20,4 +22,14 @@ export default class HostedFormService {
...
     }
+
+    async submitManualOrderPayment(data: HostedFormManualOrderData): Promise<void> {
+        const form = this._hostedForm;
+
+        if (!form) {
+            throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
+        }
+
+        await form.validate().then(() => form.submitManualOrderPayment({ data }));
```
davidchin, line 33: Avoid mixing `async/await` with `then` to keep the async syntax consistent.

packages/hosted-form-v2/src/payment/manual-order-payment-request-sender.ts
```diff
@@ -0,0 +1,47 @@
...
+        const options = {
+            headers: {
+                Accept: ContentType.Json,
+                'Content-Type': ContentType.Json,
+                'X-Payment-Session-Token': paymentSessionToken,
+            },
+            body: {
+                instrument: {
+                    type: 'card',
+                    name: instrumentFormData.cardName || '',
```
davidchin, line 30:
```suggestion
                    name: instrumentFormData.cardName ?? '',
```

packages/hosted-form-v2/src/payment/manual-order-payment-request-sender.ts
```diff
@@ -0,0 +1,47 @@
...
+                        year: Number(`20${expiryYear.trim()}`),
+                    },
+                    verification_value: instrumentFormData.cardCode ?? undefined,
+                },
+                payment_method_id: paymentMethodId,
+                form_nonce: nonce || undefined,
+            },
+        };
+
+        await this._requestSender.post<void>(`${this._paymentOrigin}/payments`, options);
```
davidchin, line 45: Do we need to read the API response? For example, would the API return something like 3DS is required and give back a redirect URL for the user to complete 3DS? reCaptcha is another example.

packages/hosted-form-v2/src/iframe-content/hosted-card-expiry-input.ts
```diff
@@ -25,7 +25,7 @@ export default class HostedCardExpiryInput extends HostedInput {
         eventPoster: IframeEventPoster<HostedInputEvent>,
         inputAggregator: HostedInputAggregator,
         inputValidator: HostedInputValidator,
-        paymentHandler: HostedInputPaymentHandler,
+        manualOrderPaymentHandler: HostedInputManualOrderPaymentHandler,
```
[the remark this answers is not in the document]
bc-peng: Hi Richa, could you please add a comment documenting the decision and reasoning for using `manualOrderPaymentHandler` instead of `paymentHandler` here?

packages/hosted-form-v2/src/hosted-field.ts
richapatel1510: Changing this to `unknown` results in build error.
Screenshot 2024-07-16 at 3 18 56 PM

davidchin: Yes, you will need to add something like [this](https://github.com/bigcommerce/checkout-sdk-js/pull/2549#discussion_r1678699052) before the return statement to narrow down the type first.

packages/hosted-form-v2/src/iframe-content/hosted-card-expiry-input.ts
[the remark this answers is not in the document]
richapatel1510: Right now this package is used for manual order only, in future when we implement it for checkout-js we can refactor the code and name it `paymentHandler` if the logic to derive the handler based on some params is easier.

packages/hosted-form-v2/src/hosted-field.ts
richapatel1510: Added the checks, but kept any since there are build errors on using `unknown` here.

davidchin (approved):
LGTM, nice work!

bc-peng (approved):
Nice work, 👏👏

PR #2551, opened 2024-06-26
fix(payment): STRIPE-385 add currencyService to payment-integration-api

PavlenkoM:
## What?
Move CurrencyService to payment-integration-api

## Why?
To reuse CurrencyService in GooglePay integration

## Testing / Proof
unit test

https://github.com/bigcommerce/checkout-sdk-js/assets/9430298/42820b0f-ab7a-4ba3-a0f3-72585d1e6576


vitalii-koshovyi (approved):

packages/payment-integration-api/src/common/utility/bind-decorator.spec.ts
```diff
@@ -1,4 +1,5 @@
-import { default as bind } from './bind-decorator';
+// eslint-disable-next-line max-classes-per-file
+import bind from './bind-decorator';
```
animesh1987, line 2: Why do we need to move this? I don't think this is the right place for his file.

PavlenkoM: Move because need this for CurrencyService
https://github.com/bigcommerce/checkout-sdk-js/pull/2551/files/7454d8b678631b01a90ec8ba27336fb06b6d41c3#diff-a7f3428af673045a6c3548fa5ef30f9604ce77734f04a373dca977466d5cccecR1

I moved `bindDecorator` from `core/.../common/utility/` to `payment-integration-api/.../common/utility/`.
Looks like it is only place for utilities in the packages. I didn't find any other package for utilities, maybe you know better place?

serhii-tkachenko (approved):

packages/payment-integration-api/src/common/utility/bind-decorator.spec.ts
bc-peng: As we don't have a utility package, `payment-integration-api` seems like the place to be.

animesh1987: I will say in this case we create a `utility` package, and move bind over there.

PavlenkoM: Ok, I will create `utility` package and move `bindDecorator ` here

PavlenkoM: done
@animesh1987 @bc-peng Can you look, please?

animesh1987 (approved):

PR #2559, opened 2024-07-02
chore(common): CHECKOUT-8410 Update to use node20

animesh1987:
## What?
Update to use node 20.

## Why?
Node 18 is already in maintenance mode, so it makes sense to move to next active version.

## Testing / Proof
- CI


bc-peng (approved):
LGTM👏👏

PR #2566, opened 2024-07-08
fix(billing): CHECKOUT-8392 Fix issue with billing address creation

animesh1987:
## What?
Fix issue with billing address creation, when a user logs in from control panel.

## Why?
We have an issue if a login happens from control panel we assign the guest cart to shopper but billing address object is not created, when the checkbox `billing address is same as shipping is unchecked`.

Which leads to an issue in billing step of checkout, since we try and use customer address id as billing address id for update. Hence add a check and for billing address.

## Testing / Proof
- CI
- Screencast

https://github.com/bigcommerce/checkout-sdk-js/assets/7134802/0a262b9c-9b49-46d1-b00e-66f67340d922

### DIFF
--- a/packages/core/src/billing/billing-address-action-creator.ts
+++ b/packages/core/src/billing/billing-address-action-creator.ts
@@ -33,6 +33,10 @@ export default class BillingAddressActionCreator {
         return (store) => {
             const state = store.getState();
             const checkout = state.checkout.getCheckout();
+            const isBillingFixExperimentEnabled =
+                state.config.getConfig()?.storeConfig.checkoutSettings.features[
+                    'CHECKOUT-8392.fix_billing_creation_in_checkout'
+                ] ?? false;
 
             if (!checkout) {
                 throw new MissingDataError(MissingDataErrorType.MissingCheckout);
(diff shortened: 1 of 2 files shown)


packages/core/src/billing/billing-address-action-creator.ts
```diff
@@ -33,6 +33,10 @@ export default class BillingAddressActionCreator {
         return (store) => {
             const state = store.getState();
             const checkout = state.checkout.getCheckout();
+            const isBillingFixExperimentEnabled =
+                state.config.getConfig()?.storeConfig.checkoutSettings.features[
+                    'CHECKOUT-8392.fix_billing_creation_in_checkout'
+                ] ?? false;
```
bc-peng, line 39: We might need to return `true` as fallback in order to safely remove an experiment in future.
Otherwise, removing the experiment will make the flow revert back to the time before the experiment rollout.

animesh1987: I did think about it, but wasn't sure. Convinced now, let me update it.

bc-peng (approved):
LGTM👏👏

richapatel1510 (approved):

PR #2572, opened 2024-07-17
feat(checkout): PI-2290 Adyen Credit Card installments implementation

vitalii-koshovyi:
## What?
Adyen Credit Card installments implementation

## Why?
Due to the feature request

## Testing / Proof

https://github.com/user-attachments/assets/0835cb4f-5caa-46a2-890d-5c4f8a87b579


packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.ts
```diff
@@ -86,6 +86,12 @@ export default class Adyenv3PaymentStrategy implements PaymentStrategy {
                 klarna_paynow: {
                     useKlarnaWidget: true,
                 },
+                card: {
+                    installmentOptions: {
+                        showInstallmentAmounts: true,
+                        ...installmentOptions,
```
serhii-tkachenko, line 92: It might be better to check if `installmentOptions` is an object first, because it might throw an error if the value will be null or undefined (or whatever value it might be except object)

serhii-tkachenko (approved):
Looks good to me except one little thing (left a comment)

PR #2574, opened 2024-07-22
feat(checkout): PI-1560 Refactor Digital River payment strategy in checkout-sdk

bc-ania:
## What?
Refactor Digital River payment strategy in checkout-sdk

## Why?
To use the new Payment API

## Testing / Proof

https://github.com/user-attachments/assets/d5e62ec2-56f6-4134-ac7f-ce189213243a

https://github.com/user-attachments/assets/314b0fa5-ee2c-4d9e-bb84-183737bebbf0

https://github.com/user-attachments/assets/69228b17-4570-40a7-9654-59bb1831dde3

https://github.com/user-attachments/assets/4d7f0038-50dd-464f-9c11-b78112ae3b98


packages/digitalriver-integration/src/digitalriver-payment-strategy.ts
```diff
@@ -43,32 +44,29 @@ export default class DigitalRiverPaymentStrategy implements PaymentStrategy {
...
         this._digitalRiverInitializeOptions = options.digitalriver;
 
-        const paymentMethod = this._store
+        const paymentMethod = this._paymentIntegrationService
             .getState()
-            .paymentMethods.getPaymentMethodOrThrow(options.methodId);
-        const { publicKey, paymentLanguage: locale } = paymentMethod.initializationData;
+            .getPaymentMethodOrThrow<DigitalRiverInitializationData>(options.methodId);
+        const { publicKey, paymentLanguage: locale } =
+            paymentMethod.initializationData as DigitalRiverInitializationData;
```
bc-nick, line 61: We need to get rid of using the as operator everywhere. Instead of it we have to create type guard in case where it is needed. 

But in your case there is no need, because you are using generic and `initializationData` alredy has an appropriate interface.

Just use it something like that: `const { ... } = paymentMethod.initializationData || {};`

packages/digitalriver-integration/src/digitalriver-payment-strategy.ts
```diff
@@ -43,32 +44,29 @@ export default class DigitalRiverPaymentStrategy implements PaymentStrategy {
...
     private _unsubscribe?: () => void;
     private _digitalRiverInitializeOptions?: DigitalRiverPaymentInitializeOptions;
-
     constructor(
-        private _store: CheckoutStore,
-        private _paymentMethodActionCreator: PaymentMethodActionCreator,
-        private _orderActionCreator: OrderActionCreator,
-        private _paymentActionCreator: PaymentActionCreator,
-        private _storeCreditActionCreator: StoreCreditActionCreator,
+        private _paymentIntegrationService: PaymentIntegrationService,
```
bc-nick, line 48: Need to get some clarification about using an underscore as the first char of a variable in the new packages.

As fas as i remember we have to wright method and variable names without them.

@animesh1987

packages/digitalriver-integration/src/digitalriver-payment-strategy.ts
bc-ania: If I do it this way: `const { ... } = paymentMethod.initializationData || {};` when I try to use publicKey later in the code, I have error ```
Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.
```

File: packages/digitalriver-integration/README.md [hunk 3]
```diff
@@ -0,0 +1,9 @@
+# digitalriver-integration
+
```
vitalii-koshovyi, line 2: please add a bit more info here

vitalii-koshovyi (approved):
Well done! Minor changes needed.

bc-peng (approved):

PR #2579, opened 2024-07-29
feat(shipping): SHIPPING-3183 Add support for channel_ids to the load shipping countries action

agchan12:
## What?
- Add support for displaying available shipping countries based on shipping method channel associations.

## Why?

To support this we need checkout to query the shipping/countries endpoint with a channel_id through as a query parameter.

## Testing / Proof

With the checkout-sdk-js and checkout-js running locally, the channel_id is piped through from checkout-sdk to checkout. The following video shows that the shipping/countries endpoint now has the query parameter channel_id passed through to it.

https://github.com/user-attachments/assets/39db2078-e888-4479-95fd-15107d74f6e2

ping @bigcommerce/team-checkout ping @bigcommerce/team-integrations

### DIFF
--- a/packages/core/src/checkout/checkout-service.ts
+++ b/packages/core/src/checkout/checkout-service.ts
@@ -528,7 +528,7 @@ export default class CheckoutService {
      * ```
      *
      * @alpha
-     * @param options - Options for loading the available shipping countries.
+     * @param query - Options for loading the available shipping countries.
      * @returns A promise that resolves to the current state.
      */
     loadPickupOptions(query: PickupOptionRequestBody): Promise<CheckoutSelectors> {
--- a/packages/core/src/shipping/shipping-country-action-creator.ts
+++ b/packages/core/src/shipping/shipping-country-action-creator.ts
@@ -1,20 +1,27 @@
 import { createAction, createErrorAction } from '@bigcommerce/data-store';
 import { Observable, Observer } from 'rxjs';
 
+import CheckoutStore from '../checkout/checkout-store';
 import { RequestOptions } from '../common/http-request';
 
 import { LoadShippingCountriesAction, ShippingCountryActionType } from './shipping-country-actions';
 import ShippingCountryRequestSender from './shipping-country-request-sender';
 
 export default class ShippingCountryActionCreator {
-    constructor(private _shippingCountryRequestSender: ShippingCountryRequestSender) {}
+    constructor(
+        private _shippingCountryRequestSender: ShippingCountryRequestSender,
+        private _store: CheckoutStore,
+    ) {}
 
     loadCountries(options?: RequestOptions): Observable<LoadShippingCountriesAction> {
+        const { checkout } = this._store.getState();
+        const { channelId } = checkout.getCheckoutOrThrow();
+
         return Observable.create((observer: Observer<LoadShippingCountriesAction>) => {
             observer.next(createAction(ShippingCountryActionType.LoadShippingCountriesRequested));
 
             this._shippingCountryRequestSender
-                .loadCountries(options)
+                .loadCountries(channelId, options)
                 .then((response) => {
                     observer.next(
                         createAction(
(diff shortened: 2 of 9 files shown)


packages/core/src/checkout/checkout-service.ts
```diff
@@ -502,7 +503,14 @@ export default class CheckoutService {
      * @returns A promise that resolves to the current state.
      */
     loadShippingCountries(options?: RequestOptions): Promise<CheckoutSelectors> {
-        const action = this._shippingCountryActionCreator.loadCountries(options);
+        const { checkout } = this._store.getState();
```
animesh1987, line 506: The change should not be here, but inside the action creator itself.

packages/core/src/shipping/shipping-country-request-sender.ts
```diff
@@ -6,8 +6,12 @@ import { CountryResponseBody } from '../geography';
 export default class ShippingCountryRequestSender {
     constructor(private _requestSender: RequestSender, private _config: { locale?: string }) {}
 
-    loadCountries({ timeout }: RequestOptions = {}): Promise<Response<CountryResponseBody>> {
-        const url = '/internalapi/v1/shipping/countries';
+    loadCountries(
+        channelId: number,
+        { timeout }: RequestOptions = {},
+    ): Promise<Response<CountryResponseBody>> {
+        const url = `/internalapi/v1/shipping/countries?channel_id=${channelId}`;
```
theromulans, line 13: To double check on this, we'll _always_ have a Channel ID?

theromulans (approved):

packages/core/src/shipping/shipping-country-request-sender.ts
agchan12: Yeah apparently, based on what Typescript is telling me

When I first load the [checkout-store](https://github.com/bigcommerce/checkout-sdk-js/pull/2579/files#diff-6a7b4a99bde46e7576bff595ea435584d8c920521ea217e9e913f6447b2179a6R18) it will load a definite [checkout](https://github.com/bigcommerce/checkout-sdk-js/blob/8c95f363d28bac904ac2b19db5096a5b4f37ee7b/packages/core/src/checkout/checkout-selector.ts#L17) or throw an exception.

And inside the [checkout](https://github.com/bigcommerce/checkout-sdk-js/blob/8c95f363d28bac904ac2b19db5096a5b4f37ee7b/packages/core/src/checkout/checkout.ts#L44) is a [mandatory channelId](https://github.com/bigcommerce/checkout-sdk-js/blob/8c95f363d28bac904ac2b19db5096a5b4f37ee7b/packages/core/src/checkout/checkout.ts#L44). It has to exist

packages/core/src/shipping/shipping-country-action-creator.ts
```diff
@@ -1,20 +1,27 @@
...
 export default class ShippingCountryActionCreator {
-    constructor(private _shippingCountryRequestSender: ShippingCountryRequestSender) {}
+    constructor(
+        private _shippingCountryRequestSender: ShippingCountryRequestSender,
+        private _store: CheckoutStore,
+    ) {}
 
     loadCountries(options?: RequestOptions): Observable<LoadShippingCountriesAction> {
+        const { checkout } = this._store.getState();
+        const { channelId } = checkout.getCheckoutOrThrow();
```
serhii-tkachenko, line 18: Are there any cases, when we can load shipping countries without having a checkout? @animesh1987 

My question is all about, does this lie by throwing an error brakes any flow?

animesh1987: This should cover all the clients, @agchan12 does it make sense to wrap this in an experiment?

animesh1987: @serhii-tkachenko we don't load countries without checkout id, as the only anomaly I could think was payment buttons but we do load checkout in the initialise methods for those.

agchan12: This will be wrapped in an experiment on the BCApp side of things. The idea is that checkout-sdk-js will send the `?channel_id={channelId}` query parameter through to the `shipping/countries` endpoint where the experiment will live.

We will only consume the `?channel_id={channel_id}` query parameter when the experiment is on.

animesh1987 (approved):

PR #2596, opened 2024-08-07
feat(payment): PI-2493 Add paymentHumanVerificationHandler method to …

bc-sebastianszafraniec:
…Payment Integration Service

## What?
Added paymentHumanVerificationHandler method to Payment Integration Service

## Testing / Proof
Tested manually

### DIFF
--- a/packages/core/src/payment-integration/default-payment-integration-service.ts
+++ b/packages/core/src/payment-integration/default-payment-integration-service.ts
@@ -20,6 +20,7 @@ import { DataStoreProjection } from '../common/data-store';
 import { CustomerActionCreator, CustomerCredentials } from '../customer';
 import { HostedFormFactory } from '../hosted-form';
 import { OrderActionCreator } from '../order';
+import { PaymentAdditionalAction } from '../payment';
 import {
     PaymentProviderCustomer,
     PaymentProviderCustomerActionCreator,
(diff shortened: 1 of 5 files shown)


packages/core/src/payment-integration/default-payment-integration-service.ts
```diff
@@ -285,4 +287,12 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
     async validateCheckout(checkout?: Checkout, options?: RequestOptions): Promise<void> {
         await this._checkoutValidator.validate(checkout, options);
     }
+
+    async handle(errorOrId: Error | string, key?: string): Promise<PaymentAdditionalAction> {
```
bc-nick, line 291: Need to think of a name for this function that is more specific to the action the function performs. @bc-sebastianszafraniec what do you think?

bc-sebastianszafraniec: @bc-nick yes, maybe it is better to use some more specific name. What would you say for `handlePaymentHumanVerification`

bc-nick: Much better for me than it is now

packages/core/src/payment-integration/default-payment-integration-service.ts
```diff
@@ -285,4 +287,15 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
     async validateCheckout(checkout?: Checkout, options?: RequestOptions): Promise<void> {
         await this._checkoutValidator.validate(checkout, options);
     }
+
+    async handlePaymentHumanVerification(
+        errorOrId: Error | string,
+        key?: string,
+    ): Promise<PaymentAdditionalAction> {
+        if (typeof errorOrId === 'string') {
```
bc-nick, line 295: I checked that we already have the same check inside of `_paymentHumanVerificationHandler.handle` method. Why did you duplicate it here?

bc-sebastianszafraniec: Without this check I was getting Typescript error here

bc-nick: Indeed, i did catch that we have a few declarations of `handle` function. But why we separated the declaration from implementation, i do not know. Anyway thanks for your updates!

bc-nick (approved):

animesh1987 (approved):

PR #2612, opened 2024-08-15
refactor(payment): PAYPAL-4578 updated ApplePay strategies with BraintreeSdk

serhii-tkachenko:
## What?
Updated apple pay strategies with `BraintreeSdk` class instead of `BraintreeIntegrationService`

## Why?
`BraintreeIntegrationService` is deprecated, so to be able to remove it in the future apple pay strategies should be updated

## Testing / Proof
Manual tests:
- successfully payed with apple pay on PDP page;
- successfully payed with apple pay on mini cart;
- successfully payed with apple pay on cart page;
- successfully payed with apple pay using button on top of checkout page;
- successfully payed with apple pay on payment step of checkout page;
(can't share videos, because macOs doesn't give an ability to pay with Apple Pay when screen recording is on)

Unit tests
CI

### DIFF
--- a/packages/apple-pay-integration/src/apple-pay-button-strategy.spec.ts
+++ b/packages/apple-pay-integration/src/apple-pay-button-strategy.spec.ts
@@ -3,8 +3,8 @@ import { getScriptLoader } from '@bigcommerce/script-loader';
 import { merge } from 'lodash';
 
 import {
-    BraintreeIntegrationService,
     BraintreeScriptLoader,
+    BraintreeSdk,
     getBraintree,
     getDataCollectorMock,
     getDeviceDataMock,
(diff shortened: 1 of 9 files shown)


bc-nick (approved):

animesh1987:
I feel we need to refactor these strategies separately to braintree ones, since with this approach we will have to pass braintree sdk everytime. Maybe its better to have a separate strategy.

serhii-tkachenko:
> I feel we need to refactor these strategies separately to braintree ones, since with this approach we will have to pass braintree sdk everytime. Maybe its better to have a separate strategy.

IMO In terms of maintenance its better to pass 1 extra class instead of creating separate strategies. I don't think it affects overall apple pay performance for other apple pay gateways... so it we might think about it when we have to inject some specific to other provider code

PR #2615, opened 2024-08-20
CHECKOUT-8300: Fix config and form field cache by ensuring input parameters are correctly compared

davidchin:
## What?
Fix the client-side cache for the `/api/storefront/checkout-settings` and `/api/storefront/form-fields` endpoints so that the cache is used when the CheckoutService#loadCheckout or CheckoutService#loadDefaultCheckout method is called multiple times.

## Why?
The client-side cache isn't working properly because the input parameter for `ConfigActionCreator` and `FormFieldActionCreator` includes the `params` field, which may contain a nested object (i.e.: `{ params: { checkoutId: id } }` and `{ params: { include: ['categoryNames'] } })`. The nested object breaks the caching as `@cacheableAction` decorator only works on shallow objects. As a result, when `ExtensionCommandType.ReloadCheckout` command is issued by an extension, we end up making repeated calls to these endpoints instead of re-using the local cache.

### Before

This video shows that local cache is not used when `ReloadCheckout` is called repeatedly.

https://github.com/user-attachments/assets/a12f7ea9-a474-4a76-9942-1a5e6b244698

### After

This video shows that local cache is used as expected when `ReloadCheckout` is called repeatedly.

https://github.com/user-attachments/assets/b9d716d8-9cc6-4275-adca-9d62f253dcba

### DIFF
--- a/packages/core/src/checkout/checkout-action-creator.ts
+++ b/packages/core/src/checkout/checkout-action-creator.ts
@@ -29,11 +29,14 @@ export default class CheckoutActionCreator {
                 of(createAction(CheckoutActionType.LoadCheckoutRequested)),
                 merge(
                     this._configActionCreator.loadConfig({
-                        ...options,
                         useCache: true,
-                        params: { ...options?.params, checkoutId: id },
+                        timeout: options?.timeout,
+                        params: { checkoutId: id },
                     }),
-                    this._formFieldsActionCreator.loadFormFields({ ...options, useCache: true }),
+                    this._formFieldsActionCreator.loadFormFields({
+                        useCache: true,
+                        timeout: options?.timeout,
+                    }),
                 ),
                 defer(() => {
                     return this._checkoutRequestSender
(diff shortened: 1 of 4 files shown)


packages/core/src/checkout/checkout-action-creator.ts
```diff
@@ -29,11 +29,14 @@ export default class CheckoutActionCreator {
                 of(createAction(CheckoutActionType.LoadCheckoutRequested)),
                 merge(
                     this._configActionCreator.loadConfig({
-                        ...options,
```
davidchin, line 32: We should be explicit about what parameters to pass into `ConfigActionCreator` and `FormFieldActionCreator` otherwise we could be passing more than needed. For instance, `params: { include: ['cart.lineItems.physicalItems.categoryNames', 'cart.lineItems.digitalItems.categoryNames'] }` is only intended for the checkout endpoint, not the settings or form fields endpoint.

animesh1987 (approved):

PR #2617, opened 2024-08-21
Revert "Revert "feat(shipping): SHIPPING-3183 Add support for channel_ids to the load shipping countries action""

agchan12:
## What?

This is a revert of the revert. i.e. we are re-raising this [PR](https://github.com/bigcommerce/checkout-sdk-js/pull/2579) 

This adds support for displaying available shipping countries based on shipping method channel associations.

## Why? 

To support this we need checkout to query the shipping/countries endpoint with a channel_id through as a query parameter.

## Testing/Proof

With the checkout-sdk-js and checkout-js running locally, the channel_id is piped through from checkout-sdk to checkout. The following video shows that the shipping/countries endpoint now has the query parameter channel_id passed through to it.

https://github.com/user-attachments/assets/39db2078-e888-4479-95fd-15107d74f6e2

ping @bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/checkout/checkout-service.spec.ts
+++ b/packages/core/src/checkout/checkout-service.spec.ts
@@ -432,7 +432,7 @@ describe('CheckoutService', () => {
             paymentMethodActionCreator,
             paymentStrategyActionCreator,
             new PickupOptionActionCreator(new PickupOptionRequestSender(requestSender)),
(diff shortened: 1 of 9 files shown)


theromulans:
Not needed?

bc-peng (approved):

PR #2631, opened 2024-08-30
feat(checkout): CHECKOUT-8587 Throw error when payment endpoint returns error/failure

richapatel1510:
## What?
Throw error when the payments endpoint returns `type: error/failure` with status 200.

## Why?
So that error can be better handled on manual order UI and appropriate message could be displayed.

## Testing / Proof

- CI checks

### DIFF
--- a/packages/hosted-form-v2/src/iframe-content/hosted-input-manual-order-payment-handler.ts
+++ b/packages/hosted-form-v2/src/iframe-content/hosted-input-manual-order-payment-handler.ts
@@ -62,18 +62,14 @@ export default class HostedInputManualOrderPaymentHandler {
                         error: { code: get(response.body, 'code') },
                     },
                 });
-            }
-
-            if (isError) {
+            } else if (isError) {
                 this._eventPoster.post({
                     type: HostedInputEventType.SubmitManualOrderFailed,
                     payload: {
                         error: { code: get(response.body, 'type') },
                     },
                 });
-            }
-
-            if (isSuccess) {
+            } else if (isSuccess) {
                 this._eventPoster.post({
                     type: HostedInputEventType.SubmitManualOrderSucceeded,
                 });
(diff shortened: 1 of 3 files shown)


packages/hosted-form-v2/src/iframe-content/hosted-input-manual-order-payment-handler.ts
```diff
@@ -44,15 +44,40 @@ export default class HostedInputManualOrderPaymentHandler {
...
+                        error: { code: get(response.body, 'type') },
+                    },
+                });
+            }
+
+            if (isSuccess) {
+                this._eventPoster.post({
+                    type: HostedInputEventType.SubmitManualOrderSucceeded,
+                });
+            }
```
bc-peng, line 80: Are we sure that failure, error, success can happen in one go?

richapatel1510: No, cannot happen in one go, however we would get only one type at a time. Updated it to use `if.. else if` instead.

packages/hosted-form-v2/src/payment/manual-order-payment-request-sender.ts
```diff
@@ -42,6 +42,6 @@ export class ManualOrderPaymentRequestSender {
             },
         };
 
-        await this._requestSender.post<void>(`${this._paymentOrigin}/payments`, options);
+        return this._requestSender.post<unknown>(`${this._paymentOrigin}/payments`, options);
```
animesh1987, line 45: Interesting fix, are we going to be validating the response in client implementations?

richapatel1510: No, we validate in payment handler, and post events only when we receive specific type https://github.com/bigcommerce/checkout-sdk-js/pull/2631/files#diff-2d86a89846731279b0343d2dde3dad8a5de52e16c8a547dc40b40408bd15ec5f

richapatel1510: Similar to the payments api implementation [here](https://github.com/bigcommerce/checkout-sdk-js/blob/b445659272efda3f381f1ed7332d8adef5da566d/packages/core/src/payment/strategies/ppsdk/step-handler/step-handler.ts#L16)

bc-peng (approved):

animesh1987 (approved):

PR #2653, opened 2024-09-17
feat(payment): PAYPAL-4611 Update Braintree LPMs strategy by adding functionality for Trustly

bc-nick:
## What?

Update Braintree LPMs strategy by adding functionality for Trustly

## Why?

To be able place an order via Braintree Trustly payment method

## Testing / Proof

All tests passed

### DIFF
--- a/packages/braintree-integration/src/braintree-local-payment-methods/braintree-local-methods-payment-strategy.ts
+++ b/packages/braintree-integration/src/braintree-local-payment-methods/braintree-local-methods-payment-strategy.ts
@@ -1,8 +1,10 @@
 import {
     BraintreeInitializationData,
     BraintreeIntegrationService,
+    BraintreeLocalPaymentMethodRedirectAction,
     LocalPaymentInstance,
     LocalPaymentsPayload,
+    NonInstantLocalPaymentMethods,
     onPaymentStartData,
     StartPaymentError,
 } from '@bigcommerce/checkout-sdk/braintree-utils';
@@ -20,6 +22,8 @@ import {
     PaymentStrategy,
 } from '@bigcommerce/checkout-sdk/payment-integration-api';
 import { LoadingIndicator } from '@bigcommerce/checkout-sdk/ui';
+
+import isPhoneNumberInstrumentLike from '../is-phone-number-instrument-like';
 
 import {
     BraintreeLocalMethods,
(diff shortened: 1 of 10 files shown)


packages/braintree-integration/src/braintree-local-payment-methods/braintree-local-methods-payment-strategy.ts
```diff
@@ -215,4 +263,27 @@ export default class BraintreeLocalMethodsPaymentStrategy implements PaymentStra
...
+    /**
+     *
+     * Utils
+     *
+     * */
+    private isNonInstantPaymentMethod(methodId: string): boolean {
+        return methodId.toUpperCase() in NonInstantLocalPaymentMethods;
+    }
+
+    private isBraintreeTrustlyRedirectError(error: unknown) {
```
serhii-tkachenko, line 276: I recommend to update this method with abstract naming, so we can use it for future implementations (for example if we will implement OXXO for Braintree or another payment method that might have such redirect flow)

```suggestion
    private isBraintreeRedirectError(error: unknown) {
```

packages/braintree-integration/src/braintree-local-payment-methods/braintree-local-methods-payment-strategy.ts
```diff
@@ -114,6 +118,50 @@ export default class BraintreeLocalMethodsPaymentStrategy implements PaymentStra
 
         this.toggleLoadingIndicator(true);
 
+        if (this.isNonInstantPaymentMethod(payment.methodId)) {
```
serhii-tkachenko, line 121: Do you like an idea to separate non instant payment method and default lpm execution logic? (you can create 2 different methods like executeWithNotInstantLPM and executeWithInstantLPM)

This change might improve execute method code readability

serhii-tkachenko (approved):
Overall looks good to me 👍

packages/braintree-integration/src/is-phone-number-instrument-like.ts
```diff
@@ -0,0 +1,14 @@
+import {
+    PaymentInstrumentPayload,
+    WithPhoneNumberInstrument,
+} from '@bigcommerce/checkout-sdk/payment-integration-api';
+
+export default function isPhoneNumberInstrumentLike(
```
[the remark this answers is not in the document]
bc-nick: Good point. I thought about it, but decided to leave it here. This applies to only one method for now.

packages/braintree-integration/src/braintree-local-payment-methods/braintree-local-methods-payment-strategy.ts
bc-nick: Make sense. Need to take a look at how it will look like

bc-peng (approved):
LGTM.👏👏

PR #2655, opened 2024-09-18
fix(checkout): PI-2633 removed modal window that appears during the frictionless 3ds flow for Adyen

vitalii-koshovyi:
## What?
The issue with an empty modal appeared during frictionless 3ds flow when the shopper didn't need to take any action.
The modal window appeared because Adyen needed to mount a hidden component to exchange all necessary information in the background through passive authentication using the shopper's device fingerprint.

When additional action is required during payment, a modal window renders to the DOM hidden by CSS, on some conditions it makes it visible.
For the implementation, I've passed the `onActionHandled` callback from the checkout-js that calls when an action, for example, a QR code or 3D Secure 2 authentication screen, is shown to the shopper. When the action type is not equal to the`3DS2Fingerprint` callback function will trigger.
From the checkout-js perspective, the `onActionHandled` call makes the modal visible by CSS.

## Why?
The issue with an empty modal appeared during frictionless 3ds flow when the shopper didn't need to take any action.

## Testing / Proof

Before fix:

https://github.com/user-attachments/assets/595d7f48-5b47-41c5-be37-cb51a90d4f11

After fix:

https://github.com/user-attachments/assets/651449c2-973c-4583-8666-defb00dd45b9


packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.ts
```diff
@@ -294,11 +294,19 @@ export default class Adyenv3PaymentStrategy implements PaymentStrategy {
...
             const { additionalActionOptions } = this._getPaymentInitializeOptions();
-            const { onBeforeLoad, containerId, onLoad, onComplete, widgetSize } =
+            const { onBeforeLoad, containerId, onLoad, onComplete, widgetSize, onActionHandled } =
                 additionalActionOptions;
             const adyenAction: AdyenAction = JSON.parse(additionalAction.action);
 
             const additionalActionComponent = this._getAdyenClient().createFromAction(adyenAction, {
+                onActionHandled: (additionalActionState) => {
+                    if (
+                        onActionHandled &&
```
serhii-tkachenko, line 304: It might be better to check that `onActionHandled` is a `function`

packages/adyen-utils/src/types.ts
```diff
@@ -1003,8 +1008,12 @@ export interface AdyenV2ThreeDS2ChallengeComponentOptions {
     onAdditionalDetails?(state: AdyenAdditionalActionState, component?: AdyenComponent): void;
     onError(error: AdyenError): void;
 }
-
+export interface ActionHandledReturnObject {
```
serhii-tkachenko, line 1011: [optional] Please, add extra empty line before and after this interface

serhii-tkachenko (approved):
Overall looks good to me 👍

PR #2674, opened 2024-10-02
CHECKOUT-8519: Switch messageformat library to be compatible with strict-dynamic CSP header

davidchin:
## What/Why?
Switch to using `intl-messageformat` instead of `messageformat` for translations, as the latter is incompatible with the CSP `script-src 'nonce-xxx' 'strict-dynamic'`  header used in checkout.

When translating messages at runtime, `messageformat` relies on `new Function` (similar to `eval`), which violates the `strict-dynamic` rule because it disallows any unsafe `eval` expressions (https://github.com/messageformat/messageformat/issues/180). In contrast, `intl-messageformat` does not use any unsafe `eval` expressions.

`intl-messageformat` can serve as a drop-in replacement for `messageformat`, as both libraries adhere to the [ICU MessageFormat](https://unicode-org.github.io/icu/userguide/format_parse/messages/) standards.

### Before
https://github.com/user-attachments/assets/1cd8f7c1-2df6-4f27-be88-1398476f552a

### After
https://github.com/user-attachments/assets/85d5949b-ab88-43f0-b51d-9eb355d319dd

### DIFF
--- a/packages/core/src/locale/language-config.ts
+++ b/packages/core/src/locale/language-config.ts
@@ -6,10 +6,28 @@ export default interface LanguageConfig {
     locale: string;
     locales: Locales;
     translations: Translations;
+    /**
+     * @hidden This property is intended for toggling an experimental change only.
+     */
+    isCspNonceExperimentEnabled?: boolean;
+}
+
+export interface TransformedLanguageConfig {
+    defaultTranslations: Translations;
+    defaultLocale?: string;
+    fallbackTranslations?: Translations;
+    fallbackLocale?: string;
+    locale: string;
+    locales: Locales;
+    translations: TransformedTranslations;
 }
 
 export interface Translations {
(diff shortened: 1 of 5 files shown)


packages/core/src/locale/language-config.ts
```diff
@@ -6,12 +6,28 @@ export default interface LanguageConfig {
     locale: string;
     locales: Locales;
     translations: Translations;
+    /** @hidden */
```
bc-peng, line 9: Just to confirm, is this comment for `typedoc`?

bc-peng (approved):

packages/core/src/locale/language-config.ts
davidchin: Yes, I will add a comment there to make it clearer. 🙂

PR #2687, opened 2024-10-08
feat(payment): PI-2777 Move itemsRequireShipping method to the utils

bc-sebastianszafraniec:
## What?
Move itemsRequireShipping method from google-pay-integration-package  to the utils so it can be use in other packages

## Testing / Proof
Tested manually


packages/payment-integration-api/src/cart/carts.mock.ts
```diff
@@ -0,0 +1,41 @@
...
+import { getDiscount } from '../discount/discounts.mock';
+
+import {
+    getCustomItem,
+    getDigitalItem,
+    getGiftCertificateItem,
+    getPhysicalItem,
+} from './line-items.mock';
+
+export default function getCart(): Cart {
```
bc-nick, line 13: Take a look at the `payment-integrations-test-utils` package. We already have this mock there.

And review all the other files pls. We need to avoid duplication and keep everything in one place

packages/payment-integration-api/src/config/config.mock.ts
```diff
@@ -0,0 +1,139 @@
+import { Config } from '../config';
+
+export default function getConfig(): Config {
```
vitalii-koshovyi, line 3: please use it from the payment-integrations-test-utils

packages/payment-integration-api/src/coupon/coupons.mock.ts
```diff
@@ -0,0 +1,21 @@
+import { Coupon } from '../coupon';
+
+export function getCoupon(): Coupon {
```
vitalii-koshovyi, line 3: please use it from the payment-integrations-test-utils

packages/payment-integration-api/src/currency/currency.mock.ts
```diff
@@ -0,0 +1,10 @@
+import { Currency } from '../currency';
+
+export function getCurrency(): Currency {
```
vitalii-koshovyi, line 3: please use it from the payment-integrations-test-utils

packages/payment-integration-api/src/discount/discounts.mock.ts
```diff
@@ -0,0 +1,8 @@
+import { Discount } from '../discount';
+
+export function getDiscount(): Discount {
```
vitalii-koshovyi, line 3: please use it from the payment-integrations-test-utils

packages/payment-integration-api/src/cart/carts.mock.ts
bc-sebastianszafraniec: @bc-nick @vitalii-koshovyi I saw that those mocks are in `payment-integrations-test-utils` but unfortunately I couldn't use it as it creates circular dependency. 
Screenshot 2024-10-10 at 15 03 31

I saw that in the past there were some duplicated mock added to `payment-integrtaion-api`, probably to overcome the same issue. I know it is not perfect solution but I am not sure if we have alternative option. What do you think?

vitalii-koshovyi: @serhii-tkachenko @PavlenkoM do you have better ideas?

vitalii-koshovyi: I believe we need to consider future refactoring and move mocks to `payment-integrtaion-api` package.|
@animesh1987 @bc-peng FYI

vitalii-koshovyi (approved):

packages/payment-integration-api/src/cart/carts.mock.ts
PavlenkoM: I don't have better idea.
Previously I had similar problem and discussed this with @bigcommerce/team-checkout , we decided that only solution to duplicate mocks.

PavlenkoM (approved):

packages/payment-integration-api/src/cart/carts.mock.ts
serhii-tkachenko: In this case it might be better to get rid of `payment-integrations-test-utils` and move them to `payment-integration-api` 👀

serhii-tkachenko (approved):

packages/payment-integration-api/src/cart/carts.mock.ts
bc-peng: Yes, I agree that it might be good for us to consider removing `test-utils` in order to avoid the circular dependencies   whenever we need to test anything requiring a cart mock in the `api`.

bc-peng (approved):
Before merging the PR, could we move all the mocks into a single folder? This will allow us to consolidate the utils package next. 🙇🏻‍♂️🙇🏻‍♂️

PR #2690, opened 2024-10-09
Restore "CHECKOUT-8519: Switch messageformat library to be compatible with strict-dynamic CSP header"

davidchin:
## What/Why?
Reverts bigcommerce/checkout-sdk-js#2682 and includes fixes to address the issues that led to the revert.

The problems were:
* The `translate` method throws an error if there are missing data for template variables.
* The `translate` method strips out `'` single quote character when used next to HTML tags because it is used as an [escape](https://unicode-org.github.io/icu/userguide/format_parse/messages/#quotingescaping) character.

### Manual tests 

This video shows that checkout works with `script-src` CSP header.

https://github.com/user-attachments/assets/aa5aeeed-2440-4571-9977-67aabcaa3c10

This screenshot shows that checkout can still be translated in a different language.

Screenshot 2024-10-10 at 7 52 40 AM

### Automated tests 

These are the new tests added to ensure that
* The `translate` method doesn't throw an error if there are missing data for template variables. - https://github.com/bigcommerce/checkout-sdk-js/pull/2690/files#diff-ecbbf79f48b829e4538ea79baeec00082cece0995095afea567b255576aceee6R72-R76
* The `translate` method can preserve `'` single quote character when used for quoting HTML attributes - https://github.com/bigcommerce/checkout-sdk-js/pull/2690/files#diff-ecbbf79f48b829e4538ea79baeec00082cece0995095afea567b255576aceee6R72-R76
* The `translate` method can still parse `'` and use it as an escape character when it is not used for quoting HTML attributes - https://github.com/bigcommerce/checkout-sdk-js/pull/2690/files#diff-ecbbf79f48b829e4538ea79baeec00082cece0995095afea567b255576aceee6R62-R66

This screenshot shows that `checkout-js` tests now pass with the new SDK release (https://github.com/bigcommerce/checkout-js/pull/2046).

Screenshot 2024-10-09 at 12 11 59 PM

### DIFF
--- a/packages/core/src/locale/language-service.spec.ts
+++ b/packages/core/src/locale/language-service.spec.ts
@@ -56,6 +56,12 @@ describe('LanguageService', () => {
             expect(langService.translate('test.thank_you_text')).toBe('<strong>Thank you<strong>');
         });
 
+        it('returns template string when values are missing for template variables', () => {
+            expect(langService.translate('test.order_number_text')).toBe(
+                'Your order number is {orderNumber}',
+            );
+        });
+
         it('pluralizes strings using ICU format', () => {
             expect(langService.translate('test.item_count_text', { count: 0 })).toBe('0 Items');
             expect(langService.translate('test.item_count_text', { count: 1 })).toBe('1 Item');
--- a/packages/core/src/locale/language-service.spec.ts
+++ b/packages/core/src/locale/language-service.spec.ts
@@ -25,6 +25,7 @@ describe('LanguageService', () => {
                 'optimized_checkout.test.continue_as_guest_action': 'Continue as guest',
                 'optimized_checkout.test.email_label': 'Email Address',
                 'optimized_checkout.test.order_number_text': 'Your order number is {orderNumber}',
+                'optimized_checkout.test.thank_you_text': '<strong>Thank you<strong>',
             },
         };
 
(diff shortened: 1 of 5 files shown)


packages/core/src/locale/language-service.spec.ts
```diff
@@ -70,7 +94,7 @@ describe('LanguageService', () => {
 
             expect(langService.translate('test.days_text', { count: 1 })).toBe('1 den');
             expect(langService.translate('test.days_text', { count: 2 })).toBe('2 dny');
-            expect(langService.translate('test.days_text', { count: 1.5 })).toBe('1.5 dne');
+            expect(langService.translate('test.days_text', { count: 1.5 })).toBe('1,5 dne');
```
richapatel1510, line 97: Curious to know why we need this change? Shouldn't it stay `1.5`?

davidchin: @richapatel1510 Great question. `1.5` should actually formatted as `1,5` for the `cs` locale. This is how number should be displayed in Czech. Displaying the decimal symbol as `.` is actually a bug. It has been fixed since `messageformat@3.0.0` (See https://github.com/messageformat/messageformat/commit/20d6fdbc9a30e49aa0bbad3a7bcc07625cb9d008). The alternative library `intl-messageformat` behaves the same way as the newer version of `messageformat`.

bc-peng (approved):

richapatel1510 (approved):

PR #2697, opened 2024-10-15
WIP fix(checkout): CHECKOUT-8264 Don't validate recaptcha if already validated

animesh1987:
## What?
Don't validate recaptcha if it's not needed or is already validated.

## Why?
- TBD

## Testing / Proof
- TBD

### DIFF
--- a/packages/core/src/spam-protection/spam-protection-action-creator.ts
+++ b/packages/core/src/spam-protection/spam-protection-action-creator.ts
@@ -1,5 +1,5 @@
-import { createAction, ThunkAction } from '@bigcommerce/data-store';
-import { concat, defer, from, of } from 'rxjs';
+import { createAction, ReadableDataStore, ThunkAction } from '@bigcommerce/data-store';
+import { concat, defer, from, Observable, of } from 'rxjs';
 import { catchError, switchMap, take } from 'rxjs/operators';
 
 import { InternalCheckoutSelectors } from '../checkout';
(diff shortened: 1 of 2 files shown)


packages/core/src/spam-protection/spam-protection-action-creator.ts
```diff
@@ -99,32 +99,53 @@ export default class SpamProtectionActionCreator {
...
+            const { shouldExecuteSpamCheck } = checkout.getCheckoutOrThrow();
+
+            const recaptchaExperiment = features['CHECKOUT-8264.recaptcha_error_fix'];
+
+            if (recaptchaExperiment) {
+                if (shouldExecuteSpamCheck) {
+                    return this._executeRecaptcha(store);
+                }
+            } else {
+                return of(createAction(SpamProtectionActionType.ExecuteActionNotNeeded, undefined));
```
davidchin, line 116: Could we change the `CustomerActionCreator` to check the `shouldExecuteSpamCheck` flag instead, similar to how the check is implemented in `PaymentStrategyActionCreator`: https://github.com/bigcommerce/checkout-sdk-js/blob/517152be10a54238079315e99dc67ac5bb438b4a/packages/core/src/payment/payment-strategy-action-creator.ts#L60-L62

This approach will reduce the risk of change as it won't impact the payment step, which already has a check in place as shown above.

animesh1987: Thanks DC. That seems good, that was my first thought as well 😄

PR #2699, opened 2024-10-15
feat(payment): PAYPAL-4737 Apple Pay button style does not apply to storefront

bc-nick:
## What?

Added styling for Apple Pay button

## Why?

To not use styling from strorefront and for consistency

## Testing / Proof

https://github.com/user-attachments/assets/e0639db2-cddb-47fa-a974-dcf35063ee69

https://github.com/user-attachments/assets/88241155-1567-4e3d-822b-113c27fbe5b5

### DIFF
--- a/packages/apple-pay-integration/src/apple-pay-button-strategy.ts
+++ b/packages/apple-pay-integration/src/apple-pay-button-strategy.ts
@@ -34,9 +34,50 @@ enum DefaultLabels {
     Shipping = 'Shipping',
 }
 
+enum ButtonStyleOption {
+    Black = 'black',
+    White = 'white',
+    WhiteBorder = 'white-border',
+}
+
 function isShippingOptions(options: ShippingOption[] | undefined): options is ShippingOption[] {
     return options instanceof Array;
 }
+
+const getApplePayButtonStyle = (option?: ButtonStyleOption): Record<string, string> => {
+    const defaultStyle: Record<string, string> = {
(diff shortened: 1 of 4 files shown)


packages/apple-pay-integration/src/apple-pay-button-strategy.ts
```diff
@@ -116,13 +156,12 @@ export default class ApplePayButtonStrategy implements CheckoutButtonStrategy {
...
         }
 
-        document.body.classList.add('apple-pay-supported');
-
         const button = document.createElement('div');
 
-        button.setAttribute('class', buttonClassName);
         button.setAttribute('role', 'button');
         button.setAttribute('aria-label', 'Apple Pay button');
+        Object.assign(button.style, getApplePayButtonStyle(styleOption));
```
serhii-tkachenko, line 163: It might be better to add a condition to run this code only if `styleOption` is available.

serhii-tkachenko (commented):
Please add unit tests for changes that you made

packages/apple-pay-integration/src/apple-pay-button-strategy.ts
bc-nick: We do not need to add a condition here to check if the `styleOption` is present or not, because if it is not, we will get the default style configuration

serhii-tkachenko (approved):

PR #2702, opened 2024-10-16
refactor(checkout): PI-2691 clenup lagacy integration spec error

vitalii-koshovyi:
## What?
Clenup legacy integration spec error

## Testing / Proof
image

### DIFF
--- a/packages/legacy-integration/.eslintrc.json
+++ b/packages/legacy-integration/.eslintrc.json
@@ -15,5 +15,5 @@
                 "@typescript-eslint/no-unsafe-member-access": "off"
             }
         }
-    ]    
+    ]
 }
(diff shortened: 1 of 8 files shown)


packages/legacy-integration/.eslintrc.json
```diff
@@ -10,10 +10,11 @@
         {
             "files": ["*.spec.ts"],
             "rules": {
+                "@typescript-eslint/no-unsafe-assignment": "off",
```
animesh1987, line 13: Why are we turning this rule off?

vitalii-koshovyi: Thanks for pointing out! 
I've added it due to the broken type in the payment-integration-service.mock.
Fixed it, removed rule where it was added, due to the type fixing.

PavlenkoM (approved):

PR #2704, opened 2024-10-17
feat(shipping): SHIPPING-3311 Update loadShippingCountries to allow for a default of no channel id

agchan12:
## What?
- Allow for a default of `null` channel ID to be sent to the the `v1/shipping/countries` endpoint

## Why?
Merchants who attempt to call the `loadShippingCountries` endpoint before initialising a checkout will run into issues with the existing `getCheckoutOrThrow()` behaviour because the checkout is not yet initialised.

With this updated code, if there is no checkout, we set the `channel id` to null and proceed to send no channel ID to the `v1/shipping/countries` endpoint. 

This will result in the existing behaviour before we introduced the `channel id` property. 

As a result of this, it will mean that merchants will view the country selector based on the shipping zone location and NOT have it be channel aware.

## Testing / Proof
Existing Unit Test

ping @animesh1987 @bc-peng

### DIFF
--- a/packages/core/src/shipping/shipping-country-request-sender.ts
+++ b/packages/core/src/shipping/shipping-country-request-sender.ts
@@ -7,10 +7,11 @@ export default class ShippingCountryRequestSender {
     constructor(private _requestSender: RequestSender, private _config: { locale?: string }) {}
 
     loadCountries(
-        channelId: number,
+        channelId: number | null,
         { timeout }: RequestOptions = {},
     ): Promise<Response<CountryResponseBody>> {
-        const url = `/internalapi/v1/shipping/countries?channel_id=${channelId}`;
+        const channelIdParam = channelId ? `?channel_id=${channelId}` : '';
+        const url = `/internalapi/v1/shipping/countries${channelIdParam}`;
 
         const headers = {
             'Accept-Language': this._config.locale,
--- a/packages/core/src/shipping/shipping-country-action-creator.spec.ts
+++ b/packages/core/src/shipping/shipping-country-action-creator.spec.ts
@@ -71,5 +71,23 @@ describe('ShippingCountryActionCreator', () => {
                     ]);
                 });
         });
+
+        describe('when checkout is undefined', () => {
+            beforeEach(() => {
+                store = createCheckoutStore({ checkout: undefined });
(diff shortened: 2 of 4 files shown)


packages/core/src/shipping/shipping-country-request-sender.ts
```diff
@@ -7,10 +7,12 @@ export default class ShippingCountryRequestSender {
     constructor(private _requestSender: RequestSender, private _config: { locale?: string }) {}
 
     loadCountries(
-        channelId: number,
+        channelId: number | null,
```
bc-peng, line 10:
```suggestion
        channelId?: number,
```

packages/core/src/shipping/shipping-country-request-sender.ts
```diff
@@ -7,10 +7,12 @@ export default class ShippingCountryRequestSender {
...
         { timeout }: RequestOptions = {},
     ): Promise<Response<CountryResponseBody>> {
-        const url = `/internalapi/v1/shipping/countries?channel_id=${channelId}`;
+        const url = `/internalapi/v1/shipping/countries${
+            channelId !== null ? `?channel_id=${channelId}` : ''
+        }`;
```
bc-peng, line 15:
```suggestion
        const channelIdParam = channelId ? `?channel_id=${channelId}` : '';
        const url = `/internalapi/v1/shipping/countries${channelIdParam}`;
```

bc-peng (commented):
Just to confirm, we don't want to enforce the use of a channel id in the request now?

packages/core/src/shipping/shipping-country-request-sender.ts
agchan12: I believe TS will complain about this in the shipping-action-country-selector.ts about it not being assignable to type number.

bc-peng: Didn't check that file.
As you said, this cannot be optional here.

agchan12:
> Just to confirm, we don't want to enforce the use of a channel id in the request now?

Yeah we won't be enforcing the channel id anymore.

bc-peng (approved):
LGTM.👏👏

theromulans:
Thanks @bc-peng!

PR #2725, opened 2024-10-31
feat(analytics): DATA-11983 Populate discount_amount field for BODL events

bc-rmalyavc:
## What?
Introduce a new field `coupon_amount` for `BODL` events

## Why?
We need to send the discount data for GA4 events. Currently checkout/order related BODL events only populate the automatic discounts on item level and discount amount is not taken into account. So that we need to add this field for BODL line items so that we can calculate the item discounts correctly for GA4 events

## Testing / Proof
- Unit tests
- Manually on dev store. Verified that the new field is being populated for all checkout/order related BODL events on item level and reflects the amount of discount given by a coupon

https://github.com/user-attachments/assets/9b6d1734-8e0b-4914-b887-99812d530e15


serhii-tkachenko (approved):
LGTM 👍

vitalii-koshovyi (approved):

bc-peng (approved):

PR #2738, opened 2024-11-21
fix(extension): CHECKOUT-8869 Fix Display Issue after Page Refresh

bc-peng:
## What?
Fix the issue where the checkout extension in the selected shipping option fails to display after a page refresh, even though it works fine on the first run.

## Why?
When an extension iframe is removed before it has finished loading, particularly in cases where there is an existing consignment, this action creates a pending extension rendering promise, which can only be resolved or rejected by either a 60-second timeout or a message from the iframe.

Related code:
https://github.com/bigcommerce/checkout-sdk-js/blob/2e6f95f3718a37ef48e1c479136a699e95e6db6f/packages/core/src/extension/resizable-iframe-creator.ts#L91

This PR resolves the issue by rejecting the promise upon detecting the removal of the iframe from the DOM.

### Manual testing

https://github.com/user-attachments/assets/3a4e0ab1-ae51-4c0a-9b9f-892e3094b376

### DIFF
--- a/packages/core/src/extension/resizable-iframe-creator.ts
+++ b/packages/core/src/extension/resizable-iframe-creator.ts
@@ -1,3 +1,4 @@
+import { DetachmentObserver, MutationObserverFactory } from '../common/dom';
 import { IFrameComponent, iframeResizer, isIframeEvent } from '../common/iframe';
 import { parseUrl } from '../common/url';
 
@@ -40,15 +41,17 @@ export default class ResizableIframeCreator {
         );
     }
 
-    private _toResizableFrame(
+    private async _toResizableFrame(
         iframe: HTMLIFrameElement,
         timeoutInterval: number,
         initCallback: () => void,
         failedCallback: () => void,
     ): Promise<IFrameComponent> {
+        const detachmentObserver = new DetachmentObserver(new MutationObserverFactory());
+
         // Can't simply listen to `load` event because it always gets triggered even if there's an error.
         // Instead, listen to the `load` inside the iframe and let the parent frame know when it happens.
-        return new Promise((resolve, reject) => {
+        const promise = new Promise<IFrameComponent>((resolve, reject) => {
             const timeout = window.setTimeout(() => {
                 failedCallback();
 
@@ -90,5 +93,7 @@ export default class ResizableIframeCreator {
 
             window.addEventListener('message', handleMessage);
         });
+
+        return detachmentObserver.ensurePresence([iframe], promise);
     }
 }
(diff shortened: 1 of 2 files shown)


packages/core/src/extension/resizable-iframe-creator.ts
```diff
@@ -90,5 +93,12 @@ export default class ResizableIframeCreator {
 
             window.addEventListener('message', handleMessage);
         });
+
+        return detachmentObserver.ensurePresence([iframe], promise);
```
bc-peng, line 97: I have tried to call `failedCallback` method in this way, but it was not ideal.
```
return detachmentObserver.ensurePresence([iframe], promise).catch((error) => {
            failedCallback();

            throw error;
        });
```
Because in our extension implementation, `failedCallback` assumes that the iframe is still present.
https://github.com/bigcommerce/checkout-sdk-js/blob/2e6f95f3718a37ef48e1c479136a699e95e6db6f/packages/core/src/extension/extension-iframe.ts#L44

davidchin: I see you are throwing an exception here. What happens on checkout side? Does the exception get caught?

bc-peng: On checkout-js side, we didn't catch it yet.

https://github.com/bigcommerce/checkout-js/blob/6fd4945e82c3d5748014e4d55d2a382a5837858d/packages/checkout-extension/src/ExtensionService.ts#L61

Is it a good idea that we catch it and ignore the error if it is `UnexpectedDetachmentError`?

animesh1987: I suppose we should still handle the exception, maybe a logger message. Atleast so we don't blow up checkout.

davidchin: Preferably unhandled promise rejections should be caught, as they are typically flagged by error tracking tools such as Sentry and Noibu, even if they don’t impact user experience.

bc-peng: Hi @animesh1987 and @davidchin, I have added a try/catch to handle errors in `failedCallback`.

On `checkout-js` side, I opened a PR to catch errors in `checkoutService.renderExtension`. This PR does not depend on SDK. https://github.com/bigcommerce/checkout-js/pull/2105

bc-peng: Updated testing video.

https://github.com/user-attachments/assets/2022684f-608f-47f7-9337-3667ecca2180

davidchin (approved):

animesh1987 (approved):

PR #2742, opened 2024-11-27
feat(payment): PAYPAL-4867 POC of headless wallet buttons

bc-nick:
## What?

POC of Headless Wallet Buttons

## Why?

According to the purpose of the project

## Testing / Proof

https://github.com/user-attachments/assets/49e4bbcc-e356-4a04-8e16-2367cdc0f2a6


packages/core/auto-export.config.json
```diff
@@ -14,6 +14,11 @@
             "inputPath": "packages/*/src/index.ts",
             "outputPath": "packages/core/src/generated/checkout-button-strategies.ts",
             "memberPattern": "^create.+ButtonStrategy$"
+        },
+        {
+            "inputPath": "packages/*/src/index.ts",
+            "outputPath": "packages/core/src/generated/checkout-headless-button-strategies.ts",
+            "memberPattern": "^create.+HeadlessButtonStrategy$"
```
bc-nick, line 21: Need to think about the right pattern for headless strategies. `test-core` **is failed** because for button strategies we have the same substring (`ButtonStrategy`) as for headless button strategies in `memberPattern` field. For this reason, we do not have a unique token. Should be replaced with one of the following examples: `HeadlessWalletStrategy`, `HeadlessStrategy` or something like that

packages/core/src/loader-cdn.ts
```diff
@@ -2,18 +2,21 @@ import { getScriptLoader } from '@bigcommerce/script-loader';
...
 
 export type CheckoutButtonBundle = typeof checkoutButtonBundle & { version: string };
+export type CheckoutHeadlessButtonBundle = typeof checkoutHeadlessButtonBundle & { version: string };
 export type EmbeddedCheckoutBundle = typeof embeddedCheckoutBundle & { version: string };
 export type HostedFormBundle = typeof hostedFormBundle & { version: string };
 export type MainBundle = typeof mainBundle & { version: string };
 
 export enum BundleType {
     CheckoutButton = 'checkout-button',
+    CheckoutHeadlessButton = 'checkout-headless-button',
```
serhii-tkachenko, line 19:
```suggestion
    HeadlessCheckoutButton = 'headless-checkout-button',
```

packages/core/src/checkout-buttons/checkout-button-initializer-options.ts
```diff
@@ -1,4 +1,6 @@
 export default interface CheckoutButtonInitializerOptions {
     host?: string;
     locale?: string;
+    storefrontJwtToken?: string;
+    siteLink?: string;
```
serhii-tkachenko, line 5: Why `siteLink` should be created when we can use `host` instead of it? As far as I remember host is equal to site link

packages/paypal-commerce-integration/src/paypal-commerce/paypal-commerce-headless-button-strategy.ts
```diff
@@ -0,0 +1,203 @@
...
+        const paypalButton = paypalSdk.Buttons(buttonRenderOptions);
+
+        if (paypalButton.isEligible()) {
+            paypalButton.render(`#${containerId}`);
+        } else {
+            this.paypalCommerceIntegrationService.removeElement(containerId);
+        }
+    }
+
+    private async onHostedCheckoutApprove(
```
serhii-tkachenko, line 112: This method should be available only in Phase 2 of this project

packages/paypal-commerce-integration/src/paypal-commerce/paypal-commerce-headless-button-strategy.ts
```diff
@@ -0,0 +1,203 @@
...
+        };
+
+        const onShippingChangeCallbacks = {
+            onShippingAddressChange: (data: ShippingAddressChangeCallbackPayload) =>
+                this.onShippingAddressChange(data),
+            onShippingOptionsChange: (data: ShippingOptionChangeCallbackPayload) =>
+                this.onShippingOptionsChange(data),
+        };
+
+        const hostedCheckoutCallbacks = {
```
serhii-tkachenko, line 90: This can be removed

packages/paypal-commerce-integration/src/paypal-commerce-credit/paypal-commerce-credit-headless-button-strategy.ts
```diff
@@ -0,0 +1,228 @@
...
+                this.onShippingAddressChange(data),
+            onShippingOptionsChange: (data: ShippingOptionChangeCallbackPayload) =>
+                this.onShippingOptionsChange(data),
+        };
+
+        const hostedCheckoutCallbacks = {
+            ...onShippingChangeCallbacks,
+            onApprove: (data: ApproveCallbackPayload, actions: ApproveCallbackActions) =>
+                this.onHostedCheckoutApprove(data, actions, methodId),
+        };
```
serhii-tkachenko, line 102: This part should be removed since it is outside of Headless WB Phase 1 scope

packages/paypal-commerce-integration/src/create-paypal-commerce-integration-service.ts
```diff
@@ -13,10 +13,10 @@ import {
 const createPayPalCommerceIntegrationService = (
     paymentIntegrationService: PaymentIntegrationService,
 ) => {
-    const { getHost } = paymentIntegrationService.getState();
+    const { getHost, getSiteLink } = paymentIntegrationService.getState();
 
     return new PayPalCommerceIntegrationService(
-        createFormPoster(),
+        createFormPoster({ host: getSiteLink() }),
```
serhii-tkachenko, line 19: Can getHost be used here instead of getSiteLink?

packages/core/src/cart/cart-action-creator.ts
```diff
@@ -0,0 +1,53 @@
...
+        cartId: string,
+        options?: RequestOptions & ActionOptions,
+    ): ThunkAction<LoadCartAction, InternalCheckoutSelectors> {
+        return (store) => {
+            return Observable.create((observer: Observer<LoadCartAction>) => {
+                const state = store.getState();
+                const jwtToken = state.config.getStorefrontJwtToken();
+
+                if (!jwtToken) {
+                    throw new MissingDataError(MissingDataErrorType.MissingPaymentToken);
```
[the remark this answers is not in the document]
bc-nick: Yeap it is not difficult to switch it for another error. We can create a new one if we do not an appropriate

packages/core/src/cart/cart-action-creator.ts
```diff
@@ -0,0 +1,53 @@
...
+                    .loadCardEntity(cartId, {
+                        headers: {
+                            Authorization: `Bearer ${jwtToken}`,
+                            'Content-Type': 'application/json',
+                        },
+                        ...options,
+                    })
+                    .then((response) => {
+                        observer.next(
+                            createAction(CartActionType.LoadCartSucceeded, response.body),
```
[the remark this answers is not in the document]
bc-nick: We can change the name if we wish, if necessary

packages/core/src/cart/cart-request-sender.ts
```diff
@@ -19,4 +47,136 @@ export default class CartRequestSender {
...
+                    site: {
+                        cart: { amount, entityId, lineItems },
+                    },
+                },
+            },
+        } = response;
+
+        const mappedLineItems: LineItemMap = {
+            // @ts-ignore
+            physicalItems: lineItems.physicalItems.map((item) => ({
```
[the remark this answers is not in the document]
bc-nick: Yes, not only digital items. We need to parse response and map it into a data regarding to `Cart` interface. I plan to make it after aligning the general functionality with Checkout team.

packages/core/src/checkout-buttons/checkout-headless-button-initializer.ts
```diff
@@ -0,0 +1,152 @@
...
+import { CheckoutStore, InternalCheckoutSelectors } from '../checkout';
+import { isElementId, setUniqueElementId } from '../common/dom';
+
+import { CheckoutButtonInitializeOptions, CheckoutButtonOptions } from './checkout-button-options';
+import CheckoutButtonSelectors from './checkout-button-selectors';
+import CheckoutHeadlessButtonStrategyActionCreator from './checkout-headless-button-strategy-action-creator';
+import createCheckoutButtonSelectors from './create-checkout-button-selectors';
+
+@bind
+export default class CheckoutHeadlessButtonInitializer {
```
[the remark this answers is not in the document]
bc-nick: I will keep that in mind

packages/core/src/checkout-buttons/checkout-button-initializer-options.ts
bc-nick: `host` is BC Control Panel
`siteLink` is the store url that we need to use making the right redirect on confirmation page

packages/paypal-commerce-integration/src/create-paypal-commerce-integration-service.ts
bc-nick: https://github.com/bigcommerce/checkout-sdk-js/pull/2742#discussion_r1867685384

packages/core/src/loader-cdn.ts
bc-nick: I will update it

packages/core/auto-export.config.json
bc-peng: Maybe we can use `HeadlessWalletStrategy` here.

packages/core/src/cart/cart-request-sender.ts
```diff
@@ -19,4 +47,136 @@ export default class CartRequestSender {
...
+                quantity: item.quantity,
+                productId: item.productEntityId,
+            })),
+        };
+
+        return {
+            ...response,
+            body: {
+                id: entityId,
+                // @ts-ignore
```
bc-peng, line 174: Could you please explain why do we have `@ts-ignore` here?

packages/core/src/cart/cart-action-creator.ts
```diff
@@ -0,0 +1,53 @@
...
+import { MissingDataError, MissingDataErrorType } from '../common/error/errors';
+
+import { CartActionType, LoadCartAction } from './cart-actions';
+import CartRequestSender from './cart-request-sender';
+
+export default class CartActionCreator {
+    constructor(private _cartRequestSender: CartRequestSender) {}
+
+    @cachableAction
+    loadCardEntity(
```
bc-peng, line 18: Is `loadCardEntity` the correct name? Or `loadCart`?

packages/core/src/checkout-buttons/create-checkout-headless-button-registry-v2.ts
```diff
@@ -0,0 +1,45 @@
...
+} from '@bigcommerce/checkout-sdk/payment-integration-api';
+
+import { ResolveIdRegistry } from '../common/registry';
+import * as defaultCheckoutHeadlessButtonStrategyFactories from '../generated/checkout-headless-button-strategies';
+
+export interface CheckoutButtonStrategyFactories {
+    [key: string]: CheckoutButtonStrategyFactory<CheckoutButtonStrategy>;
+}
+
+export default function createCheckoutHeadlessButtonStrategyRegistry(
```
davidchin, line 16: This function looks like a duplicate of `createCheckoutButtonStrategyRegistry`, except that the default value for `checkoutButtonHeadlessStrategyFactories` parameter is different. Instead of duplicating, could we just call `createCheckoutButtonStrategyRegistry` and pass it with `defaultCheckoutHeadlessButtonStrategyFactories` instead? It should work for both headless and non-headless because both implement the same interface.

packages/core/src/cart/cart-request-sender.ts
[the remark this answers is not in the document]
bc-peng: This looks fine, but we need to take care of all kinds of items in future, such as custom item.

packages/core/src/checkout-buttons/checkout-button-initializer-options.ts
bc-peng: For production, we may create a new interface for the headless wallet buttons.

packages/core/src/checkout-buttons/checkout-headless-button-initializer.ts
```diff
@@ -0,0 +1,152 @@
...
+     *     containerId: 'headlessWalletButton',
+     *     braintreepaypal: {
+     *     },
+     * });
+     * ```
+     *
+     * @param options - Options for initializing the checkout button.
+     * @returns A promise that resolves to the current state.
+     */
+    initializeHeadlessButton(
```
bc-peng, line 110: Can you explain why we don't call this method, `intialize`?

packages/core/src/payment-integration/default-payment-integration-service.ts
```diff
@@ -204,6 +205,17 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
         return buyNowCart;
     }
 
+    async loadCardEntity(
```
bc-peng, line 208:
```suggestion
    async loadCart(
```

packages/core/src/checkout-buttons/checkout-button-initializer-options.ts
davidchin: Do you mean `host` is the canonical URL, i.e.: `https://store-{hash}.mybigcommerce.com`, and `siteUrl` is the  BC Checkout's site URL (not the headless storefront site URL)?

packages/core/src/cart/cart-request-sender.ts
```diff
@@ -19,4 +47,136 @@ export default class CartRequestSender {
...
+            headers: {
+                ...options.headers,
+                'Content-Type': 'application/json',
+            },
+            body: {
+                query: graphQLQuery,
+            },
+        };
+
+        return this._requestSender
```
davidchin, line 142: It seems to me that we're making a cross-origin Ajax request from headless storefront to BC, which then returns a session token and set as a third party cookie. AFAIK, Safari blocks third party cookies. Could you test if the setup works for Safari?

Another point is, Catalyst creates carts on the server-side. Therefore, the browser does not have BC session cookie required to retrieve carts directly from the browser from BC. Could you also test if the setup works for Catalyst?

I believe the solution to both of these problems is to proxy GQL calls through Catalyst or a server-side middleware for non-Catalyst headless storefronts.

packages/core/src/checkout-buttons/checkout-button-initializer-options.ts
bc-nick: @davidchin yes, you are correct

packages/core/src/checkout-buttons/create-checkout-headless-button-registry-v2.ts
bc-nick: @davidchin yes, it should work. But based on [this](createCheckoutButtonRegistryV2) comment we need to create separate interface for Headless and at the end eventually we need this function

packages/core/src/cart/cart-action-creator.ts
bc-nick: I will switch to your version. Thanks

packages/core/src/checkout-buttons/checkout-headless-button-initializer.ts
bc-nick: What do you mean we do not call this method?

bc-peng: In other initializers, we refer to this method as initialize. I was wondering if there is a specific reason we didn't follow this pattern.

packages/core/src/cart/cart-request-sender.ts
davidchin: > For Safari's case we can avoid blocking of third party cookies by enabling/disabling the appropriate settings.

Yes, but this is not a setting we have control over, as we can't force shoppers to relax their privacy setting. I think this issue is a blocker. If we can't make it work for Safari by default, we need to explore alternative options.

> This point will be mentioned in documentation of headless wallet button.

I don't think documentation is sufficient as the solution won't be compatible with headless Catalyst, which creates carts on the server side.

It seems to me that to solve both of these problems, we shouldn't be calling the GQL API directly in the browser for headless storefronts. Instead, requests need to be routed through a server-side proxy, which in most cases will be the headless storefronts.

packages/core/src/cart/cart-request-sender.ts
```diff
@@ -19,4 +47,139 @@ export default class CartRequestSender {
...
+            .post<LoadCartResponse>(url, {
+                ...requestOptions,
+            })
+            .then(this.transformToCartResponse);
+    }
+
+    private transformToCartResponse(response: Response<LoadCartResponse>): Response<Cart> {
```
bc-nick, line 149: @davidchin @animesh1987 @bc-peng guys I would like to ask you to take a loot at this transformer much closer there are concerns I have encountered.

This transformer is going to be very large because here i am trying to transform response in an appropriate Cart interface to align with the current realisation of stored cart information.

There are a few problems from the current approach:

1. it is hard to support/maintain this code in the future
2. we do not have all required Cart fields information for properly transforming
3. double dependency with Cart interface

I would like to suggest creating a new Headless Cart reducer for avoiding bindings to Cart interface and all things above. What do you think?

animesh1987: Thanks for bringing this up, my suggestion is that we should not have this in the reducer since that's not the purpose of a reducer. I see this is to transform the response from the loadCart API to a cart here in request sender. 

Maybe we can have something similar to [this](https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/payment/payment-request-transformer.ts) and all transform logic can be placed there. 

Maybe you can also think of moving the transform method in the action creator and not in the request sender.

bc-nick: @animesh1987 Why I suggest creating a new reducer is because we can more easily manage this information. This reducer will only take the necessary information for Headless Wallet Buttons strategies

animesh1987: Hey @bc-nick what is the GQL endpoint we are calling here? Maybe we can call checkout GQL endpoint for the information you need once you have a cart.

bc-nick:
@davidchin @animesh1987 @bc-peng hi! I have updated this PR with aligned new requirements related to Headless Wallet Buttons. As you know Wallet Buttons will be working with Catalyst storefront. It means that the current approach includes a proxy request solution to the Catalyst server to avoid third-party cookie issues.

packages/core/src/cart/cart-request-sender.ts
bc-nick: @animesh1987 I come back because this question is still relevant. Unfortunately checkout GQL endpoint does not have relevant information for most of required fields. I created the draft [PR](https://github.com/bigcommerce/checkout-sdk-js/pull/2802) where you be able to see all unavailable fields by looking at the TODO messages.

PR #2744, opened 2024-12-03
feat(payment): PAYPAL-4869 Missing 'Street Name' in Delivery Address

bc-nick:
## What?

Updated shipping address with passing `address_line_2` field
Added updates name of delivery address as an additional changes

## Why?

For correct setting Shipping address

## Testing / Proof

https://github.com/user-attachments/assets/098ad6f5-fb00-4cf0-8b6a-520fc603bfda


bc-nick:
Noticed that we take delivery name from payer information but not from shipping information.
As you can see from the video delivery name is `Dedrick Kozey`, but at the CP we can see `John Doe`. Is it expected behaviour? We can take delivery name from shipping information but there is only one field `full_name` for this.

If we leave `firstName` or `lastName` as empty str the order request will failed. But we can pass a string with space for `lastName` and for `firstName` take the value from `full_name` and everything will work as expected.

serhii-tkachenko:
> Noticed that we take delivery name from payer information but not from shipping information. As you can see from the video delivery name is `Dedrick Kozey`, but at the CP we can see `John Doe`. Is it expected behaviour? We can take delivery name from shipping information but there is only one field `full_name` for this.
> 
> If we leave `firstName` or `lastName` as empty str the order request will failed. But we can pass a string with space for `lastName` and for `firstName` take the value from `full_name` and everything will work as expected.

It is not expected behaviour in my opinion, since you selected Shipping address with name of receiver, however I guess first shipping address from PayPal select was used to fill up checkout shipping info with data 🤔 

So it might be great if you fix this behaviour, since it does not look right

serhii-tkachenko (approved):
Great job! 👍

bc-dronov (approved):

bc-nick:
> Great job! 👍

@serhii-tkachenko added updates related to shipping name to avoid shipping data mismatching

serhii-tkachenko (approved):

PR #2753, opened 2024-12-16
feat(payment): CHECKOUT-8901 Add Async Payment Methods

bc-peng:
## What?
Add async payment methods support by passing payment reponse to the UI side, such as `manual-order-ui`.

## Why?
The UI side can react to an async order accordingly, such as displaying a modal.

### manual testing

https://github.com/user-attachments/assets/e65cc969-7e5e-4077-abae-fcb8b6097258

### DIFF
--- a/packages/hosted-form-v2/src/hosted-form-service.ts
+++ b/packages/hosted-form-v2/src/hosted-form-service.ts
@@ -3,6 +3,7 @@ import HostedForm from './hosted-form';
 import HostedFormFactory from './hosted-form-factory';
 import HostedFormManualOrderData from './hosted-form-manual-order-data';
 import HostedFormOptions from './hosted-form-options';
+import { HostedInputSubmitManualOrderSuccessEvent } from './iframe-content';
 
 export default class HostedFormService {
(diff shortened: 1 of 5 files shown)


packages/hosted-form-v2/src/hosted-form-service.ts
```diff
@@ -23,14 +24,17 @@ export default class HostedFormService {
...
         const form = this._hostedForm;
 
         if (!form) {
             throw new NotInitializedError(NotInitializedErrorType.PaymentNotInitialized);
         }
 
         await form.validate();
-        await form.submitManualOrderPayment({ data });
+
+        return form.submitManualOrderPayment({ data });
```
animesh1987, line 38: This method is still `async`, also what are the implications for removing the await call here?

bc-peng: The promise will be passed to the manual-order-ui side and resolved there.
I'll update the PR to keep the previous logic, which is resloving the promise here.

animesh1987 (approved):

richapatel1510 (approved):

PR #2763, opened 2025-01-16
fix(extension): CHECKOUT-8960 Fix Broadcast Interruption

bc-peng:
## What?
Introduce a new checkout service method to clear outdated poster whenever an extension is unmounted.

Handle message post failures gracefully by clearing the poster cache and logging the loss message.

## Why?
The event poster within the extension messenger is not cleared/rebuilt after an extension iframe is unmounted. 

Continuing to deliver messages through this outdated poster disrupts event broadcasting.

## Testing / Proof
https://github.com/user-attachments/assets/83c132a2-b798-4a81-868a-6d7af2ee3866

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/checkout/checkout-service.ts
+++ b/packages/core/src/checkout/checkout-service.ts
@@ -1433,6 +1433,17 @@ export default class CheckoutService {
     }
 
     /**
+     * Clear cache for a checkout extension when removing it from UI.
+     * This function should be used whenver an extension is removed from the UI.
+     *
+     * @alpha
+     * @param region - The name of an area where the extension should be presented.
+     */
+    clearExtensionCache(region: ExtensionRegion): void {
+        this._extensionMessenger.clearCacheByRegion(region);
+    }
+
+    /**
      * Manages the command handler for an extension.
      *
      * @alpha
--- a/packages/core/src/extension/extension-messenger.ts
+++ b/packages/core/src/extension/extension-messenger.ts
@@ -23,6 +23,22 @@ export class ExtensionMessenger {
         private _posters: { [extensionId: string]: IframeEventPoster<ExtensionEvent> } = {},
     ) {}
 
+    clearCacheByRegion(region: string): void {
+        const extension = this._getExtensionByRegion(region);
+
+        this.clearCacheById(extension.id);
+    }
+
+    clearCacheById(extensionId: string): void {
+        if (this._listeners[extensionId]) {
+            delete this._listeners[extensionId];
+        }
+
+        if (this._posters[extensionId]) {
+            delete this._posters[extensionId];
+        }
+    }
+
     listen<T extends keyof ExtensionCommandMap>(
         extensionId: string,
         command: T,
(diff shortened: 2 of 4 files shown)


packages/core/src/checkout/checkout-service.ts
```diff
@@ -1432,6 +1432,17 @@ export default class CheckoutService {
...
+    /**
+     * Clear cache for a checkout extension when removing it from UI.
+     * This function should be used whenver an extension is removed from the UI.
+     *
+     * @alpha
+     * @param region - The name of an area where the extension should be presented.
+     */
+    clearExtensionCache(region: ExtensionRegion): void {
+        this._extensionMessenger.clearCacheByRegion(region);
+    }
```
bc-peng, line 1444: On `checkout-js` side, this will be called as part of an extension container component's `useEffect` return.
Therefore, it will run when an extension region is unmounted.

packages/core/src/extension/extension-messenger.ts
```diff
@@ -70,37 +86,67 @@ export class ExtensionMessenger {
...
 
-        this._posters[extensionId].post(event);
+            this._posters[extensionId].post(event);
+        } catch (error) {
+            this.clearCacheById(extensionId);
+            // eslint-disable-next-line no-console
+            console.log(
+                `Unable to post event to extension(${extensionId}) because extension iframe is not mounted.\nThe event that could not be delivered:`,
+                event,
+            );
```
bc-peng, line 103: I tried to build a message queue to save messages for the extension when it’s back online, but I encountered an issue.

We use an internal command poster for sending extension messages and a normal poster for boarding, causing messages to be delivered in random order.

In my tests, saved messages often arrive at an extension before the internal poster's "extension ready" message. As a middle-ground approach, I log the lost messages for now.

bc-peng: I tried to build a message queue to save messages for the extension when it’s offline, but I encountered an issue.

We use an internal command poster for sending extension messages and a normal poster for boarding, causing messages to be delivered in random order.

In my tests, saved messages often arrive at an extension before the internal poster's "extension ready" message. As a middle-ground approach, I log the lost messages for now.

bc-peng: I tried to build a message queue to save messages for the extension when it’s offline, but I encountered an issue.

We use an internal command poster for sending internal command like extension ready and a normal poster for broadcasting, causing messages to be delivered in random order.

In my tests, saved messages often arrive at an extension before the internal poster's "extension ready" message. As a middle-ground approach, I log the lost messages for now.

packages/core/src/extension/extension-messenger.ts
```diff
@@ -70,37 +86,67 @@ export class ExtensionMessenger {
...
+        }
```
bc-peng, line 104: I tried to build a message queue to save messages for the extension when it’s offline, but I encountered an issue.

We use an internal poster for sending internal commands like extension ready and a normal poster for broadcasting, causing messages to be delivered in random order.

In my tests, saved messages often arrive at an extension before the internal poster's "extension ready" message. As a middle-ground approach, I log the lost messages for now.

animesh1987 (approved):

PR #2766, opened 2025-01-20
feat(extension): CHECKOUT-8964 Introduce GetConsignment Extension Command

bc-peng:
## What?
Introduce a new checkout extension command, `GetConsignments`

## Why?
To pass consignment data to an extension.

## Testing / Proof
Screenshot 2025-01-21 at 11 17 42 AM

### DIFF
--- a/packages/core/src/checkout/checkout-service.ts
+++ b/packages/core/src/checkout/checkout-service.ts
@@ -1449,7 +1449,7 @@ export default class CheckoutService {
      *
      * @alpha
      * @param extensionId - The ID of an extension to post the event to.
-     * @param message -The message to post to an extension.
+     * @param message - The message to post to an extension.
      */
     postMessageToExtension(extensionId: string, message: ExtensionMessage): void {
         this._extensionMessenger.post(extensionId, message);
--- a/packages/core/src/checkout/checkout-service.ts
+++ b/packages/core/src/checkout/checkout-service.ts
@@ -25,6 +25,7 @@ import {
     ExtensionActionCreator,
     ExtensionCommandMap,
     ExtensionEventBroadcaster,
+    ExtensionMessage,
     ExtensionMessenger,
     ExtensionRegion,
 } from '../extension';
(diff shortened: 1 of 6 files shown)


packages/core/src/checkout/checkout-service.ts
```diff
@@ -1443,6 +1444,17 @@ export default class CheckoutService {
...
     }
 
+    /**
+     * Posts a message to a checkout extension.
+     *
+     * @alpha
+     * @param extensionId - The ID of an extension to post the event to.
+     * @param message -The message to post to an extension.
+     */
+    postMessageToExtension(extensionId: string, message: ExtensionMessage): void {
```
animesh1987, line 1454: How come we need to expose this now?

packages/core/src/extension/extension-message.ts
```diff
@@ -0,0 +1,16 @@
+import { Consignment } from '../shipping';
+
+import { ExtensionEvent } from './extension-events';
+
+export const enum ExtensionMessageType {
+    CurrentConsignments = 'EXTENSION:CURRENT_CONSIGNMENTS',
```
animesh1987, line 6: We can just call it `GetConsignments` instead of CurrentConsignments 🤔

packages/core/src/checkout/checkout-service.ts
bc-peng: Previously, we only had a broadcaster that delivers messages from checkout-js to an extension, with the messenger wrapped inside the broadcaster. 
https://github.com/bigcommerce/checkout-sdk-js/blob/6a9b5171d8c000a7fe43702f906306f4b0e50f7b/packages/core/src/extension/extension-event-broadcaster.ts#L56
Exposing this method now allows a checkout-js command handler to send messages.

packages/core/src/extension/extension-message.ts
bc-peng: Updated.

richapatel1510 (approved):

animesh1987 (approved):

PR #2767, opened 2025-01-21
feat(payment): PI-3068 moved my account logic to hosted card v2 package

vitalii-koshovyi:
## What?
Moved my account logic to the hosted card v2 package.
1) Added StoredCardHostedFormService to the hosted fields v2 for storing credit cards on the My Account Page;
2) Added  [hosted-field.spec.ts](https://github.com/bigcommerce/checkout-sdk-js/compare/master...vitalii-koshovyi:checkout-sdk-js:PI-3068?expand=1#diff-42cc2634edd489146a40c17ba992db4b66239e1730f9b4b59403d544363350e1) test, due to the absence.
But still need to improve coverage for the submitManualOrderForm and submitStoredCardForm methods, didn't do that due to the already huge changes;
3) [ Added additional logic](https://github.com/bigcommerce/checkout-sdk-js/compare/master...vitalii-koshovyi:checkout-sdk-js:PI-3068?expand=1#diff-94b877bfbfcbbe11028cacdaf107f342764cd73b5eee4ddfa68167e1892ab53bR54) to use different iframe URLs for the storefront and CP fields usage. If you have better ideas on how to implement that, please share them with me.

## Why?
Storefront account payments could use hosted fields without importing whole SDK.

## Testing / Proof

https://github.com/user-attachments/assets/c6abacb4-c3f1-4f0b-a374-03a919aa5ade

https://github.com/user-attachments/assets/9ac77d3e-4244-42b6-b79e-9d2d363b6ff6

*logs in the console show that modified version of SDK loaded both in the microapp and in the iframe

### DIFF
--- a/packages/hosted-form-v2/src/hosted-field.ts
+++ b/packages/hosted-form-v2/src/hosted-field.ts
@@ -19,12 +19,17 @@ import HostedFormOrderData from './hosted-form-order-data';
 import {
     HostedInputEventMap,
     HostedInputEventType,
+    HostedInputStoredCardSucceededEvent,
     HostedInputSubmitErrorEvent,
     HostedInputSubmitManualOrderErrorEvent,
     HostedInputSubmitManualOrderSuccessEvent,
     HostedInputSubmitSuccessEvent,
     HostedInputValidateEvent,
 } from './iframe-content';
+import {
+    StoredCardHostedFormData,
+    StoredCardHostedFormInstrumentFields,
+} from './stored-card-hosted-form-type';
 
 export const RETRY_INTERVAL = 60 * 1000;
 export const LAST_RETRY_KEY = 'lastRetry';
@@ -35,21 +40,27 @@ export default class HostedField {
     constructor(
         private _type: HostedFieldType,
         private _containerId: string,
-        private _orderId: number,
         private _placeholder: string,
         private _accessibilityLabel: string,
         private _styles: HostedFieldStylesMap,
         private _eventPoster: IframeEventPoster<HostedFieldEvent>,
         private _eventListener: IframeEventListener<HostedInputEventMap>,
         private _detachmentObserver: DetachmentObserver,
+        private _orderId?: number,
     ) {
         this._iframe = document.createElement('iframe');
 
-        this._iframe.src = `/admin/payments/${this._orderId}/hosted-form-field?version=${LIBRARY_VERSION}`;
+        this._iframe.src = this.getFrameSrc(this._orderId);
         this._iframe.style.border = 'none';
         this._iframe.style.height = '100%';
         this._iframe.style.overflow = 'hidden';
         this._iframe.style.width = '100%';
+    }
+
+    private getFrameSrc(orderId?: number): string {
+        return typeof orderId !== 'undefined'
+            ? `/admin/payments/${this._orderId}/hosted-form-field?version=${LIBRARY_VERSION}`
+            : `/checkout/payment/hosted-field?version=${LIBRARY_VERSION}`;
     }
 
     getType(): HostedFieldType {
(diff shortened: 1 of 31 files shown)


packages/hosted-form-v2/src/hosted-field.ts
```diff
@@ -35,17 +40,20 @@ export default class HostedField {
...
         private _detachmentObserver: DetachmentObserver,
+        private _orderId?: number,
     ) {
         this._iframe = document.createElement('iframe');
 
-        this._iframe.src = `/admin/payments/${this._orderId}/hosted-form-field?version=${LIBRARY_VERSION}`;
+        this._iframe.src =
+            typeof _orderId !== 'undefined'
+                ? `/admin/payments/${this._orderId}/hosted-form-field?version=${LIBRARY_VERSION}`
+                : `/checkout/payment/hosted-field?version=${LIBRARY_VERSION}`;
```
richapatel1510, line 56:
```suggestion
            !(_orderId)
                ? `/checkout/payment/hosted-field?version=${LIBRARY_VERSION}`
                : `/admin/payments/${this._orderId}/hosted-form-field?version=${LIBRARY_VERSION}`
```

vitalii-koshovyi: I'm not sure about that, theoretically `orderId` could be equal 0, in that case yours suggestion could cause issues.
Generally, I thought to pass some kind of a flag like `isControlPannelframe` but it looks excessive.

serhii-tkachenko (approved):
LGTM 👍

PavlenkoM (approved):

packages/hosted-form-v2/src/hosted-field.ts
animesh1987: Also AFAIK orderId should never be 0 in this case, but don't mind the `typeof check`. Can also do `!!(_orderId)` which checks if the value is truthy or not.

packages/hosted-form-v2/src/hosted-form-factory.ts
```diff
@@ -23,13 +23,13 @@ export default class HostedFormFactory {
...
                     type,
                     fieldOptions.containerId,
-                    options.orderId,
                     fieldOptions.placeholder || '',
                     fieldOptions.accessibilityLabel || '',
                     options.styles || {},
                     new IframeEventPoster(host),
                     new IframeEventListener(host),
                     new DetachmentObserver(new MutationObserverFactory()),
+                    options.orderId,
```
animesh1987, line 32: Is there a reason we are moving the order here?

packages/hosted-form-v2/src/iframe-content/hosted-input-events.ts
```diff
@@ -142,3 +146,16 @@ export interface HostedInputSubmitErrorEvent {
...
     };
 }
+
+export interface HostedInputStoredCardSucceededEvent {
+    type: HostedInputEventType.StoredCardSucceeded;
+}
+
+export interface HostedInputStoredCardErrorEvent {
+    type: HostedInputEventType.StoredCardFailed;
+    payload?: {
```
animesh1987, line 156: Will there ever be a case that payload is not present?

animesh1987: Saw this https://github.com/bigcommerce/checkout-sdk-js/pull/2767/files#diff-7bea3ab0d43b6b804c7f55d7e6f0d3bd485fde61e420afbd8265aa6942554770R58 so makes sense, but shall we return a payload there?

packages/hosted-form-v2/src/hosted-field.ts
vitalii-koshovyi: `!!(_orderId)` linter will not pass it

packages/hosted-form-v2/src/hosted-form-factory.ts
vitalii-koshovyi: Yes, because it's not required param anymore, due to the usage with vaulting functionality.
By our linting rules it should be after required params.

packages/hosted-form-v2/src/iframe-content/hosted-input-events.ts
vitalii-koshovyi: Yes, we have similar comment in the initial PR. It gonna be done in separate ticket.
We need to improve error handling in general for the Stored cards functionality.

vitalii-koshovyi: We need to make strict types for the possible endpoint response errors.

packages/hosted-form-v2/src/hosted-field.ts
vitalii-koshovyi: I will create a getFrameSrc, it'll look a lot cleaner.

richapatel1510 (approved):

PR #2769, opened 2025-01-23
feat(extension): CHECKOUT-8974 Introduce Extension Query

bc-peng:
## What?
Introduce a new `query` concept that allows an extension to fetch any data from the checkout object by using a simple query key. 

## Why?
An extension can fetch checkout data immediately.

## Testing / Proof
After the extension service initialization, clicking the emoji will run:
```
console.log(await service.getConsignments(useCache));
```
`useCache` is set to `true` at first. On the second time of clicking the emoji, `useCache` will be `false`, therefore, a `GET` checkout call is made.

https://github.com/user-attachments/assets/8e87cdfb-103c-4338-bf1b-73bf2f1d2e28

### DIFF
--- a/packages/core/src/checkout/checkout-service.ts
+++ b/packages/core/src/checkout/checkout-service.ts
@@ -27,6 +27,7 @@ import {
     ExtensionEventBroadcaster,
     ExtensionMessage,
     ExtensionMessenger,
+    ExtensionQueryMap,
     ExtensionRegion,
 } from '../extension';
 import { FormFieldsActionCreator } from '../form';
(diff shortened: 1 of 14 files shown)


packages/core/src/extension/extension-commands.ts
```diff
@@ -4,6 +4,10 @@ export type ExtensionCommand =
     | SetIframeStyleCommand
     | GetConsignmentsCommand;
 
+export enum InstantDataCommandType {
+    Consignments = 'consignments',
+}
+
```
bc-peng, line 10: Introducing a new set of data commands that work on top of the existing command and message types, allowing developers to write `extensionService.get('consignments')` to fetch data.

packages/core/src/checkout/checkout-service.ts
```diff
@@ -1469,7 +1470,24 @@ export default class CheckoutService {
...
+    /**
+     * Manages the query handler for an extension.
+     *
+     * @alpha
+     * @param extensionId - The ID of the extension sending the query.
+     * @param query - The query to be handled.
+     * @param handler - The handler function for the extension query.
+     * @returns A function that, when called, will deregister the query handler.
+     */
+    handleExtensionQuery<T extends keyof ExtensionQueryMap>(
```
bc-peng, line 1485: This is a copy of the method, `handleExtensionCommand`.

packages/core/src/extension/extension-commands.ts
```diff
@@ -35,13 +29,8 @@ export interface SetIframeStyleCommand {
     };
 }
 
-export interface GetConsignmentsCommand {
-    type: ExtensionCommandType.GetConsignments;
-}
-
```
bc-peng, line 41: Moved to `extension-queries.ts`.

animesh1987 (approved):

PR #2772, opened 2025-01-24
feat(payment): PI-3099 update Adyen SDK to Remove iDEAL bank selection dropdown

bc-ania:
## What?
Update Adyen SDK to remove iDEAL bank selection dropdown for Adyen

## Why?
iDEAL have updated their central technical infrastructure.
[according to official documentation](https://docs.adyen.com/payment-methods/ideal/).

## Testing / Proof
https://github.com/user-attachments/assets/30b901f6-dd8c-4cef-aa16-9cbb8b2ebd63

Screenshot 2025-01-27 at 12 36 24

image


vitalii-koshovyi (changes requested):
I believe we should update Adyen-web package version, they have changed implementation of the Ideal on their side.
Why do we need an experiment for it?

vitalii-koshovyi (approved):

PR #2776, opened 2025-01-30
WIP chore(checkout): CHECKOUT-8264 Reset recaptcha if initialize is called twice

animesh1987:
## What?
Reset recaptcha if recaptcha is instantiated twice instead of throwing an error.

## Why?
If we validate recaptcha for apple pay button and if a shopper tries to create a new account, then an error is thrown regarding recaptcha client already created.
In order to fix this, reset recaptcha client and remove the element so recaptcha can be triggered again.

## Testing / Proof
- CI 

#### Experiment off

https://github.com/user-attachments/assets/4a07fee8-d019-4c99-95ce-0220e61c3a78

#### Experiment on

https://github.com/user-attachments/assets/ace96df5-dc89-4317-bcd9-9a42b1e027d2

### DIFF
--- a/packages/core/src/spam-protection/google-recaptcha.ts
+++ b/packages/core/src/spam-protection/google-recaptcha.ts
@@ -83,7 +83,7 @@ export default class GoogleRecaptcha {
             });
     }
 
-    reset(containerId: string): any {
+    reset(containerId: string): void {
         const element = document.getElementById(containerId);
 
         element?.remove();
--- a/packages/core/src/spam-protection/google-recaptcha.ts
+++ b/packages/core/src/spam-protection/google-recaptcha.ts
@@ -66,16 +66,28 @@ export default class GoogleRecaptcha {
     }
 
     load(containerId: string, sitekey: string): Promise<void> {
-        return this.googleRecaptchaScriptLoader.load().then((recaptcha) => {
-            if (recaptcha) {
-                this._event$ = this._memoized(
-                    recaptcha,
-                    sitekey,
-                    document.getElementById(containerId),
-                );
-                this._recaptcha = recaptcha;
-            }
-        });
+        return this.googleRecaptchaScriptLoader
+            .load()
+            .then((recaptcha) => {
+                if (recaptcha) {
+                    this._event$ = this._memoized(
+                        recaptcha,
+                        sitekey,
+                        document.getElementById(containerId),
+                    );
+                    this._recaptcha = recaptcha;
+                }
+            })
+            .catch((err) => {
+                throw err;
+            });
+    }
+
+    reset(containerId: string): any {
+        const element = document.getElementById(containerId);
+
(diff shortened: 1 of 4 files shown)


packages/core/src/spam-protection/spam-protection-action-creator.ts
```diff
@@ -22,14 +22,28 @@ export default class SpamProtectionActionCreator {
...
         options?: SpamProtectionOptions,
     ): ThunkAction<SpamProtectionAction, InternalCheckoutSelectors> {
-        return (store) =>
-            concat(
+        return (store) => {
+            const state = store.getState();
+            const isRecaptchaResetExperimentEnabled =
+                state.config.getConfig()?.storeConfig.checkoutSettings.features[
+                    'CHECKOUT-8264.recaptcha_error_fix'
+                ] ?? false;
```
bc-peng, line 30: Perhpas we can build the similar function in sdk and ensure the default value is `true`.
```
export default function isExperimentEnabled(
    checkoutSettings: CheckoutSettings | undefined,
    experimentName: string,
): boolean {
    return Boolean(checkoutSettings?.features[experimentName] ?? true);
}
```

packages/core/src/spam-protection/google-recaptcha.ts
```diff
@@ -66,16 +66,28 @@ export default class GoogleRecaptcha {
...
+                        recaptcha,
+                        sitekey,
+                        document.getElementById(containerId),
+                    );
+                    this._recaptcha = recaptcha;
+                }
+            })
+            .catch((err) => {
+                throw err;
+            });
```
bc-peng, line 83: I see we're adding this block here.
Could you please explain how the error will be handled?

packages/core/src/spam-protection/spam-protection-action-creator.ts
animesh1987: SDK is different than JS, as folks don't fork it so we should be ok in that regard.

packages/core/src/spam-protection/google-recaptcha.ts
animesh1987: This should be on the clients or just throw. It was previously unhandled so made this change.

packages/core/src/spam-protection/spam-protection-action-creator.ts
bc-peng: On a second thought, when a forked custom checkout-js uses a lower SDK version, if we remove the experiment here it will return `false`.
Do we want to return `true` instead?

animesh1987: Good point, I tried to keep this in existing sync with the logic in other places. Might be better to create the function you mentioned above.

bc-peng (approved):

PR #2785, opened 2025-02-07
feat(payment): PAYPAL-5067 ID creation updates

bc-nick:
## What?

ID creation updates

## Why?

To avoid situation with duplicate buttons

## Testing / Proof

Before

https://github.com/user-attachments/assets/8ad663a5-4f43-43b4-9d7d-262ebdf4a7b2

After

https://github.com/user-attachments/assets/ae586c9e-366a-4857-b837-8030a2371019

### DIFF
--- a/packages/core/src/common/dom/set-unique-element-id.ts
+++ b/packages/core/src/common/dom/set-unique-element-id.ts
@@ -13,7 +13,10 @@ export default function setUniqueElementId(selector: string, idPrefix: string): 
 
     return Array.prototype.slice.call(containers).map((container: HTMLElement) => {
         if (!container.id) {
-            container.id = uniqueId(idPrefix);
+            const getUniqId = (id: string): string =>
+                document.getElementById(id) ? getUniqId(uniqueId(idPrefix)) : id;
+
+            container.id = getUniqId(uniqueId(idPrefix));
         }
 
         return container.id;
--- a/packages/core/src/common/dom/set-unique-element-id.ts
+++ b/packages/core/src/common/dom/set-unique-element-id.ts
@@ -2,6 +2,12 @@ import { uniqueId } from 'lodash';
 
 import { InvalidArgumentError } from '../error/errors';
 
+const getUniqId = (idPrefix?: string): string => {
+    const id = uniqueId(idPrefix);
+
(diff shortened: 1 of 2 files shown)


serhii-tkachenko (approved):

packages/core/src/common/dom/set-unique-element-id.ts
```diff
@@ -13,7 +13,10 @@ export default function setUniqueElementId(selector: string, idPrefix: string):
 
     return Array.prototype.slice.call(containers).map((container: HTMLElement) => {
         if (!container.id) {
-            container.id = uniqueId(idPrefix);
+            const getUniqId = (id: string): string =>
+                document.getElementById(id) ? getUniqId(uniqueId(idPrefix)) : id;
```
animesh1987, line 17: Can we declare this outside `setUniqueElementId` method please?

animesh1987: Also seems does it make sense to add the check in `initializeButton` method in checkout-button-initializer.

bc-nick: > Can we declare this outside `setUniqueElementId` method please?

Would it be ok if I move it to the same file before the `setUniqueElementId` is declared? Or create another file for this?

bc-nick: > Also seems does it make sense to add the check in `initializeButton` method in checkout-button-initializer.

Check is already added by using `setUniqueElementId` in `initializeButton`

bc-nick: @animesh1987 what do you think?

packages/core/src/common/dom/set-unique-element-id.ts
```diff
@@ -13,7 +13,10 @@ export default function setUniqueElementId(selector: string, idPrefix: string):
...
+
+            container.id = getUniqId(uniqueId(idPrefix));
```
bc-peng, line 19: Can we put `uniqueId(idPrefix)` inside the `getUniqId` function so that `uniqueId(idPrefix)` only appears once in the code?

animesh1987 (approved):

PR #2787, opened 2025-02-10
feat(payment): PI-2875 Google Pay coupons handling

bc-dronov:
## What?
Added coupons handling on the Google Pay payment sheet.

Cart flow:

https://github.com/user-attachments/assets/2419842b-0807-4f87-9eec-c09add1ebc0f

Mini cart flow:

https://github.com/user-attachments/assets/8cd5c30d-c58c-4f96-b6a3-a7eeb0739985

PDP flow:

https://github.com/user-attachments/assets/78046f5f-c557-44b7-8a8e-5923780553d3

Customer step flow:

https://github.com/user-attachments/assets/20e16cb0-b400-4197-bf3e-69710c9907a4

Payment step flow:

https://github.com/user-attachments/assets/9e6ec42e-ff93-408e-9d6d-2b7ec834650e

Error showing:

https://github.com/user-attachments/assets/3a1bf300-8d5a-4221-83cc-9e1b29359a76

Removing Coupon:

https://github.com/user-attachments/assets/af332f2c-8f8e-4be9-b68f-c2ec320bd149

## Why?
Because we don't handle coupons adding through the Google Pay payment pop up.

## Testing / Proof
All tests have been passed / manual testing.

## Rollout/Rollback
Covered by experiment:
`PI-2875.googlepay_coupons_handling
`

### DIFF
--- a/packages/core/src/payment-integration/default-payment-integration-service.ts
+++ b/packages/core/src/payment-integration/default-payment-integration-service.ts
@@ -17,6 +17,7 @@ import { BillingAddressActionCreator } from '../billing';
 import { CartRequestSender } from '../cart';
 import { Checkout, CheckoutActionCreator, CheckoutStore, CheckoutValidator } from '../checkout';
 import { DataStoreProjection } from '../common/data-store';
+import CouponActionCreator from '../coupon/coupon-action-creator';
 import { CustomerActionCreator, CustomerCredentials } from '../customer';
 import { HostedFormFactory } from '../hosted-form';
 import { OrderActionCreator } from '../order';
@@ -54,6 +55,7 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
         private _customerActionCreator: CustomerActionCreator,
         private _cartRequestSender: CartRequestSender,
         private _storeCreditActionCreator: StoreCreditActionCreator,
+        private _couponActionCreator: CouponActionCreator,
         private _spamProtectionActionCreator: SpamProtectionActionCreator,
         private _paymentProviderCustomerActionCreator: PaymentProviderCustomerActionCreator,
         private _shippingCountryActionCreator: ShippingCountryActionCreator,
@@ -211,6 +213,24 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
         await this._store.dispatch(
             this._storeCreditActionCreator.applyStoreCredit(useStoreCredit, options),
         );
+
+        return this._storeProjection.getState();
+    }
+
+    async applyCoupon(
+        coupon: string,
+        options?: RequestOptions,
+    ): Promise<PaymentIntegrationSelectors> {
+        await this._store.dispatch(this._couponActionCreator.applyCoupon(coupon, options));
+
+        return this._storeProjection.getState();
+    }
+
+    async removeCoupon(
+        coupon: string,
+        options?: RequestOptions,
+    ): Promise<PaymentIntegrationSelectors> {
+        await this._store.dispatch(this._couponActionCreator.removeCoupon(coupon, options));
 
         return this._storeProjection.getState();
     }
--- a/packages/google-pay-integration/src/gateways/google-pay-gateway.ts
+++ b/packages/google-pay-integration/src/gateways/google-pay-gateway.ts
@@ -21,17 +21,22 @@ import isGooglePayCardNetworkKey from '../guards/is-google-pay-card-network-key'
 import {
     CallbackIntentsType,
     CallbackTriggerType,
+    ErrorReasonType,
     ExtraPaymentData,
     GooglePayCardDataResponse,
     GooglePayCardNetwork,
     GooglePayCardParameters,
+    GooglePayError,
     GooglePayFullBillingAddress,
     GooglePayGatewayParameters,
     GooglePayInitializationData,
     GooglePayMerchantInfo,
+    GooglePayPaymentDataRequest,
     GooglePayRequiredPaymentData,
     GooglePaySetExternalCheckoutData,
     GooglePayTransactionInfo,
+    HandleCouponsOut,
+    IntermediatePaymentData,
     ShippingOptionParameters,
     TotalPriceStatusType,
 } from '../types';
@@ -138,23 +143,36 @@ export default class GooglePayGateway {
     }
 
     getCallbackTriggers(): { [key: string]: CallbackTriggerType[] } {
+        const state = this._paymentIntegrationService.getState();
+        // TODO remove this experiment usage after we make sure that coupons handling works fine
+        const isGooglePayCouponsExperimentOn =
+            state.getStoreConfigOrThrow().checkoutSettings.features[
+                'PI-2875.googlepay_coupons_handling'
+            ] || false;
+
         const availableTriggers = [
             CallbackTriggerType.INITIALIZE,
             CallbackTriggerType.SHIPPING_ADDRESS,
             CallbackTriggerType.SHIPPING_OPTION,
+            ...(isGooglePayCouponsExperimentOn ? [CallbackTriggerType.OFFER] : []),
         ];
+
         const initializationTrigger = [CallbackTriggerType.INITIALIZE];
         const addressChangeTriggers = [
             CallbackTriggerType.INITIALIZE,
             CallbackTriggerType.SHIPPING_ADDRESS,
         ];
         const shippingOptionsChangeTriggers = [CallbackTriggerType.SHIPPING_OPTION];
+        const offerChangeTriggers = isGooglePayCouponsExperimentOn
+            ? [CallbackTriggerType.OFFER]
+            : [];
 
         return {
             availableTriggers,
             initializationTrigger,
             addressChangeTriggers,
             shippingOptionsChangeTriggers,
+            offerChangeTriggers,
         };
     }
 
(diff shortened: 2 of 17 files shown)


packages/google-pay-integration/src/gateways/google-pay-gateway.ts
```diff
@@ -329,6 +338,123 @@ export default class GooglePayGateway {
...
+                    message: e.message,
+                    intent: CallbackTriggerType.OFFER,
+                };
+            }
+
+            return error;
+        }
+
+        if (isCartPage) {
+            const couponCodeInput = document.getElementById('couponcode');
```
vitalii-koshovyi, line 421: should we have a ability to pass those selectors?
I believe there could be potential problems with custom themes.

packages/google-pay-integration/src/gateways/google-pay-gateway.ts
```diff
@@ -329,6 +338,123 @@ export default class GooglePayGateway {
...
+
+            if (couponCodeInput) {
+                (couponCodeInput as HTMLInputElement).value = code;
+
+                const form = document.querySelector('.coupon-form');
```
vitalii-koshovyi, line 426: same here

vitalii-koshovyi:
Have you tested coupon code removal from the Google Pay modal window?

bc-dronov:
> Have you tested coupon code removal from the Google Pay modal window?

Yes, and it works fine. I have recorded a video with coupon removal.

packages/google-pay-integration/src/gateways/google-pay-gateway.ts
bc-dronov: If we don't have these selectors because of a custom theme, we can still have the coupon applied/removed because I'm using our generic 'applyCoupon/removeCoupon' actions. The purpose of this code is to show the user on the cart page the updated price after the coupon is applied/removed.
Screenshot 2025-03-17 at 11 49 51

If the user reloads the page, the price will be displayed as expected.

packages/google-pay-integration/src/gateways/google-pay-gateway.ts
bc-dronov: Same answer as above.

packages/google-pay-integration/src/gateways/google-pay-gateway.ts
bc-nick, line 421: Is this element exist only in Cart Page?

packages/google-pay-integration/src/gateways/google-pay-gateway.ts
```diff
@@ -329,6 +338,123 @@ export default class GooglePayGateway {
...
+
+        await Promise.all(couponsPromise);
+
+        if (withClickOnTheCartPage) {
+            const removeCouponLink = document.querySelector(
+                'a[href*="/cart.php?action=removecoupon"]',
+            );
+
+            if (removeCouponLink) {
+                (removeCouponLink as HTMLAnchorElement).click();
```
bc-nick, line 453: Why are you removing coupons like this?

packages/google-pay-integration/src/google-pay-button-strategy.spec.ts
```diff
@@ -48,7 +49,7 @@ describe('GooglePayButtonStrategy', () => {
     let scriptLoader: GooglePayScriptLoader;
     let formPoster: FormPoster;
     let mockReturnedPaymentDataChangedValue: NewTransactionInfo & NewShippingOptionParameters;
-
+    let gateway;
```
bc-nick, line 52: Add interface pls

packages/google-pay-integration/src/google-pay-button-strategy.ts
```diff
@@ -216,11 +219,24 @@ export default class GooglePayButtonStrategy implements CheckoutButtonStrategy {
                         );
                     }
 
+                    const { newOfferInfo = undefined, error: couponsError = undefined } =
+                        offerChangeTriggers.includes(callbackTrigger)
+                            ? await this._googlePayPaymentProcessor.handleCoupons(offerData, true)
+                            : {};
+
+                    if (couponsError) {
```
bc-nick, line 227: The same

packages/google-pay-integration/src/google-pay-customer-strategy.ts
```diff
@@ -135,6 +137,15 @@ export default class GooglePayCustomerStrategy implements CustomerStrategy {
                         );
                     }
 
+                    const { newOfferInfo = undefined, error: couponsError = undefined } =
+                        offerChangeTriggers.includes(callbackTrigger)
+                            ? await this._googlePayPaymentProcessor.handleCoupons(offerData)
+                            : {};
+
+                    if (couponsError) {
```
bc-nick, line 145: The same

packages/google-pay-integration/src/gateways/google-pay-gateway.ts
```diff
@@ -329,6 +338,123 @@ export default class GooglePayGateway {
...
+            return {
+                newOfferInfo: appliedCoupons,
+                error,
+            };
+        }
+
+        const code = redemptionCodes[redemptionCodes.length - 1];
+        const appliedCouponError = await this.applyCoupon(code, isCartPage);
+
+        if (appliedCouponError) {
```
bc-nick, line 375: You can avoid this checking by using `appliedCouponError` only without redefinition of error

```
{
    newOfferInfo: this.getAppliedCoupons(),
    error: await this.applyCoupon(code, isCartPage),
}
```

packages/google-pay-integration/src/gateways/google-pay-gateway.ts
```diff
@@ -329,6 +338,123 @@ export default class GooglePayGateway {
...
+        if (!redemptionCodes.length) {
+            await this.removeAllCoupons(appliedCoupons, isCartPage);
+
+            return {
+                newOfferInfo: this.getAppliedCoupons(),
+            };
+        }
+
+        // Validate for applied coupons to make sure we have only one applied
+        if (appliedCoupons.offers.length) {
```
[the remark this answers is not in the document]
bc-dronov: In this case `appliedCoupons.offers` is an array of coupons which is already applied. 
So when `handleCoupons` function is triggered, we check if there is a new coupons wants to be applied, if yes, then we check if we already have some coupons applied (in our case `appliedCoupons.offers.length >= 2` not correct, because we have to go even with `appliedCoupons.offers.length === 1`)

packages/google-pay-integration/src/gateways/google-pay-gateway.ts
bc-dronov: Yes, this element exists only in cart page. This code is only for updating coupons on the cart page in real time, without having to reload the page.
Screenshot 2025-03-17 at 14 56 59

packages/google-pay-integration/src/gateways/google-pay-gateway.ts
bc-dronov: This is for the cart page to synchronise google pay price with the price displayed on the cart page, without having to reload the page. Same reason as above.

https://github.com/user-attachments/assets/be8da1c3-02e9-499e-8b5b-086cb00abd5a

packages/google-pay-integration/src/google-pay-button-strategy.ts
bc-dronov: Please check this reply: https://github.com/bigcommerce/checkout-sdk-js/pull/2787#discussion_r1998761259

packages/google-pay-integration/src/gateways/google-pay-gateway.ts
bc-dronov: Done

packages/google-pay-integration/src/google-pay-button-strategy.spec.ts
bc-dronov: Done

packages/core/src/payment-integration/default-payment-integration-service.ts
```diff
@@ -215,6 +217,24 @@ export default class DefaultPaymentIntegrationService implements PaymentIntegrat
         return this._storeProjection.getState();
     }
 
+    async applyCoupon(
```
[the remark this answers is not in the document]
bc-dronov: Done

packages/google-pay-integration/src/google-pay-button-strategy.ts
```diff
@@ -216,11 +219,24 @@ export default class GooglePayButtonStrategy implements CheckoutButtonStrategy {
                         );
                     }
 
+                    const { newOfferInfo = undefined, error: couponsError = undefined } =
+                        offerChangeTriggers.includes(callbackTrigger)
+                            ? await this._googlePayPaymentProcessor.handleCoupons(offerData, true)
```
bc-nick, line 224: Secondary parameter can lead to some misunderstandings because this strategy is working on PDP and Cart Page (including mini cart). I would like to propose you to remove this parameter and detect Cart Page only using [this](https://github.com/bigcommerce/checkout-sdk-js/pull/2787/files/067d24c13e485646b572083e5b7b7825cca194bb#diff-344a206df74418e9cd10b2f25c10d6803c010e5b5b85133731267f1120321618R421) part of code. If we do not have such an element it will mean that current page is not Cart Page.

packages/google-pay-integration/src/gateways/google-pay-gateway.ts
bc-nick: Take a look https://github.com/bigcommerce/checkout-sdk-js/pull/2787#discussion_r1999109060

packages/google-pay-integration/src/google-pay-payment-strategy.ts
```diff
@@ -189,13 +188,28 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
 
+                    const { offerChangeTriggers } =
+                        this._googlePayPaymentProcessor.getCallbackTriggers();
+
+                    const { newOfferInfo = undefined, error: couponsError = undefined } =
+                        offerChangeTriggers.includes(callbackTrigger)
+                            ? await this._googlePayPaymentProcessor.handleCoupons(offerData)
+                            : {};
+
+                    if (couponsError) {
```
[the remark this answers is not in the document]
bc-dronov: Thanks for mentioning it, but in my opinion, it is better to keep the possibility to add another errors, then make redundant refactoring later. 
Especially, I would like to create a ticket for displaying a shipping related errors. So it is make sense for now.

packages/google-pay-integration/src/google-pay-button-strategy.ts
bc-dronov: That's a good point, let me check it locally. But we also have to keep in mind that the merchant may have a custom theme without these selectors.

packages/google-pay-integration/src/google-pay-payment-strategy.ts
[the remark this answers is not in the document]
bc-nick: Wouldn't it be easier to do it this way instead of adding `if` statement. What do you think?

```
const error = couponsError || shippingError || anotherError;

return {
  ...(error && {
      error,
  }),
}
```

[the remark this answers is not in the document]
bc-dronov: This one looks better, let me check it too.

packages/google-pay-integration/src/google-pay-button-strategy.ts
```diff
@@ -216,11 +219,24 @@ export default class GooglePayButtonStrategy implements CheckoutButtonStrategy {
                         );
                     }
 
+                    const { newOfferInfo = undefined, error: couponsError = undefined } =
```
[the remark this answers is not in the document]
bc-dronov: Yes, because typescript returns an error without it.

packages/google-pay-integration/src/types.ts
```diff
@@ -192,16 +226,24 @@ export interface IntermediatePaymentData {
...
     paymentDataCallbacks?: {
         onPaymentDataChanged(
             intermediatePaymentData: IntermediatePaymentData,
-        ): Promise<(NewTransactionInfo & NewShippingOptionParameters) | void>;
+        ): onPaymentDataChangedOut;
     };
 }
 
+export type onPaymentDataChangedOut = Promise<
+    | (NewTransactionInfo & NewShippingOptionParameters & NewOfferInfo & { error?: GooglePayError })
```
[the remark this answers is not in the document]
bc-dronov: Our linter is configured to put this operator at the beginning. So we can't remove it.

packages/google-pay-integration/src/gateways/google-pay-gateway.ts
bc-nick: @bc-dronov I have some doubt about this functionality too. Do we really need to update UI after coupon was applied?

In my opinion there is no need to update UI to see total sum changes after coupon was applied by using GP feature.
Total sum can be updated by changing shipping option but we do not update UI in this case. The same situation with changing shipping address. This actions have to be isolated because we are working in scope of Google Pay functionality.  

@vitalii-koshovyi what do you think?

vitalii-koshovyi: @bc-nick this logic already implemented in the theme. I believe @bc-dronov has implemented price change when shopper applies coupon within GooglePay modal. Am I correct?

vitalii-koshovyi: @bc-dronov I believe we should implement such functionality, so theme developers can customize those selectors if that's possible.
It's not a good idea be bound to the hardcoded selectors within checkout-sdk, in my opinion.

packages/google-pay-integration/src/google-pay-payment-strategy.ts
```diff
@@ -185,12 +188,46 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
         );
     }
 
+    protected async _isSignedIn(): Promise<boolean> {
+        let signedIn = false;
+
+        try {
+            signedIn = !!(await this._googlePayPaymentProcessor.getNonce(this._getMethodId()));
```
bc-nick, line 195: We have different implementation of nonce for a few strategies. Can you please take a look at the few following strategies (google-pay-adyenv3-gateway, google-pay-adyenv3-gateway, google-pay-braintree-gateway) to be sure that this changes is applicable for each of them pls?

vitalii-koshovyi (approved):
Discussed in slack with @bc-dronov . Product manager approved a solution just for Cornerstone.
Solution for custom selectors will be developed later.

animesh1987 (approved):

PR #2788, opened 2025-02-11
feat(payment): PAYPAL-5113 added test coverage for braintree-utils

andriiVitvitskyi1990:
## What?
Added test coverage for braintree-utils

## Why?
To increase test coverage

## Testing / Proof
Screenshot 2025-02-11 at 14 39 01
Screenshot 2025-02-11 at 14 38 36

### DIFF
--- a/packages/braintree-utils/src/braintree-integration-service.spec.ts
+++ b/packages/braintree-utils/src/braintree-integration-service.spec.ts
@@ -86,6 +86,32 @@ describe('BraintreeIntegrationService', () => {
         localStorage.clear();
     });
 
+    describe('#getPaypal', () => {
+        it('throws an error if PayPal SDK cannot be loaded or created', async () => {
+            jest.spyOn(braintreeScriptLoader, 'loadPaypal').mockRejectedValue(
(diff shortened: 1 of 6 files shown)


packages/braintree-utils/src/braintree-integration-service.spec.ts
```diff
@@ -359,6 +385,39 @@ describe('BraintreeIntegrationService', () => {
 
             expect(braintreeIntegrationService.mapToLegacyShippingAddress(props)).toEqual(expects);
         });
+
+        it('returns empty string if name is empty', () => {
+            const details = {
```
bc-nick, line 390: Could you please use already declare the similar variable (detailsMock) in upper scope?

And take info from it for the expecting result value

serhii-tkachenko (approved):

packages/braintree-utils/src/braintree-integration-service.spec.ts
andriiVitvitskyi1990: I created test scope variable with different fields values, that's why i didn't use existing one

PR #2794, opened 2025-02-19
fix(shipping): CHECKOUT-8999 Remove duplicate call for shipping option

animesh1987:
## What?
Avoid duplicate call for shipping option if shipping option id passed is same.

## Why?
Avoid duplicate call for shipping option if shipping option id that is passed same as the one that is already selected. This can happen sometimes with various subscriptions on clients from SDK that watch for consignment changes.

## Testing / Proof
- CI
- Screencast

#### Experiment off

https://github.com/user-attachments/assets/b651844d-8a1f-440c-b9f1-838933ad2900

#### Experiment on

https://github.com/user-attachments/assets/275fc799-e0b6-43bc-a0e1-0bad80980089

### DIFF
--- a/packages/utility/src/is-experiment-enabled/is-experiment-enabled.ts
+++ b/packages/utility/src/is-experiment-enabled/is-experiment-enabled.ts
@@ -1,8 +1,7 @@
-import { CheckoutSettings } from '@bigcommerce/checkout-sdk/payment-integration-api';
+export interface Features {
+    [featureName: string]: boolean | undefined;
+}
 
-export default function isExperimentEnabled(
-    checkoutSettings: CheckoutSettings | undefined,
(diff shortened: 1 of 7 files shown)


packages/utility/src/is-experiment-enabled/is-experiment-enabled.ts
```diff
@@ -0,0 +1,7 @@
+export interface Features {
```
animesh1987, line 1: We need to move shared interfaces in a models package. Feels odd now that most shared models are in `payment-integration-api` package, which causes circular dependency issues.

bc-peng (approved):
LGTM. Nicely done.👏👏

richapatel1510 (approved):

PR #2800, opened 2025-03-05
chore(checkout): CHECKOUT-9052 Throw new error type for missing shipping method

animesh1987:
## What?
Throw new error type for missing shipping method so checkout-js can render the correct header in error modal

## Why?
If a shipping method is invalid or missing, in that case on order submit an error modal is shown with error title being `Something went wrong`. This is an attempt at having a more explicit title.

## Testing / Proof
- CI
- Screencast

Screenshot 2025-03-05 at 10 28 03 pm


packages/core/src/order/errors/missing-shipping-method-error.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Checkout prevents consumers from placing their orders when a merchant wishes
+ * to be able to block transactions if the automated tax provider cannot be reached.
+ */
+export default class MissingShippingMethodError extends StandardError {
+    constructor(message?: string) {
+        super(
+            message ||
+                'A shipping method is required for your order. Please review your shipping details and select a shipping option.',
```
bc-mackxu, line 11: No sure about FF, but do you need to do translated here?

bc-shawnwang (approved):
thanks for the quick fix

packages/core/src/order/errors/missing-shipping-method-error.ts
```diff
@@ -0,0 +1,17 @@
+import { StandardError } from '../../common/error/errors';
+
+/**
+ * Checkout prevents consumers from placing their orders when a merchant wishes
+ * to be able to block transactions if the automated tax provider cannot be reached.
+ */
```
bc-shawnwang, line 6: this comment is a irelevant?

richapatel1510 (approved):

bc-mackxu (approved):

packages/core/src/order/order-request-sender.ts
```diff
@@ -75,6 +75,10 @@ export default class OrderRequestSender {
                     throw new CartConsistencyError();
                 }
 
+                if (error.body.type === 'missing_shipping_method') {
+                    throw new MissingShippingMethodError(error.body.detail);
```
bc-shawnwang, line 79: good to use server side message directly since it's translated

bc-shawnwang (approved):

bc-peng (approved):

PR #2802, opened 2025-03-05
feat(payment): PAYPAL-4935 added CartActionCreator for handling/storing cart information

bc-nick:
## What?

Added `CartActionCreator` for handling/storing cart information
Based on [POC](https://github.com/bigcommerce/checkout-sdk-js/pull/2742)

## Why?

In scope of Headless Wallet Buttons project

## Testing / Proof

Manual testing

### DIFF
--- a/packages/core/src/cart/cart-action-creator.ts
+++ b/packages/core/src/cart/cart-action-creator.ts
@@ -14,19 +14,19 @@ export default class CartActionCreator {
     constructor(private _cartRequestSender: CartRequestSender) {}
 
     @cachableAction
-    loadCard(
+    loadCart(
         cartId: string,
         options?: RequestOptions & ActionOptions,
     ): ThunkAction<LoadCartAction, InternalCheckoutSelectors> {
         return (store) => {
-            return Observable.create((observer: Observer<LoadCartAction>) => {
+            return new Observable((observer: Observer<LoadCartAction>) => {
                 const state = store.getState();
                 const host = state.config.getHost();
 
                 observer.next(createAction(CartActionType.LoadCartRequested, undefined));
 
                 this._cartRequestSender
-                    .loadCard(cartId, host, options)
+                    .loadCart(cartId, host, options)
                     .then((response) => {
                         observer.next(
                             createAction(CartActionType.LoadCartSucceeded, response.body),
--- a/packages/core/src/cart/cart-action-creator.ts
+++ b/packages/core/src/cart/cart-action-creator.ts
@@ -18,12 +18,15 @@ export default class CartActionCreator {
         cartId: string,
         options?: RequestOptions & ActionOptions,
     ): ThunkAction<LoadCartAction, InternalCheckoutSelectors> {
-        return () => {
-            return new Observable((observer: Observer<LoadCartAction>) => {
+        return (store) => {
+            return Observable.create((observer: Observer<LoadCartAction>) => {
+                const state = store.getState();
+                const host = state.config.getHost();
+
                 observer.next(createAction(CartActionType.LoadCartRequested, undefined));
 
                 this._cartRequestSender
-                    .loadCart(cartId, options)
+                    .loadCart(cartId, host, options)
                     .then((response) => {
                         observer.next(
                             createAction(CartActionType.LoadCartSucceeded, response.body),
--- a/packages/core/src/cart/cart-action-creator.ts
+++ b/packages/core/src/cart/cart-action-creator.ts
@@ -1,0 +1,42 @@
+import { createAction, createErrorAction, ThunkAction } from '@bigcommerce/data-store';
+import { Observable, Observer } from 'rxjs';
+
+import { RequestOptions } from '@bigcommerce/checkout-sdk/payment-integration-api';
+
+import { InternalCheckoutSelectors } from '../checkout';
+import { cachableAction } from '../common/data-store';
+import ActionOptions from '../common/data-store/action-options';
+
+import { CartActionType, LoadCartAction } from './cart-actions';
+import CartRequestSender from './cart-request-sender';
+
+export default class CartActionCreator {
+    constructor(private _cartRequestSender: CartRequestSender) {}
+
+    @cachableAction
+    loadCard(
+        cartId: string,
+        options?: RequestOptions & ActionOptions,
+    ): ThunkAction<LoadCartAction, InternalCheckoutSelectors> {
+        return (store) => {
+            return Observable.create((observer: Observer<LoadCartAction>) => {
+                const state = store.getState();
+                const host = state.config.getHost();
+
+                observer.next(createAction(CartActionType.LoadCartRequested, undefined));
+
+                this._cartRequestSender
+                    .loadCard(cartId, host, options)
+                    .then((response) => {
+                        observer.next(
+                            createAction(CartActionType.LoadCartSucceeded, response.body),
+                        );
+                        observer.complete();
+                    })
+                    .catch((response) => {
+                        observer.error(createErrorAction(CartActionType.LoadCartFailed, response));
+                    });
+            });
+        };
+    }
+}
--- a/packages/core/src/cart/cart-action-creator.ts
+++ b/packages/core/src/cart/cart-action-creator.ts
@@ -18,15 +18,12 @@ export default class CartActionCreator {
         cartId: string,
         options?: RequestOptions & ActionOptions,
     ): ThunkAction<LoadCartAction, InternalCheckoutSelectors> {
-        return (store) => {
+        return () => {
             return new Observable((observer: Observer<LoadCartAction>) => {
-                const state = store.getState();
-                const host = state.config.getHost();
-
                 observer.next(createAction(CartActionType.LoadCartRequested, undefined));
 
                 this._cartRequestSender
-                    .loadCart(cartId, host, options)
+                    .loadCart(cartId, options)
                     .then((response) => {
                         observer.next(
                             createAction(CartActionType.LoadCartSucceeded, response.body),
--- a/packages/core/src/cart/cart-action-creator.ts
+++ b/packages/core/src/cart/cart-action-creator.ts
@@ -1,4 +1,6 @@
 import { createAction, createErrorAction, ThunkAction } from '@bigcommerce/data-store';
+import { Response } from '@bigcommerce/request-sender';
+import { merge } from 'lodash';
 import { Observable, Observer } from 'rxjs';
 
 import { RequestOptions } from '@bigcommerce/checkout-sdk/payment-integration-api';
@@ -7,8 +9,10 @@ import { InternalCheckoutSelectors } from '../checkout';
 import { cachableAction } from '../common/data-store';
 import ActionOptions from '../common/data-store/action-options';
 
+import Cart from './cart';
 import { CartActionType, LoadCartAction } from './cart-actions';
 import CartRequestSender from './cart-request-sender';
+import { GQLCartResponse, GQLCurrencyResponse, GQLRequestResponse, mapToCart } from './gql-cart';
 
 export default class CartActionCreator {
     constructor(private _cartRequestSender: CartRequestSender) {}
@@ -19,24 +23,49 @@ export default class CartActionCreator {
         options?: RequestOptions & ActionOptions,
     ): ThunkAction<LoadCartAction, InternalCheckoutSelectors> {
         return (store) => {
-            return Observable.create((observer: Observer<LoadCartAction>) => {
+            return new Observable((observer: Observer<LoadCartAction>) => {
                 const state = store.getState();
-                const host = state.config.getHost();
+                const gqlUrl = state.config.getGQLRequestUrl();
 
                 observer.next(createAction(CartActionType.LoadCartRequested, undefined));
 
                 this._cartRequestSender
-                    .loadCart(cartId, host, options)
-                    .then((response) => {
-                        observer.next(
-                            createAction(CartActionType.LoadCartSucceeded, response.body),
-                        );
-                        observer.complete();
+                    .loadCart(cartId, gqlUrl, options)
+                    .then((cartResponse) => {
+                        return this._cartRequestSender
+                            .loadCartCurrency(
+                                cartResponse.body.data.site.cart.currencyCode,
+                                gqlUrl,
+                                options,
+                            )
+                            .then((currencyResponse) => {
+                                observer.next(
+                                    createAction(
+                                        CartActionType.LoadCartSucceeded,
+                                        this.transformToCartResponse(
+                                            merge(cartResponse, currencyResponse),
+                                        ),
+                                    ),
+                                );
+                                observer.complete();
+                            });
                     })
                     .catch((response) => {
                         observer.error(createErrorAction(CartActionType.LoadCartFailed, response));
                     });
             });
         };
+    }
+
+    private transformToCartResponse(
+        response: Response<GQLRequestResponse<GQLCartResponse & GQLCurrencyResponse>>,
+    ): Cart {
+        const {
+            body: {
+                data: { site },
+            },
+        } = response;
+
+        return mapToCart(site);
     }
 }
--- a/packages/core/src/cart/cart-request-sender.spec.ts
+++ b/packages/core/src/cart/cart-request-sender.spec.ts
@@ -7,6 +7,7 @@ import {
 
 import { CartSource } from '@bigcommerce/checkout-sdk/payment-integration-api';
 
+import { GQL_REQUEST_URL } from '../common/gql-request';
 import { ContentType, SDK_VERSION_HEADERS } from '../common/http-request';
 import { getResponse } from '../common/http-request/responses.mock';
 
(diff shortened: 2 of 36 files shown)


packages/core/src/cart/cart-action-creator.ts
```diff
@@ -0,0 +1,42 @@
...
+import { cachableAction } from '../common/data-store';
+import ActionOptions from '../common/data-store/action-options';
+
+import { CartActionType, LoadCartAction } from './cart-actions';
+import CartRequestSender from './cart-request-sender';
+
+export default class CartActionCreator {
+    constructor(private _cartRequestSender: CartRequestSender) {}
+
+    @cachableAction
```
animesh1987, line 16: Do we need to cache this action?

packages/core/src/cart/cart-action-creator.ts
```diff
@@ -0,0 +1,42 @@
...
+    loadCard(
```
animesh1987, line 17:
```suggestion
    loadCart(
```

packages/core/src/cart/cart-action-creator.ts
```diff
@@ -0,0 +1,42 @@
...
+        cartId: string,
+        options?: RequestOptions & ActionOptions,
+    ): ThunkAction<LoadCartAction, InternalCheckoutSelectors> {
+        return (store) => {
+            return Observable.create((observer: Observer<LoadCartAction>) => {
```
animesh1987, line 22: I think `Observable.create` is deprecated. You might need to use `new Observable` here?

packages/core/src/cart/cart-action-creator.ts
```diff
@@ -0,0 +1,42 @@
...
+                const state = store.getState();
+                const host = state.config.getHost();
+
+                observer.next(createAction(CartActionType.LoadCartRequested, undefined));
+
+                this._cartRequestSender
+                    .loadCard(cartId, host, options)
```
animesh1987, line 29:
```suggestion
                    .loadCart(cartId, host, options)
```

animesh1987 (commented):
I have a general question around is it needed to add methods to `loadCart`, since we already gave `loadCheckout`. Cart is a sub entity of checkout so if we fetch checkout we can extract the cart object from there.

packages/core/src/cart/cart-action-creator.ts
bc-nick: @animesh1987 we need to cache to avoid redundant requests in case when more that one wallet button is initialized

packages/core/src/cart/cart-action-creator.ts
bc-nick: Make sense, thanks

bc-nick:
> I have a general question around is it needed to add methods to `loadCart`, since we already gave `loadCheckout`. Cart is a sub entity of checkout so if we fetch checkout we can extract the cart object from there.

@animesh1987 yeap, you are right. We need to be able to get some information from `CheckoutStoreState` for a few wallet buttons, because they are using `consignment`, `checkout` and other information from the store to initialize and execute the payment process. I will be able to add these changes after I will finish part of main core functionality related to `createHeadlessCheckoutWalletInitializer`. These changes will be easier to add when we have the basic functionality.

packages/core/src/cart/cart-request-sender.spec.ts
```diff
@@ -75,4 +78,35 @@ describe('CartRequestSender', () => {
             });
         });
     });
+
+    describe('#loadCard', () => {
```
bc-peng, line 82:
```suggestion
    describe('#loadCart', () => {
```

bc-peng (commented):
To make things future-proof, we should consider doing loadCheckout instead of loadCart. This change could be incorporated into this PR.

bc-nick:
> To make things future-proof, we should consider doing loadCheckout instead of loadCart. This change could be incorporated into this PR.

@bc-peng thanks for your comment. I want to separate this functionality to another PR because of huge changes/updates, so this part of the work needs to be separated. 
I was able to update POC ([last commit](https://github.com/bigcommerce/checkout-sdk-js/pull/2742/commits/e940e56e505153779c916be27686e9459c9d5832)) with providing functionality related to `loadCheckout` implementation for headless storefronts, take a look pls. Ignore conflicts, they are easily fixed.

But this PR would be better to leave as it is.

cc @animesh1987

packages/core/src/cart/headless-cart/mocks/headless-cart.mock.ts
serhii-tkachenko, line 1: You can transform this mock to constants, since there is no calculations inside functions

packages/core/src/cart/headless-cart/map-to-cart.ts
```diff
@@ -0,0 +1,47 @@
...
+            id: item.entityId,
+            code: item.code,
+            couponType: item.couponType,
+            discountedAmount: item.discountedAmount.value,
+            // TODO:: there is no info about displayName field
+            displayName: '',
+        })),
+        // TODO:: information about email field can be pulled from Billing Address or Shipping Address (https://developer.bigcommerce.com/docs/storefront/cart-checkout/guide/graphql-storefront#get-checkout)
+        email: '',
+        // TODO:: there is no info about customerId field
```
serhii-tkachenko, line 44:
```suggestion
            // Info:: there is no info about displayName field
            displayName: '',
        })),
        // Info:: information about email field can be pulled from Billing Address or Shipping Address (https://developer.bigcommerce.com/docs/storefront/cart-checkout/guide/graphql-storefront#get-checkout)
        email: '',
        // Info:: there is no info about customerId field
```

packages/core/src/cart/headless-cart/map-to-cart.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    let headlessCartResponse: Cart | undefined;
+
+    beforeEach(() => {
+        headlessCartResponse = mapToCart(getHeadlessCartResponse().data.site);
+    });
+
+    it('maps to internal cart', () => {
+        const cart = getCart();
+
+        // TODO:: data is not yet fully compatible due to lack of information
```
serhii-tkachenko, line 18: When do this information will be available?

packages/core/src/cart/headless-cart/map-to-cart-line-item.ts
```diff
@@ -0,0 +1,62 @@
...
+        id: entityId,
+        name,
+        quantity,
+        productId: productEntityId,
+        brand,
+        couponAmount: couponAmount.value,
+        discountAmount: discountedAmount.value,
+        discounts: discounts.map((discount) => ({
+            discountedAmount: discount.discountedAmount.value,
+            // TODO:: discount item does not have name field in response body when making request using REST API, but there is name in interface, for a while set name as entityID
```
serhii-tkachenko, line 37:
```suggestion
            // Info:: discount item does not have name field in response body when making request using REST API, but there is name in interface, for a while set name as entityID
```

packages/core/src/cart/headless-cart/map-to-cart-line-item.ts
```diff
@@ -0,0 +1,62 @@
...
+        url,
+        variantId: variantEntityId,
+        options: selectedOptions?.map((option) => ({
+            name: option.name,
+            nameId: option.entityId,
+            value: option.value,
+            valueId: option.valueEntityId,
+        })),
+
+        // TODO:: we do not have any information regarding to fields below in the GraphQL Storefront doc
```
serhii-tkachenko, line 56:
```suggestion
        // Info:: we do not have any information regarding to fields below in the GraphQL Storefront doc
```

packages/core/src/cart/headless-cart/headless-cart.ts
```diff
@@ -0,0 +1,141 @@
...
+export interface HeadlessLineItems {
+    physicalItems: HeadlessPhysicalItem[];
+    digitalItems: HeadlessDigitalItem[];
+    customItems: HeadlessCustomItem[];
+    giftCertificates?: HeadlessGiftCertificates[];
+}
+
+export default interface HeadlessCartResponse {
+    cart?: {
+        amount: BaseFieldFragment; // cart amount;
```
serhii-tkachenko, line 104: This comment can be removed

packages/core/src/cart/cart-request-sender.ts
```diff
@@ -19,4 +21,41 @@ export default class CartRequestSender {
...
+            params: {
+                cartId,
+            },
+        };
+
+        return this._requestSender
+            .get<HeadlessCartRequestResponse>(url, {
+                ...requestOptions,
+            })
+            .then(this.transformToCartResponse);
```
serhii-tkachenko, line 44: Lets keep one style for handling promises if you are using async

```suggestion
        const response = await this._requestSender
            .get<HeadlessCartRequestResponse>(url, {
                ...requestOptions,
            });
        
        return this.transformToCartResponse(response);    
```

serhii-tkachenko (approved):
Overall looks good to me 👍

packages/core/src/cart/headless-cart/map-to-cart.spec.ts
bc-nick: Need to update graphql request with providing this information. All missing information are not required for current implementation, so we can get rid of it by setting default values

packages/core/src/cart/headless-cart/headless-cart.ts
```diff
@@ -0,0 +1,141 @@
+interface BaseFieldFragment {
+    value: number;
+}
+
+export interface HeadlessLineItem {
```
animesh1987, line 5: Is it a headlessCart or a GQL cart response?

packages/core/src/cart/cart-action-creator.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    @cachableAction
+    loadCart(
+        cartId: string,
+        options?: RequestOptions & ActionOptions,
+    ): ThunkAction<LoadCartAction, InternalCheckoutSelectors> {
+        return (store) => {
+            return new Observable((observer: Observer<LoadCartAction>) => {
+                const state = store.getState();
+                const host = state.config.getHost();
```
animesh1987, line 24: `getHost` reads from config.meta. Do you know when we set the host?

packages/core/src/cart/cart-request-sender.ts
```diff
@@ -19,4 +21,41 @@ export default class CartRequestSender {
...
         return this._requestSender.post(url, { body, headers, timeout });
     }
+
+    async loadCart(
+        cartId: string,
+        host?: string,
+        options?: RequestOptions,
+    ): Promise<Response<Cart | undefined>> {
+        const path = 'cart-information';
+        const url = host ? `${host}/${path}` : `/${path}`;
```
animesh1987, line 31: If there is no host, then are we calling `https://<some-domain>/cart-information`? Where is this endpoint implemented?

packages/core/src/cart/headless-cart/headless-cart.ts
```diff
@@ -0,0 +1,141 @@
...
+        value: number;
+    };
+    salePrice: {
+        value: number;
+    };
+    sku: string;
+    url: string;
+    variantEntityId: number;
+    selectedOptions: Array<{
+        __typename: string;
```
animesh1987, line 44: Do we need this?

packages/core/src/cart/headless-cart/headless-cart.ts
bc-nick: This is GraphQL cart response

packages/core/src/cart/cart-action-creator.ts
bc-nick: For catalyst this param is not required. For default host will be taken as **window.location.origin** when we are making an request. But I set it up directly

packages/core/src/cart/headless-cart/headless-cart.ts
animesh1987: Can we rename it accordingly, since headless cart doesn't necessarily mean GQL. Something like `CartQueryResponse` or `GQLCart`.

packages/core/src/cart/cart-action-creator.ts
animesh1987: If we don't need it maybe we can add it when it's needed, or else it just adds confusion.

packages/core/src/cart/cart-request-sender.ts
animesh1987: Why are we not calling cart GQL get directly?

packages/core/src/cart/headless-cart/headless-cart.ts
bc-nick: This field is available by default. So I added it to interface

animesh1987: If we don't need it, maybe we shall omit it.

packages/core/src/cart/cart-request-sender.ts
bc-nick: Because we have a problem with third party cookies if we try to make request directly. To avoid this problem, we should make the request as server to server

packages/core/src/cart/headless-cart/headless-cart.ts
bc-nick: @animesh1987 In this case it will be like `CartQueryLineItem`. Are you ok with that?

bc-nick: In this case i have to rename all files including interfaces and folder

packages/core/src/cart/cart-request-sender.ts
animesh1987: Do we know who owns this endpoint and how it works?

packages/core/src/cart/headless-cart/headless-cart.ts
animesh1987: Yes please.

packages/core/src/cart/cart-request-sender.ts
davidchin: Ideally, `checkout-sdk` shouldn't hardcode any endpoints exposed through Catalyst, as Catalyst is a reference implementation for headless merchants. This means it can be customised, and the endpoint may not exist or may be located at a different path. Additionally, we have other non-Catalyst headless storefronts.

I suggest allowing the client to specify the URL when initialising the SDK, rather than hardcoding it. For convenience, we can provide a default value (e.g.: `/cart-information`) if the client does not specify one. Furthermore, we should document the expected contract of this endpoint in DevDocs so that custom headless storefronts can have their own implementation.

bc-nick: I have provided ability to specify this URL if it is necessary. Client can provide a host that will be applied with the base path.

@davidchin @animesh1987

davidchin: Thanks Nick, can you point to the line that allows clients to change the default URL `/cart-information` to a different path?

bc-nick: @davidchin here is the [line](https://github.com/bigcommerce/checkout-sdk-js/pull/2802/commits/7d966d5b1460e972cc69099924c1927756918a34#diff-cf8ccb432c580b6c926e2e04561916a63354a20f8ff882a0e27c829ed80198f4R31)

As far as I understood we have to provide ability to change host by client but default path will be the same. It means client is not be able to change `api/wallet-buttons/cart-information` but only `host`.  We have other endpoints that we will work with in a similar fashion such as `api/wallet-buttons/get-initialization-data`, `api/wallet-buttons/get-customer`, `api/wallet-buttons/load-checkout` and so on. Do I understand you correctly?

bc-nick: We have to provide a default value for each of the endpoints, and it doesn't matter if the host is set or not. But based on your suggestion we have to create config for each endpoint to handle it properly.

For example create a config as following:
```
const graphQLRequestConfig = {
  GET_CHECKOUT: 'get-checkout-path' ,
  GET_CART_INFORMATION: 'get-cart-path',
  ...
}
```

We should save this config to the store during wallet button initialization, saving it to [configMetaState](https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/config/config-state.ts#L10) as an example.

If we do not have `graphQLRequestConfig.GET_CART_INFORMATION`, for example, we will set up default url. @davidchin what do you think?

davidchin: > As far as I understood we have to provide ability to change host by client but default path will be the same.

I was referring to the path rather than the host, as requests are expected to be sent to the same host (e.g.: Catalyst storefront) rather than a different one. Depending on the headless implementation, the path may differ from the default path.

> We have to provide a default value for each of the endpoints

I wasn't aware that we needed to expose so many REST endpoints from Catalyst. Taking a step back, I believe we should aim to keep the Catalyst API layer as thin as possible. The reasoning is similar to the above: Catalyst serves as a reference implementation for headless storefronts. Once it is forked, BC has no control over it, meaning we can't update the APIs. Imagine if a new payment wallet requires an additional endpoint or payload, we wouldn't be able to retroactively add it to existing headless Catalyst storefronts.

Instead, could we create a single endpoint that proxies GraphQL requests to the our Storefront GraphQL API? The proxy endpoint would only need to perform a session check to ensure the caller has permission to access the requested data. This approach would be more scalable IMO, as Catalyst would be responsible only for session validation, while BC would handle the actual requests.

bc-nick: I see what you mean. I am not sure that we can make it using one enpoint unfortunately. We have a list of endpoints and we have to define each of them separately. Perhaps one possible way to solve this problem is to create a config that will contain a list of endpoints as i referred [here](https://github.com/bigcommerce/checkout-sdk-js/pull/2802#discussion_r2022442851). @davidchin what do you think?

bc-nick: @davidchin @animesh1987 I've updated the PR with changes related to our last conversation about specifying a URL when initializing the wallet buttons and setting a default value if it wasn't specified. Take a look please :pray:

PR #2811, opened 2025-03-20
feat(checkout): CHECKOUT-8517 Introduce consignment discount interface

bc-mackxu:
## What?
Introduce consignment discount interface

## Why?
To able read consignment discount in checkout-js

## Testing / Proof
exiting test

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/shipping/consignment-discount.ts
+++ b/packages/core/src/shipping/consignment-discount.ts
@@ -1,0 +1,22 @@
+enum ConsignmentDiscountType {
+    Automatic = 'AUTOMATIC',
+    Coupon = 'COUPON',
+}
+
+interface ConsignmentDiscountBase {
+    id: number;
+    amount: number;
+    type: ConsignmentDiscountType;
+}
+
+interface ConsignmentAutomaticDiscount extends ConsignmentDiscountBase {
+    type: ConsignmentDiscountType.Automatic;
+}
+
+interface ConsignmentCouponDiscount extends ConsignmentDiscountBase {
+    type: ConsignmentDiscountType.Coupon;
+    couponId: number;
+    couponCode: string;
+}
+
+export type ConsignmentDiscount = ConsignmentAutomaticDiscount | ConsignmentCouponDiscount;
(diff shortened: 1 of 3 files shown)


packages/core/src/shipping/consignment-discount.ts
```diff
@@ -0,0 +1,22 @@
...
+    type: ConsignmentDiscountType.Automatic;
+}
+
+interface ConsignmentCouponDiscount extends ConsignmentDiscountBase {
+    type: ConsignmentDiscountType.Coupon;
+    couponId: number;
+    couponCode: string;
+}
+
+export type ConsignmentDiscount = ConsignmentAutomaticDiscount | ConsignmentCouponDiscount;
```
bc-peng, line 22: Your solution is easy to understand. I'd go brutal like below. 
```
export type ConsignmentDiscount = 
    | { id: number; amount: number; type: 'AUTOMATIC' }
    | { id: number; amount: number; type: 'COUPON'; couponId: number; couponCode: string };
```

bc-peng (approved):

packages/core/src/shipping/consignment-discount.ts
animesh1987: I think what @bc-mackxu has is good since then we can use `ConsignmentCouponDiscount` and `ConsignmentAutomaticDiscount` can be used separately as well.

The only change I will suggest is in Base interface have `type` as a generic, so we don't need to specify it again in others.

bc-mackxu: Having enum issue when mocking data in checkout-js. I changed to fixed string in SDK, like this `type: 'AUTOMATIC'`

animesh1987: Can you try if this works
```
interface ConsignmentDiscountBase<T> {
    id: number;
    amount: number;
    type: T;
}

interface ConsignmentAutomaticDiscount extends ConsignmentDiscountBase<'AUTOMATIC'> {
}

interface ConsignmentCouponDiscount extends ConsignmentDiscountBase<'COUPON'> {
    couponId: number;
    couponCode: string;
}

```

bc-mackxu: Yes, this is working.

animesh1987 (approved):

PR #2817, opened 2025-03-31
feat(checkout): CHECKOUT-8521 Introduce OrderShippingConsignmentDiscount interface

bc-mackxu:
## What?
Introduce OrderShippingConsignmentDiscount interface

## Why?
To able read the field in checkout-js

## Testing / Proof
- test

### DIFF
--- a/packages/core/src/order/order.ts
+++ b/packages/core/src/order/order.ts
@@ -84,6 +84,12 @@ export interface OrderConsignment {
     shipping: OrderShippingConsignment[];
 }
 
+export interface OrderShippingConsignmentDiscount {
+    id: number;
+    amount: number;
+    code: string | null;
+}
+
 export interface OrderShippingConsignment {
     lineItems: Array<{
         id: number;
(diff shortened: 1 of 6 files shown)


packages/core/src/order/order.ts
```diff
@@ -84,6 +84,12 @@ export interface OrderConsignment {
     shipping: OrderShippingConsignment[];
 }
 
+export interface OrderShippingConsignmentDiscount {
+    id: number;
+    amount: number;
+    code: string | null;
```
animesh1987, line 90: Is it null or undefined?

bc-mackxu: The GET order API is returning null or string.

https://github.com/bigcommerce/bigcommerce/blob/f7b7ee97eca5d23d2408767bf22434413926e0d3/app/Http/Storefront/Api/Order/Consignment/Shipping/AppliedShippingDiscountResponseTransformer.php#L24

image

PR #2826, opened 2025-04-08
feat(checkout): CHECKOUt-8521 Mark discounts field as optional

bc-mackxu:
## What?
Mark discounts field as optional

## Why?
The `discounts` field is not consistently returned yet, as the related API experiment hasn't fully rolled out.

## Testing / Proof
CI

### DIFF
--- a/packages/core/src/order/order.ts
+++ b/packages/core/src/order/order.ts
@@ -126,5 +126,5 @@ export interface OrderShippingConsignment {
         name: string;
         value: string | null;
     }>;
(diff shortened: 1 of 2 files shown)


richapatel1510 (approved):

bc-mackxu:
@animesh1987 Here is the experiment `PROJECT-6643.enable_shipping_discounts_in_orders`

bc-mackxu:
As discussed, we decided to change and use the same flag with API side. When experiment is enabled, the field will be presented. No need to mark as optional

PR #2833, opened 2025-04-15
feat(payment): PI-3825 Added new link for the storefront hosted fields

vitalii-koshovyi:
## What?
Added new link for the storefront hosted fields

## Why?
Due to the new approach of hosted fields

## Testing / Proof
image

### DIFF
--- a/packages/hosted-form-v2/src/hosted-field.ts
+++ b/packages/hosted-form-v2/src/hosted-field.ts
@@ -60,7 +60,7 @@ export default class HostedField {
     private getFrameSrc(orderId?: number): string {
         return typeof orderId !== 'undefined'
             ? `/admin/payments/${this._orderId}/hosted-form-field?version=${LIBRARY_VERSION}`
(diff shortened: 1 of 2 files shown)


packages/hosted-form-v2/src/hosted-field.ts
```diff
@@ -60,7 +60,7 @@ export default class HostedField {
     private getFrameSrc(orderId?: number): string {
```
animesh1987, line 60: Looks ok for now, but maybe we need to think of passing a `client` property here by which we can serve the relevant source.

animesh1987 (approved):

packages/hosted-form-v2/src/hosted-field.ts
vitalii-koshovyi: @animesh1987 Yes, agree. When we want to implement hosted fields v2 on the checkout, this change would be mandatory.

PR #2835, opened 2025-04-15
feat(payment): PI-3225 added loadingIndicator to the google-pay-payment-strategy

bc-dronov:
## What?
Added Loading Indicator to the google-pay-payment-strategy

## Why?
Because of the Sentry issue. 
User clicks on the Place order button before our requests are done, so we have to show loader in this case.

## Testing / Proof
Manual tests

https://github.com/user-attachments/assets/654d9c8b-91cb-4e2f-a2e2-7e1f98cba21a

### DIFF
--- a/packages/google-pay-integration/src/google-pay-payment-strategy.ts
+++ b/packages/google-pay-integration/src/google-pay-payment-strategy.ts
@@ -16,6 +16,7 @@ import {
     PaymentMethodFailedError,
     PaymentStrategy,
 } from '@bigcommerce/checkout-sdk/payment-integration-api';
+import { DEFAULT_CONTAINER_STYLES, LoadingIndicator } from '@bigcommerce/checkout-sdk/ui';
 
 import GooglePayPaymentInitializeOptions, {
     WithGooglePayPaymentInitializeOptions,
@@ -35,6 +36,8 @@ import {
 } from './types';
 
 export default class GooglePayPaymentStrategy implements PaymentStrategy {
+    private _loadingIndicator: LoadingIndicator;
+    private _loadingIndicatorContainer?: string;
     private _paymentButton?: HTMLElement;
     private _clickListener?: (event: MouseEvent) => unknown;
     private _methodId?: keyof WithGooglePayPaymentInitializeOptions;
@@ -42,7 +45,11 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
     constructor(
         protected _paymentIntegrationService: PaymentIntegrationService,
         protected _googlePayPaymentProcessor: GooglePayPaymentProcessor,
-    ) {}
+    ) {
+        this._loadingIndicator = new LoadingIndicator({
+            containerStyles: DEFAULT_CONTAINER_STYLES,
+        });
+    }
 
     async initialize(
         options?: PaymentInitializeOptions & WithGooglePayPaymentInitializeOptions,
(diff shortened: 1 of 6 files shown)


packages/google-pay-integration/src/google-pay-payment-strategy.ts
```diff
@@ -35,14 +36,23 @@ import {
 } from './types';
 
 export default class GooglePayPaymentStrategy implements PaymentStrategy {
+    loadingIndicator: LoadingIndicator;
```
PavlenkoM, line 39:
```suggestion
    private _loadingIndicator: LoadingIndicator;
```

packages/google-pay-integration/src/google-pay-payment-strategy.ts
```diff
@@ -113,6 +123,14 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
         return Promise.resolve();
     }
 
+    toggleLoadingIndicator(isLoading: boolean): void {
```
PavlenkoM, line 126:
```suggestion
    private _toggleLoadingIndicator(isLoading: boolean): void {
```

packages/ui/src/loading-indicator/loading-indicator.ts
```diff
@@ -34,9 +34,10 @@ export default class LoadingIndicator {
         this.container.appendChild(this.indicator);
     }
 
-    show(parentId?: string): void {
-        if (parentId) {
-            const parent = document.getElementById(parentId);
+    show(container?: string | HTMLElement): void {
```
PavlenkoM, line 37: I am not sure that we need to add a new type as argument. Because it will make logic more complex with additional type checks.
If you need to add overlay to the whole checkout app you can use id `checkout-page-container`. I this case you don't need to add any changes to loading-indicator, and cover this changes by tests.

PavlenkoM: Lets try to add loading indicator to `body` by default in case if `container?: string` is optional.
In this case we can keep only one argument type and no need type guards.
• if no `containerId` provided - add loader to body element.
• if `containerId` is provided - make as it is, find element by ID and add loader to it.

bc-dronov: Updated

packages/google-pay-integration/src/google-pay-payment-strategy.ts
bc-dronov: Updated

packages/google-pay-integration/src/google-pay-payment-strategy.ts
```diff
@@ -35,14 +36,24 @@ import {
...
     private _clickListener?: (event: MouseEvent) => unknown;
     private _methodId?: keyof WithGooglePayPaymentInitializeOptions;
 
     constructor(
         protected _paymentIntegrationService: PaymentIntegrationService,
         protected _googlePayPaymentProcessor: GooglePayPaymentProcessor,
-    ) {}
+    ) {
+        this._loadingIndicator = new LoadingIndicator({
+            containerStyles: {
```
vitalii-koshovyi, line 50: would be better to use it from the constant.
it could be a constant on the LoadingIndicator package level.
I see that it's a default styles for loading indicators

bc-dronov: Updated

vitalii-koshovyi (approved):

PR #2836, opened 2025-04-17
chore(common): update node to 22

animesh1987:
## What?
- Update node to 22
- Drop validate commit CI step

## Why?
Update to latest node version.

## Testing / Proof
- CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ aliases:
   - &node_executor
       executor:
         name: node/node
-        node-version: "20.6.0"
+        node-version: "22.13.0"
 
 version: 2.1
 
@@ -66,7 +66,7 @@ jobs:
 
   build:
     <<: *node_executor
-    resource_class: medium+
+    resource_class: large
     steps:
       - ci/pre-setup
       - node/npm-install
@@ -132,7 +132,6 @@ workflows:
       - test-packages
       - lint
       - build
-      - ci/validate-commits
 
       # Only build and push CDN artifact when commits are merged to canonical repo
       - build_cdn:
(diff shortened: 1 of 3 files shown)


snaderiBC (approved):
👏 Noice!

.circleci/config.yml
```diff
@@ -66,7 +66,7 @@ jobs:
 
   build:
     <<: *node_executor
-    resource_class: medium+
+    resource_class: large
```
bc-peng, line 69: Could you please explain the findings behind this change?

animesh1987: Had to increase the size as CI build kept failing due to CPU being maxxed out. Will need to optimize this build as well and then we can lower the class to medium+ again.

bc-peng (approved):

.circleci/config.yml
```diff
@@ -132,7 +132,6 @@ workflows:
       - test-packages
       - lint
       - build
-      - ci/validate-commits
```
davidchin, line 135: So we don't want to validate commit messages anymore?

animesh1987: The present tool for validate messages requires jira information, which seems to not align with our documentation for public repos.

richapatel1510 (approved):

.circleci/config.yml
davidchin: Ok, I recommend updating `validate-commits` to disallow JIRA references for public repos. Otherwise, if we can't enforce it programatically, we will end up with inconsistent commit messages and not able to achieve our goal.

animesh1987: Yes that sounds like a reasonable way forward 👍

PR #2843, opened 2025-04-22
feature(payment): create blank package bigcommerce-payments-integration

serhiiFilkovskyi:
## What?
Create blank package for  bigcommerce-payments-integration  with scope of integration

## Why?
To create separate package for BCP (bigcommerce payments) features

## Testing / Proof
Checkout build command passed successfully


serhii-tkachenko (approved):

packages/bigcommerce-payments-integration/README.md
```diff
@@ -0,0 +1,11 @@
+# bigcommerce-payments-integration
+
+This package contains the integration layer for the BigCommerce provider.
```
bc-peng, line 3: If possible, could you please explain bit more about the purpose of this package?
We have packages like `payment-integration-api` already, `bigcommerce-payments-integration` may confuse developers.

bc-peng: Same questions for the rest 2 related PRs.
We may need to align better on the names of the three packages.

serhii-tkachenko: bigcommerce-payments is a new payment provider, so based on our rules for package creation (and long discussions with payments core team), we decided to use this naming.

serhii-tkachenko: I think in this case it might we can thing about changing `payment-integration-api` package name to `integration-api` since it contains logic that may be used in different (non payment integrations) packages

animesh1987 (approved):

bc-peng (approved):

PR #2847, opened 2025-04-23
feat(payment): upload Stripe client with custom options

bc-dronov:
## What?
Upload Stripe client with custom options

## Why?
Because we need to upload Stripe Link v2 with other options

## Testing / Proof
Manual tests


packages/stripe-integration/src/stripe-upe/stripe-upe-script-loader.ts
```diff
@@ -2,6 +2,8 @@ import { ScriptLoader } from '@bigcommerce/script-loader';
 
 import { PaymentMethodClientUnavailableError } from '@bigcommerce/checkout-sdk/payment-integration-api';
 
+import { StripeConfigurationOptions } from '../stripev3/stripev3';
```
PavlenkoM, line 5: we should not use Stripe V3 interfaces for Stripe UPE, please copy it to UPE with necessary options.

PavlenkoM (approved):
LGTM 👍

PR #2848, opened 2025-04-25
feat(payment): Updated braintree venmo config

bc-nick:
## What?

Updated braintree venmo config

## Why?

This parameter allows you to provide a web-based login experience for customers that don't have the Venmo app installed on their mobile device. When tokenization is initiated and the Venmo app is not installed, customers will see a popup to authenticate themselves by logging into their Venmo account.

## Testing / Proof

Local testing

### DIFF
--- a/packages/core/src/payment/strategies/braintree/braintree-venmo-payment-strategy.ts
+++ b/packages/core/src/payment/strategies/braintree/braintree-venmo-payment-strategy.ts
@@ -1,4 +1,5 @@
 import { PaymentMethodFailedError } from '@bigcommerce/checkout-sdk/payment-integration-api';
+import { isExperimentEnabled } from '@bigcommerce/checkout-sdk/utility';
 
 import { CheckoutStore, InternalCheckoutSelectors } from '../../../checkout';
 import { MissingDataError, MissingDataErrorType } from '../../../common/error/errors';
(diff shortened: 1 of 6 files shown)


packages/core/src/payment/strategies/braintree/braintree-venmo-payment-strategy.ts
```diff
@@ -93,9 +94,18 @@ export default class BraintreeVenmoPaymentStrategy implements PaymentStrategy {
...
+        const isBraintreeVenmoWebFallbackSupport = isExperimentEnabled(
+            features,
+            'PAYPAL-5406.braintree_venmo_web_fallback_support',
+        );
+
         try {
             this._braintreePaymentProcessor.initialize(clientToken);
-            this._braintreeVenmoCheckout = await this._braintreePaymentProcessor.getVenmoCheckout();
+            this._braintreeVenmoCheckout = await this._braintreePaymentProcessor.getVenmoCheckout({
+                mobileWebFallBack: isBraintreeVenmoWebFallbackSupport,
```
serhii-tkachenko, line 107: So when experiment is disabled `mobileWebFallBack` will not be supported?
Or maybe in this case we should not provide any config to getVenmoCheckout? WDYT?

bc-nick: Yes, will not. I added exp here to be able to turn off if something bad happened. We have some problem with venmo testing, I want to be sure that this field will not lead to specific behaviour and in this case we will be able to fix it by turning off this exp. If everything work as expected and problem related to this changes will be fixed I will remove related code and pass this parameter to where we set up this config. So lets leave it as it is.

serhii-tkachenko (approved):
Looks good to me! 👍

PR #2855, opened 2025-05-09
PoC HWB

bc-nick:
## What?

PoC related to Headless Wallet Button implementation

Catalyst: https://github.com/bigcommerce/catalyst/pull/2298

## Why?

  Wallet payment buttons today only work in pre-set store placements and can't be added to headless storefronts. This limits BC merchants and system integrators who want wallet buttons on custom or headless sites.                                                                                                                                                                                        
                                                                                                                                                                                                         
This PR lays the groundwork on the SDK side so that wallet payment buttons can be initialized regardless of predefined placement locations, providing merchants with a convenient way to add wallet buttons to “headless” platforms.

This PR delivers the final approach for the Headless Wallet Button implementation. It introduces a new wallet-button-integration package and extends the existing bundle with a new wallet-button entry point that exposes access to the button strategy initialization process.                                                                                                                               
                                                                                                                                                                                                         
Changes:
1. `packages/core/src/wallet-buttons` — registers wallet button strategies, builds the `wallet-button` bundle entry point, and exposes `WalletButtonInitializer` for initializing/deinitializing wallet buttons by method ID.                                                                                                                                                                               
2. `packages/wallet-button-integration` — a new package providing WalletButtonIntegrationService, a shared service with common methods (billing address updates, payment order intent creation, redirect-to-checkout URL retrieval) that strategy packages can reuse.                                                                                                                               
3. `packages/paypal-commerce-integration` — adds `PaypalCommerceWalletService` and the corresponding wallet strategy (`PaypalCommerceWalletStrategy`), the first consumer of the new integration service, handling PayPal SDK loading, payment tokenization/submission, and button style validation.

## Testing / Proof
...

### DIFF
--- a/packages/wallet-button-integration/src/billing/billing-address-request-sender.ts
+++ b/packages/wallet-button-integration/src/billing/billing-address-request-sender.ts
@@ -1,0 +1,162 @@
+import { RequestSender, Response } from '@bigcommerce/request-sender';
+
+import { GraphQLRequestOptions } from '../graphql-request-options';
+
+import {
+    AddressRequestBody,
+    BillingAddressResponse,
+    BillingAddressResponseBody,
+    BillingAddressUpdateRequestBody,
+} from './billing-address';
+
+export default class BillingAddressRequestSender {
+    constructor(private readonly requestSender: RequestSender) {}
+
+    async updateBillingAddress(
+        graphQLEndpoint: string,
+        checkoutId: string,
+        address: BillingAddressUpdateRequestBody,
+        options?: GraphQLRequestOptions,
+    ): Promise<Response<BillingAddressResponse>> {
+        const document = `
+            mutation UpdateCheckoutBillingAddressMutation(
+                $input: UpdateCheckoutBillingAddressInput!
+            ) {
+                checkout {
+                    updateCheckoutBillingAddress(input: $input) {
+                        checkout {
+                            billingAddress {
+                                address1
+                                address2
+                                city
+                                company
+                                countryCode
+                                email
+                                entityId
+                                firstName
+                                lastName
+                                phone
+                                postalCode
+                                stateOrProvince
+                                stateOrProvinceCode
+                            }
+                        }
+                    }
+                }
+            }
+        `;
+
+        const { id, ...billingAddressFields } = address;
+
+        const requestOptions: GraphQLRequestOptions = {
+            headers: {
+                ...options?.headers,
+                'Content-Type': 'application/json',
+            },
+            body: {
+                ...options?.body,
+                document,
+                variables: JSON.stringify({
+                    input: {
+                        checkoutEntityId: checkoutId,
+                        addressEntityId: address.id,
+                        data: {
+                            address: {
+                                ...billingAddressFields,
+                            },
+                        },
+                    },
+                }),
+            },
+        };
+
+        const response = await this.requestSender.post<BillingAddressResponseBody>(
+            `${window.location.origin}/${graphQLEndpoint}`,
+            requestOptions,
+        );
+
+        return this.transformToBillingAddressResponse(response);
+    }
+
+    async addBillingAddress(
+        graphQLEndpoint: string,
+        checkoutId: string,
+        address: AddressRequestBody,
+        options?: GraphQLRequestOptions,
+    ): Promise<Response<BillingAddressResponse>> {
+        const document = `
+            mutation AddCheckoutBillingAddressMutation(
+                $input: AddCheckoutBillingAddressInput!
+            ) {
+                checkout {
+                    addCheckoutBillingAddress(input: $input) {
+                        checkout {
+                            billingAddress {
+                                address1
+                                address2
+                                city
+                                company
+                                countryCode
+                                email
+                                entityId
+                                firstName
+                                lastName
+                                phone
+                                postalCode
+                                stateOrProvince
+                                stateOrProvinceCode
+                            }
+                        }
+                    }
+                }
+            }
+        `;
+
+        const requestOptions: GraphQLRequestOptions = {
+            headers: {
+                ...options?.headers,
+                'Content-Type': 'application/json',
+            },
+            body: {
+                ...options?.body,
+                document,
+                variables: JSON.stringify({
+                    input: {
+                        checkoutEntityId: checkoutId,
+                        data: {
+                            address: {
+                                ...address,
+                            },
+                        },
+                    },
+                }),
+            },
+        };
+
+        const response = await this.requestSender.post<BillingAddressResponseBody>(
+            `${window.location.origin}/${graphQLEndpoint}`,
+            requestOptions,
+        );
+
+        return this.transformToBillingAddressResponse(response);
+    }
+
+    private transformToBillingAddressResponse(
+        response: Response<BillingAddressResponseBody>,
+    ): Response<BillingAddressResponse> {
+        const {
+            body: {
+                data: {
+                    site: {
+                        checkout: { billingAddress },
+                    },
+                },
+            },
+        } = response;
+
+        return {
+            ...response,
+            body: billingAddress,
+        };
+    }
+}
--- a/packages/wallet-button-integration/src/payment/payment-request-sender.ts
+++ b/packages/wallet-button-integration/src/payment/payment-request-sender.ts
@@ -1,0 +1,168 @@
+import { RequestSender, Response } from '@bigcommerce/request-sender';
+
+import { PaymentMethodCancelledError } from '@bigcommerce/checkout-sdk/payment-integration-api';
+
+import { GraphQLRequestOptions } from '../graphql-request-options';
+
+import {
+    CreatePaymentOrderIntentInputData,
+    CreatePaymentOrderIntentResponse,
+    CreatePaymentOrderIntentResponseBody,
+    CreateRedirectToCheckoutResponse,
+    CreateRedirectToCheckoutResponseBody,
+    RedirectToCheckoutUrlInputData,
+} from './payment';
+import PaymentOrderCreationError from './payment-order-intent-creation-error';
+
+export class PaymentRequestSender {
+    constructor(private readonly requestSender: RequestSender) {}
+
+    async createPaymentOrderIntent(
+        graphQLEndpoint: string,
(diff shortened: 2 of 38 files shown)


packages/headless-wallet-button-integration/src/headless-wallet-button-integration-request-service.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export default class HeadlessWalletButtonIntegrationRequestService {
+    constructor(private readonly requestSender: RequestSender) {}
+
+    updateBillingAddress(
+        checkoutId: string,
+        address: BillingAddressUpdateRequestBody,
+        options?: GraphQLRequestOptions,
+    ): Promise<Response<BillingAddress>> {
+        const path = 'wallet/graphql';
```
bc-nick, line 42: This endpoint will be changed to the appropriate one

packages/headless-wallet-button-integration/src/headless-wallet-button-integration-request-service.ts
```diff
@@ -0,0 +1,117 @@
...
+                    postalCode: string;
+                    stateOrProvince: string;
+                    stateOrProvinceCode: string;
+                };
+            };
+        };
+    };
+}
+
+export default class HeadlessWalletButtonIntegrationRequestService {
```
bc-nick, line 34: This class will be included all required graphql requests for wallet button initialization process, such as: `updateBillingAddress/addBillingAddress`, `updateShippingAddress/addSippingAddress`, `updateShippingOption, getShippingCoutryList`

packages/headless-wallet-button-integration/project.json
```diff
@@ -0,0 +1,23 @@
+{
+    "root": "packages/headless-wallet-button-integration",
```
bc-nick, line 2: Service will be renamed to `wallet-button-integration`

packages/paypal-commerce-integration/src/paypal-commerce-request-sender.ts
```diff
@@ -69,4 +72,140 @@ export default class PayPalCommerceRequestSender {
...
+    ): Promise<string> {
+        const path = 'wallet/graphql';
+
+        const graphQLQuery = `
+          mutation {
+              cart {
+                createCartRedirectUrls(
+                  input: {
+                    paymentWalletData: {
+                      initializationId: 51,
```
bc-nick, line 161: As far as I understood `initializationId` will be taken from session on Catalyst side initializationId was set using `createPaymentOrderIntent`. But in this case we have to modify graphql query while querying that query.

packages/headless-wallet-button-integration/src/headless-wallet-button-integration-request-service.ts
```diff
@@ -0,0 +1,117 @@
+import { RequestSender, Response } from '@bigcommerce/request-sender';
+
+import {
+    BillingAddress,
```
davidchin, line 4: I believe these types come from REST API interface, which may differ from GraphQL, therefore we can't reuse the types. I suggest defining separate ones for GraphQL. Instead of manually defining them, consider using tools such as `graphql-codegen` that can automatically generate the types based on a GraphQL schema.

packages/headless-wallet-button-integration/src/headless-wallet-button-integration-request-service.ts
bc-nick: Yes, sure I am working on it. Thanks

packages/core/src/checkout-buttons/headless-checkout-wallet-initializer.ts
```diff
@@ -0,0 +1,75 @@
...
+        this._store.subscribe((state) => {
+            this._state = createCheckoutButtonSelectors(state);
+        });
+    }
+
+    getState(): CheckoutButtonSelectors {
+        return this._state;
+    }
+
+    subscribe(
```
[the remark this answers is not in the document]
bc-nick: @davidchin the reason why I decided to add it is to give an opportunity retrieve relevant information about current initialisation statuses of available wallet buttons. It can be useful when we want to make some actions based on these statuses. Maybe would be better to leave it as it is?

packages/paypal-commerce-integration/src/paypal-commerce-request-sender.ts
koa75: Hi @davidchin We discussed this case in a private chat. When we create a payment intent, we should not return the initializationId to the client part, but save it in the session. This data is needed so that all requests that pass InitializationId take them on the server part (not the client part) and add them to the necessary mutations (for example, createCartRedirectUrls). Therefore, I had a concern in using graphQL proxies. This was easy to do using the rest API proxy

koa75: Maybe the security team will allow us to use initializationId on the client side, since PayPal checks which merchant is sending this request. We discussed this issue in PR https://github.com/bigcommerce/storefront/pull/3651#discussion_r1998122267

packages/headless-wallet-button-integration/src/headless-wallet-button-integration-service.ts
```diff
@@ -0,0 +1,15 @@
+import { Response } from '@bigcommerce/request-sender';
+
+import {
+    BillingAddress,
+    BillingAddressUpdateRequestBody,
+    RequestOptions,
+} from '@bigcommerce/checkout-sdk/payment-integration-api';
+
+export default interface HeadlessWalletButtonIntegrationService {
```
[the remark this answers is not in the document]
bc-nick: Yeap, I understand how it works. I just wanted to follow to the general approach. I have an idea how we can simplify it. Working on it, thanks

packages/paypal-commerce-integration/src/paypal-commerce-request-sender.ts
bc-nick: @davidchin hi! What are your thoughts regarding @koa75 concerns? Can security team allow us to use `initializationId` on the client side?

packages/headless-wallet-button-integration/src/headless-wallet-button-integration-service.ts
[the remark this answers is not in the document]
bc-nick: @davidchin I have updated this implementation according to your comments. Take a look pls

packages/paypal-commerce-integration/src/paypal-commerce-request-sender.ts
animesh1987: @bc-donfran thoughts on above?

davidchin: Thanks @koa75. I actually didn't know what `initializationId` was. My original was referring to the fact that the query was constructed with a hardcoded value instead of a variable, not whether the value could be exposed to the browser.

For our SF REST API, the value is stored against BC session. However, since SF GQL needs to work without BC session, we need somewhere to store the value once it's generated for a checkout session. Rather than exposing the value to Catalyst via our public API, and having to propagate the value back to BC during the checkout redirect, could we just store the token temporarily against a cart ID in the backend (e.g.: Redis). Then, when we need the value for payment processing, we could just retrieve it in the backend by cart ID without needing the client to pass anything?

bc-donfran: I rather the value not be hardcoded and if it was hardcoded would we really need to display it client side. Also would different initialisation id do different actions it was modified (unsure if there is certain values does certain actions).

davidchin: @koa75 I wasn't referring to Catalyst session. What I meant is that there's no need to return `initializationId` to Catalyst at all, since Catalyst doesn't actually use it. It's only returned so that Catalyst can pass the ID back to BC during the redirect. If the `createPaymentWalletIntent` mutation links the `initializationId` with the `cartId` on BC side, then we should be able to retrieve the `initializationId` using the `cartId` upon redirect, eliminating the need to pass it from Catalyst.

koa75: > If the createPaymentWalletIntent mutation links the initializationId with the cartId on BC side, then we should be able to retrieve the initializationId using the cartId upon redirect, eliminating the need to pass it from Catalyst.

Yes, I completely agree with you. But now initializationId is stored in the session, so to fix this we need to spend a lot of time on spike and refactoring to move initializationId from session to cache. We will try to fix this.

packages/wallet-button-integration/src/billing/billing-address-request-sender.ts
```diff
@@ -0,0 +1,162 @@
...
+    BillingAddressResponse,
+    BillingAddressResponseBody,
+    BillingAddressUpdateRequestBody,
+} from './billing-address';
+
+export default class BillingAddressRequestSender {
+    constructor(private readonly requestSender: RequestSender) {}
+
+    async updateBillingAddress(
+        graphQLEndpoint: string,
```
davidchin, line 16: This can be passed as a constructor parameter, since it doesn't change after initialisation. The same pattern can be applied to other request senders.

packages/wallet-button-integration/src/payment/payment-request-sender.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    async getRedirectToCheckoutUrl(
+        graphQLEndpoint: string,
+        inputData: RedirectToCheckoutUrlInputData,
+        options?: GraphQLRequestOptions,
+    ): Promise<Response<CreateRedirectToCheckoutResponseBody>> {
+        const document = `
+            mutation CheckoutRedirectMutation($input: CreateCartRedirectUrlsInput!) {
+                cart {
+                    createCartRedirectUrls(input: $input) {
```
davidchin, line 114: This mutation doesn't belong to the `payment` namespace as it is used to generate a cart redirect URL.

davidchin (commented):
Nice work @bc-nick! Thank your for patience. I don't have anything to add besides these small comments. I think you're good to break up this POC PR into smaller ones and write tests for each.

PR #2858, opened 2025-05-13
feat(checkout): Skeleton for Web Worker-based Checkout Extension

bc-peng:
## What?
This is a skeleton for building checkout extension using web worker.

## Why?
Support building checkout extension using web worker.

## Testing / Proof
CI.

### DIFF
--- a/packages/core/src/checkout/checkout-service.ts
+++ b/packages/core/src/checkout/checkout-service.ts
@@ -1425,7 +1425,6 @@ export default class CheckoutService {
      * @param region - The name of an area where the extension should be presented.
      * @returns A promise that resolves to the current state.
      */
-    // Keep this interface?
     async renderExtension(container: string, region: ExtensionRegion): Promise<CheckoutSelectors> {
         const action = this._extensionActionCreator.renderExtension(container, region);
         const state = await this._dispatch(action, { queueId: 'extensions' });
@@ -1453,7 +1452,6 @@ export default class CheckoutService {
      * @param extensionId - The ID of an extension to post the event to.
      * @param message - The message to post to an extension.
      */
-    // Keep this interface?
     postMessageToExtension(extensionId: string, message: ExtensionMessage): void {
         this._extensionMessenger.post(extensionId, message);
     }
@@ -1467,7 +1465,6 @@ export default class CheckoutService {
      * @param handler - The handler function for the extension command.
      * @returns A function that, when called, will deregister the command handler.
      */
-    // Keep this interface?
     handleExtensionCommand<T extends keyof ExtensionCommandMap>(
         extensionId: string,
         command: T,
@@ -1485,7 +1482,6 @@ export default class CheckoutService {
      * @param handler - The handler function for the extension query.
      * @returns A function that, when called, will deregister the query handler.
      */
-    // Keep this interface?
     handleExtensionQuery<T extends keyof ExtensionQueryMap>(
         extensionId: string,
         query: T,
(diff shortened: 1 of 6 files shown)


packages/core/src/checkout/checkout-service.ts
bc-peng, line 1: ## Topic 1
Should we aim to maintain the `CheckoutService` interface, as it would minimize the workload for developers switching between different types of extensions?

packages/core/src/extension/extension-action-creator.ts
```diff
@@ -67,12 +68,21 @@ export class ExtensionActionCreator {
 
                     observer.next(createAction(ExtensionActionType.RenderExtensionRequested));
 
-                    const iframe = new ExtensionIframe(container, extension, {
-                        cartId,
-                        parentOrigin: parseUrl(checkoutLink).origin,
-                    });
```
bc-peng, line 73: ## Topic 2.

When rendering an extension, we need to keep a reference for the web worker to send and receive messages.

We have two options: either add the web worker to the existing extensionMessenger or create a new extensionWorker for this purpose. 

In either case, we need to access and add the worker reference here.

packages/core/src/extension/extension-messenger.ts
```diff
@@ -46,6 +46,7 @@ export class ExtensionMessenger {
         }
     }
 
+    // TODO: Add support for handling worker commands
```
bc-peng, line 49: ## Topic 3

The existing iframe-based messaging mechanism does not work for web workers due to the way messages are delivered. 

```javascript
// From worker
worker.onmessage = (event) => {
    console.log('Host got message:', event.data);
};

// To worker
worker.postMessage('hello world');
```
Therefore, we need to update listen and post methods for the worker.
Each one can be a single ticket.

packages/core/src/extension/extension-service.ts
```diff
@@ -20,6 +20,7 @@ import {
 } from './extension-message';
 import { ExtensionQueryType } from './extension-queries';
 
+// TO BE DISCUSSED: Shall we create a new ExtensionService for worker?
```
bc-peng, line 23: ## Topic 4

The extension initialization on the worker side differs from that of the iframe. 

In the iframe, we treat the iframe resizing process as a handshake. 

Given there is no visuals in the worker, should we create a new handshake mechanism, as well as `listen` and `post`?

packages/core/src/extension/extension.ts
```diff
@@ -16,4 +17,5 @@ export const enum ExtensionRegion {
     ShippingSelectedShippingMethod = 'shipping.selectedShippingMethod',
     SummaryAfter = 'summary.after',
     SummaryLastItemAfter = 'summary.lastItem.after',
+    GlobalWebWorker = 'global.webworker',
```
bc-peng, line 20: TBD, it depends on the BE PR.

packages/core/src/extension/extension-messenger.ts
bc-peng: Or should we avoid using `extensionMessenger` and use `extensionWorker` directly instead?

packages/core/src/extension/extension-webworker.ts
bc-peng, line 1: Terminating a worker is simple as `worker.terminate();`.
We might don't need to create worker in this way, just a factory method should be enough.
In our context, we need to clean up listeners for a worker when termintaing one, that does not happen here.

PR #2876, opened 2025-05-26
feat(checkout): Add worker extension messenger and worker-event-poster

richapatel1510:
## What?
Add worker-extension-messenger and worker-event-poster.

## Why?
To support handling life cycle of worker event poster.

## Testing / Proof

Added unit tests.

**Manually testing**

https://github.com/user-attachments/assets/2d7c5b20-ddb0-4047-805b-f65eaaa4ec6d

### DIFF
--- a/packages/core/src/checkout/checkout-service.ts
+++ b/packages/core/src/checkout/checkout-service.ts
@@ -29,6 +29,7 @@ import {
     ExtensionMessenger,
     ExtensionQueryMap,
     ExtensionRegion,
+    WorkerExtensionMessenger,
 } from '../extension';
 import { FormFieldsActionCreator } from '../form';
 import { CountryActionCreator } from '../geography';
@@ -106,6 +107,7 @@ export default class CheckoutService {
         private _subscriptionsActionCreator: SubscriptionsActionCreator,
         private _formFieldsActionCreator: FormFieldsActionCreator,
         private _extensionActionCreator: ExtensionActionCreator,
+        private _workerExtensionMessenger: WorkerExtensionMessenger,
     ) {
         this._errorTransformer = createCheckoutServiceErrorTransformer();
     }
@@ -1426,7 +1428,11 @@ export default class CheckoutService {
      * @returns A promise that resolves to the current state.
      */
     async renderExtension(container: string, region: ExtensionRegion): Promise<CheckoutSelectors> {
-        const action = this._extensionActionCreator.renderExtension(container, region);
+        const action = this._extensionActionCreator.renderExtension(
+            container,
+            region,
+            this._workerExtensionMessenger,
+        );
         const state = await this._dispatch(action, { queueId: 'extensions' });
 
         this._extensionEventBroadcaster.listen();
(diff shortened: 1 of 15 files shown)


packages/core/src/extension/worker-extension-messenger.ts
```diff
@@ -0,0 +1,29 @@
...
+    add(extensionId: string, worker: Worker): void {
+        this._workers[extensionId] = worker;
+    }
+
+    post(extensionId: string, message: ExtensionMessage): void {
+        if (!this._workers[extensionId]) {
+            throw new Error(`Worker with extensionId ${extensionId} not found`);
+        }
+
+        const workerPoster = new WorkerEventPoster(this._workers[extensionId], extensionId);
```
bc-peng, line 21:
```suggestion
        const workerPoster = new WorkerEventPoster<ExtensionCommand, ExtensionCommandOrQueryContext>(this._workers[extensionId], extensionId);
```
Maybe we should initialize it this way. Then define new command/query for translation purpose.

bc-peng: The rest looks good to me.

packages/core/src/common/worker/worker-event-poster.ts
```diff
@@ -0,0 +1,23 @@
+import { IframeEvent } from '../iframe';
+
+export interface IframeEventPostOptions<
+    TSuccessEvent extends IframeEvent,
+    TErrorEvent extends IframeEvent,
+> {
+    errorType?: TErrorEvent['type'];
+    successType?: TSuccessEvent['type'];
+}
```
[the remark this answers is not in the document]
richapatel1510: Removed this, was unused since we don't have method overload for post method here.

packages/core/src/checkout/checkout-service.ts
```diff
@@ -1426,7 +1428,11 @@ export default class CheckoutService {
      * @returns A promise that resolves to the current state.
      */
     async renderExtension(container: string, region: ExtensionRegion): Promise<CheckoutSelectors> {
-        const action = this._extensionActionCreator.renderExtension(container, region);
+        const action = this._extensionActionCreator.renderExtension(
+            container,
+            region,
+            this._workerExtensionMessenger,
```
animesh1987, line 1434: Should this be exposed here? Or handled inside extension action creator?

packages/core/src/extension/extension.ts
```diff
@@ -3,7 +3,7 @@ export interface Extension {
     name: string;
     region: ExtensionRegion;
     url: string;
-    type?: 'iframe' | 'worker';
+    type?: ExtensionType;
```
animesh1987, line 6: Maybe an old thing, shall we remove optional from type?

packages/core/src/checkout/checkout-service.ts
richapatel1510: We need the `_workerExtensionMessenger` reference in the action creator. Other way for achieving this was to add `this._workerExtensionMessenger`  to the checkout store state and then access it from there, but since the reference is already available here, we decided to pass it to `renderExtension` method directly from here.

animesh1987 (approved):

PR #2878, opened 2025-05-26
feat(payment): added canMakePayments method as additional verification of payment capability

bc-nick:
## What?

Added canMakePayments method as additional verification of payment capability

## Why?

For additional check of apple pay payment capability

## Testing / Proof

Manual testing

### DIFF
--- a/packages/apple-pay-integration/src/apple-pay-button-strategy.ts
+++ b/packages/apple-pay-integration/src/apple-pay-button-strategy.ts
@@ -16,7 +16,6 @@ import {
     PaymentIntegrationService,
     PaymentMethod,
     PaymentMethodCancelledError,
-    PaymentMethodFailedError,
     ShippingOption,
     StoreConfig,
 } from '@bigcommerce/checkout-sdk/payment-integration-api';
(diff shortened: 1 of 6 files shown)


packages/apple-pay-integration/src/apple-pay-button-strategy.ts
```diff
@@ -125,6 +126,12 @@ export default class ApplePayButtonStrategy implements CheckoutButtonStrategy {
 
         assertApplePayWindow(window);
 
+        if (!this._sessionFactory.canMakePayment()) {
+            throw new PaymentMethodFailedError(
+                'This device is not capable of making Apple Pay payments',
```
serhii-tkachenko, line 131: The message that could be shown for customer should be translated. I think it might be great to add console.error before throwing an error, and then throw an error with specific key via onError callback (if it is available)

serhii-tkachenko (approved):
Overall looks good to me 👍

packages/apple-pay-integration/src/apple-pay-button-strategy.ts
bc-nick: We do not need to show this error to customer. Otherwise we will be faced with UX issues that can confuse customer a bit.

PR #2886, opened 2025-06-03
feat(checkout): Add listening methods to worker extension messenger

bc-peng:
## What?
Implement listening methods for the worker extension messenger.

## Why?
Enable worker extension to use command and query.

### Manual testing
- The sample worker extension commands `checkout-js` to toggle the loading indication every second.
- The sample worker retrieves the current consignment details.

https://github.com/user-attachments/assets/345e312f-001b-44b5-be56-2f7291d32eb0

### DIFF
--- a/packages/core/src/checkout/checkout-service.spec.ts
+++ b/packages/core/src/checkout/checkout-service.spec.ts
@@ -1580,7 +1580,7 @@ describe('CheckoutService', () => {
         });
 
         it('posts a message to an extension', async () => {
-            jest.spyOn(extensionMessenger, 'post');
+            jest.spyOn(extensionMessenger, 'post').mockImplementation(() => Promise.resolve());
 
             const message = {
                 type: ExtensionMessageType.GetConsignments,
--- a/packages/core/src/extension/extension-messenger.ts
+++ b/packages/core/src/extension/extension-messenger.ts
@@ -41,6 +41,14 @@ export class ExtensionMessenger {
     }
 
     clearCacheById(extensionId: string): void {
+        const extension = this._getExtensionById(extensionId);
+
+        if (extension.type === ExtensionType.Worker) {
+            this._workerExtensionMessenger.clearCacheById(extension.id);
+
+            return;
+        }
+
         if (this._commandListeners[extensionId]) {
             delete this._commandListeners[extensionId];
         }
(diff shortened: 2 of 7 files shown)


packages/core/src/checkout/checkout-service.spec.ts
```diff
@@ -1580,7 +1580,7 @@ describe('CheckoutService', () => {
         });
 
         it('posts a message to an extension', async () => {
-            jest.spyOn(extensionMessenger, 'post');
+            jest.spyOn(extensionMessenger, 'post').mockImplementation(() => Promise.resolve());
```
bc-peng, line 1583: Running `checkoutService.postMessageToExtension('xxx', message)` will throw an error immediately, hence adding mock implementation.

packages/core/src/extension/worker-extension-messenger.ts
```diff
@@ -1,18 +1,111 @@
...
+    stopListen(extensionId: string): void {
+        if (this._commandListeners[extensionId]) {
+            this._commandListeners[extensionId].stopListen();
+        }
+
+        if (this._queryListeners[extensionId]) {
+            this._queryListeners[extensionId].stopListen();
+        }
+    }
+
```
bc-peng, line 108: We may need to refactor the `post` method and add unit tests in another PR.

richapatel1510: https://bigcommercecloud.atlassian.net/browse/CHECKOUT-9248

packages/core/src/extension/extension-messenger.ts
```diff
@@ -127,6 +147,14 @@ export class ExtensionMessenger {
     }
 
     stopListen(extensionId: string): void {
+        const extension = this._getExtensionById(extensionId);
+
+        if (extension.type === ExtensionType.Worker) {
```
animesh1987, line 152: As discussed offline, these should be addresses in a future refactor to split the messenger?

bc-peng: The refactoring would make this file as a proxy, and here will be changed into below:
```
if (extension.type === ExtensionType.Worker) {
        this._workerExtensionMessenger.stopListen(extensionId);
    } else {
        this._iframeExtensionMessenger.stopListen(extensionId);
}
```

PR #2892, opened 2025-06-05
fix(payment): Google Pay payment step custom fields added

bc-dronov:
## What?
When user interacts with a Google Pay payment sheet, we update billing address. In case if merchant has required custom fields we will recieve an error, because we don't pass customFields on billing aggress updating. This PR fix this issue.

## Testing / Proof
Before:

https://github.com/user-attachments/assets/af272935-b9fa-4c93-9776-2a15f0cffc09

After:

https://github.com/user-attachments/assets/280f5b8e-b113-438a-9d1d-afcec3340e07

### DIFF
--- a/packages/google-pay-integration/src/gateways/google-pay-gateway.ts
+++ b/packages/google-pay-integration/src/gateways/google-pay-gateway.ts
@@ -78,10 +78,11 @@ export default class GooglePayGateway {
             company = '',
             phone = '',
             email,
+            customFields = [],
         } = this._paymentIntegrationService.getState().getBillingAddress() || {};
 
         return {
-            ...this._mapToAddressRequestBody(billingAddress, company, phone),
+            ...this._mapToAddressRequestBody(billingAddress, company, phone, customFields),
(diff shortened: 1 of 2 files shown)


packages/google-pay-integration/src/gateways/google-pay-gateway.ts
```diff
@@ -78,11 +78,13 @@ export default class GooglePayGateway {
             company = '',
             phone = '',
             email,
+            customFields = [],
         } = this._paymentIntegrationService.getState().getBillingAddress() || {};
 
         return {
             ...this._mapToAddressRequestBody(billingAddress, company, phone),
             email: email || response.email,
+            customFields,
```
bc-nick: I would like to suggest to pass this parameter to `_mapToAddressRequestBody` function where we return object based on `AddressRequestBody` interface. And cover it by a test to make sure we don't face a similar problem in the future

bc-dronov: Updated

PR #2925, opened 2025-07-14
chore(common): Optimise build times for sdk

animesh1987:
## What?
Optimise build times for sdk.

## Why?
SDK compilation is very slow, almost taking 3-4 mins for build, 1 min for bundle watch. Switch to esbuild-loader to get faster build times.

Have also added an analysis doc and what the next potential changes should be.

## Testing / Proof
- CI
- Screenshots

##### npm run build
Screenshot 2025-07-14 at 4 04 01 pm

##### npm run bundle:watch
Screenshot 2025-07-14 at 4 05 35 pm

##### npm run build
Screenshot 2025-07-14 at 3 50 43 pm

##### npm run bundle:watch
Screenshot 2025-07-14 at 3 51 26 pm


WEBPACK_SPEED_ANALYSIS_RESULTS.md
```diff
@@ -0,0 +1,197 @@
+# Webpack Build Speed Analysis Results
```
davidchin, line 1: Do we need this file? I'd say it makes more sense for this to be a blog post rather than a markdown file in the repo.

package.json
```diff
@@ -26,13 +26,15 @@
   "homepage": "https://github.com/bigcommerce/checkout-sdk-js",
   "scripts": {
     "prepare": "check-node-version --node '>=22' --npm '>=10'",
-    "build": "npm run bundle && npm run bundle-dts",
+    "build": "npx nx reset && npm run bundle && npm run bundle-dts",
```
davidchin, line 29: I don't think we should reset the cache on every build, otherwise it defeats the purpose of having cache in the first place. We only need to reset cache when we are iterating and testing changes to the build script.

packages/core/project.json
```diff
@@ -6,7 +6,11 @@
         "build": {
             "executor": "@nrwl/workspace:run-commands",
             "options": {
-                "command": "webpack --config webpack.config.js"
+                "parallel": false,
+                "commands": [
+                    "webpack --config webpack.config.js",
+                    "nx run core:build-dts"
```
davidchin: I'm not sure why you need to add `nx run core:build-dts`. I can see that `npm run build` runs both `npm run bundle` and `npm run bundle-dts`. Therefore if `nx run core:build-dts` is added to the `build` task, `build-dts` will be run twice when `npm run build` is executed.

scripts/webpack/package-loader-rule.js
```diff
@@ -1,22 +1,38 @@
...
         test: /\.[tj]s$/,
         include: packageSrcPath,
         loader: 'ts-loader',
     });
 
+    // New esbuild-loader rules for all packages
+    esbuildSrcPackages.push({
+        test: /\.[tj]s$/,
+        include: packageSrcPath,
+        exclude: /node_modules/,
```
davidchin: `packageSrcPath` seems to already exclude `node_modules`?

scripts/webpack/package-loader-rule.js
```diff
@@ -1,22 +1,38 @@
...
 const tsSrcPackages = [];
+const esbuildSrcPackages = [];
 const aliasMap = {};
 
-for (const [packageName, packagePath] of Object.entries(projects)) {
-    const packageSrcPath =  path.join(__dirname, '../../', `${packagePath}/src`);
+Object.entries(projects).forEach(([packageName, packagePath]) => {
+    const packageSrcPath = path.join(__dirname, '../../', `${packagePath}/src`);
 
+    // Original ts-loader rules (kept for fallback)
```
davidchin: Why do we need a fallback? When would the fallback be triggered?

webpack-common.config.js
```diff
@@ -69,36 +70,46 @@ const babelEnvPreset = [
     '@babel/preset-env',
     {
         corejs: 3,
-        targets: ['defaults', 'ie 11'],
+        targets: ['defaults'], // Removed IE 11 support
         useBuiltIns: 'usage',
     },
 ];
 
+// Keep babel-loader rules for fallback/comparison (also exclude node_modules)
```
davidchin: I'm not sure why we need to fallback or comparison. If we switch over to use esbuild, I don't think we need Babel anymore.

WEBPACK_SPEED_ANALYSIS_RESULTS.md
animesh1987: This is a working doc for now, since there are follow up items that I want to leverage from this doc. I wanted to break the work before round 2 of improvements as we have achieved a decent state.

package.json
animesh1987: There were issues with the optimisations and nx cache hence I had to reset cache here.

packages/core/project.json
animesh1987: Let me remove this and test again.

scripts/webpack/package-loader-rule.js
animesh1987: This is just for now, once this change is rolled out and is stable I will clean this up.

webpack-common.config.js
animesh1987: This is just for now, once this change is rolled out and is stable I will clean this up.

scripts/webpack/package-loader-rule.js
animesh1987: I have removed this.

scripts/webpack/package-loader-rule.js
```diff
@@ -1,22 +1,30 @@
...
 
-    tsSrcPackages.push({
+    // New esbuild-loader rules for all packages
+    esbuildSrcPackages.push({
         test: /\.[tj]s$/,
         include: packageSrcPath,
-        loader: 'ts-loader',
+        loader: 'esbuild-loader',
+        options: {
+            target: 'es2018', // Modern target without IE 11
```
davidchin, line 18: Let's align the version with `checkout-js` first as our first step, which is `es2015`

PR #2930, opened 2025-07-17
refactor(payment): updated ppcp fastlane implementation to use cookies instead of local storage

serhii-tkachenko:
## What?
Updated ppcp fastlane implementation to use cookies instead of local storage

## Why?
Local storage solution is not quite good in terms of security, since it is not possible to set expiration date for temporary data

## Testing / Proof
Unit tests
Manual tests
CI

Now session id that we use to manage Fastlane authentication check stored in cookies:
Screenshot 2025-07-17 at 10 21 57

And as you can see, it is not present in local storage anymore:
Screenshot 2025-07-17 at 11 35 18

### DIFF
--- a/packages/paypal-commerce-utils/src/paypal-commerce-fastlane-utils.ts
+++ b/packages/paypal-commerce-utils/src/paypal-commerce-fastlane-utils.ts
@@ -7,7 +7,7 @@ import {
     PaymentMethodClientUnavailableError,
     UntrustedShippingCardVerificationType,
 } from '@bigcommerce/checkout-sdk/payment-integration-api';
-import { BrowserStorage } from '@bigcommerce/checkout-sdk/storage';
+import { CookieStorage } from '@bigcommerce/checkout-sdk/storage';
 
 import {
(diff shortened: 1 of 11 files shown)


andriiVitvitskyi1990 (approved):

packages/paypal-commerce-utils/src/paypal-commerce-fastlane-utils.ts
```diff
@@ -7,7 +7,7 @@ import {
     PaymentMethodClientUnavailableError,
     UntrustedShippingCardVerificationType,
 } from '@bigcommerce/checkout-sdk/payment-integration-api';
-import { BrowserStorage } from '@bigcommerce/checkout-sdk/storage';
+import { CookieStorage } from '@bigcommerce/checkout-sdk/storage';
```
bc-nick, line 10: I think would be better to use existed solution for handling cookies from here `packages/core/src/common/storage/browser-storage.ts`. We use it with whole project if we need to use cookies, What do you think?

bc-nick (approved):

PR #2932, opened 2025-07-17
feat(payment): Create Braintree SDK version manager

bc-nick:
## What?

Create Braintree SDK version manager

## Why?

To handle Braintree's version based on experiments

## Testing / Proof

Manual testing

### DIFF
--- a/packages/braintree-utils/src/braintree-sdk-version-manager.ts
+++ b/packages/braintree-utils/src/braintree-sdk-version-manager.ts
@@ -1,0 +1,13 @@
+import { PaymentIntegrationService } from '@bigcommerce/checkout-sdk/payment-integration-api';
+import { BRAINTREE_SDK_STABLE_VERSION } from './braintree-sdk-verison';
+
+export default class BraintreeSDKVersionManager {
+    // TODO: remove ts-ignore when paymentIntegrationService will be used to determine braintree version by experiment name
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    constructor(private paymentIntegrationService: PaymentIntegrationService) {}
+
+    getSDKVersion() {
+        return BRAINTREE_SDK_STABLE_VERSION;
+    }
+}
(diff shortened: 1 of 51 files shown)


packages/braintree-utils/src/braintree-sdk-version-manager.ts
```diff
@@ -0,0 +1,13 @@
+import { PaymentIntegrationService } from '@bigcommerce/checkout-sdk/payment-integration-api';
+import { BRAINTREE_SDK_STABLE_VERSION } from './braintree-sdk-verison';
+
+export default class BraintreeSDKVersionManager {
+    // TODO: remove ts-ignore when paymentIntegrationService will be used to determine braintree version by experiment name
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    constructor(private paymentIntegrationService: PaymentIntegrationService) {}
```
bc-peng, line 8: I see `private paymentIntegrationService`  leads to changes in `core`, could you please explain your idea of having `private paymentIntegrationService` here? Payment methods are supposed to be able to access `paymentIntegrationService` directly.

bc-nick: @bc-peng I left a related comment that should clear that up. I do not no what exactly I have to add because it is simple class to handle braintree sdk version based on experiments that we get via `paymentIntegrationService` API, like that: `getState().getStoreConfig()?.checkoutSettings.features`

This changes related to braintree only and will be removed from `core` as soon as we move all braintree related strategies from `core` to a separate package.

serhii-tkachenko: Hey @bc-peng,
We planned to remove Braintree integration related code in Q3, so changes in core will not be there for a long period of time. `paymentIntegrationService` is needed in BraintreeSDKVersionManager so we can get access to features list to switch Braintree SDK version smoothly, since there are 3d-party services that should update their versions in their apps on our release date (experiment switch).

We already made a brainstorm session to avoid providing paymentIntegrationService on core side of BraintreeSDKVersionManager usage, however Braintree code in core uses BraintreeScriptLoader from braintree-utils package and it might be tricky to use checkoutStore for access to features, since BraintreeScriptLoader class already used in braintree-integration strategies..

Please let me know if you need additional context for this.

andriiVitvitskyi1990 (approved):

PR #2940, opened 2025-07-22
feat(payment): Added liability shift processing for BCP FL

andriiVitvitskyi1990:
## What?
Create order after liability shift passed when 3ds is on for BCP

## Why?
When BCP FL 3ds check is on and liability shift will be "YES" it means that 3ds check passed successfully and only after that create order otherwise we will execute unnecessary operation which will create order that will not be used

## Testing / Proof

https://github.com/user-attachments/assets/bb74b8fb-4dff-4f4a-bcea-cbd9ecc4bb32

https://github.com/user-attachments/assets/b269848b-f169-4739-854f-04e9af8747c8


packages/bigcommerce-payments-integration/src/bigcommerce-payments-fastlane/bigcommerce-payments-fastlane-payment-strategy.ts
```diff
@@ -57,6 +59,7 @@ export default class BigCommercePaymentsFastlanePaymentStrategy implements Payme
         options: PaymentInitializeOptions & WithBigCommercePaymentsFastlanePaymentInitializeOptions,
     ): Promise<void> {
         const { methodId, bigcommerce_payments_fastlane } = options;
+        this.methodId = methodId;
```
bc-nick: It is better to move this line after the `methodId` checking (64 line)

andriiVitvitskyi1990: fixed

PR #2948, opened 2025-07-28
refactor(payment): updated BigCommercePaymentsFastlane strategies implementation to use cookies instead of local storage

serhii-tkachenko:
## What?
updated BigCommercePaymentsFastlane strategies implementation to use cookies instead of local storage

## Why?
Cookies usage for session id storage is better, because it is possible to set proper expiration date

## Testing / Proof
Unit tests
Manual tests
CI

### DIFF
--- a/packages/bigcommerce-payments-utils/src/bigcommerce-payments-fastlane-utils.ts
+++ b/packages/bigcommerce-payments-utils/src/bigcommerce-payments-fastlane-utils.ts
@@ -7,7 +7,7 @@ import {
     PaymentMethodClientUnavailableError,
     UntrustedShippingCardVerificationType,
 } from '@bigcommerce/checkout-sdk/payment-integration-api';
-import { BrowserStorage } from '@bigcommerce/checkout-sdk/storage';
+import { CookieStorage } from '@bigcommerce/checkout-sdk/storage';
 
 import {
     PayPalFastlane,
@@ -27,7 +27,7 @@ import {
 export default class BigCommercePaymentsFastlaneUtils {
     private window: PayPalHostWindow;
 
-    constructor(private browserStorage: BrowserStorage) {
+    constructor() {
         this.window = window;
     }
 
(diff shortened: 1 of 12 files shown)


packages/bigcommerce-payments-utils/src/bigcommerce-payments-fastlane-utils.ts
```diff
@@ -98,24 +98,31 @@ export default class BigCommercePaymentsFastlaneUtils {
...
+        const cookiesOption = {
+            expires,
+            secure: true,
+        };
+
+        CookieStorage.set('bc-fastlane-sessionId', sessionId, cookiesOption);
+    }
+
+    removeStorageSessionId(): void {
+        CookieStorage.remove('bc-fastlane-sessionId', { secure: true });
```
bc-nick: Do we need to provide `{ secure: true }` when we removing cookie by name?

packages/bigcommerce-payments-utils/src/create-bigcommerce-payments-fastlane-utils.ts
```diff
@@ -1,7 +1,5 @@
-import { BrowserStorage } from '@bigcommerce/checkout-sdk/storage';
-
 import BigCommercePaymentsFastlaneUtils from './bigcommerce-payments-fastlane-utils';
 
 export default function createBigCommercePaymentsFastlaneUtils(): BigCommercePaymentsFastlaneUtils {
-    return new BigCommercePaymentsFastlaneUtils(new BrowserStorage('paypalFastlane'));
+    return new BigCommercePaymentsFastlaneUtils();
```
bc-nick, line 4: Suggestion

Maybe would be better to provide cookie name and set it as a private variable to use whole class?

serhii-tkachenko: Thanks @bc-nick for suggestion!

I think it might be better to keep Cookies class as abstract as possible and does not provide any cookies names on private level there. In this case we can set a cookie name as a private variable in BigCommercePaymentsFastlaneUtils strategy, however I didn't see much sense of it for now, since all implementation is located almost in one place.

packages/bigcommerce-payments-utils/src/bigcommerce-payments-fastlane-utils.ts
serhii-tkachenko: Good catch 👍 
I just checked it once again. It is not necessary to provide secure param, when removing cookie. I will remove it

PR #2982, opened 2025-08-20
fix(payment): PI-4266 Apple Pay fails on product pages with reCAPTCHA…

bc-sebastianszafraniec:
… enabled

## What?
Added verifyCheckoutSpamProtection method call for scenarios when customer want's to pay with ApplePay smart payment button from pdp and quickview.

## Why?
As we have spam protection check for scenarios when customer want's to pay with ApplePay smart payment button from cart and minicart https://github.com/bigcommerce/checkout-sdk-js/blob/b7a2a6f8ac607f5c3d070961f294f3e434cc6e68/packages/apple-pay-integration/src/apple-pay-button-strategy.ts#L111 , we don't have it for pdp and quickview because then checkout doesn't exist yet. In this PR there is added verifyCheckoutSpamProtection method call for such a scenarios. VerifyCheckoutSpamProtection is called after checkout is created and loaded. In payment flow it happens after customer clicks ApplePay payment button and before authenticating by finger print.

## Testing / Proof
Unit tests passed.
Tested manually. Unfortunately we can't complete the order when recording the screen so I can't attache any video.
When reCaptcha enabled, completed orders using ApplePay smart payment button from:
- Pdp 
- quickview
- cart
- minicart

https://github.com/user-attachments/assets/37dd1848-74a1-47f8-ae57-356d8f362f6d

Zrzut ekranu 2025-08-20 o 18 32 27

### DIFF
--- a/packages/apple-pay-integration/src/apple-pay-button-strategy.ts
+++ b/packages/apple-pay-integration/src/apple-pay-button-strategy.ts
@@ -357,6 +357,13 @@ export default class ApplePayButtonStrategy implements CheckoutButtonStrategy {
             );
 
             await this._paymentIntegrationService.loadCheckout(buyNowCart.id);
+
+            const state = this._paymentIntegrationService.getState();
+            const { shouldExecuteSpamCheck } = state.getCheckoutOrThrow();
+
+            if (shouldExecuteSpamCheck) {
+                await this._paymentIntegrationService.verifyCheckoutSpamProtection();
+            }
         } catch (error) {
             throw new BuyNowCartCreationError();
         }
--- a/packages/apple-pay-integration/src/apple-pay-button-strategy.spec.ts
+++ b/packages/apple-pay-integration/src/apple-pay-button-strategy.spec.ts
@@ -428,6 +428,37 @@ describe('ApplePayButtonStrategy', () => {
             }
         });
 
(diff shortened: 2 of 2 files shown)


packages/apple-pay-integration/src/apple-pay-button-strategy.ts
```diff
@@ -357,6 +357,13 @@ export default class ApplePayButtonStrategy implements CheckoutButtonStrategy {
             );
 
             await this._paymentIntegrationService.loadCheckout(buyNowCart.id);
+
+            const state = this._paymentIntegrationService.getState();
+            const { shouldExecuteSpamCheck } = state.getCheckoutOrThrow();
+
+            if (shouldExecuteSpamCheck) {
+                await this._paymentIntegrationService.verifyCheckoutSpamProtection();
+            }
```
bc-illia-rezvin, line 366: I have a little question. What if spam check is executed, but, for some reason, fails? Will it stop (due to exception), or will retry indefinitely? How will it look like from the shopper perspective?

bc-sebastianszafraniec: Looks like we have maximum number of tries which is 28 (7000/250) . It retries in 250ms intervals.
https://github.com/bigcommerce/checkout-sdk-js/blob/7cb528000d1e0e234518df9aceadc7dedbf1bf91/packages/core/src/spam-protection/google-recaptcha.ts#L17

bc-illia-rezvin (approved):

bc-peng (approved):

PR #2983, opened 2025-08-22
feat(payment): PAYPAL-5660 add Braintree PayPal message render condition

bc-yaroslav-zhmutskyi:
## What?
Add condition for Braintree PayPal message render
...

## Why?
We have situations when Pay Later method (and button) on checkout page is disabled but BNPL messaging is still enabled
...

## Testing / Proof
Manual
Unit
...

### DIFF
--- a/packages/braintree-integration/src/braintree-paypal/braintree-paypal-payment-strategy.ts
+++ b/packages/braintree-integration/src/braintree-paypal/braintree-paypal-payment-strategy.ts
@@ -267,7 +267,7 @@ export default class BraintreePaypalPaymentStrategy implements PaymentStrategy {
     }
 
     private async loadPaypalCheckoutInstance() {
-        const { clientToken, initializationData } = this.paymentMethod || {};
+        const { clientToken, initializationData, id: paymentMethodId } = this.paymentMethod || {};
 
         if (!clientToken) {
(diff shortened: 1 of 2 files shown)


packages/braintree-integration/src/braintree-paypal/braintree-paypal-payment-strategy.ts
```diff
@@ -285,12 +285,20 @@ export default class BraintreePaypalPaymentStrategy implements PaymentStrategy {
                 isCreditEnabled: initializationData?.isCreditEnabled,
             };
 
+            const isBraintreePaypalCredit = paymentMethodId === 'braintreepaypalcredit';
```
bc-nick, line 288: `isBraintreePaypalCredit` always be set to `false` as each strategy only works with one payment method

bc-yaroslav-zhmutskyi: But we don't have braintree-paypal-credit-payment-strategy. And this one works for both BT PayPal and BT PayPal credit

bc-nick: Yeap, your are right. Did not notice it. Well done!

Screenshot 2025-08-22 at 11 35 50

PR #3008, opened 2025-09-19
feat(payment): PAYPAL-5386 remove new apple pay version param

bc-yaroslav-zhmutskyi:
## What/Why?
Remove param for new apple pay version

## Rollout/Rollback
Revert

## Testing
No need

### DIFF
--- a/packages/apple-pay-integration/src/apple-pay-button-initialize-options.ts
+++ b/packages/apple-pay-integration/src/apple-pay-button-initialize-options.ts
@@ -13,11 +13,6 @@ export default interface ApplePayButtonInitializeOptions {
     requiresShipping?: boolean;
 
     /**
-     * Enabling a new version of Apple Pay with using Apple Pay SDK
-     */
-    isWebBrowserSupported?: boolean;
-
-    /**
      * The options that are required to initialize Buy Now functionality.
      */
     buyNowInitializeOptions?: {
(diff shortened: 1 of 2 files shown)


docs/interfaces/ApplePayButtonInitializeOptions.md
```diff
@@ -12,7 +12,6 @@ DOM. When a customer clicks on it, it will trigger Apple sheet.
 ### Properties
 
 - [buyNowInitializeOptions](ApplePayButtonInitializeOptions.md#buynowinitializeoptions)
-- [isWebBrowserSupported](ApplePayButtonInitializeOptions.md#iswebbrowsersupported)
```
bc-nick: You do not need to update these docs because it will be handled by default during bumping version of checkout-sdk

packages/apple-pay-integration/src/apple-pay-button-initialize-options.ts
```diff
@@ -12,11 +12,6 @@ export default interface ApplePayButtonInitializeOptions {
      */
     requiresShipping?: boolean;
 
-    /**
-     * Enabling a new version of Apple Pay with using Apple Pay SDK
-     */
-    isWebBrowserSupported?: boolean;
```
bc-nick, line 18: Why don't you remove this parameter from the Apple Pay button strategy?

bc-yaroslav-zhmutskyi: There is no such parameter at Apple Pay button strategy. You already delete it [commit](https://github.com/bigcommerce/checkout-sdk-js/commit/07c7cb6cb1dedfb52cd80b7ff1b3314e287ce21e#diff-1e21ac0b56385d124ac395b7ea974cf5c4b5bfe4975ad290180a53c6e6cfb4ab)

bc-nick: My bad. I had the old one master.

PR #3217, opened 2026-04-14
chore(checkout): CHECKOUT-9950 Add experiment aware request sender

animesh1987:
## What/Why?
Add experiment aware request sender to append subfolder locale before making API requests.

## Rollout/Rollback
- revert this PR

## Testing
- CI
- Screencast

- When a store has subfolder for base path

https://github.com/user-attachments/assets/dde17fe4-3b3c-48d8-99c5-dca72e0d0bdf

- When a store does not have subfolder for base path

https://github.com/user-attachments/assets/fd3fa065-77ee-40a1-aafc-997699b749d5

- When a store does not have subfolder for base path but checkout is on subfolder

https://github.com/user-attachments/assets/4d63ca01-937b-4976-b6c2-58a7c118c627

- For a store with no subfolder and experiment on

https://github.com/user-attachments/assets/b0420662-490e-47d0-bf44-efcd9e6375d1

- For a store with no subfolder and experiment off

https://github.com/user-attachments/assets/11de7b89-a6f4-409d-b966-f63b1ce41b52

---

> [!NOTE]
> **Medium Risk**
> Touches core checkout networking by routing many mutating API calls through a new URL-prefixing wrapper gated by a feature flag; mistakes could cause checkout write requests to hit the wrong endpoint in some configurations.
> 
> **Overview**
> **Adds `ExperimentAwareRequestSender`** to conditionally prefix *mutating* request URLs (`post`/`put`/`patch`/`delete`) with the store `links.baseUrl` when the `CHECKOUT-9950.update_sf_checkout_url_for_subfolder` feature flag is enabled, while leaving `get`/`sendRequest` unchanged and skipping prefixing for absolute URLs or when `baseUrl` ends with `/checkout`.
> 
> **Routes most checkout request senders through the new wrapper** in `createCheckoutService` (order, checkout, consignments, customer, coupons, instruments, shipping, etc.) so write operations can target subfolder storefront URLs; adds unit coverage for the prefixing and guard conditions. Also extends `StoreLinks` in both `core` and `payment-integration-api` configs to include optional `baseUrl`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0df4477db41ee603f793c0ed0fb0766ea179fa2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/checkout/create-checkout-service.ts
+++ b/packages/core/src/checkout/create-checkout-service.ts
@@ -1,9 +1,10 @@
-import { createRequestSender } from '@bigcommerce/request-sender';
+import { createRequestSender, RequestSender } from '@bigcommerce/request-sender';
 import { createScriptLoader } from '@bigcommerce/script-loader';
 
 import { BillingAddressActionCreator, BillingAddressRequestSender } from '../billing';
 import { createDataStoreProjection } from '../common/data-store';
 import { ErrorActionCreator, ErrorLogger } from '../common/error';
+import { ExperimentAwareRequestSender } from '../common/http-request';
 import { getDefaultLogger } from '../common/log';
 import { getEnvironment } from '../common/utility';
 import { ConfigActionCreator, ConfigRequestSender, ConfigState, ConfigWindow } from '../config';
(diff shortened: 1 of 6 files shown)


packages/core/src/checkout/create-checkout-service.ts
```diff
@@ -113,12 +114,17 @@ export default function createCheckoutService(options?: CheckoutServiceOptions):
     const { locale = '', shouldWarnMutation = true } = options || {};
     const requestSender = createRequestSender({ host: options && options.host });
     const store = createCheckoutStore({ config }, { shouldWarnMutation });
+    const experimentRequestSender = new ExperimentAwareRequestSender(requestSender, {
```
davidchin, line 117: Could we just pass the `baseUrl` as the `host` property to `createRequestSender` if experiment is on?

bc-peng (approved):

packages/core/src/checkout/create-checkout-service.ts
animesh1987: I had a look at this, but then we need to pass the experiment in different ways tied to a template.

davidchin (approved):

PR #3226, opened 2026-04-21
chore(checkout): CHECKOUT-9720 Pass version to api from store credit API

animesh1987:
## What/Why?
Pass version to api from store credit API

## Rollout/Rollback
- revert this PR

## Testing
- CI
- Screencast

#### version not from client

https://github.com/user-attachments/assets/c115e7f3-b092-4a1e-b893-60320722a03d

---

> [!NOTE]
> **Medium Risk**
> Changes the request contract for applying/removing store credit by adding a `version` payload, which could break or alter behavior if the backend/client expectations diverge. Scope is limited to store credit flows and related typings/tests.
> 
> **Overview**
> Adds optimistic concurrency support to store credit operations by introducing `version` on `Checkout` and `RequestOptions`, then sending that version in the request body for `POST`/`DELETE /store-credit`.
> 
> Updates `StoreCreditActionCreator` to default the version from the current checkout state (unless explicitly provided) and adjusts unit tests/mocks to assert the version is propagated through `CheckoutService` → request sender.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff8d8336cb9e0fc8913e4d062c690ef7ebd1e0ff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/checkout/checkout-service.spec.ts
+++ b/packages/core/src/checkout/checkout-service.spec.ts
@@ -1350,7 +1350,7 @@ describe('CheckoutService', () => {
 
             expect(storeCreditRequestSender.applyStoreCredit).toHaveBeenCalledWith(
                 getCheckout().id,
-                options,
+                { ...options, version: getCheckout().version },
             );
         });
 
@@ -1361,7 +1361,7 @@ describe('CheckoutService', () => {
 
             expect(storeCreditRequestSender.removeStoreCredit).toHaveBeenCalledWith(
                 getCheckout().id,
-                options,
+                { ...options, version: getCheckout().version },
             );
         });
     });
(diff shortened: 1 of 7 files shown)


packages/core/src/checkout/checkout-service.ts
```diff
@@ -1247,9 +1247,14 @@ export default class CheckoutService {
      */
     applyStoreCredit(
         useStoreCredit: boolean,
+        version?: number,
```
davidchin: I think this would be a breaking change as it changes the order of params. What if we add an optional version property to `options` instead?

animesh1987: Yeah that was my first approach, but then I reverted to this. Thanks for sanity checking, I will update this.

davidchin (approved):

bc-peng (approved):

PR #3248, opened 2026-05-14
feat(checkout): CHECKOUT-9820 implement fetching b2b checkout config for PO

richapatel1510:
## What/Why?

Adds a new `loadPoConfig()` SDK method (plus `getPoConfig(), isLoadingPoConfig(), getLoadPoConfigError()` accessors) so the checkout-js `usePoConfig` hook can read the B2B store's purchase order (PO) configuration — whether the PO field is enabled, its label, and whether it's required.

Fetches from `GET {b2bApiSettings.baseUrl}/api/v2/store-configs/checkout`, authenticated with the B2B token from state (with a BC JWT fallback for the `authToken` header when the B2B token hasn't been loaded yet).

## Rollout/Rollback

Revert this PR

## Testing

- CI checks

### DIFF
--- a/packages/core/src/po-config/po-config-request-sender.ts
+++ b/packages/core/src/po-config/po-config-request-sender.ts
@@ -1,0 +1,58 @@
+import { RequestSender, Response } from '@bigcommerce/request-sender';
+
+import { RequestOptions, SDK_VERSION_HEADERS } from '../common/http-request';
+
+export interface PoConfigStoreSetting {
+    id?: number;
+    value: string;
+    type?: string;
+}
+
+export interface PoConfigResponseData {
+    checkoutPaymentPurchaseEnableExtra: PoConfigStoreSetting;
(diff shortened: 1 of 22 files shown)


packages/core/src/po-config/po-config-request-sender.ts
```diff
@@ -0,0 +1,58 @@
...
+                    timeout: options?.timeout,
+                    params: { app_client_id: appClientId },
+                    headers: SDK_VERSION_HEADERS,
+                },
+            );
+
+            bcToken = body.token;
+        }
+
+        return this._requestSender.get(`${b2bBaseUrl}/api/v2/store-configs/checkout`, {
```
bc-peng, line 48: I see, we may not need to do this as PO config may be surfaced at the initializing checkout like `extraFields`.

richapatel1510:
This is not required anymore. Will raise separate PR to get these config from checkoutSettings

PR #3250, opened 2026-05-14
feat(payment): PAYPAL-6469 wallet-button-integration package with core bundle integration

bc-yaroslav-zhmutskyi:
## What/Why?
Scaffolded the @bigcommerce/checkout-sdk/wallet-button-integration package and wire it into the core bundle for wallet buttons. 

- The WalletButtonIntegrationService will be an empty class at this stage — concrete methods will be added in follow-up tickets.
- Empty PaypalCommerceWalletStrategy stub added to packages/paypal-commerce-integration to prevent npm run build-cdn errors on an empty generated file

## Rollout/Rollback
Revert

## Testing
CI Passed

---

> [!NOTE]
> **Medium Risk**
> Introduces a new CDN bundle/type and new strategy auto-generation paths, which affects build outputs and module loading behavior despite minimal runtime logic.
> 
> **Overview**
> Adds a new `wallet-button` entry point to the core SDK: webpack now builds a dedicated `wallet-button` bundle, and `loader`/`loader-cdn` can load it via the new `BundleType.WalletButton`.
> 
> Introduces a new `wallet-buttons` module in `core` with a `createWalletButtonInitializer` API, registry wiring to auto-generated `create*WalletStrategy` factories, and basic initializer/registry tests (including multi-container initialization).
> 
> Scaffolds a new `@bigcommerce/checkout-sdk/wallet-button-integration` package (Nx project, tsconfig/jest/lint) exposing a minimal `WalletButtonIntegrationService` and factory type, and adds a stub `createPayPalCommerceWalletStrategy`/`PayPalCommerceWalletStrategy` export so wallet strategy generation/builds don’t fail. Also updates `auto-export`/`extend-interface` configs, TS path mappings, workspace registration, and CODEOWNERS accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f5c83904ee769452b1c3f436aca1e66d0ecf98df. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/bundles/wallet-button.ts
+++ b/packages/core/src/bundles/wallet-button.ts
@@ -1,0 +1,1 @@
+export { createWalletButtonInitializer } from '../wallet-buttons';
--- a/packages/core/src/wallet-buttons/create-wallet-button-initializer.ts
+++ b/packages/core/src/wallet-buttons/create-wallet-button-initializer.ts
@@ -1,0 +1,22 @@
+import { createWalletButtonIntegrationService } from '@bigcommerce/checkout-sdk/wallet-button-integration';
+
+import * as walletButtonStrategyFactories from '../generated/wallet-button-strategies';
+
+import createWalletButtonStrategyRegistry from './create-wallet-button-strategy-registry';
+import WalletButtonInitializer from './wallet-button-initializer';
+import { WalletButtonInitializerOptions } from './wallet-buttons';
+
+export default function createWalletButtonInitializer(
+    options: WalletButtonInitializerOptions,
+): WalletButtonInitializer {
+    const { graphQLEndpoint } = options;
+
+    const walletButtonIntegrationService = createWalletButtonIntegrationService(graphQLEndpoint);
+
+    const registryV2 = createWalletButtonStrategyRegistry(
+        walletButtonIntegrationService,
+        walletButtonStrategyFactories,
+    );
+
+    return new WalletButtonInitializer(registryV2);
+}
--- a/packages/core/src/wallet-buttons/create-wallet-button-strategy-registry.spec.ts
+++ b/packages/core/src/wallet-buttons/create-wallet-button-strategy-registry.spec.ts
@@ -1,0 +1,49 @@
+import {
+    CheckoutButtonStrategy,
+    toResolvableModule,
+} from '@bigcommerce/checkout-sdk/payment-integration-api';
+import { WalletButtonIntegrationService } from '@bigcommerce/checkout-sdk/wallet-button-integration';
+
+import createWalletButtonStrategyRegistry from './create-wallet-button-strategy-registry';
+
+describe('createWalletButtonStrategyRegistry', () => {
+    let walletButtonIntegrationService: WalletButtonIntegrationService;
+    let testStrategy: CheckoutButtonStrategy;
+
+    beforeEach(() => {
+        walletButtonIntegrationService = new WalletButtonIntegrationService('/graphql');
+        testStrategy = {
+            initialize: jest.fn(),
+            deinitialize: jest.fn(),
+        };
+    });
+
+    it('creates registry with factories pre-registered', () => {
+        const registry = createWalletButtonStrategyRegistry(walletButtonIntegrationService, {
+            createTestStrategy: toResolvableModule(() => testStrategy, [{ id: 'test' }]),
+        });
+        const strategy = registry.get({ id: 'test' });
+
+        expect(strategy).toEqual(testStrategy);
+    });
+
+    it('skips non-resolvable modules', () => {
+        const registry = createWalletButtonStrategyRegistry(walletButtonIntegrationService, {
+            notResolvable: (() => ({})) as any,
+        });
+
+        expect(() => registry.get({ id: 'notResolvable' })).toThrow();
+    });
+
+    it('registers multiple resolve IDs for a single factory', () => {
+        const registry = createWalletButtonStrategyRegistry(walletButtonIntegrationService, {
+            createTestStrategy: toResolvableModule(
+                () => testStrategy,
+                [{ id: 'test' }, { id: 'mock' }],
+            ),
+        });
+
+        expect(registry.get({ id: 'test' })).toEqual(testStrategy);
+        expect(registry.get({ id: 'mock' })).toEqual(testStrategy);
+    });
+});
--- a/packages/core/src/wallet-buttons/index.ts
+++ b/packages/core/src/wallet-buttons/index.ts
@@ -1,0 +1,6 @@
+export { default as createWalletButtonInitializer } from './create-wallet-button-initializer';
+export {
+    BaseWalletButtonInitializeOptions,
+    WalletButtonMethodType,
+    WalletButtonOptions,
+} from './wallet-button-options';
--- a/packages/core/src/wallet-buttons/wallet-button-initializer.spec.ts
+++ b/packages/core/src/wallet-buttons/wallet-button-initializer.spec.ts
@@ -1,0 +1,78 @@
+import { CheckoutButtonStrategy } from '@bigcommerce/checkout-sdk/payment-integration-api';
+
+import { ResolveIdRegistry } from '../common/registry';
+
+import WalletButtonInitializer from './wallet-button-initializer';
+import {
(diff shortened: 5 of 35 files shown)


animesh1987:
hey @bc-yaroslav-zhmutskyi what is the purpose of doing this?

bc-yaroslav-zhmutskyi:
> the purpose

This PR is the first part of Headless Wallet Button implementation based on [POC](https://github.com/bigcommerce/checkout-sdk-js/pull/2855). It set ups @bigcommerce/checkout-sdk/wallet-button-integration package and wire it into the `core` bundle for wallet buttons.

bc-nick:
> hey @bc-yaroslav-zhmutskyi what is the purpose of doing this?

Hi @animesh1987! Getting back to the Headless development process: as the first step in its implementation, the agreed approach, which was implemented as a [POC](https://github.com/bigcommerce/checkout-sdk-js/pull/2855), was separated into several tickets. This PR does not introduce any new logic. It simply implements a previously approved solution.

packages/core/src/wallet-buttons/create-wallet-button-strategy-registry.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+describe('createWalletButtonStrategyRegistry', () => {
+    let walletButtonIntegrationService: WalletButtonIntegrationService;
+
+    beforeEach(() => {
+        walletButtonIntegrationService = new WalletButtonIntegrationService('/graphql');
+    });
+
+    it('creates registry with factories pre-registered', () => {
+        const testStrategy = {} as CheckoutButtonStrategy;
```
bc-nick: I would suggest you to move this variable and make it as `let` to reuse it whenever we need. Clean state you can do using `afterEach`.

packages/core/src/wallet-buttons/wallet-button-initializer.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+        registry = new ResolveIdRegistry();
+        registry.register({ id: 'testmethod' }, () => strategy);
+
+        initializer = new WalletButtonInitializer(registry);
+    });
+
+    describe('initializeWalletButton()', () => {
+        it('initializes strategy with provided options', async () => {
+            const options = {
+                methodId: 'testmethod' as any,
```
bc-nick: Why don't you use `BaseWalletButtonInitializeOptions` here and other places?

packages/core/src/wallet-buttons/index.ts
```diff
@@ -0,0 +1,6 @@
+export { default as createWalletButtonInitializer } from './create-wallet-button-initializer';
+export {
+    BaseWalletButtonInitializeOptions,
+    WalletButtonMethodType,
+    WalletButtonOptions,
+} from './wallet-button-options';
```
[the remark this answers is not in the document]
bc-nick: It is not necessary now

packages/core/src/bundles/wallet-button.ts
```diff
@@ -0,0 +1 @@
+export { createWalletButtonInitializer } from '../wallet-buttons';
```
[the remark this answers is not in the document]
bc-nick: Can be exported when needed

bc-nick (approved):
LGTM

animesh1987:
Hey @bc-nick can you please link the decision to create a separate button integration package. We should aim for ADR in future for project like this.

bc-nick:
> Hey @bc-nick can you please link the decision to create a separate button integration package. We should aim for ADR in future for project like this.

@animesh1987 hi! I believe this (https://github.com/bigcommerce/checkout-sdk-js/pull/2855#pullrequestreview-2915750968) information will be enough to cover your request.

packages/wallet-button-integration/src/wallet-button-integration-service.ts
```diff
@@ -0,0 +1,7 @@
+export default class WalletButtonIntegrationService {
+    constructor(private graphQLEndpoint: string) {}
+
+    getGraphQLEndpoint(): string {
```
davidchin, line 4: Does the service need to expose the GQL endpoint? If it handles the calls internally via its own methods, we should keep the endpoint private to minimise the public API surface area.

packages/core/src/wallet-buttons/create-wallet-button-initializer.ts
```diff
@@ -0,0 +1,22 @@
...
+export default function createWalletButtonInitializer(
+    options: WalletButtonInitializerOptions,
+): WalletButtonInitializer {
+    const { graphQLEndpoint } = options;
+
+    const walletButtonIntegrationService = createWalletButtonIntegrationService(graphQLEndpoint);
+
+    const registryV2 = createWalletButtonStrategyRegistry(
+        walletButtonIntegrationService,
+        walletButtonStrategyFactories,
```
davidchin, line 18: We want to avoid bundling wallet strategies within the core bundle, as that would include all strategies even for providers that are not enabled. Instead, we should load them on demand. You can refer to how payment strategies are loaded for reference: https://github.com/bigcommerce/checkout-sdk-js/blob/6420723321de4adeab3e6ae2fde2ccdc8d3826d7/packages/core/src/checkout/create-checkout-service.ts#L157-L161

Afterwards, you can configure `auto-export.config.json` to auto export wallet strategies to the client side.

packages/core/src/wallet-buttons/wallet-button-initializer.ts
```diff
@@ -0,0 +1,33 @@
+import { bindDecorator as bind } from '@bigcommerce/checkout-sdk/utility';
+
+import { isElementId, setUniqueElementId } from '../common/dom';
+
+import { BaseWalletButtonInitializeOptions, WalletButtonOptions } from './wallet-button-options';
+import WalletButtonRegistryV2 from './wallet-button-strategy-registry-v2';
+
+@bind
+export default class WalletButtonInitializer {
```
davidchin, line 9: We already have an existing initializer with similar functionality. Do you plan to deprecate `CheckoutButtonInitializer` in favour of `WalletButtonInitializer`? The long-term plan isn't clear - it would be helpful to know if these are intended to coexist or if one will eventually supersede the other.

bc-nick: @davidchin hi! `CheckoutButtonInitializer` is not fully compatible for our purposes and brings non-readable code with a lot of mapping mechanism and unnecessary functionality that can lead to code breaking. This initializer makes code difficult to read and understand if we want to use it for HWB purposes.

At the beginning it was the main idea to re-use `CheckoutButtonInitializer` as far as i remember for this project but we eventually decided that a different initializer, one that is much simpler, would be a better fit for HWB.

packages/wallet-button-integration/src/wallet-button-integration-service.ts
bc-nick: Hi @davidchin! This method will be removed in the next PR as we need to keep this endpoint fully private. By now it is served as a temporary placeholder. No worries at all.

packages/core/src/wallet-buttons/create-wallet-button-initializer.ts
bc-nick: @bc-yaroslav-zhmutskyi take a look pls

bc-nick: @davidchin the main concern here, in my opinion, that we are using `sdk` through CDN not as a package in **Catalyst**. It really makes sense when we want to avoid package version updates every time we need to keep everything consistent in Catalyst.

Since we do not use `sdk` as a package and this implementation, which allows strategies to be loaded separately, was added after this approach had already been agreed, we did not take this into account.

In this case i believe that in the future we will be able to add this feature of separating packages in a separate PR. What do you think?

davidchin: We can still load the wallet strategies lazily via the CDN. However, you're right, we don't currently have any examples for you to follow. Feel free to leave it out for now, and we can revisit it later if the file becomes too large and starts negatively impacting LCP.

davidchin (approved):

PR #3253, opened 2026-05-18
fix(checkout): CHECKOUT-9967 surface invalid_billing_address error on order submission

bc-mackxu:
## What/Why?

The order submission endpoint can return `{ type: "invalid_billing_address" }` on a 400, but `OrderRequestSender.submitOrder` had no mapping for it. The raw `RequestError` propagated to checkout-js, which had no way to identify the error type and rendered the generic "Something's gone wrong" modal.

This PR adds an `InvalidBillingAddressError` (mirroring `InvalidShippingAddressError`) and throws it from `submitOrder` when the backend returns `type: "invalid_billing_address"`.

## Rollout/Rollback
Revert

## Testing

- New unit tests for the error class and the `submitOrder` mapping (including `body.detail` propagation).
- `jest order/order-request-sender|order/errors/invalid-(billing|shipping)-address-error` → 20/20 pass.
- Manual: have the order endpoint return `{ status: 400, type: "invalid_billing_address" }` and confirm the SDK rejects with `error.type === "invalid_billing_address"`.

### DIFF
--- a/packages/core/src/order/errors/invalid-billing-address-error.spec.ts
+++ b/packages/core/src/order/errors/invalid-billing-address-error.spec.ts
@@ -1,0 +1,15 @@
+import InvalidBillingAddressError from './invalid-billing-address-error';
+
+describe('InvalidBillingAddressError', () => {
+    it('returns error name', () => {
+        const error = new InvalidBillingAddressError('error');
+
+        expect(error.name).toBe('InvalidBillingAddressError');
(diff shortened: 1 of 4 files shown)


bc-mackxu:
Closed — the root cause was identified in BCApp. A PR has been created to address it: https://github.com/bigcommerce/bigcommerce/pull/68226

PR #3254, opened 2026-05-18
feat(checkout): CHECKOUT-9821 Add Loading B2B Payment Methods

bc-peng:
## What/Why?

Key changes:

* New `payment/company-payment-method-action`.
* New `CheckoutService` API.
   * New `loadCompanyPaymentMethods(options?)` method, dispatched on its own `companyPaymentMethods` queue.
   * New selectors: 
     * `getCompanyPaymentMethods()`
     * `getLoadCompanyPaymentMethodsError()`
     * `isLoadingCompanyPaymentMethods()`
* Update `Cart` type
   * Added optional `companyId?: number | null` to the `Cart` interface

## Rollout/Rollback

CI.

### Before applying the filter.
Screenshot 2026-05-18 at 16 51 25
Screenshot 2026-05-18 at 16 54 14

### After applying the filter.
Screenshot 2026-05-18 at 16 51 00

---

> [!NOTE]
> **Medium Risk**
> Adds a new checkout data slice and service API that calls B2B endpoints and gates behavior on customer/company/token presence, which could impact payment method availability and error/loading handling.
> 
> **Overview**
> Adds support for loading a B2B company’s allowed payment-method allow-list into checkout state. This introduces a new `b2bCompanyPaymentMethods` store slice (actions/reducer/selector/state), wires it into `CheckoutService` via `loadB2BCompanyPaymentMethods()` (queued under `b2bCompanyPaymentMethods`), and exposes corresponding data/error/loading selectors.
> 
> Updates the `Cart` type to include optional `companyId` (used to fetch company methods) and adds request-sender logic/tests to call the B2B `/api/v2/companies/{companyId}/payments` endpoint and normalize `isEnabled` values.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 921371e6d24c1f5ce586e23acdb4a8bfff8f9a68. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/payment/b2b-company-payment-method-request-sender.ts
+++ b/packages/core/src/payment/b2b-company-payment-method-request-sender.ts
@@ -1,0 +1,35 @@
+import { RequestSender, Response } from '@bigcommerce/request-sender';
+
+import { RequestOptions } from '../common/http-request';
+
+export interface B2BCompanyPaymentMethodsResponseBody {
+    data: Array<{
+        code: string;
+        name: string;
+        isEnabled: '1' | '0';
+        paymentId: number;
+    }>;
+}
+
+export default class B2BCompanyPaymentMethodRequestSender {
+    constructor(private _requestSender: RequestSender) {}
+
+    async getB2BCompanyPaymentMethods(
+        companyId: number,
+        b2bToken: string,
+        b2bBaseUrl: string,
+        options?: RequestOptions,
+    ): Promise<Response<B2BCompanyPaymentMethodsResponseBody>> {
+        return this._requestSender.get<B2BCompanyPaymentMethodsResponseBody>(
+            `${b2bBaseUrl}/api/v2/companies/${companyId}/payments`,
+            {
+                timeout: options?.timeout,
+                credentials: false,
+                headers: {
+                    authToken: b2bToken,
+                    Authorization: `Bearer ${b2bToken}`,
+                },
+            },
+        );
+    }
+}
--- a/packages/core/src/cart/cart.ts
+++ b/packages/core/src/cart/cart.ts
@@ -9,6 +9,7 @@ import LineItemMap from './line-item-map';
 export default interface Cart {
     id: string;
     customerId: number;
+    companyId?: number | null;
     currency: Currency;
     email: string;
     isTaxIncluded: boolean;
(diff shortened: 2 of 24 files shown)


packages/core/src/payment/b2b-company-payment-method-request-sender.ts
```diff
@@ -0,0 +1,35 @@
...
+        options?: RequestOptions,
+    ): Promise<Response<B2BCompanyPaymentMethodsResponseBody>> {
+        return this._requestSender.get<B2BCompanyPaymentMethodsResponseBody>(
+            `${b2bBaseUrl}/api/v2/companies/${companyId}/payments`,
+            {
+                timeout: options?.timeout,
+                credentials: false,
+                headers: {
+                    authToken: b2bToken,
+                    Authorization: `Bearer ${b2bToken}`,
```
bc-peng, line 30: I see the current B2B checkout uses both to make a call.

Image

animesh1987:
Overall good work @bc-peng, I think we need to rethink this approach about creating a new selector and method. We have two options here that we can either keep the existing setup or add a check in loadPaymentMethods where we check for a company and return b2b available methods.

I believe in the end both b2b or b2c methods should conform to the same interface so we can have a simpler client side code which can build against one interface.

My preference would be Option 2 — extending loadPaymentMethods with a company check, possibly passed via options property 🤔 . We can pass the b2b api response through a transformer which acts as an ACL between b2c and payment method for now. Eventually when B2B SF API is ready, then we can remove the transformer and it should work seamlessly.

bc-peng:
> Overall good work @bc-peng, I think we need to rethink this approach about creating a new selector and method. We have two options here that we can either keep the existing setup or add a check in loadPaymentMethods where we check for a company and return b2b available methods.
> 
> I believe in the end both b2b or b2c methods should conform to the same interface so we can have a simpler client side code which can build against one interface.
> 
> My preference would be Option 2 — extending loadPaymentMethods with a company check, possibly passed via options property 🤔 . We can pass the b2b api response through a transformer which acts as an ACL between b2c and payment method for now. Eventually when B2B SF API is ready, then we can remove the transformer and it should work seamlessly.

Just noticed a nice benefit of Option 2.
checkout-js only needs a one-file change to use the SDK updates. 😂

bc-peng:
Closing as we'd like to move with another approach. https://github.com/bigcommerce/checkout-sdk-js/pull/3255

PR #3272, opened 2026-05-28
feat(payment): PI-5297 Adyen Sending Extra Payment Details within "paymentData"

bc-ania:
## What/Why?

Merchant has reached out to report that in Adyen Klarna, the payment will show as authorized but in BigCommerce the order is showing as declined/incomplete.
Additional context: “Adyen responded with error for Authorization requests - Gateway authorize response message = Structure of CardDetails contains the following unknown fields: [paymentData, details] and code = 702, thus we considered it as not successful and the order was declined.
In fact, Authorization did happen and later we received authorization success webhook, but we don't have restore mechanism for declined orders. The issue is that we occasionally send extra `details` data within paymentData, which Adyen rejects to proceed.

The solution: don't send `details` if it's an empty object.

## Rollout/Rollback

Revert this PR

## Testing

TBA

---

> [!NOTE]
> **Medium Risk**
> Touches Klarna authorization and additional-action payment submission; scope is limited to three method IDs and only removes empty `details`, but incorrect stripping could break Klarna when `details` is required.
> 
> **Overview**
> Fixes **Adyen Klarna** checkouts where authorization succeeds at Adyen but BigCommerce marks the order declined because the gateway rejects payloads that include an empty **`details`** object inside **`paymentData`**.
> 
> The v3 strategy now centralizes Klarna detection via **`_isKlarnaPaymentMethod`** and strips **`details`** with **`_removeEmptyDetails`** when it is missing or has no keys. That sanitization runs on the initial **`credit_card_token`** payload and on **`onAdditionalDetails`** nonces during extra actions, while non-Klarna flows are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5e3a444d9295bf4925918503a48b25fdfa492375. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.ts
+++ b/packages/adyen-integration/src/adyenv3/adyenv3-payment-strategy.ts
@@ -115,7 +115,6 @@ export default class Adyenv3PaymentStrategy implements PaymentStrategy {
                 'es-PE': { 'creditCard.expiryDateField.title': 'Fecha de caducidad' },
             },
         });
-
         this.paymentComponent = await this._mountPaymentComponent(paymentMethod);
 
         if (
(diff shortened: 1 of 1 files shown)


bc-ania:
cursor review

PR #3274, opened 2026-05-29
fix(payments): reinitialize expired Stipe checkout session

PavlenkoM:
## What/Why?
When the Stripe Checkout (OCS) session expires, the backend rejects submitPayment with an authorization_expired error code. Previously this bubbled up as a generic payment failure and the strategy stayed bound to the now-dead session, so every subsequent attempt failed the same way until the shopper reloaded the page.

This change detects the expired-session error and silently re-initializes the strategy in the background while still surfacing the original error to the caller, so the next payment attempt uses a fresh Stripe Checkout session.

## Rollout/Rollback
Rollback this PR

## Testing
Before:

https://github.com/user-attachments/assets/b7fb80c3-d641-41da-9f63-e8fb7c07935a

After:

https://github.com/user-attachments/assets/315cedb0-d775-4192-82f2-a03254e086e3

---

> [!NOTE]
> **Medium Risk**
> Touches payment error handling and async re-init on the Stripe CS path; behavior is scoped to `authorization_expired` but incorrect reinit could affect subsequent payment attempts without a page reload.
> 
> **Overview**
> When Stripe Checkout (OCS) returns **`authorization_expired`** on `submitPayment`, the CS payment strategy now treats it as an expired session instead of a dead-end failure.
> 
> **`StripeIntegrationService`** gains **`isSessionExpiredError`** (matches `authorization_expired`). **`StripeCSPaymentStrategy`** stores **`stripeocs`** init options on initialize and clears them on deinitialize. In **`_processAdditionalAction`**, that error triggers a **fire-and-forget** **`deinitialize` → `initialize`** with **cloned** cached options and the current gateway/method ids, while **still rethrowing** the original error so the shopper sees the failure but the **next** attempt can use a fresh session.
> 
> Specs cover rethrow, background reinit order, cloned options, and non-`RequestError` paths. Link v2 button/customer specs adjust the Stripe element **`on`** mock so async handler rejections are swallowed (test stability).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a3f4cc588b65128fbf2a17cb75ffcbbbfb34f919. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/stripe-integration/src/stripe-cs/stripe-cs-payment-strategy.spec.ts
+++ b/packages/stripe-integration/src/stripe-cs/stripe-cs-payment-strategy.spec.ts
@@ -1294,7 +1294,7 @@ describe('StripeOCSPaymentStrategy', () => {
                 ).rejects.toBe(sessionExpiredErrorResponse);
 
                 // Background reinit is fire-and-forget — flush microtasks before asserting.
-                await new Promise((resolve) => setImmediate(resolve));
+                await new Promise((resolve) => setTimeout(resolve, 0));
 
                 expect(deinitializeSpy).toHaveBeenCalledTimes(1);
                 expect(initializeSpy).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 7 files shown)


bc-wpietrzak (approved):
LGTM

bc-ania (approved):

PR #3281, opened 2026-06-05
fix(payments): STRIPE-1511 use Stripe session id for submitPayment

PavlenkoM:
## What/Why?
When the Stripe Checkout Session was created on the BE, the strategy was passing the BE-issued clientToken to submitPayment. The live Stripe Checkout Session has its own canonical ID — and that's what the BE actually needs to look up the session and confirm payment.

This change makes the strategy prefer the live session ID from stripeActions.getSession() when calling submitPayment, falling back to clientToken only when no live session ID is available. The result: payment confirmation now matches the session shoppers actually saw, eliminating the class of mismatches that occur when the BE-side session diverges from the client-side one (re-initialization, expiry recovery, etc.).

Alongside that, a few small cleanups inside execute():
  - Store-credit application and Stripe-actions loading now run in parallel — one fewer round-trip in the critical path.
  - clientToken is now read from fresh state (after loadPaymentMethod runs), so any token refresh during the flow is respected.

## Rollout/Rollback
Rollback this PR

## Testing

https://github.com/user-attachments/assets/57e35bde-d2d2-487a-84cd-f54477593569

Screenshot 2026-06-05 at 12 47 16

---

> [!NOTE]
> **Medium Risk**
> Changes which token the backend receives on payment submit—affects checkout payment confirmation for Stripe OCS; scope is limited to one strategy with new unit tests.
> 
> **Overview**
> **Stripe Checkout Session (CS) payment** now sends the **live Stripe session id** from `stripeActions.getSession()` as the `credit_card_token` on the first `submitPayment`, using the backend `clientToken` only when that id is missing. That aligns confirmation with the session the shopper actually used and avoids mismatches after re-init or expiry recovery.
> 
> In **`execute()`**, store-credit application and loading Stripe actions run **in parallel** via `Promise.all`, and **`clientToken` is read from state after** `loadPaymentMethod` (inside `_updateCheckoutSessionData`) so refreshed tokens are used. **`_updateStripeShopperData`** and **`_updateCheckoutSessionData`** now take a shared **`StripeCheckoutSessionActions`** instance instead of calling `_getStripeActionsOrThrow()` again inside each helper; **`initialize()`** loads actions once before updating shopper data on the element.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7acff3c142ab4a5c3c91ecc17e4bdc2ada73b980. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/stripe-integration/src/stripe-cs/stripe-cs-payment-strategy.spec.ts
+++ b/packages/stripe-integration/src/stripe-cs/stripe-cs-payment-strategy.spec.ts
@@ -952,6 +952,83 @@ describe('StripeOCSPaymentStrategy', () => {
             });
         });
 
+        it('prefers live stripe checkout session id over clientToken on first submitPayment', async () => {
+            jest.spyOn(stripeScriptLoader, 'getStripeCheckout').mockReturnValue(
+                Promise.resolve({
+                    ...getStripeCheckoutInstanceMock(),
+                    loadActions: () =>
+                        Promise.resolve({
+                            type: StripeLoadActionsResultType.SUCCESS,
+                            actions: {
+                                ...getStripeCheckoutSessionActionsMock(),
+                                getSession: jest.fn(() =>
+                                    Promise.resolve({
+                                        id: 'cs_live_session_id',
+                                    } as StripeCheckoutSession),
+                                ),
+                            },
+                        }),
+                }),
+            );
+
+            await stripeCSPaymentStrategy.initialize(stripeOptions);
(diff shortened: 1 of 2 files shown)


PavlenkoM:
cursor review

henrikh-kovalenko (approved):

bc-wpietrzak (approved):
LGTM

PR #3293, opened 2026-06-16
feat(checkout): CHECKOUT-10080 Save Company Address

bc-peng:
## What/Why?

Save company address as part of the persisting B2B metadata action.

- Each API call only saves one address.
- No `checkout-js` changes needed.

## Rollout/Rollback

Revert.

### Save billing and shipping with two API calls.
Screenshot 2026-06-15 at 17 45 37
Screenshot 2026-06-15 at 17 45 44

### Save billing and shipping with one API call.
Screenshot 2026-06-15 at 17 55 00

---

> [!NOTE]
> **Medium Risk**
> Touches post-order B2B orchestration and external address persistence; a failed save now fails the whole metadata action before order extra fields.
> 
> **Overview**
> Extends **non-invoice** `persistB2BMetadata` so checkout can persist billing/shipping addresses to the B2B company address book when the shopper opts in via `shouldSaveAddress`.
> 
> When the cart has a `companyId`, the action builds one or more `CreateCompanyAddressPayload` entries from billing and consignment shipping addresses (plus `extraInfo.addressExtraFields`), **deduplicates** matching shipping rows and **merges** billing/shipping into a single payload when the address and extra fields match. It then calls the new `POST .../companies/{id}/addresses` helper **before** `submitOrderExtraFields`; failures block the rest of the metadata flow. Invoice checkout skips this path; no `companyId` or no opt-in skips API calls.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7cbfa93f6ece36fe8942d3e29c968fd0bcfa9459. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/payment/b2b-post-order-action-creator.ts
+++ b/packages/core/src/payment/b2b-post-order-action-creator.ts
@@ -2,19 +2,26 @@ import { createAction, ThunkAction } from '@bigcommerce/data-store';
 import { concat, defer, of } from 'rxjs';
 import { catchError } from 'rxjs/operators';
 
-import { Address } from '../address';
+import { Address, isAddressEqual } from '../address';
 // TODO: CHECKOUT-9979 remove this import before delivery
 import { resolveB2bBaseUrl } from '../b2b-dev-tools';
+import { BillingAddress } from '../billing';
 import { Checkout, InternalCheckoutSelectors } from '../checkout';
 import { throwErrorAction } from '../common/error';
 import { MissingDataError, MissingDataErrorType } from '../common/error/errors';
+import { Consignment } from '../shipping';
 
 import {
     B2BPostOrderActionType,
     PersistB2BMetadataAction,
     PersistB2BMetadataOptions,
 } from './b2b-post-order-actions';
-import B2BPostOrderRequestSender, { QuoteOrderedPayload } from './b2b-post-order-request-sender';
+import B2BPostOrderRequestSender, {
+    AddOrderExtraFieldsPayload,
+    B2BExtraField,
+    CreateCompanyAddressPayload,
+    QuoteOrderedPayload,
+} from './b2b-post-order-request-sender';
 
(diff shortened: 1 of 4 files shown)


packages/core/src/payment/b2b-post-order-action-creator.ts
```diff
@@ -50,6 +58,94 @@ function mapToQuoteOrderedPayload(
...
+            rawAddress: billingAddress,
+            mappedAddress: mapToCompanyAddress(
+                billingAddress,
+                { isBilling: 1, isShipping: 0 },
+                addressExtraFields?.billingAddressExtraFields ?? [],
+            ),
+        });
+    }
+
+    const shippingExtraFields = addressExtraFields?.shippingAddressExtraFields ?? [];
```
bc-peng, line 115: Currently, we only have one set of shipping address extra fields.
If we want to support multi-shipping, we simply need to change here to have the right address extra fields for each new address.

packages/core/src/payment/b2b-post-order-request-sender.ts
```diff
@@ -54,6 +54,36 @@ export interface QuoteOrderedPayload {
...
+    extraFields?: B2BExtraField[];
+}
+
+interface CreateCompanyAddressResponseBody {
+    data: {
+        addressId: string;
+    };
+    code: number;
+}
+
```
richapatel1510, line 86: Do we need to use this `addressId` to pass it to postOrder API ?

richapatel1510 (approved):
LGTM! 👏 
Just added one small question.

packages/core/src/payment/b2b-post-order-request-sender.ts
bc-peng: No, the `addressId` is discarded in the current B2B checkout.

PR #3267, opened 2026-05-27
feat(payment): PI-5216 show full list of payment methods when GooglePay direct-pay flow payment is processed

bc-wpietrzak:
## What/Why?

Currently, in direct pay flow for Google Pay, when the shopper clicks 'Pay' in the modal and the payment is being processed, the payment method list collapses and only shows Google Pay (just like it does for sign-in/sign-out flow).
According to the project requirements, we should be showing the whole list of the methods during this step.

## Rollout/Rollback

Revert

## Testing

Before, the payment method list collapses to Google Pay only:

https://github.com/user-attachments/assets/71df1427-7a50-44c2-94c7-1e06f52bad19

After, we keep the whole payment method list:

https://github.com/user-attachments/assets/cd5aa2bc-021a-422e-ae09-3b695695a901

---

> [!NOTE]
> **Medium Risk**
> Touches the Google Pay direct-pay checkout path and error recovery; behavior change is localized but affects payment-step UI and post-failure state refresh.
> 
> **Overview**
> Google Pay **direct pay** (`_interactWithPaymentSheetAndPay`) no longer calls `loadCheckout()` before `loadPaymentMethod` and `execute` on the happy path, so the checkout UI keeps showing **all payment methods** while payment is processing instead of collapsing to Google Pay only.
> 
> On failure, `execute` is wrapped in **try/catch**: `loadCheckout()` runs once to refresh checkout state, then the error is rethrown (container flow still surfaces `onError` and hides loading via existing handlers). Specs now assert no `loadCheckout` on success, `loadCheckout` on execute failure, and use spies on internal helpers; ESLint allows `no-underscore-dangle` in `*.spec.ts`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f127eb9a19798a20f80f2a88af56500ace96bd4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3247, opened 2026-05-13
feat(payment): PI-5215 [FE][Google Pay] Implement container mode in GooglePayPaymentStrategy

bc-wpietrzak:
## What/Why?

Add container mode support to the Google Pay payment strategy. 
When `container` is provided (instead of walletButton), a Google Pay button is rendered inside the target element, in place of the Place Order button.

Note: this PR only contains changes related to rendering Google Pay button in the SDK; `checkout-js` part is still in progress. Both recordings in Testing section have experiment `PI-5111.google_pay_direct_pay_on_click` on, so stores with disabled experiment won't experience strange behaviour between closing of Google Pay modal and order confirmation.

## Rollout/Rollback

Revert

## Testing

When `container` value is provided:

https://github.com/user-attachments/assets/1c106821-14f3-4fb3-b96f-b59f6bbb560d

When `container` value is not provided (defaults to using current implementation, `walletButton`):

https://github.com/user-attachments/assets/1cf75b6c-07b8-42b0-b4ea-5137b7de89b4

---

> [!NOTE]
> **Medium Risk**
> Changes Google Pay checkout entrypoint by supporting a new container-rendered button and direct-pay flow, affecting initialization, click handling, and teardown paths. Risk is moderate due to new UI/DOM integration and error-handling paths that can impact conversion if misconfigured.
> 
> **Overview**
> Adds an alternative *container mode* to `GooglePayPaymentStrategy`: when `container` is provided (instead of `walletButton`), the strategy renders a branded Google Pay button into the target element and uses that button to launch the payment sheet and proceed with the direct-pay checkout flow.
> 
> Extends `GooglePayPaymentInitializeOptions` with `container`, button styling options (`buttonColor`, `buttonSizeMode`, `buttonType`), and an `onInit(renderButton)` hook to defer rendering until the container exists. Updates deinitialization and centralizes Google Pay widget error handling (including mapping `CANCELED` to `PaymentMethodCancelledError`), with comprehensive new unit coverage for container initialization, click behavior, and cleanup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35133aa1add017f08b39489d0d10632feda2f5c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3299, opened 2026-06-25
fix(payment): fix GooglePay strategy spec crashing under Jest 29 / Node 22

PavlenkoM:
## What/Why?

The `google-pay-integration` test suite was failing on `master` with `Jest worker encountered N child process exceptions, exceeding retry limit` — unrelated to any feature work.

Root cause: the CHECKOUT-10061 toolchain upgrade (TypeScript 4.7→5.8, **Jest 29 + Node 22**, bigcommerce/checkout-sdk-js#3273) made the worker crash on unhandled promise rejections; Jest 26 previously tolerated them. `google-pay-payment-strategy.spec.ts` had two distinct problems, the first masking the second:

1. **Leaked unhandled rejections from fire-and-forget click handlers.** Several tests trigger an async click handler via `button.click()` and never await it. When the handler reports an error via `onError` and re-throws, the rejection is never handled → it crashes the worker. This was amplified by the `button` being created once in `beforeAll` and shared across all tests with no `deinitialize` in between, so click listeners accumulated and a later `button.click()` fired an *earlier* test's throwing handler. (The newer container-mode tests added in bigcommerce/checkout-sdk-js#3247 already avoid this by capturing and awaiting the handler.)
2. **A pre-existing assertion failure masked by the crash.** PI-5216 (bigcommerce/checkout-sdk-js#3267) mocked `_completeCheckoutFlow` to a no-op in the direct-pay describe, but that method is the only place `LoadingHide` is called on the success path — so `should hide loading indicator after completing checkout` could never pass.

This PR fixes the suite with **test-only changes — no production code is touched**:
- Create a fresh `button` per test (`beforeAll` → `beforeEach`, removed in `afterEach`) so click listeners can't accumulate across tests.
- For the two tests whose flow rejects (`onError`, `getNonce`), capture the registered click handler and `await expect(...).rejects.toThrow(...)` instead of fire-and-forget `button.click()` — matching the existing #3247 container-test pattern.
- For `should hide loading indicator after completing checkout`, restore the real `_completeCheckoutFlow` (jsdom only logs the unimplemented `window.location.replace`, it doesn't throw) so the test asserts genuine production behaviour rather than a stub.

## Rollout/Rollback

No production impact — changes are limited to `packages/google-pay-integration/src/google-pay-payment-strategy.spec.ts`. No experiments, feature flags, or database migrations. Rollback is a straight revert of this commit.

## Testing

- `npx nx run google-pay-integration:test` → **71 suites / 414 tests pass**, run repeatedly with no flaky worker crashes (the failure was timing-sensitive).
- `npx eslint packages/google-pay-integration/src/google-pay-payment-strategy.spec.ts` → clean.
- Verified each change is independently necessary: with the fresh button alone, reverting `onError` to `button.click()` still crashes; the `_completeCheckoutFlow` test now exercises the real method (confirmed it calls `LoadingHide` via production code, not a stub).

Before:
Screenshot 2026-06-25 at 14 23 20

After:
Screenshot 2026-06-25 at 14 24 42

---

> [!NOTE]
> **Low Risk**
> Changes are confined to one spec file with no production code, payment runtime, or deployment impact.
> 
> **Overview**
> **Test-only** updates to `google-pay-payment-strategy.spec.ts` so the suite stops crashing Jest workers after the Jest 29 / Node 22 upgrade (unhandled promise rejections are no longer tolerated).
> 
> Wallet button setup now uses **`beforeEach`** to create and append a fresh DOM node and **`afterEach`** to remove it, instead of a shared **`beforeAll`** button that let click listeners stack across tests.
> 
> For flows that **reject** (`onError` on widget failure, coupons path when `getNonce` fails), tests no longer fire **`button.click()`** without awaiting; they pull the registered **`click`** handler from **`addEventListener`** mocks and **`await expect(...).rejects.toThrow(...)`**, matching the existing container-button pattern.
> 
> In the direct-pay-on-click block, **`should hide loading indicator after completing checkout`** restores the real **`_completeCheckoutFlow`** (drops the no-op spy) so **`LoadingHide`** is asserted on the actual success path instead of a stub that never hid loading.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a86996546d67b8c8530506ba3e28d42c01c4b00. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/google-pay-integration/src/google-pay-payment-strategy.spec.ts
+++ b/packages/google-pay-integration/src/google-pay-payment-strategy.spec.ts
@@ -113,7 +113,7 @@ describe('GooglePayPaymentStrategy', () => {
         };
     });
 
-    beforeAll(() => {
+    beforeEach(() => {
         button = document.createElement('div');
 
         button.id = BUTTON_ID;
@@ -125,8 +125,7 @@ describe('GooglePayPaymentStrategy', () => {
     });
 
     afterEach(() => {
-        (button.addEventListener as jest.Mock).mockClear();
-        (button.removeEventListener as jest.Mock).mockClear();
+        button.remove();
     });
 
     describe('#initialize', () => {
@@ -449,9 +448,11 @@ describe('GooglePayPaymentStrategy', () => {
 
             await strategy.initialize(options);
 
-            button.click();
+            const [, clickHandler] = (button.addEventListener as jest.Mock).mock.calls.find(
+                ([eventName]) => eventName === 'click',
+            );
 
-            await new Promise((resolve) => process.nextTick(resolve));
+            await expect(clickHandler(new MouseEvent('click'))).rejects.toThrow(internalError);
 
             expect(LoadingHide).toHaveBeenCalled();
             expect(rejectedInitializeWidgetMock).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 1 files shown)


PavlenkoM:
cursor review

bc-ania (approved):

bc-wpietrzak (approved):

PR #3303, opened 2026-06-30
feat(checkout): CHECKOUT-10140 strip unnecessary company address fields

richapatel1510:
## What/Why?

Strip company address-book-only flags (`isShipping, isBilling, isDefaultShipping, isDefaultBilling`) before passing it to checkout APIs.

## Rollout/Rollback

Revert this PR.

## Testing

- CI checks

---

> [!NOTE]
> **Medium Risk**
> Touches checkout billing and consignment API payloads; behavior is narrowly scoped to omitting invalid fields and is well tested, but mistakes could affect address submission in checkout.
> 
> **Overview**
> Adds **`mapToAddressRequestBody`** to remove address-book-only booleans (`isShipping`, `isBilling`, `isDefaultShipping`, `isDefaultBilling`) from address-like objects before they hit storefront checkout APIs, while leaving real request fields (including `id`, `type`, `country`, `shouldSaveAddress`) unchanged.
> 
> **Billing** create/update and **consignment** create/update now run addresses through this mapper (including nested `address` / `shippingAddress` on consignments) instead of posting a selected `CustomerAddress` verbatim. Unit tests cover the mapper and the request senders’ outgoing bodies.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3f895645dc386130664a69798b1096ad0dd808e7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/address/map-to-address-request-body.ts
+++ b/packages/core/src/address/map-to-address-request-body.ts
@@ -2,7 +2,7 @@ import { omit } from 'lodash';
 
 import { CustomerAddress } from '../customer';
 
-import Address from './address';
+import Address, { AddressRequestBody } from './address';
 
 const CUSTOMER_ADDRESS_METADATA_FIELDS: Array<keyof CustomerAddress> = [
     'isShipping',
@@ -19,6 +19,6 @@ const CUSTOMER_ADDRESS_METADATA_FIELDS: Array<keyof CustomerAddress> = [
  * callers that previously forwarded a selected `CustomerAddress` straight to an
  * update call.
  */
-export default function mapToAddressRequestBody<T extends Partial<Address>>(address: T): T {
-    return omit(address, CUSTOMER_ADDRESS_METADATA_FIELDS) as T;
+export default function mapToAddressRequestBody(address: Partial<Address>): AddressRequestBody {
+    return omit(address, CUSTOMER_ADDRESS_METADATA_FIELDS) as AddressRequestBody;
 }
--- a/packages/core/src/address/map-to-address-request-body.ts
+++ b/packages/core/src/address/map-to-address-request-body.ts
@@ -1,0 +1,24 @@
+import { omit } from 'lodash';
+
+import { CustomerAddress } from '../customer';
+
+import Address from './address';
+
+const CUSTOMER_ADDRESS_METADATA_FIELDS: Array<keyof CustomerAddress> = [
+    'isShipping',
+    'isBilling',
+    'isDefaultShipping',
+    'isDefaultBilling',
+];
+
+/**
+ * Strips `CustomerAddress`-only boolean flags (`isShipping`, `isBilling`,
+ * `isDefaultShipping`, `isDefaultBilling`) from an address-like object so it can be
+ * sent to the API. Every other field — including `id`, `type`, `country` and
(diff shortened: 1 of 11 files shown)


richapatel1510:
cursor review

packages/core/src/address/map-to-address-request-body.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Strips `CustomerAddress`-only boolean flags (`isShipping`, `isBilling`,
+ * `isDefaultShipping`, `isDefaultBilling`) from an address-like object so it can be
+ * sent to the API. Every other field — including `id`, `type`, `country` and
+ * `shouldSaveAddress` — is preserved, so this is a behaviour-safe replacement for
+ * callers that previously forwarded a selected `CustomerAddress` straight to an
+ * update call.
+ */
+export default function mapToAddressRequestBody<T extends Partial<Address>>(address: T): T {
```
animesh1987: Do we need to type the return as generic? Since we can return a shape as address?

richapatel1510: The generic is intentional. `omit` only strips the 4 metadata flags; every other field on the input survives at runtime, so `T → T` keeps the caller's concrete type intact. If we returned a fixed shape (e.g. `Partial<Address>`), callers would lose their extra fields from the type even though they're still present at runtime — e.g. the billing caller passes `Partial<BillingAddressUpdateRequestBody>` and then spreads the result expecting its billing-specific fields (id & email) to still be there, and the consignment caller assigns the result straight back into an `AddressRequestBody` field without a cast. A fixed return type would break both of those.

packages/core/src/billing/billing-address-action-creator.ts
```diff
@@ -107,13 +108,15 @@ export default class BillingAddressActionCreator {
...
+                const sanitizedAddress = mapToAddressRequestBody(address);
+
                 // If email is not present in the address provided by the client, then
                 // fall back to the stored email as it could have been set separately
                 // using a convenience method. We can't rely on billingAddress having
                 // an ID to consider that there's a preexisting email, as billingAddress
                 // object from Order doesn't have an ID.
                 const billingAddressRequestBody = {
-                    ...address,
+                    ...sanitizedAddress,
```
bc-peng: What's the issue you see if we only strip those values in the request sender and pretend all these optional values are part of the API contract?

richapatel1510: Good idea! Moved it to request sender https://github.com/bigcommerce/checkout-sdk-js/pull/3303/commits/65f3b7ee4f96491fea9b475e9eacba511c3f924b

packages/core/src/address/map-to-address-request-body.ts
richapatel1510: After moving it to request sender we can now type it to `AddressRequestBody`
https://github.com/bigcommerce/checkout-sdk-js/pull/3303/commits/65f3b7ee4f96491fea9b475e9eacba511c3f924b

animesh1987 (approved):

bc-maxy (approved):

PR #3304, opened 2026-07-03
fix(payment): PAYPAL-6718 BT google pay refresh 3ds state before verification

bc-yaroslav-zhmutskyi:
## What
Fixes the Braintree Google Pay 3DS checkout error "Cannot call verifyCard while existing authentication is in progress." Before starting a new 3DS verification, the Braintree Google Pay gateway now calls `cancelVerifyCard` to release any authentication left in progress on the reused `threeDSecure` instance.
## Why
BraintreeSdk.getBraintreeThreeDS() caches and reuses a single `threeDSecure` instance. verifyCard sets an internal "authentication in progress" lock that is only released when its promise settles (after the full 3DS challenge). It can be left set on the shared instance when a challenge is interrupted. Calling `cancelVerifyCard()` before each verification clears the stale lock so every attempt starts clean.

## Rollout/Rollback
Revert

#### Before

[3DSInProgressError.webm](https://github.com/user-attachments/assets/aa4294ef-42a2-4228-a2b7-f5b65f0bc27b)

#### After

[withChanges3ds.webm](https://github.com/user-attachments/assets/67562241-2e1f-41c3-8924-ebfc1faa33a7)

Diirect pay(PI-5111.google_pay_direct_pay_on_click=true)

[screen-capture.webm](https://github.com/user-attachments/assets/7cd6cb09-a65d-4702-92fb-045b182ffe28)

---

> [!NOTE]
> **Medium Risk**
> Touches payment 3DS flow and concurrent nonce retrieval; behavior is scoped to Braintree Google Pay with tests, but incorrect cancel timing could affect liability shift or challenge UX.
> 
> **Overview**
> Fixes Braintree Google Pay checkout failures when **3DS** is enabled and a second `getNonce` runs while an earlier `verifyCard` is still in flight on the shared cached `threeDSecure` instance.
> 
> `GooglePayBraintreeGateway` now tracks overlapping verifications with `_verificationsInProgress`. Before starting another challenge it awaits **`cancelVerifyCard`** (errors swallowed when nothing is in progress), then wraps `verifyCard` in **try/finally** so the counter decrements when the promise settles. The 3DS helper type is aligned to **`BraintreeThreeDSecure`**.
> 
> New unit tests cover first verification without cancel, cancel on retry while one is pending, cancel when multiple overlap, and propagation when `verifyCard` rejects without calling cancel.
> 
> The Stripe CS strategy change is formatting only (blank line before `render?.()`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 244fb7d013a972396009b0b9edceb16ab3250e94. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/google-pay-integration/src/gateways/google-pay-braintree-gateway.ts
+++ b/packages/google-pay-integration/src/gateways/google-pay-braintree-gateway.ts
@@ -1,7 +1,7 @@
 import {
     BraintreeGooglePayment,
-    BraintreeGooglePayThreeDSecure,
     BraintreeSdk,
+    BraintreeThreeDSecure,
 } from '@bigcommerce/checkout-sdk/braintree-utils';
 import {
     CancellablePromise,
@@ -27,6 +27,7 @@ export default class GooglePayBraintreeGateway extends GooglePayGateway {
     private _braintreeGooglePayment?: BraintreeGooglePayment;
     private _service: PaymentIntegrationService;
     private _methodId = GooglePayKey.BRAINTREE;
+    private _verificationsInProgress = 0;
 
     constructor(service: PaymentIntegrationService, private _braintreeSdk: BraintreeSdk) {
         super('braintree', service);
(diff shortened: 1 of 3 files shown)


packages/google-pay-integration/src/gateways/google-pay-braintree-gateway.ts
```diff
@@ -154,12 +154,14 @@ export default class GooglePayBraintreeGateway extends GooglePayGateway {
...
 
-    private _braintreePresent3DSChallenge(
-        threeDSecure: BraintreeGooglePayThreeDSecure,
+    private async _braintreePresent3DSChallenge(
+        threeDSecure: BraintreeThreeDSecure,
         amount: number,
         nonce: string,
         bin: string,
     ) {
+        await this._cancelBraintree3DSVerification(threeDSecure);
```
bc-nick: You try to clear stale every time even when there is nothing to stale. As far as I understood this issue can be reproduced on the second attempt since it is the case when stale exists.

Try creating a variable called `isProgress`, which will help us determine whether we need to call `cancelVerifyCard` or if we can skip this step.

bc-nick (approved):
LGTM! Thanks for updates

serhii-tkachenko (approved):
LGTM! 👍

PR #3308, opened 2026-07-07
feat(checkout): CHECKOUT-10142 Add getB2BContext

bc-peng:
## What/Why?

Reflect API changes:
- The internal order endpoint returns  `billingAddressId?: number;` and `shippingAddressId?: number;` when an order is placed with existing company addresses.

Add `getB2BContext` internal selector to return all B2B post order metadata, including:
- billingAddressId?: number;
- shippingAddressId?: number;
- receiptId?: string;
  
## Rollout/Rollback

CI.

#### Invoice flow

https://github.com/user-attachments/assets/4818895f-d00b-4703-8d38-cf3c44e1f593

---

> [!NOTE]
> **Medium Risk**
> Removes `getB2BReceiptId` in favor of `getB2BContext`, which is a breaking change for SDK consumers; behavior affects B2B checkout/order flows.
> 
> **Overview**
> Introduces a unified **`B2BContext`** (`billingAddressId`, `shippingAddressId`, `receiptId`) for B2B post-order metadata, aligned with the internal order API when company addresses are used at checkout.
> 
> **Order state** now persists `b2bContext` from submit/finalize order responses (and clears stale context when a later order omits it). **`OrderSelector#getB2BContext`** reads it from order meta.
> 
> On the public checkout API, **`getB2BReceiptId` is replaced by `getB2BContext`**, which merges order B2B fields with the existing B2B post-order receipt id when either is present.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 01009dd54872dbd00832df3a5929694032e5bfa0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/checkout/checkout-store-selector.ts
+++ b/packages/core/src/checkout/checkout-store-selector.ts
@@ -12,7 +12,7 @@ import { Customer } from '../customer';
 import { Extension, ExtensionRegion } from '../extension';
 import { FormField } from '../form';
 import { Country } from '../geography';
-import { Order } from '../order';
+import { B2BContext, Order } from '../order';
 import { PaymentMethod } from '../payment';
 import { CardInstrument, PaymentInstrument } from '../payment/instrument';
 import { PaymentProviderCustomer } from '../payment-provider-customer';
@@ -67,11 +67,11 @@ export default interface CheckoutStoreSelector {
     getB2BToken(): string | undefined;
 
     /**
-     * Gets the B2B receipt id produced after persisting order metadata.
+     * Gets the B2B context of the current order.
      *
-     * @returns The B2B receipt id string if it has been persisted, otherwise undefined.
+     * @returns The B2B context if it is available, otherwise undefined.
      */
-    getB2BReceiptId(): string | undefined;
+    getB2BContext(): B2BContext | undefined;
 
     /**
      * Gets the shipping address of the current checkout.
@@ -521,9 +521,20 @@ export function createCheckoutStoreSelectorFactory(): CheckoutStoreSelectorFacto
         (getToken) => clone(getToken),
     );
 
-    const getB2BReceiptId = createSelector(
+    const getB2BContext = createSelector(
+        ({ order }: InternalCheckoutSelectors) => order.getB2BContext,
         ({ b2bPostOrder }: InternalCheckoutSelectors) => b2bPostOrder.getReceiptId,
-        (getReceiptId) => clone(getReceiptId),
+        (getB2BContext, getReceiptId) =>
+            clone(() => {
+                const b2bContext = getB2BContext();
+                const receiptId = getReceiptId();
+
+                if (!b2bContext && receiptId === undefined) {
+                    return;
+                }
+
+                return { ...b2bContext, receiptId };
+            }),
     );
 
     const isPaymentDataRequired = createSelector(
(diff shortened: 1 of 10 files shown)


bc-peng:
cursor review

packages/core/src/order/order-reducer.ts
```diff
@@ -51,12 +51,16 @@ function metaReducer(
...
+            return objectSet(
+                objectMerge(meta, {
+                    ...action.meta,
+                    callbackUrl: action.payload && action.payload.order.callbackUrl,
+                    orderToken: action.payload && action.payload.order.token,
+                    payment: action.payload && action.payload.order && action.payload.order.payment,
+                }),
+                'b2bContext',
+                action.payload?.b2bContext,
+            );
```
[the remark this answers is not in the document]
bc-peng: Here we intentionally avoid merging `b2bContext` in order to always use the returned values from the endpoint.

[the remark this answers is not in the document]
bc-peng: Will check with the team about the return of finalizing order endpoint, `/internalapi/v1/checkout/order/${orderId}`.

packages/core/src/checkout/checkout-store-selector.ts
```diff
@@ -521,9 +521,20 @@ export function createCheckoutStoreSelectorFactory(): CheckoutStoreSelectorFacto
...
+        (getB2BContext, getReceiptId) =>
+            clone(() => {
+                const b2bContext = getB2BContext();
+                const receiptId = getReceiptId();
+
+                if (!b2bContext && receiptId === undefined) {
+                    return;
+                }
+
+                return { ...b2bContext, receiptId };
```
bc-maxy, line 536: Should we fall back to b2bContext.receiptId in case receiptId from getReceiptId() is undefined?

packages/core/src/order/b2b-context.ts
```diff
@@ -0,0 +1,5 @@
+export default interface B2BContext {
```
bc-maxy:
```suggestion
export interface B2BContext {
```

packages/core/src/checkout/checkout-store-selector.ts
bc-peng: The return is defined in this way.

```
B2BContext {
    billingAddressId?: number;
    shippingAddressId?: number;
    receiptId?: string;
}
```

So the consumer side always checks in the following way.
```
        const b2bContext = checkoutService.getState().data.getB2bContext();

        if (invoiceRedirect && b2bContext?.receiptId) {
```

Maybe I didn't quite get your idea, I feel `b2bContext.receiptId` doesn't change the way of the usage above?

packages/core/src/order/b2b-context.ts
bc-peng: I prefer your suggested style. Here is just following the existing pattern.

Updated. [03ba8d9](https://github.com/bigcommerce/checkout-sdk-js/pull/3308/commits/03ba8d98b1f216306a085bf771a6f723841f135a)

richapatel1510 (approved):

packages/core/src/checkout/checkout-store-selector.ts
bc-maxy: I was just curious if we need to do something like this:

return { ...b2bContext, receiptId: receiptId ?? b2bContext?.receiptId };

And fall back in some situations where b2bContext is defined, but maybe unnecessary. Just curious, not a blocker 😄

PR #3315, opened 2026-07-15
chore: Upgrade to TypeScript 6.0.3 and API Extractor 7.x

bc-0dp:
## What/Why?

Upgrade the SDK from TypeScript 5.8 to 6.0.3 (via a 5.9 checkpoint), ahead of the eventual TypeScript 7 (native) move. TS 6.0 is a transitional release; its breaking changes are handled here. Also migrates the declaration rollup from `@microsoft/api-extractor` 6.3.0 (bundled TS 3.1) to 7.58.9.

Key changes:

- `tsconfig.base.json`: `"ignoreDeprecations": "6.0"` (silences `baseUrl`, `moduleResolution` node10, and `target` es5 pending the TS 7 migration) and `"types": ["*"]` to restore the pre-6.0 automatic `@types` global inclusion.
- ts-loader now uses an explicit `configFile` with `onlyCompileBundledFiles`, so the webpack build type-checks only bundled files, not the whole repo (TS 6.0 defaults `strict: true` and `types: []`).
- `build-dts` passes `--rootDir ..` (TS 6.0 no longer infers rootDir for the multi-package emit) and strips api-extractor 7.x `export { }` markers, so consumers keep importing the SDK's bare-declared types, matching the long-standing published `.d.ts` shape.
- Migrated 6 api-extractor configs and the `auto-export` generator to the 7.x schema.
- `.npmrc` `legacy-peer-deps=true` works around `@bigcommerce/eslint-config`'s stale optional `typescript` peer (`^4 || ^5`); added `@bigcommerce/eslint-plugin` as a direct devDep so ESLint resolves it.

Reviewer note: the `docs/` diff includes api-extractor 7.x numeric disambiguation renames (e.g. `Omit.md` becomes `Omit_2.md`, `PaymentProviderCustomerType` becomes `PaymentProviderCustomer_2`). Content is unchanged; only the internal collision labels differ.

## Rollout/Rollback

Tooling and build-config change only. No runtime feature flags, experiments, or migrations. The published JS bundles are behavior-equivalent (Babel still targets the existing browserslist); only the `.d.ts` structure and build config change. Rollback is a straight revert of this branch.

## Testing

- SDK, all green under TS 6.0.3: `lint` (52 projects), `test` (core 2478 tests plus 51 other projects), `build`, `bundle-dts`, `docs` (587 files), `build-cdn`.
- Consumer (checkout-js): `typecheck` output is byte-identical to the published-SDK baseline (214 pre-existing errors on its `master`, 0 new). `TS2459` "declared locally but not exported" count is 0, confirming the `export { }` strip. `build` passes; the test suite reports 1324 pass with 2 pre-existing failures unrelated to the SDK.
- Sandbox UAT: full checkout flow works end to end (cart, payment, order placement, order confirmation) against the TS6 build.

---

> [!NOTE]
> **Medium Risk**
> Changes affect the published `.d.ts` pipeline and monorepo typecheck scope; consumer typecheck was validated against checkout-js but any downstream relying on exact declaration formatting could still be sensitive.
> 
> **Overview**
> Upgrades the monorepo from TypeScript 5.8 to **6.0.3** (with a 5.9 checkpoint) and **@microsoft/api-extractor** from 6.x to **7.58.9**, as prep for a future TypeScript 7 move. Runtime JS is unchanged (Babel/browserslist); impact is build tooling and published **`.d.ts`** shape.
> 
> **TypeScript 6** defaults (`strict`, empty `types`) are offset in `tsconfig.base.json` via `ignoreDeprecations: "6.0"` and `types: ["*"]`. Webpack **ts-loader** now uses an explicit `configFile` and `onlyCompileBundledFiles` so the bundle build type-checks only entry graphs, not the whole repo.
> 
> **Declaration rollup** passes `--rootDir ..` on `tsc` (TS 6 no longer infers rootDir for multi-package emit), updates six api-extractor configs plus the `auto-export` generator to the 7.x schema, and strips api-extractor 7’s `export { }` markers so consumers keep the same bare-declared import surface. **hosted-form-v2** follows the same `build-dts` pattern.
> 
> **npm/ESLint**: `legacy-peer-deps` and a direct `@bigcommerce/eslint-plugin` devDep address stale TypeScript peer ranges on `@bigcommerce/eslint-config`. Generated **docs** only rename collision suffixes (e.g. `Omit_2.md`); content is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb1229f743843b5a4155a2cf682997a72c266095. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/hosted-form-v2/project.json
```diff
@@ -26,8 +26,9 @@
                 "cwd": "packages/hosted-form-v2",
                 "parallel": false,
                 "commands": [
-                    "tsc --outDir ../../temp --declaration --emitDeclarationOnly",
+                    "tsc --outDir ../../temp --rootDir .. --declaration --emitDeclarationOnly",
                     "api-extractor run --config api-extractor/hosted-form-v2-iframe-content.json & api-extractor run --config api-extractor/hosted-form-v2-iframe-host.json",
+                    "find ../../dist -maxdepth 1 -name 'hosted-form-v2-iframe-*.d.ts' -type f -exec perl -i -ne 'print unless /^export \\{.*\\}\\s*$/' {} +",
```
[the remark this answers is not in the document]
animesh1987: @bc-0dp can you please check if this is relevant.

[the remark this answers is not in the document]
bc-0dp: Confirmed relevant. The iframe-content extractor was backgrounded with no trailing `wait`, so the strip step and `rm -rf ../../temp` could run before it finished. Fixed in 1aedde434 by adding `& wait` to match the core `build-dts` target. Verified: build passes and both `hosted-form-v2-iframe-*.d.ts` are produced with `export {}` stripped.

packages/core/api-extractor/checkout-sdk.json
```diff
@@ -1,22 +1,64 @@
 {
+    "$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
+    "mainEntryPointFilePath": "<projectFolder>/../../temp/core/src/bundles/checkout-sdk.d.ts",
+    "bundledPackages": [],
     "compiler": {
-        "configType": "tsconfig",
-        "rootFolder": "."
+        "overrideTsconfig": {
+            "compilerOptions": {
+                "target": "es5",
```
bc-maxy: Any reason we use es5 and not es6 in all overrides?

package.json
```diff
@@ -107,10 +107,11 @@
     "@babel/core": "^7.6.2",
     "@babel/preset-env": "^7.6.2",
     "@bigcommerce/eslint-config": "^2.13.1",
+    "@bigcommerce/eslint-plugin": "^1.4.1",
```
bc-maxy: any reason for adding this dependency?

bc-maxy (commented):
Looks good to me overall, thanks for doing it, just curious about the es5 comment

packages/core/api-extractor/checkout-sdk.json
bc-0dp: @bc-maxy Not at all, I was just focused on the other parts. I've switched to es6. Thanks for flagging

package.json
bc-0dp: `@bigcommerce/eslint-plugin` is already a dependency of `@bigcommerce/eslint-config` and is required by its `plugin:@bigcommerce/recommended` config. I promoted it to a direct devDependency because the `.npmrc` `legacy-peer-deps=true` (needed since eslint-config's optional `typescript` peer is `^4 || ^5` and rejects TS 6) stops npm from hoisting it to the top level, which breaks ESLint's plugin resolution.

I did it this way so we don't have to update `bigcommerce/eslint-config` ahead of this feature. Once we're happy with this release, we can widen eslint-config's `typescript` peer to include `^6` and then drop both the `.npmrc` line and this devDependency.

Hope that makes sense

bc-maxy: Thanks, that makes sense!

bc-maxy (approved):
Looks good from my side

.npmrc
```diff
@@ -0,0 +1,5 @@
+# @bigcommerce/eslint-config@2.13.1 declares a stale optional peer
+# `typescript@"^4.0.0 || ^5.0.0"` that excludes TypeScript 6.x, even though the
+# underlying @typescript-eslint 8.x toolchain it wraps already supports TS 6.
+# Ignore optional peer conflicts until eslint-config publishes a TS 6 peer range.
+legacy-peer-deps=true
```
richapatel1510: Can we instead add this to overrides in package.json and avoid ignoring peer dependencies for all the packages?
`"overrides": { "@bigcommerce/eslint-config": { "typescript": "$typescript" } }`

packages/core/project.json
```diff
@@ -91,9 +91,10 @@
                 "cwd": "packages/core",
                 "parallel": false,
                 "commands": [
-                    "tsc --outDir ../../temp --declaration --emitDeclarationOnly",
+                    "tsc --outDir ../../temp --rootDir .. --declaration --emitDeclarationOnly",
                     "api-extractor run --config api-extractor/checkout-sdk.json & api-extractor run --config api-extractor/checkout-button.json & api-extractor run --config api-extractor/embedded-checkout.json & api-extractor run --config api-extractor/internal-mappers.json & wait",
                     "find src/generated/integrations -name 'api-extractor.json' | xargs -I {} -P 8 sh -c 'cd \"$(dirname \"{}\")\" && npx api-extractor run --config api-extractor.json'",
+                    "find ../../dist/types -name '*.d.ts' -type f -exec perl -i -ne 'print unless /^export \\{.*\\}\\s*$/' {} +",
```
richapatel1510, line 97: I ran `build-dts` on this branch to check the new output — the regex here catches more than the trailing `export { }`. API Extractor 7 also emits real exports in this form (no semicolon), and the raw output has `export { createTimeout }` in both `checkout-sdk.d.ts` and `checkout-button.d.ts`. The perl step deletes those too, so createTimeout disappears from the published typings. Funnily enough, v7 was actually fixing this — the current typings have the import but no export, and this strip reverts the fix.

Anchoring it to the empty marker fixes it:


`perl -i -ne 'print unless /^export \{\s*\}\s*$/'`

Same change needed in hosted-form-v2's `project.json`.

packages/hosted-form-v2/project.json
```diff
@@ -26,8 +26,9 @@
                 "cwd": "packages/hosted-form-v2",
                 "parallel": false,
                 "commands": [
-                    "tsc --outDir ../../temp --declaration --emitDeclarationOnly",
-                    "api-extractor run --config api-extractor/hosted-form-v2-iframe-content.json & api-extractor run --config api-extractor/hosted-form-v2-iframe-host.json",
+                    "tsc --outDir ../../temp --rootDir .. --declaration --emitDeclarationOnly",
+                    "api-extractor run --config api-extractor/hosted-form-v2-iframe-content.json & api-extractor run --config api-extractor/hosted-form-v2-iframe-host.json & wait",
```
richapatel1510: `A & B & wait` always exits 0 (`wait` with no args ignores the jobs' statuses), so if either extractor run fails, the build still goes green, perl rewrites the stale committed `.d.ts`, and `rm -rf` temp cleans up the evidence. Backgrounding both is good (it fixes the race with the `rm`), just capture the exit codes:

```
api-extractor run --config ...content.json & pid1=$!
api-extractor run --config ...host.json & pid2=$!
wait $pid1 && wait $pid2

```

richapatel1510 (commented):
Since dist/ typings only get regenerated at release, I built this branch locally to see the v7 output before it ships. Everything's green, and the diff vs. committed typings is ~1,700 lines of harmless alias-renaming churn — the upgrade itself looks good. The only real issues are the two inline comments (perl strip deleting real exports, and & wait swallowing extractor failures). With the anchored regex, the output comes out correct.

packages/core/project.json
bc-0dp: @richapatel1510 I think I need a bit of guidance on this one.

When I try your suggestion `perl -i -ne 'print unless /^export \{\s*\}\s*$/'`, only 20 typedoc files are output vs 590 - any `export { … }` block (even `export { createTimeout }`) flips the entry file into explicit-export mode, so typedoc stops documenting the bare declares.

If I apply the narrow strip to the types only (a hybrid), `createTimeout()` gets exported and the docs stay at 590. Consumer imports are unaffected either way.

If I keep the broad strip as-is, `createTimeout()` isn't exported - but am I wrong in saying it's also not exported on main today?

So I'm leaning toward the hybrid. Does that work for you, or would you rather handle it differently?

packages/hosted-form-v2/project.json
bc-0dp: Fixed! I capture each extractor's PID and wait on them individually (… & pid1=$!; … & pid2=$!; wait $pid1 && wait $pid2) so a failing run now propagates a non-zero exit. Applied the same in core/project.json

.npmrc
bc-0dp: Done! Switched to the scoped override and removed .npmrc's legacy-peer-deps.

Once I dropped the direct `@bigcommerce/eslint-plugin` devDep (which only existed to work around legacy-peer-deps un-hoisting), your `@bigcommerce/eslint-config` override cascades to the nested plugin and covers both peers, so that single override is all that's needed. Happy Days!

packages/core/project.json
richapatel1510: Hey @bc-0dp ,

Dug into this properly and you're right — keeping the broad strip is correct. Any surviving `export { ... } `statement (even from my suggested regex) flips the `.d.ts` out of implicit-export mode, and I confirmed with `tsc` that the narrow version breaks imports of most public types, while your version matches the exact surface published today. So: as-is is right, withdrawing my suggestion.

On createTimeout — confirmed it's already missing on main, so nothing for this PR. (For a follow-up: `"bundledPackages": ["@bigcommerce/request-sender"]` in `checkout-sdk.json` makes API Extractor inline it and would fix it properly.)

richapatel1510 (approved):
Nice work! 👏 👏 
Thanks for doing this 🙇

serhiiFilkovskyi (approved):

PR #3317, opened 2026-07-16
fix(checkout): CHECKOUT-10222 Update InternalOrder Interface

bc-peng:
## What/Why?

Update `InternalOrder` Interface to correctly save `shippingAddressId` and `billingAddressId`.

## Rollout/Rollback

Revert.

### Manual Testing (CFF needs to be removed)
1. Getting ids from the order endpoint.
Screenshot 2026-07-16 at 14 13 02

2. Submit ids to the B2B API.
Screenshot 2026-07-16 at 14 13 23

---

> [!NOTE]
> **Medium Risk**
> Touches checkout order state and B2B address ID persistence; wrong mapping could break B2B billing/shipping selection after submit.
> 
> **Overview**
> B2B **billing** and **shipping** address IDs are now modeled on **`InternalOrder`** as optional **`b2bMetadata`** (`shippingAddressId` / `billingAddressId`, nullable) instead of a separate **`b2bContext`** field on the submit-order response payload.
> 
> On **submit/finalize success**, **`order-reducer`** derives **`meta.b2bContext`** from **`order.b2bMetadata`** via **`mapToB2BContext`** (nulls become omitted optional fields), so selectors and existing meta shape stay the same while the action payload matches the API. **`InternalOrderResponseData`** no longer declares **`b2bContext`**; tests put **`b2bMetadata`** on the nested **order**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 20b6199fde09508b8de9d41a05350ca69ec2bdab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/order/internal-order.ts
+++ b/packages/core/src/order/internal-order.ts
@@ -52,6 +52,12 @@ export default interface InternalOrder {
     isDownloadable: boolean;
     isComplete: boolean;
     callbackUrl?: string;
+    b2bMetadata?: InternalOrderB2BMetadata;
+}
+
+export interface InternalOrderB2BMetadata {
+    shippingAddressId: number | null;
+    billingAddressId: number | null;
 }
 
 export interface InternalIncompleteOrder {
--- a/packages/core/src/order/index.ts
+++ b/packages/core/src/order/index.ts
@@ -6,6 +6,7 @@ export { B2BContext } from './b2b-context';
 export {
     default as InternalOrder,
     InternalIncompleteOrder,
+    InternalOrderB2BMetadata,
     InternalOrderPayment,
 } from './internal-order';
 export { default as InternalOrderRequestBody } from './internal-order-request-body';
--- a/packages/core/src/order/internal-order-responses.ts
+++ b/packages/core/src/order/internal-order-responses.ts
@@ -1,7 +1,6 @@
 import { InternalResponseBody } from '../common/http-request';
 import { InternalCustomer } from '../customer';
 
-import { B2BContext } from './b2b-context';
 import InternalOrder from './internal-order';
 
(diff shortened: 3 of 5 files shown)


packages/core/src/order/internal-order.ts
```diff
@@ -52,6 +52,12 @@ export default interface InternalOrder {
     isDownloadable: boolean;
     isComplete: boolean;
     callbackUrl?: string;
+    b2bMetadata?: InternalOrderB2BMetadata;
```
bc-peng, line 55: Checked both source code and actual prod API return, confirmed this is the API interface implemented.

bc-peng: In my manual testing, I was unable to get Ids from API even using an existing address.

Further debugging with B2B support may be needed. For this PR, we have confirmed the interface change is accurate.

Image

bc-maxy (approved):

packages/core/src/order/internal-order.ts
valentindellangela: It's working fine for me 
image

did you use CFF as well?

bc-peng: Yes, I did use CFF for the testing.
Is CFF cancelling the address match?

bc-peng: Tested again without CFF. Things work.

Updated screenshots in the PR description.

Thanks @valentindellangela for checking.

richapatel1510: Is that not expected to work when CFF are setup?

PR #3323, opened 2026-07-22
feat(checkout): SHIPPING-4147 fallback to full country list if shippa…

bc-vincent-zhao:
…ble country endpoint fails with 500

## What/Why?

2 weeks ago there was an incident where `/internalapi/v1/shipping/countries` started to return 500 due to geo data issue, which resulted in storefront checkout being completed blocked for the impacted stores, given no countries are returned and can be selected during shipping step.

During incident retro, team proposed an idea that it's probably better to have a degraded experience than entirely blocking checkout, if we fallback to the default full country list when shippable countries endpoint fail. 

As an example: if a store configured "AU and NZ" as 2 shippable countries with shipping provider configured for them, and when `/internalapi/v1/shipping/countries` fails and we fallback to return all country list 
- for shoppers in AU and NZ, they can still select AU and NZ and continue with checkout (just a longer list to choose from -- degraded experience)
- for shoppers in other countries, they can select their country but end up with a "no shipping available for that region" message, so not worse than not seeing their country in the first place.

So this PR is to introduce that fallback call to `/stores/countries` when `/internalapi/v1/shipping/countries` fails. 

NOTE: this is just POC at this stage to gather feedback. 

## Rollout/Rollback
- rollback checkout-sdk version
- revert PR

## Testing
Manual testing

Step 1: setup custom checkout with checkout-sdk and checkout.js running locally and point my vm store to it
Screenshot 2026-07-24 at 3 56 57 pm

Step 2: check page when `/shipping/countries` still returns normal 200
Screenshot 2026-07-24 at 3 57 41 pm

Step 3: throw 500 for the endpoint locally and see checkout falls back to full countries list 

```
➜  bigcommerce git:(master) ✗ git diff
diff --git a/lib/Store/InternalApi/V1/Shipping/Geography/CountriesController.php b/lib/Store/InternalApi/V1/Shipping/Geography/CountriesController.php
index e4ae18f1809..d942f293fc6 100644
--- a/lib/Store/InternalApi/V1/Shipping/Geography/CountriesController.php
+++ b/lib/Store/InternalApi/V1/Shipping/Geography/CountriesController.php
@@ -45,6 +45,7 @@ class CountriesController extends ResourceController
      */
     public function indexAction(): array
     {
+        throw InternalApiException::create500("something goes wrong");
``` 
Screenshot 2026-07-24 at 4 02 25 pm

---

> [!NOTE]
> **Medium Risk**
> Changes checkout shipping country loading on server errors and may show countries that are not actually shippable until shipping options fail later; scope is limited to the shipping countries load path with tests.
> 
> **Overview**
> When loading **shipping countries** fails with a **5xx** from the shippable-countries endpoint, checkout now **falls back** to the full **store countries** list (`CountryRequestSender`) instead of failing with no countries—so shoppers can still pick a country during shipping (degraded UX vs. a blocked checkout).
> 
> `ShippingCountryActionCreator` takes a new `CountryRequestSender` dependency; **non-5xx** errors and **timeouts** still fail without calling the fallback. If the fallback also fails, the **original** shipping-countries error is surfaced. Service wiring in `create-checkout-service`, payment integration, and tests cover success, 5xx fallback, no fallback on 4xx/timeout, and fallback failure.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3ef9424e469b05a30c0f82bc29271eea1ccb6e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/checkout/checkout-service.spec.ts
+++ b/packages/core/src/checkout/checkout-service.spec.ts
@@ -474,7 +474,11 @@ describe('CheckoutService', () => {
             paymentMethodActionCreator,
             paymentStrategyActionCreator,
             new PickupOptionActionCreator(new PickupOptionRequestSender(requestSender)),
-            new ShippingCountryActionCreator(shippingCountryRequestSender, store),
+            new ShippingCountryActionCreator(
+                shippingCountryRequestSender,
+                store,
+                countryRequestSender,
+            ),
             shippingStrategyActionCreator,
             signInEmailActionCreator,
             spamProtectionActionCreator,
--- a/packages/core/src/shipping/shipping-country-action-creator.spec.ts
+++ b/packages/core/src/shipping/shipping-country-action-creator.spec.ts
@@ -8,7 +8,8 @@ import { CheckoutStore } from '../checkout';
 import { getCheckoutStoreState } from '../checkout/checkouts.mock';
 import createCheckoutStore from '../checkout/create-checkout-store';
 import { getErrorResponse, getResponse } from '../common/http-request/responses.mock';
-import { getCountries } from '../geography/countries.mock';
+import { CountryRequestSender } from '../geography';
+import { getAustralia, getCountries } from '../geography/countries.mock';
 
 import ShippingCountryActionCreator from './shipping-country-action-creator';
 import { ShippingCountryActionType } from './shipping-country-actions';
@@ -16,21 +17,32 @@ import ShippingCountryRequestSender from './shipping-country-request-sender';
 
 describe('ShippingCountryActionCreator', () => {
     let requestSender: ShippingCountryRequestSender;
+    let countryRequestSender: CountryRequestSender;
     let shippingCountryActionCreator: ShippingCountryActionCreator;
     let errorResponse: Response<ErrorResponseBody>;
     let response: Response<any>;
+    let fallbackResponse: Response<any>;
     let store: CheckoutStore;
 
     beforeEach(() => {
         store = createCheckoutStore(getCheckoutStoreState());
         response = getResponse({ data: getCountries() });
+        fallbackResponse = getResponse({ data: [getAustralia()] });
         errorResponse = getErrorResponse();
 
         requestSender = new ShippingCountryRequestSender(createRequestSender(), {});
+        countryRequestSender = new CountryRequestSender(createRequestSender(), {});
 
         jest.spyOn(requestSender, 'loadCountries').mockReturnValue(Promise.resolve(response));
+        jest.spyOn(countryRequestSender, 'loadCountries').mockReturnValue(
+            Promise.resolve(fallbackResponse),
+        );
 
-        shippingCountryActionCreator = new ShippingCountryActionCreator(requestSender, store);
+        shippingCountryActionCreator = new ShippingCountryActionCreator(
+            requestSender,
+            store,
+            countryRequestSender,
+        );
     });
 
     describe('#loadCountries()', () => {
@@ -49,7 +61,7 @@ describe('ShippingCountryActionCreator', () => {
                 });
         });
 
-        it('emits error actions if unable to load countries', () => {
+        it('emits error actions if unable to load countries with a non-5xx error', () => {
             jest.spyOn(requestSender, 'loadCountries').mockReturnValue(
                 Promise.reject(errorResponse),
             );
@@ -61,6 +73,7 @@ describe('ShippingCountryActionCreator', () => {
                 .pipe(catchError(errorHandler), toArray())
                 .subscribe((actions) => {
                     expect(errorHandler).toHaveBeenCalled();
+                    expect(countryRequestSender.loadCountries).not.toHaveBeenCalled();
                     expect(actions).toEqual([
                         { type: ShippingCountryActionType.LoadShippingCountriesRequested },
                         {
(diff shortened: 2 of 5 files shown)


packages/core/src/shipping/shipping-country-action-creator.spec.ts
```diff
@@ -72,12 +85,96 @@ describe('ShippingCountryActionCreator', () => {
                 });
         });
 
+        it('does not fall back to /store/countries on a timeout response', () => {
```
theromulans: 🍻 We wouldn't want to fallback in a timeout scenario?

One thing that is notable about the `shipping/countries` endpoint is the higher level of complexity required to answer the request - I could certainly see this endpoint as more likely to experience a timeout, simply given more the moving parts (and opportunity for failure) involved.

packages/core/src/checkout/checkout-service.spec.ts
```diff
@@ -470,7 +470,11 @@ describe('CheckoutService', () => {
             paymentMethodActionCreator,
             paymentStrategyActionCreator,
             new PickupOptionActionCreator(new PickupOptionRequestSender(requestSender)),
-            new ShippingCountryActionCreator(shippingCountryRequestSender, store),
+            new ShippingCountryActionCreator(
+                shippingCountryRequestSender,
+                store,
+                countryRequestSender,
```
theromulans, line 480: 👁️ Curious: does this change imply two separate calls to the same `store/countries` endpoint (one at the shipping step, one at the billing step), when we're dealing with a fallback scenario?

Probably not a big concern if it does, given this only speaks to a very limited scenario, more curious on the end results.

packages/core/src/shipping/shipping-country-action-creator.spec.ts
bc-vincent-zhao: great question, I limited it to 500 error first to limit scope given this change is for incident retro action and I'm not sure if checkout has any retry mechanism for timeout errors. If checkout doesn't have retry on timeout and is happy to do fallback on timeout, can definitely add that as well.

packages/core/src/checkout/checkout-service.spec.ts
bc-vincent-zhao: good question, I'm not sure tbh, I will ask Claude to find out

animesh1987: We already make call to both billing and shipping as part of shipping step loading on client side. So maybe can check if the data is there already or not?

animesh1987: https://github.com/bigcommerce/checkout-js/blob/master/packages/core/src/app/shipping/Shipping.tsx#L88

bc-vincent-zhao: from Claude analysis

```
  How it works: ShippingCountryActionCreator and CountryActionCreator both operate on the same CheckoutStore instance (confirmed — create-checkout-service.ts creates one store and passes it into both). Billing countries live in state.countries (CountryState { data?:
  Country[]; isLoading; loadError }), separate from state.shippingCountries. So if the billing step already called loadBillingCountries() successfully before the shipping-countries outage hits, this._store.getState().countries.getCountries() would already return that
  Country[] — no need to fire a second network request in the fallback.

  Caveats:
  - getCountries() returns undefined if billing countries were never loaded (e.g. shopper hits shipping step first, or a first-party integration never calls billing countries at all) — the fallback would still need to fall through to a real
  _countryRequestSender.loadCountries() call in that case.
  - No staleness/TTL concept exists in this slice — once loaded, it's cached for the session with no expiry. That's actually fine for country lists (effectively static), so reusing it isn't introducing a new staleness risk.
  - No race condition risk in practice: checking data !== undefined naturally skips a same-tick in-flight load (data stays undefined until success).

  Trade-off to weigh: this turns a one-line "always call the fallback sender" into "check store cache, fall through to network only if empty" — a small but real complexity bump (reading a second selector, reasoning about the undefined case) to save a request that, per our
  last exchange, only happens during an already-alerted-on incident window. Given you were already comfortable with the redundant call as acceptable, I'd lean toward not adding this unless you want the extra request eliminated on principle. Want me to implement the
  store-reuse check, or leave it as-is?
  ```
  
can billing / shipping steps change ordering ? if billing always happens before shipping then it might be a quick change, but if there's chance shipping can happen before billing, I'm more inclined to keep the current code to fire the request again instead of trying to share the data between those 2. 

@animesh1987 what's your view on this  ?

File: packages/core/src/shipping/shipping-country-action-creator.spec.ts (hunk 1 above)
Comment (theromulans) [reply packages/core/src/shipping/shipping-country-action-creator.spec.ts]: Great, I'd be leaning this way 👍 I could see this avoiding an incident in the future.

File: packages/core/src/checkout/checkout-service.spec.ts (hunk 2 above)
Comment (bc-vincent-zhao) [reply packages/core/src/checkout/checkout-service.spec.ts:480]: from my local testing I see we do shipping step first then billing

File: packages/core/src/shipping/shipping-country-action-creator.spec.ts (hunk 1 above)
Comment (bc-vincent-zhao) [reply packages/core/src/shipping/shipping-country-action-creator.spec.ts]: asking claude to explore a bit and it came back with 

```
⏺ No retry mechanism exists anywhere — neither in @bigcommerce/request-sender itself (its Timeout class just calls request.abort() once
  and rejects; no retry field in RequestOptions) nor anywhere in this SDK (grepped for retry/backoff patterns — the only ones found
  are unrelated: reCAPTCHA polling, embedded-checkout cookie retry, hosted-iframe reload). A timeout on /shipping/countries rejects
  immediately with status === 0, straight to .catch().
  
    One caveat: nothing in this SDK sets a default timeout for this call — RequestOptions.timeout is entirely caller-supplied. So a
  timeout only happens at all if the checkout-js frontend (separate repo) explicitly passes a Timeout when calling
  loadShippingCountries(). Worth confirming with the team whether that's actually configured in the storefront today — otherwise
  "timeout" is a theoretical gap rather than an active one.
```

I will defer whether we should cater for timeout in this case to checkout SME.

**Comment by bc-peng** [comment]
LGTM. Shall we merge the PR?

File: packages/core/src/shipping/shipping-country-action-creator.spec.ts [hunk 3]
```diff
@@ -72,12 +85,96 @@ describe('ShippingCountryActionCreator', () => {
...
+                    .subscribe((actions) => {
+                        expect(actions).toEqual([
+                            { type: ShippingCountryActionType.LoadShippingCountriesRequested },
+                            {
+                                type: ShippingCountryActionType.LoadShippingCountriesSucceeded,
+                                payload: fallbackResponse.body.data,
+                            },
+                        ]);
+                    });
+            });
```
bc-maxy: I don't think the assertions inside of subscribe(callback) actually gate any tests?
Awaiting the observable would guarantee it but maybe I'm missing something (not too familiar with RxJS design yet 😄 ):
```
const actions = await shippingCountryActionCreator
        .loadCountries()
        .pipe(toArray())
        .toPromise();

    expect(actions).toEqual([...]);
```
What do you think?

bc-maxy (commented):
I have a comment about tests but the rest looks good to me

packages/core/src/shipping/shipping-country-action-creator.spec.ts
bc-vincent-zhao: great observation @bc-maxy, I completely missed that. 

I've updated the whole test file to use this pattern and Claude double confirms it does fails when asserting deliberate wrong response 

```
⏺ Update(packages/core/src/shipping/shipping-country-action-creator.spec.ts)
  ⎿  Added 1 line, removed 7 lines
      124                      .pipe(toArray())
      125                      .toPromise();
      126
      127 -                expect(actions).toEqual([
      128 -                    { type: ShippingCountryActionType.LoadShippingCountriesRequested },
      129 -                    {
      130 -                        type: ShippingCountryActionType.LoadShippingCountriesSucceeded,
      131 -                        payload: fallbackResponse.body.data,
      132 -                    },
      133 -                ]);
      127 +                expect(actions).toEqual([{ type: 'THIS_IS_DELIBERATELY_WRONG' }]);
      128              });
      129
      130              it('calls the fallback without a channelId, passing through the given options', async () => {



         - Expected  -  1
         + Received  + 22

           Array [
             Object {
         -     "type": "THIS_IS_DELIBERATELY_WRONG",
         +     "type": "LOAD_SHIPPING_COUNTRIES_REQUESTED",
         +     "payload": Array [
         +       Object {
         +         "code": "AU",
         +         "hasPostalCodes": true,
         +         "name": "Australia",
         +         "requiresState": true,
         +         "subdivisions": Array [
         +           Object {
         +             "code": "NSW",
         +             "name": "New South Wales",
         +           },
         +           Object {
         +             "code": "VIC",
         +             "name": "Victoria",
         +           },
         +         ],
         +       },
         +     ],
         +     "type": "LOAD_SHIPPING_COUNTRIES_SUCCEEDED",
             },
           ]

           125 |                     .toPromise();
           126 |
         > 127 |                 expect(actions).toEqual([{ type: 'THIS_IS_DELIBERATELY_WRONG' }]);
               |                                 ^
           128 |             });
           129 |
           130 |             it('calls the fallback without a channelId, passing through the given options', async () => {

           at src/shipping/shipping-country-action-creator.spec.ts:127:33
           at fulfilled (src/shipping/shipping-country-action-creator.spec.ts:5:58)
     Test Suites: 1 failed, 1 total
     Tests:       1 failed, 6 skipped, 7 total
     Snapshots:   0 total
     Time:        6.898 s
 ```

packages/core/src/checkout/checkout-service.spec.ts
bc-vincent-zhao: I think we're ok to not share the data between those 2 steps for now, given the benefit of it probably doesn't worth the complexity for it.

packages/core/src/shipping/shipping-country-action-creator.spec.ts
bc-maxy: Thank you for verifying and fixing it!

bc-maxy (approved):
Looks good from my side ✅

PR #3326, opened 2026-07-24
fix(checkout): CHECKOUT-10262 Consistently Send shouldSaveAddress Value

bc-peng:
## What/Why?

The checkout API currently defaults `shouldSaveAddress` to `true`.
We decided to handle this on the FE by consistently sending `false` as the default value.

## Rollout/Rollback

Revert.

## Testing

CI.

---

> [!NOTE]
> **Medium Risk**
> Changes checkout address save behavior for any caller that relied on omitting the field and inheriting API default true; explicit opt-in to save still works.
> 
> **Overview**
> Checkout address payloads sent through **`mapToAddressRequestBody`** now **always include `shouldSaveAddress`**, defaulting to **`false`** when the source address omits it—countering the API’s default of `true`.
> 
> **`shouldSaveAddress`** is moved from the **`Address`** type onto **`AddressRequestBody`** so it lives with the request shape. Explicit **`true`** (or any set value) is still passed through unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c5c1c356ea8a6103fa393ef68a2b4115ba39c7e5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/address/map-to-address-request-body.ts
+++ b/packages/core/src/address/map-to-address-request-body.ts
@@ -25,12 +25,19 @@ export default function mapToAddressRequestBody(
 ): Partial<AddressRequestBody> {
     const { b2b, ...requestBody } = address;
 
+    // The API currently defaults shouldSaveAddress to true. We decided to handle this on the FE
+    // by consistently sending false as the default on 24 July 2026.
+    const requestBodyWithSaveFlag = {
+        ...requestBody,
+        shouldSaveAddress: requestBody.shouldSaveAddress ?? false,
+    };
+
     if (!b2b) {
-        return requestBody;
+        return requestBodyWithSaveFlag;
     }
 
     return {
-        ...requestBody,
-        extraFields: requestBody.extraFields ?? b2b.extraFields,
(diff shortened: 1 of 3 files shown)


packages/core/src/address/map-to-address-request-body.ts
```diff
@@ -25,12 +25,19 @@ export default function mapToAddressRequestBody(
 ): Partial<AddressRequestBody> {
     const { b2b, ...requestBody } = address;
 
+    // The API currently defaults shouldSaveAddress to true. We decided to handle this on the FE
+    // by consistently sending false as the default on 24 July 2026.
```
richapatel1510, line 29: Do we want to keep this comment here? OR move to decisions log?

bc-peng: Thought it would be easier to read the comment here in the future as this also has impact on B2C flows.
I'll add ADR anyway now.

richapatel1510 (approved):

PR #3349, opened 2026-08-10
feat(payment): PAYPAL-6856 added error logger for BCP FL

andriiVitvitskyi1990:
## What/Why?

Added error logger for BCP Fastlane

## Rollout/Rollback
Revert

## Testing
Unit tests

---

> [!NOTE]
> **Low Risk**
> Changes are limited to optional error logging on already soft-failing Fastlane paths; payment and checkout behavior remain non-blocking.
> 
> **Overview**
> Adds an optional **`onErrorLog`** callback so integrators can record BigCommerce Payments Fastlane failures without changing the existing non-blocking checkout behavior.
> 
> On the **customer** strategy, Fastlane SDK initialization errors and guest authentication failures during `executePaymentMethodCheckout` now invoke `handleErrorLog` instead of silently swallowing errors (replacing the prior TODO). The **payment** strategy does the same when the PayPal authentication flow fails during initialize. Checkout still continues—`continueWithCheckoutCallback` and initialization still resolve—when these paths fail.
> 
> `onErrorLog` is documented on **`BigCommercePaymentsFastlanePaymentInitializeOptions`** and is only called when it is a function; invalid or missing callbacks are ignored. Unit specs cover success paths, forwarding, and soft-failure behavior for both strategies.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3e0e664f43a3702939c6d3bb7cd43a0f00757877. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/bigcommerce-payments-integration/src/bigcommerce-payments-fastlane/bigcommerce-payments-fastlane-customer-strategy.ts
+++ b/packages/bigcommerce-payments-integration/src/bigcommerce-payments-fastlane/bigcommerce-payments-fastlane-customer-strategy.ts
@@ -22,6 +22,9 @@ import BigCommercePaymentsFastlaneCustomerInitializeOptions, {
 } from './bigcommerce-payments-fastlane-customer-initialize-options';
 
 export default class BigCommercePaymentsFastlaneCustomerStrategy implements CustomerStrategy {
+    private options?: CustomerInitializeOptions &
+        WithBigCommercePaymentsFastlaneCustomerInitializeOptions;
+
     constructor(
         private paymentIntegrationService: PaymentIntegrationService,
         private bigCommercePaymentsSdk: PayPalSdkHelper,
@@ -33,6 +36,8 @@ export default class BigCommercePaymentsFastlaneCustomerStrategy implements Cust
             WithBigCommercePaymentsFastlaneCustomerInitializeOptions,
     ): Promise<void> {
         const { methodId, bigcommerce_payments_fastlane } = options;
+
+        this.options = options;
 
         if (!methodId) {
             throw new InvalidArgumentError(
@@ -61,9 +66,9 @@ export default class BigCommercePaymentsFastlaneCustomerStrategy implements Cust
                 isTestModeEnabled,
                 this.getFastlaneStyles(methodId, bigcommerce_payments_fastlane),
             );
-        } catch (_) {
-            // TODO: add logger to be able to debug issues if there any
+        } catch (error) {
             // Info: Do not throw anything here to avoid blocking customer from passing checkout flow
+            this.handleErrorLog(error);
         }
 
         return Promise.resolve();
@@ -120,9 +125,9 @@ export default class BigCommercePaymentsFastlaneCustomerStrategy implements Cust
 
             try {
                 await this.runPayPalAuthenticationFlowOrThrow(methodId);
-            } catch (_) {
-                // TODO: add logger to be able to debug issues if there any
+            } catch (error) {
                 // Info: Do not throw anything here to avoid blocking customer from passing checkout flow
+                this.handleErrorLog(error);
             }
         }
 
@@ -234,5 +239,13 @@ export default class BigCommercePaymentsFastlaneCustomerStrategy implements Cust
             isFastlaneStylingEnabled ? fastlaneStyles : {},
             bigcommerce_payments_fastlane?.styles,
         );
+    }
+
+    private handleErrorLog(error: unknown): void {
+        const { onErrorLog } = this.options || {};
+
+        if (onErrorLog && typeof onErrorLog === 'function') {
+            onErrorLog(error);
+        }
     }
 }
--- a/packages/bigcommerce-payments-integration/src/bigcommerce-payments-fastlane/bigcommerce-payments-fastlane-payment-initialize-options.ts
+++ b/packages/bigcommerce-payments-integration/src/bigcommerce-payments-fastlane/bigcommerce-payments-fastlane-payment-initialize-options.ts
@@ -74,6 +74,11 @@ export default interface BigCommercePaymentsFastlanePaymentInitializeOptions {
      * no matter what strategy was initialised first
      */
     styles?: PayPalFastlaneStylesOption;
+
+    /**
+     * Method that will only log errors with no-blocking flow
+     */
+    onErrorLog?: (error: unknown) => void;
 }
 
 export interface WithBigCommercePaymentsFastlanePaymentInitializeOptions {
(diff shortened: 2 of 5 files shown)


packages/bigcommerce-payments-integration/src/bigcommerce-payments-fastlane/bigcommerce-payments-fastlane-customer-strategy.ts
```diff
@@ -235,4 +240,12 @@ export default class BigCommercePaymentsFastlaneCustomerStrategy implements Cust
             bigcommerce_payments_fastlane?.styles,
         );
     }
+
+    private handleErrorLog(error: unknown): void {
+        const { onErrorLog } = this.options || {};
+
+        if (onErrorLog && typeof onErrorLog === 'function') {
+            onErrorLog(error);
```
bc-nick, line 248: The option better to call just `onError`

packages/bigcommerce-payments-integration/src/bigcommerce-payments-fastlane/bigcommerce-payments-fastlane-payment-initialize-options.ts
```diff
@@ -74,6 +74,11 @@ export default interface BigCommercePaymentsFastlanePaymentInitializeOptions {
      * no matter what strategy was initialised first
      */
     styles?: PayPalFastlaneStylesOption;
+
+    /**
+     * Method that will only log errors with no-blocking flow
```
bc-nick, line 79: This function does not log anything. It is used to escalate an error to a higher level and than customer can decide how it should be handled by himself (Custom checkout)

packages/bigcommerce-payments-integration/src/bigcommerce-payments-fastlane/bigcommerce-payments-fastlane-payment-strategy.ts
```diff
@@ -543,4 +544,13 @@ export default class BigCommercePaymentsFastlanePaymentStrategy implements Payme
             this.bigcommerce_payments_fastlane.onError(error);
         }
     }
+
+    private handleErrorLog(error: unknown): void {
+        if (
+            this.bigcommerce_payments_fastlane?.onErrorLog &&
+            typeof this.bigcommerce_payments_fastlane.onErrorLog === 'function'
+        ) {
+            this.bigcommerce_payments_fastlane.onErrorLog(error);
```
bc-nick, line 553: the same

packages/bigcommerce-payments-integration/src/bigcommerce-payments-fastlane/bigcommerce-payments-fastlane-customer-strategy.ts
andriiVitvitskyi1990: I think we should leave it onErrorLog, because for instance we have onError callback in other payment strategies and it actually process errors and throwing them. In other side this callback logs errors and it named just like that

packages/bigcommerce-payments-integration/src/bigcommerce-payments-fastlane/bigcommerce-payments-fastlane-payment-initialize-options.ts
andriiVitvitskyi1990: It created 
```
const errorLogger = useMemo(
    () => createErrorLogger(
        { sentry: sentryConfig },
        { errorTypes: ['UnrecoverableError'], publicPath, sampleRate: sentrySampleRate || 0.1 },
    ),
    [],
);
```
So it logs to sentry on production
```
export default function createErrorLogger(serviceConfig?, options?): ErrorLogger {
    if (serviceConfig && serviceConfig.sentry) {
        return new SentryErrorLogger(serviceConfig.sentry, {
            ...options,
            consoleLogger: new ConsoleErrorLogger(options),
        });
    }
    if (process.env.NODE_ENV === 'test') return new NoopErrorLogger();
    return new ConsoleErrorLogger(options);
}
```

packages/bigcommerce-payments-integration/src/bigcommerce-payments-fastlane/bigcommerce-payments-fastlane-payment-strategy.ts
andriiVitvitskyi1990: same as above

packages/bigcommerce-payments-integration/src/bigcommerce-payments-fastlane/bigcommerce-payments-fastlane-customer-strategy.ts
bc-nick: Why do not use `onError` to through error and for log them at the same time? `onErrorLog` callback is redundant

andriiVitvitskyi1990: How it can be redundant if i pass it from checkout js. Maybe you think about different one?

andriiVitvitskyi1990: The reason we add this is that we have some flows where we can't throw error to not block payment, but we want to know if some errors occurs

bc-nick: Because for all these purposes use can use `onError` and do not create an additional callback. It means `onErrorLog` is redundant when everything can be managed with `onError`

andriiVitvitskyi1990: onError throws an error, in our case we don't want to throw

bc-nick: you meant we do not need to show popup in this case?

andriiVitvitskyi1990: Yes, do not need to show popup and stop code execution.

bc-nick: Okay, now it makes more sense for me. Thanks for explanation!

bc-nick (approved):

PR #3356, opened 2026-08-12
fix(payment): PAYPAL-6983 Attempting to use PayPal Checkout on cart page results in user being returned to cart page with no order placed

bc-pavlo-khodakivskyi:
## What/Why?

Fix for `Attempting to use PayPal Checkout on cart page results in user being returned to cart page with no order placed`
https://bigcommercecloud.atlassian.net/browse/PAYPAL-6873

## Rollout/Rollback

Revert

## Testing

https://github.com/user-attachments/assets/45a68c15-ed48-49ee-8a2f-284e57bd01de

---

> [!NOTE]
> **Medium Risk**
> Touches payment checkout timing and bot-protection gating; mis-timed calls could block legitimate PayPal clicks or skip required verification.
> 
> **Overview**
> **Fixes PayPal Checkout on the cart page** when ReCAPTCHA spam protection is required: orders were failing and shoppers were sent back to the cart because spam verification never ran on that path.
> 
> After the default checkout is loaded during PayPal Commerce button **initialize** (not Buy Now), the strategy now calls `verifyCheckoutSpamProtection` when `checkout.shouldExecuteSpamCheck` is true. **Buy Now** still skips spam check on initialize (no checkout session yet).
> 
> For **Buy Now**, the same check runs in `paypal-button-creation-service` inside `createOrder`, after `loadCheckout` for the new buy-now cart, so spam protection runs before `createOrder` proceeds.
> 
> Specs cover spam check on/off, and that Buy Now initialize does not verify spam protection even when the flag is true.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4fa29c74cb11e83100f9e8ae236b0c1dbeb770ff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/paypal-commerce-integration/src/paypal-commerce/paypal-commerce-button-strategy.ts
+++ b/packages/paypal-commerce-integration/src/paypal-commerce/paypal-commerce-button-strategy.ts
@@ -66,6 +66,11 @@ export default class PayPalCommerceButtonStrategy implements CheckoutButtonStrat
             // Info: default checkout should not be loaded for BuyNow flow,
             // since there is no checkout session available for that.
             await this.paymentIntegrationService.loadDefaultCheckout();
+
+            const checkout = this.paymentIntegrationService.getState().getCheckoutOrThrow();
+            if (checkout.shouldExecuteSpamCheck) {
+                await this.paymentIntegrationService.verifyCheckoutSpamProtection();
+            }
         }
 
         // Info: we are using provided currency code for buy now cart,
--- a/packages/paypal-commerce-integration/src/paypal-commerce/paypal-commerce-button-strategy.spec.ts
+++ b/packages/paypal-commerce-integration/src/paypal-commerce/paypal-commerce-button-strategy.spec.ts
@@ -11,6 +11,7 @@ import {
     getBuyNowCart,
     getBuyNowCartRequestBody,
     getCart,
+    getCheckout,
     getConfig,
     getConsignment,
     getShippingOption,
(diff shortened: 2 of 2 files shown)


packages/paypal-commerce-integration/src/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -68,12 +68,20 @@ export default class PayPalCommerceButtonStrategy implements CheckoutButtonStrat
...
 
+        let currencyCode: string | undefined;
+
         // Info: we are using provided currency code for buy now cart,
         // because checkout session is not available before buy now cart creation,
         // hence application will throw an error on getCartOrThrow method call
-        const currencyCode = isBuyNowFlow
-            ? paypalcommerce.currencyCode
-            : this.paymentIntegrationService.getState().getCartOrThrow().currency.code;
+        if (isBuyNowFlow) {
```
bc-nick: I think spam protection should be added for Buy Now flow as well but in different place since we do not have cart until we click on smart payment button on PD page.

You need to verify do we need to call `verifyCheckoutSpamProtection` based on `shouldExecuteSpamCheck` value when it is BuyNow flow in cart creation method. Take a look at the `paypal-button-creation-service.ts` where we are working with `createOrder` callback for button initialisation.

packages/paypal-commerce-integration/src/paypal-commerce/paypal-commerce-button-strategy.ts
```diff
@@ -75,6 +75,10 @@ export default class PayPalCommerceButtonStrategy implements CheckoutButtonStrat
             ? paypalcommerce.currencyCode
             : this.paymentIntegrationService.getState().getCartOrThrow().currency.code;
 
+        if (this.paymentIntegrationService.getState().getCheckoutOrThrow().shouldExecuteSpamCheck) {
```
bc-nick: Fo PDP we do not have checkout and the code will never be executed for Buy Now flow instead of it will cause an initialisation error due to getCheckoutOrThrow that will stop initialisation error.

bc-nick (approved):
LGTM

PR #3357, opened 2026-08-13
feat(checkout): CHECKOUT-10199 deinitialize the initialized strategy on method lost after reload

richapatel1510:
## What/Why?

- If a payment methods reload removes the method the shopper had selected (changing billing country usually), its component unmounts and calls deinitializePayment. That looked the method up in state, didn't find it, and threw MissingPaymentMethod — so you'd get an error modal for no good reason, and the strategy never actually got torn down, so its iframe stuck around.
- Fix is to hang on to the method we initialized with and use that when it's no longer in state. Same strategy instance, proper teardown.
- Kept it tight on purpose: only a method that was initialized and then vanished behaves differently. Anything unknown or never initialized throws like it always did.
- Also rewrote the "throws if method not loaded" test — it was green even when nothing threw.
- The checkout-js half of this ticket drops a temporary suppression hack and needs this released first. https://github.com/bigcommerce/checkout-js/pull/3248

## Rollout/Rollback

Revert this PR.

## Testing

- CI checks

---

> [!NOTE]
> **Medium Risk**
> Touches payment strategy initialize/deinitialize lifecycle during checkout; scoped change with tests, but incorrect cache or teardown could leave strategies running or block cleanup.
> 
> **Overview**
> Fixes teardown when **payment methods are reloaded** and a method disappears from store while its strategy is still initialized.
> 
> `PaymentStrategyActionCreator` now keeps an **`_initializedMethodsCache`** keyed by `methodId` or `methodId.gatewayId`. Entries are added only after **initialize succeeds** and removed after **deinitialize succeeds**, so unknown methods still throw `MissingDataError` and failed teardowns can be retried.
> 
> **Deinitialize** resolves the `PaymentMethod` from the store first, then falls back to the cache so `registry.getByMethod` still gets the original object. If the method was never initialized, behavior is unchanged (no actions, no `deinitialize` call).
> 
> Specs cover removed methods after init, gateway cache keys, and the unknown-method path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 58a84836222b4ff2db7aeafb30b6789ef44bd147. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/payment/payment-strategy-action-creator.spec.ts
+++ b/packages/core/src/payment/payment-strategy-action-creator.spec.ts
@@ -455,11 +455,147 @@ describe('PaymentStrategyActionCreator', () => {
         });
 
         it('throws error if payment method has not been loaded', async () => {
+            expect.assertions(2);
+
             try {
                 await from(actionCreator.deinitialize({ methodId: 'unknown' })(store)).toPromise();
             } catch (action) {
                 expect((action as Action).payload).toBeInstanceOf(MissingDataError);
             }
+
+            expect(strategy.deinitialize).not.toHaveBeenCalled();
+        });
+
+        describe('when the payment method is removed after initialization', () => {
+            let storeWithMethodRemoved: CheckoutStore;
+
+            beforeEach(async () => {
+                jest.spyOn(strategy, 'initialize').mockReturnValue(
+                    Promise.resolve(store.getState()),
+                );
+
+                // Must run against a store where the method is not already initialized,
+                // otherwise `initialize` short-circuits and nothing is cached.
+                await from(
+                    actionCreator.initialize({
+                        methodId: method.id,
+                        gatewayId: method.gateway,
+                    })(createCheckoutStore(state)),
+                ).toPromise();
+
+                // Do not `merge` the payment methods data - lodash merges arrays
+                // element-wise, which would leave the method in place.
+                storeWithMethodRemoved = createCheckoutStore({
+                    ...merge({}, state, {
+                        paymentStrategies: {
+                            data: { [method.id]: { isInitialized: true } },
+                        },
+                    }),
+                    paymentMethods: { ...state.paymentMethods, data: [] },
+                });
+            });
+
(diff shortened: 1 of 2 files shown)


richapatel1510:
cursor review

packages/core/src/payment/payment-strategy-action-creator.spec.ts
```diff
@@ -455,11 +455,147 @@ describe('PaymentStrategyActionCreator', () => {
         });
 
         it('throws error if payment method has not been loaded', async () => {
+            expect.assertions(2);
```
bc-peng, line 458: TIL.

packages/core/src/payment/payment-strategy-action-creator.ts
```diff
@@ -45,6 +45,13 @@ import { PaymentStrategy } from './strategies';
 export default class PaymentStrategyActionCreator {
     private _paymentStrategyWidgetActionCreator: PaymentStrategyWidgetActionCreator;
 
+    /**
+     * Methods with a live strategy. A payment methods reload can drop a method from
+     * state while its strategy is still running, and tearing that strategy down needs
+     * the original method object to resolve it.
+     */
+    private _initializedMethodsCache = new Map<string, PaymentMethod>();
```
bc-peng, line 53: Could you please explain when can a checkout flow have more than one initialized method?
Is there a possibility that this cache could become stale?

richapatel1510: It can have more than one initialized methods in the normal wallet-buttons case such as PayPal, Google Pay, Apple Pay, etc., hence a Map rather than a single cached method. There is no enforcement that only 1 method can be initialized.

The cache is only a fallback for teardown, and it's updated on the same actions that update `isInitialized`, so it can't drift out of sync with state.

bc-peng (approved):
Nice design!👏👏

PR #3364, opened 2026-08-19
feat(payment): PAYMENTS-11912 Use channel's SF origin for hosted field iframe src in headless SF

Tharaae:
## What/Why?
The hosted card fields  currently don't work on a headless storefront. This  change makes them work, without changing anything for existing Stencil storefronts.

When the SDK renders hosted card fields, it creates an iframe pointing at `/account/stored-instruments/hosted-fields`. That path was hardcoded with no host on it, so the browser resolved it against whatever page it was on. On a stencil storefront, that's fine as the page and that route live on the same domain, so it just works. On a headless storefront, the page is on the merchant's own domain which has no such route. So, the iframe gives 404 error and no card fields appear.

This PR adds a new optional `storefrontHost` param to be used as the origin serving the hosted-fields route only to build the iframe src. It defaults to `''` so anything not passing it keeps the exact relative URL it has today, keeping stencil SF scenario untouched.

## Rollout/Rollback
Revert

## Testing
Unit tests
Screen record
https://github.com/user-attachments/assets/633e1ebe-2321-45ae-acec-cce3fdbd7c33

---

> [!NOTE]
> **Medium Risk**
> Changes payment iframe URL construction for stored cards; wrong storefrontHost could break fields on headless while defaults preserve existing Stencil behavior.
> 
> **Overview**
> Adds an optional **`storefrontHost`** so stored-card hosted fields can load the iframe from the channel storefront origin while **`host`** still targets the payments origin for `postMessage` validation.
> 
> **`HostedField`** now prefixes the shopper-account iframe path with `storefrontHost` when set; admin Control Panel URLs stay same-origin and ignore it. When `storefrontHost` is omitted or empty, behaviour matches today (relative `/account/stored-instruments/hosted-fields` resolved against the page).
> 
> The parameter is threaded through **`createStoredCardHostedFormService`**, **`StoredCardHostedFormService`**, and **`HostedFormFactory.create`**, with docs clarifying the split between payments vs storefront origins.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb5b3761da8e26be1b02833c76674a78b2bdcd0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/hosted-form-v2/src/hosted-field.ts
+++ b/packages/hosted-form-v2/src/hosted-field.ts
@@ -47,6 +47,7 @@ export default class HostedField {
         private _eventListener: IframeEventListener<HostedInputEventMap>,
         private _detachmentObserver: DetachmentObserver,
         private _orderId?: number,
+        private _storefrontHost = '',
     ) {
         this._iframe = document.createElement('iframe');
 
@@ -60,7 +61,7 @@ export default class HostedField {
     private getFrameSrc(orderId?: number): string {
         return typeof orderId !== 'undefined'
             ? `/admin/payments/${this._orderId}/hosted-form-field?version=${LIBRARY_VERSION}`
-            : `/account/stored-instruments/hosted-fields?version=${LIBRARY_VERSION}`;
+            : `${this._storefrontHost}/account/stored-instruments/hosted-fields?version=${LIBRARY_VERSION}`;
     }
 
     getType(): HostedFieldType {
--- a/packages/hosted-form-v2/src/hosted-form-factory.ts
+++ b/packages/hosted-form-v2/src/hosted-form-factory.ts
@@ -8,7 +8,13 @@ import HostedForm from './hosted-form';
 import HostedFormOptions from './hosted-form-options';
 
 export default class HostedFormFactory {
-    create(host: string, options: HostedFormOptions): HostedForm {
+    /**
+     * @param host - (Payments origin) Used for postMessage target/source validation, because the
+     * hosted-fields route redirects there, making it the iframe's actual document origin.
+     * @param storefrontHost - (Channel's storefront origin) Serving the hosted-fields route,
+     * used only for the iframe src. Only needed by headless storefronts, where it differs from the page's own origin.
+     */
+    create(host: string, options: HostedFormOptions, storefrontHost = ''): HostedForm {
         const fieldTypes = Object.keys(options.fields) as HostedFieldType[];
         const fields = fieldTypes.reduce<HostedField[]>((result, type) => {
             const fields = options.fields;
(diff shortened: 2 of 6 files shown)


bc-alfreid (approved):
LGTM

packages/hosted-form-v2/src/hosted-form-factory.ts
```diff
@@ -8,7 +8,13 @@ import HostedForm from './hosted-form';
...
 
 export default class HostedFormFactory {
-    create(host: string, options: HostedFormOptions): HostedForm {
+    /**
+     * @param host - (Payments origin) Used for postMessage target/source validation, because the
+     * hosted-fields route redirects there, making it the iframe's actual document origin.
+     * @param storefrontHost - (Channel's storefront origin) Serving the hosted-fields route,
+     * used only for the iframe src. Only needed by headless storefronts, where it differs from the page's own origin.
+     */
+    create(host: string, options: HostedFormOptions, storefrontHost = ''): HostedForm {
```
bc-afeng, line 17: Not sure if it would be better. Would it be possible to add `storeFrontHost` to `HostedFormOptions` instead (as maybe `storefrontHost?: string`) and that way we could do `options.storefrontHost`?

Tharaae: I'd rather keep it off `HostedFormOptions`, for two reasons:

1. `storefrontHost `is fixed for the life of the page, the same way `host` (payments origin) already is; it's not per-render/per-order data the way `fields`, `styles`, and `orderId` of `HostedFormOptions` are. It seemed to belong next to `host`, not next to those.

2. `HostedFormOptions` is exported public API, supplied fresh on each `initialize()` call. If `storefrontHost` lived there, it should be passed on every call. I'd rather not do that.

packages/hosted-form-v2/src/hosted-field.ts
```diff
@@ -47,6 +47,12 @@ export default class HostedField {
         private _eventListener: IframeEventListener<HostedInputEventMap>,
         private _detachmentObserver: DetachmentObserver,
         private _orderId?: number,
+        // Channel's storefront origin serving the hosted-fields route, used ONLY to build the iframe src.
+        // (This is NOT the same as the `host` given to IframeEventPoster/IframeEventListener which is payments host
+        // that redirects to the payment provider, so the iframe's final document origin)
+        // This is needed only by headless storefronts, where it differs from the page's own origin.
+        // Empty keeps the relative-URL behavior, which resolves same-origin on a Stencil storefront.
```
bc-peng:
```suggestion
```
The code appears to be self-explainable enough.

packages/hosted-form-v2/src/hosted-form-factory.ts
bc-peng: Good question.
I had the same thought, but I agree with Tharaae on this one. 
- The lifecycle is correct, and
- V1 `hostedFormFactory` already has `checkoutId` as a third parameter.

bc-peng (approved):

bc-chenli (approved):

PR #3382, opened 2026-08-31
feat(payment): Stripe CS checkout session sync and store credit optimization

PavlenkoM:
### Stripe checkout session sync (`runServerUpdate`)

Changes on the BigCommerce side (e.g. price updates) were not reflected in the
mounted Stripe payment element, because Stripe was never told to re-fetch its
checkout session after our backend mutated it.

- `StripeCheckoutSessionActions.runServerUpdate` is now properly typed:
  `runServerUpdate(update: () => Promise<unknown>): Promise<StripeCheckoutSessionActionResult>`.
- `initialize()` runs a session sync after shopper data is pushed, so the
  element renders against the latest session state.
- `execute()` wraps the config request (`loadPaymentMethod`) in
  `runServerUpdate`, so the session update triggered on the BE side is pulled
  into the Stripe element before order submission.
- The result is checked: a `{ type: 'error' }` response throws a typed
  `PaymentMethodFailedError` **before** `submitOrder`, instead of silently
  proceeding against an unsynced session.

### Store credit request optimization

`execute()` previously re-applied store credit on every order whenever
`isStoreCreditApplied` was `true` — a blocking `POST .../store-credit`
round trip that is a server-side no-op (the flag is persisted on the cart, so
it can be `true` even for guest carts, and the BE early-returns when the state
already matches). Hosted checkout applies/removes store credit eagerly when the
shopper toggles the checkbox, so the strategy-level call added latency to every
Place Order click without changing anything.

- New shared `StripeIntegrationService.applyStoreCreditIfNeeded(useStoreCredit?)`,
  used by both `stripe-cs` and `stripe-ocs` (keeps the same-gateway strategies
  in sync): the API is called only when `useStoreCredit` is explicitly provided
  in the order request **and** differs from the current checkout state.
- Hosted checkout does not pass the flag → one blocking request removed from
  every order submission.
- Headless consumers passing `useStoreCredit` now get correct semantics,
  including credit **removal** on `false` (previously impossible).
- `useStoreCredit` is still passed through to `submitOrder` in the order body
  (pinned by tests).

### Safer initialization failure handling

A failure during `initialize()` (e.g. a failed session sync) now tears the
strategy down via `deinitialize()` before reporting through `onError`. A
subsequent `execute()` throws `NotInitializedError` instead of creating an
order against a strategy whose payment element never mounted.

## Rollout/Rollback
- No experiment/feature flag; ships with the next SDK release and reaches
  storefronts as checkout consumes the new version.
- Rollback: revert this PR / pin the previous SDK version. No BE changes or
  migrations; the store-credit and session-sync behavior is fully client-side.

### Stripe OCS

https://github.com/user-attachments/assets/bb3760a6-32a8-4be9-ab15-2066e4abe849

https://github.com/user-attachments/assets/8d6b7875-9de4-4a1e-b128-9bbda7fdd07f

https://github.com/user-attachments/assets/7cb146e5-2725-4a27-81fc-3af1d645e879

### Stripe CS

https://github.com/user-attachments/assets/0491b4d4-1624-474f-8d4f-94bc2a7f105e

https://github.com/user-attachments/assets/91747779-888b-499f-bc09-a900100b32c7

https://github.com/user-attachments/assets/62d17a24-d44a-485b-8d1c-8fee9285109e

---

> [!NOTE]
> **Medium Risk**
> Changes affect Stripe CS payment initialization, execute order flow, and store credit timing on place order; session sync failures now block submission, which is safer but alters failure modes at checkout.
> 
> **Overview**
> **Stripe Checkout Session (CS)** now keeps the mounted payment element aligned with BigCommerce backend changes by calling Stripe’s **`runServerUpdate`** after shopper data on **initialize**, and wrapping **`loadPaymentMethod`** in **`runServerUpdate`** on **execute**. Failed syncs raise **`PaymentMethodFailedError`** before **`submitOrder`**. Failed **initialize** now **`deinitialize()`**s first so a later **execute** throws **`NotInitializedError`** instead of submitting against an unmounted element.
> 
> **Store credit** is centralized in **`StripeIntegrationService.applyStoreCreditIfNeeded(useStoreCredit?)`**, used by both **stripe-cs** and **stripe-ocs**. Store credit is applied or removed only when **`useStoreCredit`** is explicitly set and differs from checkout state—avoiding redundant **`applyStoreCredit`** calls on every place order when hosted checkout already synced the cart.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a9f7c7689fc5f7726dc012204c027b37c82f3d99. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/stripe-integration/src/stripe-cs/stripe-cs-payment-strategy.spec.ts
+++ b/packages/stripe-integration/src/stripe-cs/stripe-cs-payment-strategy.spec.ts
@@ -13,7 +13,6 @@ import {
 import {
     getBillingAddress,
     getCart,
-    getCheckout,
     getErrorPaymentResponseBody,
     getResponse,
     getShippingAddress,
@@ -216,6 +215,32 @@ describe('StripeOCSPaymentStrategy', () => {
             expect(togglePreloaderMock).toHaveBeenCalled();
             expect(stripeScriptLoader.getStripeClient).toHaveBeenCalled();
             expect(stripeIntegrationService.mountElement).toHaveBeenCalled();
+        });
+
+        it('triggers stripe checkout session server update on initialize', async () => {
+            const runServerUpdateMock = jest.fn(async (update: () => Promise<unknown>) => {
+                await update();
+
+                return { type: StripeLoadActionsResultType.SUCCESS };
+            });
+
+            jest.spyOn(stripeScriptLoader, 'getStripeCheckout').mockReturnValue(
+                Promise.resolve({
+                    ...getStripeCheckoutInstanceMock(),
+                    loadActions: () =>
+                        Promise.resolve({
+                            type: StripeLoadActionsResultType.SUCCESS,
+                            actions: {
+                                ...getStripeCheckoutSessionActionsMock(),
+                                runServerUpdate: runServerUpdateMock,
+                            },
+                        }),
+                }),
+            );
+
+            await stripeCSPaymentStrategy.initialize(stripeOptions);
+
+            expect(runServerUpdateMock).toHaveBeenCalledTimes(1);
         });
 
         it('should call render when payment element ready event fires', async () => {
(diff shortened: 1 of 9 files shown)


PavlenkoM:
cursor review

bc-wpietrzak (approved):
LGTM

PR #3389, opened 2026-09-02
fix(payment): PAYPAL-7021 removed logger to avoid error spamming

andriiVitvitskyi1990:
## What/Why?
Removed logger to avoid `MissingDataError` error spamming

## Rollout/Rollback
Revert

## Testing
Unit test

---

> [!NOTE]
> **Low Risk**
> Narrow change to error-logging behavior during optional Fastlane init; checkout flow and logging for other errors are unchanged.
> 
> **Overview**
> **Braintree Fastlane customer `initialize`** still swallows Fastlane init failures so checkout can continue, but it no longer forwards **`MissingDataError` / `MissingPaymentMethod`** (e.g. when no client token is available) to the optional **`onErrorLog`** callback. Other initialization errors are still logged the same way.
> 
> A unit test asserts that **`onErrorLog` is not invoked** when Fastlane init rejects with that missing-payment-method error.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc72a222a52accdeb9111414331cbfd40d82e407. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/braintree-integration/src/braintree-fastlane/braintree-fastlane-customer-strategy.ts
+++ b/packages/braintree-integration/src/braintree-fastlane/braintree-fastlane-customer-strategy.ts
@@ -8,6 +8,8 @@ import {
     CustomerStrategy,
     ExecutePaymentMethodCheckoutOptions,
     InvalidArgumentError,
+    MissingDataError,
+    MissingDataErrorType,
     PaymentIntegrationService,
     PaymentMethod,
     RequestOptions,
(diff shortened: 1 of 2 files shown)


packages/braintree-integration/src/braintree-fastlane/braintree-fastlane-customer-strategy.ts
```diff
@@ -62,7 +62,6 @@ export default class BraintreeFastlaneCustomerStrategy implements CustomerStrate
             }
         } catch (error) {
             // Info: Do not throw anything here to avoid blocking customer from passing checkout flow
-            this.handleErrorLog(error);
```
serhii-tkachenko, line 65: How will you debug the issue if something new occur related Fastlane implementation? I am asking, because your change completely removes the code we can stick to during debug session

andriiVitvitskyi1990: You are right, i will extract exact error

serhiiFilkovskyi (approved):

PR #3397, opened 2026-09-07
refactor(payment): PAYPAL-5774 update BCP payment strategy by providing PaypalUtilsService

bc-yaroslav-zhmutskyi:
## What/Why?
Replace BigCommercePaymentsIntegrationService with PaypalUtilsService from paypal-utils package
Added BCP namespace to `paypal-utils/paypal-sdk-script-loader`

## Rollout/Rollback
Revert

## Testing
Manual

[bcp_paypal_button.webm](https://github.com/user-attachments/assets/58ea4e1e-6d03-41a7-8489-b6ad658dd20f)

---

> [!NOTE]
> **Medium Risk**
> Changes touch checkout PayPal button initialization and SDK loading/isolation on `window`, which can affect coexisting PayPal integrations if namespaces clash; behavior should match prior flows but warrants payment checkout regression testing.
> 
> **Overview**
> **BigCommerce Payments checkout** now depends on shared **`PayPalIntegrationService`** from `paypal-utils` instead of the local **`BigCommercePaymentsIntegrationService`** for SDK load, button styling, and order creation. The strategy factory injects **`createPayPalIntegrationService`**, and unit tests were updated to mock the PayPal utils types and service.
> 
> In **`paypal-utils`**, PayPal script loading is **namespace-aware**: method IDs starting with `bigcommerce_payments` load the SDK with **`data-namespace: bigCommercePaymentsPayPalSDK`** and resolve the instance via new **`getPayPalSdkModule`** and a **`PayPalSdkNamespace`** enum (other SDK variants now use the enum too). Constants were consolidated under **`paypal-constants`** (replacing `paypal-commerce-constants` exports).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 39b98ad5617331766f9f0c972450ecd15f3be803. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/paypal-utils/src/paypal-constants.ts
+++ b/packages/paypal-utils/src/paypal-constants.ts
@@ -4,6 +4,20 @@ export const LOADING_INDICATOR_STYLES = {
 };
 
 /**
+ * PayPalSdkNamespace maps each PayPal SDK module to the
+ * window global it registers under (also used as the `data-namespace` attribute)
+ */
+export enum PayPalSdkNamespace {
+    PayPal = 'paypal',
+    BigCommercePaymentsPayPalSDK = 'bigCommercePaymentsPayPalSDK',
+    PayPalFastlaneSdk = 'paypalFastlaneSdk',
+    PayPalFastlane = 'paypalFastlane',
+    PayPalGooglePay = 'paypalGooglePay',
+    PayPalApms = 'paypalApms',
+    PayPalMessages = 'paypalMessages',
+}
+
+/**
  * PayPal SDK supported locales
  * https://developer.paypal.com/sdk/js/configuration/#link-locale
  */
--- a/packages/paypal-utils/src/paypal-sdk-script-loader.ts
+++ b/packages/paypal-utils/src/paypal-sdk-script-loader.ts
@@ -34,7 +34,9 @@ export default class PayPalSdkScriptLoader {
         initializesOnCheckoutPage?: boolean,
         forceLoad?: boolean,
     ): Promise<PayPalSDK> {
-        if (!this.window.paypal || forceLoad) {
+        const isBigCommercePayments = this.isBigCommercePaymentsMethod(paymentMethod.id);
+
+        if (!this.getLoadedPayPalSdk(isBigCommercePayments) || forceLoad) {
             const paypalSdkScriptConfig = this.getPayPalSdkScriptConfigOrThrow(
                 paymentMethod,
                 currencyCode,
@@ -43,13 +45,15 @@ export default class PayPalSdkScriptLoader {
             );
 
             await this.loadPayPalSdk(paypalSdkScriptConfig);
-
-            if (!this.window.paypal) {
-                throw new PaymentMethodClientUnavailableError();
-            }
         }
 
-        return this.window.paypal;
+        const paypalSdk = this.getLoadedPayPalSdk(isBigCommercePayments);
+
+        if (!paypalSdk) {
+            throw new PaymentMethodClientUnavailableError();
+        }
+
+        return paypalSdk;
     }
 
     async getPayPalFastlaneSdk(
@@ -143,6 +147,16 @@ export default class PayPalSdkScriptLoader {
         }
 
         return this.window.paypalMessages;
+    }
+
+    private isBigCommercePaymentsMethod(methodId: string): boolean {
+        return methodId.startsWith('bigcommerce_payments');
+    }
+
+    private getLoadedPayPalSdk(isBigCommercePayments: boolean): PayPalSDK | undefined {
+        return isBigCommercePayments
+            ? this.window.bigCommercePaymentsPayPalSDK
+            : this.window.paypal;
     }
 
     /**
@@ -254,6 +268,9 @@ export default class PayPalSdkScriptLoader {
             attributes: {
                 'data-partner-attribution-id': attributionId,
                 'data-client-token': clientToken,
+                ...(this.isBigCommercePaymentsMethod(id) && {
+                    'data-namespace': 'bigCommercePaymentsPayPalSDK',
+                }),
             },
         };
     }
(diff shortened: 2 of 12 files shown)


andriiVitvitskyi1990 (approved):

packages/paypal-utils/src/paypal-sdk-script-loader.ts
```diff
@@ -34,7 +34,9 @@ export default class PayPalSdkScriptLoader {
         initializesOnCheckoutPage?: boolean,
         forceLoad?: boolean,
     ): Promise<PayPalSDK> {
-        if (!this.window.paypal || forceLoad) {
+        const isBigCommercePayments = this.isBigCommercePaymentsMethod(paymentMethod.id);
+
+        if (!this.getLoadedPayPalSdk(isBigCommercePayments) || forceLoad) {
```
bc-nick: We do not know does this instance exists or not at this moment. Better to rename it to `getPayPalSdkModule`.

packages/paypal-utils/src/paypal-sdk-script-loader.ts
```diff
@@ -43,13 +45,15 @@ export default class PayPalSdkScriptLoader {
             );
 
             await this.loadPayPalSdk(paypalSdkScriptConfig);
+        }
 
-            if (!this.window.paypal) {
-                throw new PaymentMethodClientUnavailableError();
-            }
+        const paypalSdk = this.getLoadedPayPalSdk(isBigCommercePayments);
```
bc-nick: Here you invoke this function one more time. I think better to reduce it to one call at the beginning of this method.

packages/paypal-utils/src/paypal-sdk-script-loader.ts
```diff
@@ -145,6 +149,16 @@ export default class PayPalSdkScriptLoader {
         return this.window.paypalMessages;
     }
 
+    private isBigCommercePaymentsMethod(methodId: string): boolean {
+        return methodId.startsWith('bigcommerce_payments');
+    }
+
+    private getLoadedPayPalSdk(isBigCommercePayments: boolean): PayPalSDK | undefined {
+        return isBigCommercePayments
+            ? this.window.bigCommercePaymentsPayPalSDK
```
bc-nick: Lets create a const for `bigCommercePaymentsPayPalSDK` and reuse it where needed.

packages/paypal-utils/src/paypal-sdk-script-loader.ts
```diff
@@ -145,6 +149,16 @@ export default class PayPalSdkScriptLoader {
         return this.window.paypalMessages;
     }
 
+    private isBigCommercePaymentsMethod(methodId: string): boolean {
+        return methodId.startsWith('bigcommerce_payments');
+    }
+
+    private getLoadedPayPalSdk(isBigCommercePayments: boolean): PayPalSDK | undefined {
```
bc-nick: I would recommend you to update this function to a more common approach since we will have take into account others modules here that will require the same changes.

    private getPayPalSdkModule<K extends keyof PayPalHostWindow>(
        namespace: K,
    ): PayPalHostWindow[K] {
        return this.window[namespace];
    }

Generic should be at the separate file in `utils` folder

packages/paypal-utils/src/paypal-sdk-script-loader.ts
bc-yaroslav-zhmutskyi: We should keep two calls because on line [#47](https://github.com/bigcommerce/checkout-sdk-js/pull/3397/files#diff-da5015542f7ea3fec0e932b1876231478ac10d6873fcba28c683af7dad740c62R47) we call `loadPayPalSdk` and after that we should call this function once more

packages/paypal-utils/src/utils/get-paypal-sdk-module.ts
```diff
@@ -0,0 +1,8 @@
+import { PayPalSdkNamespace } from '../paypal-constants';
+import { PayPalHostWindow } from '../paypal-types';
+
+export default function getPayPalSdkModule<K extends PayPalSdkNamespace>(
+    namespace: K,
+): PayPalHostWindow[K] {
+    return (window as PayPalHostWindow)[namespace];
```
bc-nick, line 7: Do we really need use `as` here?

bc-yaroslav-zhmutskyi: Yes we need this, window have standard `Window` type and we need to convert it

bc-nick: I got you. Thanks. I was wondering since if we use it as a method in class it should work without `as`  due to window is defined there. I am ok with both ways, let you decide leave it as function or move it as method in order to avoid `as`.

packages/paypal-utils/src/paypal-constants.ts
```diff
@@ -3,6 +3,19 @@ export const LOADING_INDICATOR_STYLES = {
     'z-index': '1000',
 };
 
+/**
+ * PayPalSdkNamespace maps each PayPal SDK module to the
+ * window global it registers under (also used as the `data-namespace` attribute)
+ */
+export enum PayPalSdkNamespace {
+    PayPal = 'paypal',
+    BigCommercePayments = 'bigCommercePaymentsPayPalSDK',
```
bc-nick: Could you update PayPalHostWindow with this constance for each of existing module? And better to use the same value name for the each corresponding key. It increase readability and understanding.

```
BigCommercePaymentsPayPalSDK = 'bigCommercePaymentsPayPalSDK',
PaypalFastlaneSdk = 'paypalFastlaneSdk',
...
```

packages/paypal-utils/src/utils/get-paypal-sdk-module.ts
bc-yaroslav-zhmutskyi: Lets keep it as function, will be easier to use in other utils in case if needed

bc-nick (approved):
LGTM! Thanks a lot for updates!

PR #3403, opened 2026-09-09
feat(payment): PI-5643 [Google Pay] show error on 3ds challenge failure

bc-wpietrzak:
## What/Why?

For Google Pay, when a shopper fails a 3DS challenge and is redirected back to checkout, no error was shown. The order was left silently unfinalized, and the shopper would only see an error when they tried to place an order again. The root cause is that `finalizeOrder()` was only called when `getPaymentStatus()` reported `FINALIZE`. On a declined 3DS challenge, the payment record's step rolls back to `INITIALIZE` instead of advancing, so `finalize()` treated the redirect as "nothing to do".

`GooglePayPaymentStrategy`'s `finalize` now also finalizes when `getPaymentStatus()` reports `INITIALIZE`. This can't produce a false positive on a fresh, never-attempted checkout, as `getPaymentStatus()` only ever reports `INITIALIZE`/`FINALIZE` once a payment record already exists for the method — before any attempt, it reports `undefined`. On `finalize` failure, the strategy now invalidates the cached payment method, allowing the shopper to try to place an order again with a fresh token.

checkout-js part of the fix: https://github.com/bigcommerce/checkout-js/pull/3330

## Rollout/Rollback

Revert

## Testing

Before - no error is shown after redirect (tested with Checkout.com)

https://github.com/user-attachments/assets/158980d8-d2fc-4239-b907-dc477e13f5a5

After - after failed 3DS, an error is shown after the redirect back to the checkout; previous token is invalidated, so the shopper has to reselect the card (or select a new one) to try to place an order again (tested with Checkout.com)

https://github.com/user-attachments/assets/00abfc5a-dff8-46ba-9a31-d5cc6eba43ce

---

> [!NOTE]
> **Medium Risk**
> Changes Google Pay payment finalization and token invalidation behind a feature flag, but touches checkout payment state after 3DS redirects where incorrect behavior could block retries or hide errors.
> 
> **Overview**
> Fixes silent checkout after a **failed Google Pay 3DS redirect** by extending the base `GooglePayPaymentStrategy` when feature flag **`PI-5643.google_pay_handle_unsuccessful_3ds_check`** is enabled.
> 
> **`finalize`** no longer always no-ops: it calls `finalizeOrder` when an order exists and payment status is **`FINALIZE` or `INITIALIZE`** (covering declined 3DS where status rolls back). **`finalizeOrder` errors are rethrown** instead of being swallowed, and a failed finalize triggers **`loadPaymentMethod`** to drop a stale token so the shopper must pick payment again. With the flag off, behavior stays **`OrderFinalizationNotRequiredError`**.
> 
> **Submit failures** route through **`_handleSubmitPaymentFailure`**: redirect-style additional actions unchanged, but when additional action handling **throws** (hard decline) and the flag is on, the cached method is invalidated the same way.
> 
> **BigCommerce Payments** and **PayPal Commerce** Google Pay strategies inherit the submit-path helper but **override `finalize`** to always reject finalization (their flow does not use this redirect finalize path). Specs cover flag on/off, redirect vs hard decline, and finalize success/failure.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd394a57e879e0d52acaa53f5fc7a119df822869. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bc-wpietrzak:
cursor review

bc-peng (approved):

packages/google-pay-integration/src/google-pay-payment-strategy.ts
```diff
@@ -128,11 +130,36 @@ export default class GooglePayPaymentStrategy implements PaymentStrategy {
...
+            return Promise.reject(new OrderFinalizationNotRequiredError());
+        }
+
+        const state = this._paymentIntegrationService.getState();
+        const order = state.getOrder();
+        const status = state.getPaymentStatus();
+
+        if (
+            order &&
+            (status === PaymentStatusTypes.FINALIZE || status === PaymentStatusTypes.INITIALIZE)
```
bc-maxy: It's intentional right that we try finalizeOrder not only on PaymentStatusTypes.FINALIZE but also on PaymentStatusTypes.INITIALIZE?

bc-maxy (approved):

PavlenkoM (approved):

PR #3407, opened 2026-09-13
fix(common): GH-3121 Export RequestError from checkout-sdk bundle

### LINKED ISSUE #3121: `RequestError` is not exported
**Describe the bug**
The `RequestError` class is no longer exported, although it [still appears in the type declaration](https://github.com/bigcommerce/checkout-sdk-js/blob/43a34598423f20366f8eeb5656850601bbdc992c/dist/types/checkout-sdk.d.ts#L7860). This bug was introduced in [version 1.803.0](https://github.com/bigcommerce/checkout-sdk-js/compare/v1.802.0...v1.803.0).

**To Reproduce**
With `"type": "module"`:

```js
import { RequestError } from "@bigcommerce/checkout-sdk";
```

```
import { RequestError } from "@bigcommerce/checkout-sdk";
         ^^^^^^^^^^^^
SyntaxError: Named export 'RequestError' not found. The requested module '@bigcommerce/checkout-sdk' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:
```

With `"type": "commonjs"`:

```js
const { RequestError } = require("@bigcommerce/checkout-sdk");

console.log(RequestError);
```

Logs `undefined`.

**Expected behavior**
`RequestError` should be available as a top-level export to avoid breaking changes.

**Desktop (please complete the following information):**
 - OS: N/A
 - Browser N/A
 - Version N/A

**Smartphone (please complete the following information):**
 - Device: N/A
 - OS: N/A
 - Browser N/A
 - Version N/A

**Additional context**

cpruijsen:
`RequestError` stopped being exported in 1.803.0 while remaining in the published type declarations,
so code that imports it type-checks and then fails at runtime with an undefined binding. It is the
error type consumers need in order to tell a request failure from any other rejection, so there is
no equivalent they can reach for instead.

This re-exports it from `packages/core/src/bundles/checkout-sdk.ts`, the bundle entry the other
public error types already go through, which puts the runtime export back in agreement with the
declaration file.

Fixes #3121

---

> [!NOTE]
> **Low Risk**
> Adds a missing public re-export and a unit test; no behavioral change to error handling inside the SDK.
> 
> **Overview**
> Restores **`RequestError`** as a **runtime** export from the `checkout-sdk` bundle entry, re-exporting it from `../common/error/errors` alongside the other public surface. That aligns the published bundle with the type declarations again so consumers can import the constructor and use `instanceof RequestError` (or equivalent) when handling failed checkout API calls instead of hitting an undefined binding at runtime.
> 
> Adds a small **`checkout-sdk.spec.ts`** regression test that asserts the bundle’s `RequestError` is the same constructor as the core implementation and that instances keep the expected `name`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit acefb9a5a6563d8045d024dbe2b60d97195c5ac9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bc-maxy (approved):
LGTM, thanks for contributing!

PR #3416, opened 2026-09-22
fix(checkout): CHECKOUT-10425 Support B2B gateway filtering

bc-peng:
Jira: [CHECKOUT-10425](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-10425)

## What/Why?

Checkout FE's B2B filter matches on method `id` only before.
Using `gateway` as the second identifier is the gap this PR closes.

## Rollout/Rollback

No new flags or migrations. The filter only runs when the existing `b2bPaymentMethodFilterType` capability is set, so blast radius is limited to B2B stores already on company payment filtering. Rollback is a straight revert of this commit.

The change only ever widens the match, so no method that renders today can stop rendering. The reverse is possible by design: a company record whose code is a bare method id shared by several gateways (e.g. `credit_card`) now enables all of them. That is the legacy behaviour and is covered by a test so it is not changed by accident.

## Testing

`packages/core/src/payment/b2b-company-payment-method-filter-transformer.spec.ts` — added coverage for gateway matching (allow-listed, disabled record, two providers sharing `credit_card`, shared-id fan-out) and for a legacy provider matching its raw code.

Verified the new tests actually pin the change by reverting the source file to its prior state and re-running: 5 failed / 10 passed. With the fix in place, 12/12 pass. `tsc --noEmit` on `packages/core` and ESLint are both clean.

```
npx jest --config packages/core/jest.config.js --testPathPattern b2b-company-payment-method-filter-transformer
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[CHECKOUT-10425]: https://bigcommercecloud.atlassian.net/browse/CHECKOUT-10425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes which payment methods appear at B2B checkout when company filtering is enabled; the match logic only widens visibility but affects payment selection for those stores.
> 
> **Overview**
> Fixes B2B checkout showing **no payment methods** when company allow-lists use gateway codes (e.g. `bluesnapdirect`) while storefront methods use a shared id like `credit_card` plus `gateway`.
> 
> **`filterPaymentMethodsByB2BCompanyAllowList`** now keeps a method if the enabled B2B `code` matches **`PaymentMethod.id`**, **`PaymentMethod.gateway`**, or a **legacy alias** (e.g. `quickbooks` → `qbmsv2`)—instead of only comparing id via the legacy map. That restores parity with the legacy B2B filter: gateway-backed providers can match, two `credit_card` gateways can be split by enabled gateway code, and legacy providers can match both raw and mapped codes.
> 
> Specs add coverage for gateway allow/disabled cases, shared `credit_card` ids, and raw legacy codes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1d97dd54f147c8618a3a632b1a9cb94468d512de. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bc-peng:
cursor review

valentindellangela (approved):

PR #3417, opened 2026-09-22
fix(payment): PAYPAL-7076 send Google Pay billing address on PayPal approve

bc-dmytro-khromov:
## What/Why?

Google Pay transactions show `--` in the Name column of PayPal's merchant reports
([PAYPAL-7076](https://bigcommercecloud.atlassian.net/browse/PAYPAL-7076)).

The cause is not a missing field but a discarded one. `GooglePayGateway#getCardParameters`
already requests `billingAddressRequired: true` with `format: 'FULL'`, so Google Pay returns a
complete billing address on every payment. The BCP and PPCP strategies then override `execute()`
and rebuild `paymentMethodData.info` by hand from `initializationData.card_information` —
`cardNetwork` and `cardDetails` only — dropping the address before it ever reaches PayPal.

This adds a shared `getApprovalBillingAddress` helper used by both strategies. It reads the
billing address back out of checkout state, re-encodes it as a `GooglePayFullBillingAddress`, and
passes it as the top-level `billingAddress` argument to `confirmOrder`; PayPal's JS SDK copies
that onto `paymentMethodData.info.billingAddress` before issuing the `ApproveGooglePayPayment`
mutation. Reading from state is faithful rather than a guess — the base strategy has already
written Google Pay's own billing address there via `updateBillingAddress`, so the helper is the
exact inverse of `GooglePayGateway#_mapToAddressRequestBody`.

**The field is all-or-nothing.** PayPal maps it onto
`payment_source.google_pay.card.billing_address` and, once that object exists, rejects the whole
request without a country code:

```
/payment_source/google_pay/card/billing_address/country_code   MISSING_REQUIRED_PARAMETER
```

So the helper omits `billingAddress` entirely unless both a name and a country code are present:
a partial address fails the payment outright rather than degrading.

Scope: only these two strategies build this payload. Braintree Google Pay runs on the base
`GooglePayPaymentStrategy.execute()`, which never constructs `paymentMethodData.info`, and is
unaffected.

## Rollout/Rollback

No experiment, no feature flag, no migration. Rollback is a plain revert.

## Testing

Verified end to end against PayPal sandbox:

- **before** — Google Pay order completes, transaction shows `--` in the Name column;
- **after** — same flow, buyer name present in PayPal's transactions list.

**BCP**

https://github.com/user-attachments/assets/9daf71d7-e6a6-4c35-840b-0e80448faeee

**PPCP**

https://github.com/user-attachments/assets/32305379-8aab-43cf-b0f9-7c2d6ef281e3

[PAYPAL-7076]: https://bigcommercecloud.atlassian.net/browse/PAYPAL-7076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes the PayPal Google Pay `confirmOrder` payload during checkout; omission rules limit breakage, but incorrect mapping could still fail approvals or misreport payer data.
> 
> **Overview**
> Fixes PayPal merchant reports showing `--` for Google Pay buyer names by **including billing address on `Googlepay().confirmOrder`** for BigCommerce Payments and PayPal Commerce Google Pay flows.
> 
> A shared **`getApprovalBillingAddress`** helper rebuilds the address from checkout state (already populated from Google Pay) and passes it as top-level `billingAddress` on the approve payload; PayPal’s SDK maps that onto `paymentMethodData.info.billingAddress`. The field is **omitted unless both name and country code** are present, because PayPal rejects partial `billing_address` objects.
> 
> SDK typings in **`bigcommerce-payments-utils`** and **`paypal-commerce-utils`** gain optional `billingAddress` on `ConfirmOrderConfig`. Unit tests cover full address passthrough and omission when name or country is missing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5d3073e3384ed9017efa8a44a21fb4adb3ca2535. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


andriiVitvitskyi1990 (approved):
LGTM!

packages/google-pay-integration/src/google-pay-bigcommerce-payments/google-pay-bigcommerce-payments-payment-strategy.spec.ts
```diff
@@ -244,6 +245,77 @@ describe('BigCommercePaymentsGooglePayPaymentStrategy', () => {
             expect(sdk.Googlepay().initiatePayerAction).toHaveBeenCalled();
         });
 
+        describe('purchaser billing address in the approveGooglePayPayment payload', () => {
```
koa75, line 248: Nice test coverage for the missing name and missing country code cases! One small suggestion: could we also add a test where getBillingAddress() returns undefined? The helper handles this case, but it isn't covered by a test yet.

packages/google-pay-integration/src/get-approval-billing-address.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Rebuilds the Google Pay billing address to send with the PayPal approve request.
+ *
+ * Google Pay returns a full billing address on the payment data response, but the PayPal
+ * strategies rebuild `paymentMethodData.info` from the stored card information and drop it,
+ * so PayPal's merchant reports show "--" in the Name column. The base strategy has already
+ * written that same address into checkout state via `updateBillingAddress`, so this is the
+ * inverse of `GooglePayGateway#_mapToAddressRequestBody`.
+ */
+export default function getApprovalBillingAddress(
```
koa75, line 14: Since this helper contains the main logic of the fix, would it make sense to give it its own get-approval-billing-address.spec.ts file? Right now it's only tested indirectly through the two strategies. A direct spec would make it easier to cover edge cases, for example a name that consists only of spaces.

packages/google-pay-integration/src/google-pay-bigcommerce-payments/google-pay-bigcommerce-payments-payment-strategy.spec.ts
bc-dmytro-khromov: The undefined case is covered in the new `get-approval-billing-address.spec.ts`, alongside the helper's other guards.

packages/google-pay-integration/src/get-approval-billing-address.ts
bc-dmytro-khromov: Agreed, it deserved its own spec. `get-approval-billing-address.spec.ts` now covers the full mapping, first name only, blank phone (key omitted rather than sent empty), no billing address, whitespace-only name, and missing country code.

koa75 (approved):

